PR #2, opened 2018-08-09
Better newtypes

ltratt:
This is a scary looking PR, but its basic idea is simple: to use newtypes in a better way in grmtools. Previously we had a struct that looked roughly like this:

```rust
struct NTIdx { v: u16 }
```

which then exposed an API which made it look like `NTIdx` was storing `u32`s (this was my attempt at future proofing). It was a bad idea, and made the code a lot more convoluted. This PR moves to using a tuple struct with a user specified storage size, along the lines of:

```rust
struct NTIdx<StorageT> (StorageT)
```

This is much easier to work with, and allows the user to choose the storage size that works best for them. [If they don't choose, things default to `u32` which should be big enough for even the craziest of crazy grammars -- if someone has more than 4294967295 productions in a grammar, they probably have lots of other problems.] It will also brings better static type checking and will make the compile-time generation of constants (using one of the `builder.rs` APIs) possible without forcing the user to pull in additional dependencies.

The basic motivation is explained in https://github.com/softdevteam/grmtools/commit/ea7fc39976587253aa3496f05556a27ee39a5f5e: that commit is also, unfortunately, borderline unreviewable, because so many changes are needed to make the system compile given the changes required (threading `StorageT` through causes ripples throughout the system).

The several following commits (https://github.com/softdevteam/grmtools/commit/a8a4b54a18b4709f4465e8841539c572e77f8ef1, https://github.com/softdevteam/grmtools/commit/ae799bc02d54524073eb1c6eadb63358e6e3c2ca, https://github.com/softdevteam/grmtools/commit/be9ffe0d7b3f76488dcfda6d2eee21e860bc04a4) are in the same vein but have a much easier job, because https://github.com/softdevteam/grmtools/commit/ea7fc39976587253aa3496f05556a27ee39a5f5e provides most (not all, but most) of the necessary infrastructure.

There are then a few tidy-up commits at the end.

I think this PR is useful as it stands, and you can get a little taste of that in https://github.com/softdevteam/grmtools/commit/536a889ca3339d6643ad460a6083a6418d5698e8, but I hope for more to come later. However, this PR is big enough already, so I think it's best to try and review this (as best we can), and do more stuff in later PRs.


cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -287,60 +309,60 @@ impl YaccGrammar {
...
-    pub fn nonterm_name(&self, i: NTIdx) -> &str {
+    pub fn nonterm_name(&self, i: NTIdx<StorageT>) -> &str {
         &self.nonterm_names[usize::from(i)]
     }
 
     /// Return an iterator which produces (in no particular order) all this grammar's valid `NTIdx`s.
-    pub fn iter_nonterm_idxs(&self) -> Box<Iterator<Item=NTIdx>> {
-        Box::new((0..self.nonterms_len).map(NTIdx::from))
+    pub fn iter_nonterm_idxs(&self) -> Box<dyn Iterator<Item=NTIdx<StorageT>>> {
+        Box::new((0..self.nonterms_len).map(|x| NTIdx::from(x)))
```
snim2: I think you can remove the closure here: `map(NTIDx::from)` or some such?

lrtable/src/lib/mod.rs
```diff
@@ -69,7 +74,7 @@ impl From<u32> for StIdx {
 
 impl From<usize> for StIdx {
     fn from(v: usize) -> Self {
-        if v > u16::max_value() as usize {
+        if v > usize::from(u16::max_value()) {
```
snim2: Just below this (but not in the diff) there are a couple more `as u16` that could be `u16::from`

ltratt: I've only done the ones that Clippy complained about. My plan is to do a much bigger tidy up of the integer conversions in a later PR (which is now, hopefully, much easier than it was before).

cfgrammar/src/lib/yacc/grammar.rs
ltratt: I did try this after Clippy complained, but for reasons I don't fully understand this leads to an error:

```
   --> cfgrammar/src/lib/yacc/grammar.rs:332:9
    |
113 | impl<StorageT: PrimInt + Unsigned> YaccGrammar<StorageT> {
    |      --------- help: consider adding an explicit lifetime bound `StorageT: 'static`...
...
332 |         Box::new((0..self.nonterms_len).map(NTIdx::from))
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: ...so that the type `std::iter::Map<std::ops::Range<u32>, fn(u32) -> idxnewtype::NTIdx<StorageT> {<idxnewtype::NTIdx<StorageT> as std::convert::From<u32>>::from}>` will meet its required lifetime bounds
   --> cfgrammar/src/lib/yacc/grammar.rs:332:9
    |
332 |         Box::new((0..self.nonterms_len).map(NTIdx::from))
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

To me the code looks equivalent before and after...

cfgrammar/src/lib/u32struct.rs
```diff
@@ -71,6 +76,53 @@ macro_rules! u32struct {
...
+
+        impl<T: PrimInt + Unsigned> From<$n<T>> for u32 where T: Unsigned {
+            fn from(st: $n<T>) -> Self {
+                num_traits::cast(st.0).unwrap()
+            }
+        }
+
+        impl<T: PrimInt + Unsigned> TryFrom<$n<T>>
+                                for u8
+                              where T: Unsigned, usize: From<T>
```
snim2: Should the `for` and the `where` be aligned here?

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -101,14 +104,33 @@ pub struct YaccGrammar {
...
-    /// that references the user defined start rule.
-    pub fn new(yacc_kind: YaccKind, ast: &ast::GrammarAST) -> YaccGrammar {
+    /// As we're compiling the `YaccGrammar`, we add a new start rule (which we'll refer to as `^`,
+    /// though the actual name is a fresh name that is guaranteed to be unique) that references the
+    /// user defined start rule.
+    pub fn new_with_storaget(yacc_kind: YaccKind, s: &str)
+                          -> Result<Self, YaccGrammarError>
+    {
+        let ast = match yacc_kind {
+            YaccKind::Original | YaccKind::Eco => {
```
snim2: If there is only one "arm" in this `match`, is there any reason not to say `match yacc_kind {  _ => ... };`?

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -273,7 +295,7 @@ impl YaccGrammar {
             }
         }
 
-        YaccGrammar{
+        Ok(YaccGrammar{
```
snim2: Is there a reason for not calling `YaccGrammar::new` here?

cfgrammar/src/lib/u32struct.rs
ltratt: Oops, somehow I pressed return randomly here (dunno how). Fixed.

cfgrammar/src/lib/yacc/grammar.rs
ltratt: Yes, because we *might* in future add `YaccKind`s which can't use this code path, and then we'll get a static type error reminding us to fix things.

lrpar/src/lib/mf.rs
```diff
@@ -612,17 +624,19 @@ pub(crate) fn simplify_repairs(all_rprs: &mut Vec<Vec<ParseRepair>>)
...
+impl<StorageT: Hash + PrimInt + Unsigned> Dist<StorageT> {
+    pub(crate) fn new<F>
+                     (grm: &YaccGrammar<StorageT>,
+                      sgraph: &StateGraph<StorageT>,
+                      stable: &StateTable<StorageT>,
+                      term_cost: F)
+                   -> Dist<StorageT>
                    where F: Fn(TIdx) -> u8
     {
         // This is an extension of dist from the KimYi paper: it also takes into account reductions
```
snim2: Should this comment be indented another couple of spaces?

cfgrammar/src/lib/yacc/grammar.rs
ltratt: This is basically `YaccGrammar::new` (well, `YaccGrammar::new_with_storaget`, but same difference), so we don't an infinitely recursive call here :)

lrpar/src/lib/mf.rs
ltratt: It's indented to 4 spaces already?

cfgrammar/src/lib/yacc/grammar.rs
snim2: OK

cfgrammar/src/lib/yacc/grammar.rs
snim2: OK

lrpar/src/lib/mf.rs
snim2: I think you're right. For some reason it looks really wrong in GH.

cfgrammar/src/lib/yacc/grammar.rs
snim2: Yes, I couldn't get this to work either :(

ltratt: I feel like this might be a type inference bug, but honestly I'm not sure.

lrtable/src/lib/mod.rs
snim2: Clippy complains about the ones I mentioned, but I think it's reasonable to defer that to a separate PR, since you haven't touched those lines here.

snim2:
OK, please squash.

ltratt:
Squashed.

PR #5, opened 2018-08-18
Clear up integer types

ltratt:
This PR is a fairly huge clean-up of integer type usage in the grmtools repo. The basic problem is that integer types in Rust don't quite match what we want. One can infallibly (i.e. guaranteed to succeed) convert `u8` to `usize`, and `u16` to `usize`; but `u32` (and bigger) can't be infallibly converted to `usize`. I can't explain why Rust does this: it makes sense that `u8` is infallibly convertible to `usize`, but if `u16` is also infallibly convertible, why isn't `u32`? I'm not aware of many 16 bit machines lurking around, and I doubt many Rust programs would run on them without changes to the library (certainly) and (possibly) the language. Anyway, them's the rules...

With those rules in mind, conversions involving `as` are problematic: anything from `u32` to `usize` is fundamentally undefined. Also, and at least as troubling, `as` hides problems: `usize` to (say) `u32` might be defined, but we might truncate a `usize` which would overflow a `u32` and not notice.

This PR therefore has as its ultimate goal auditing everything involving `as`. It aims to remove as many `as` conversions as possible; and, for those that have to remain, to guarantee that they are safe. This is not a small job. The basic tactic is to gradually migrate the repo to use the newtypes we define in more places (and to reduce the number of places that instantiate the newtypes by introducing new `iter_` methods that are easily audited e.g. https://github.com/softdevteam/grmtools/commit/385a8acb4077a402861fc3cdfdf23fa6debc614a). This gradually allows us to remove some of the undefined conversion code entirely (e.g. https://github.com/softdevteam/grmtools/commit/610526db34fb233aab9cb11ee37e059c1fde4796).

There are a lot of commits here, but each is the minimal unit which compiles and passes all tests. They're not entirely linear, but in this sort of conversion, it's impossible to entirely predict the path that will need to be taken. Hopefully they make some sort of sense; the important commits all have detailed explanations.

The good news is that this PR moves the code from having over 70 unaudited `as` conversions to having only 3 left. Those 3 are in lrtable, which needs a little bit of an overhaul to really finish the job (it should probably have more newtypes than it currently does). That's big enough that I think that's best left for another PR: this one is already big enough!


lrpar/src/lib/builder.rs
```diff
@@ -125,9 +148,9 @@ pub fn process_file<StorageT, TokId, P, Q>(inp: P,
     outs.push_str(&format!("use lrpar::{{Node, parse_rcvry, ParseError, reconstitute, RecoveryKind}};
 use lrlex::Lexeme;
 
-pub fn parse(lexemes: &Vec<Lexeme<{tn}>>)
-          -> Result<Node<{storaget}, {tn}>, (Option<Node<{storaget}, {tn}>>, Vec<ParseError<{storaget}, {tn}>>)>
-{{", storaget=StorageT::type_name(), tn=TokId::type_name()));
+pub fn parse(lexemes: &Vec<Lexeme<{storaget}>>)
```
snim2: What does `Lexeme<{storaget}>` mean?

ltratt: It's turned into whatever type the user specified. So if you specified `u8` this would generate code `pub fn parse(lexemes: &Vec<Lexeme<u8>>)`.

lrtable/src/lib/stategraph.rs
```diff
@@ -59,6 +59,15 @@ where usize: AsPrimitive<StorageT>
         StateGraph{states, edges}
     }
 
+    /// Return an iterator which produces (in order from `0..self.nonterms_len()`) all this
+    /// grammar's valid `NTIdx`s.
+    pub fn iter_stidxs(&self) -> Box<dyn Iterator<Item=StIdx>>
+    {
+        // We can use as_ safely, because we know that we're only generating integers from
+        // 0..self.states.len() which we've already checked fits within StIdxStorageT.
```
snim2: Backquotes?

ltratt: It's a non-doc comment, so it's not markdown formatted.

snim2:
This generally looks fine. A couple of comments about commit messages:

In the commit message for 385a8acb4077a402861fc3cdfdf23fa6debc614a you state that one idiom is being replaced by another but it would help the reader to know why.

The commit message for 4e83a17578c8dd002d6ec9e5716abf2106d58d94 has a typo in the third sentence "That makes it very hard to deal code that uses integers of different sizes". 

Also, I see Clippy still has quite a bit to say about this repo (although not all of the clippy feedback is helpful) but since this PR is already huge, perhaps it's not worth following up here.

Please squash!

ltratt:
I've fixed both commit messages. Unless you have suggestions, I'd prefer to keep the commits separated, otherwise this just stuff ends up being completely mind-boggling.

Most of the remaining Clippy feedback is either noise or, in some cases, wrong. It would be nice if the `clippy.toml` file allowed one to turn such warnings off, but it seems one would have to use `cfg` tricks to do so. I must admit to being a bit reluctant to have those in the code, but perhaps I should just swallow my pride...

snim2:
Well, the clippy directives can go in mod.rs or lib.rs, so they don't have to be scattered around the crate. I don't think it's *so* bad!

PR #6, opened 2018-08-22
Better builder API

ltratt:
The API for statically generating grammars was, previously, pretty bad: in particular, it was a bit fiddly to use, not very flexible, and not very extendible. This PR moves it to a struct-based system that solves these problems as best as I know how to. It also provides more error checking, and a safe way to avoid recompiling grammars when possible. The commits are a bit of a hodge-podge, but hopefully each one makes clear what it's doing.

### DIFF
--- a/lrpar/src/lib/builder.rs
+++ b/lrpar/src/lib/builder.rs
@@ -35,13 +35,15 @@ use std::convert::AsRef;
 use std::env::{current_dir, var};
 use std::error::Error;
 use std::fmt::Debug;
-use std::fs::{File, read_to_string};
+use std::fs::{self, File, read_to_string};
 use std::hash::Hash;
 use std::io::Write;
 use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
 
+use cfgrammar::Grammar;
 use cfgrammar::yacc::{YaccGrammar, YaccKind};
+use filetime::FileTime;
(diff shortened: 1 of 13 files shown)


lrpar/src/lib/builder.rs
```diff
@@ -152,10 +154,30 @@ where StorageT: 'static + Debug + Hash + PrimInt + Serialize + TypeName + Unsign
...
                                       .collect::<HashMap<_, _>>();
+        let imc = self.gen_ids_map_cache(&grm);
+
+        // We don't need to go through the full rigmarole of generating an output file if all of
+        // the following are true: the output file exists; it is newer than the input file; and the
+        // rule IDs map cache hasn't changed. The last of these might be surprising, but it's
+        // vital: we don't know what the IDs map might be from one run to the next, and it might
+        // change for reasons beyond lrpar's control. If it does change, that means that the lexer
+        // and lrpar would get out of sync, so we have to play it safe and regenerate in such cases.
+        if let Ok(ref inmd) = fs::metadata(&inp) {
```
snim2: It seems odd to me to have so many nested `if`s rather than using `&&`

ltratt: You can't chain `if let`s :/ Which is very annoying.

snim2:
OK, this looks uncontroversial. Do you want to squash?

ltratt:
I think the commits are best left as they are.

PR #10, opened 2018-08-30
The recoverer needs to be part of the rebuild cache.

ltratt:
The rebuild cache is a very useful performance feature, because it stops us spending time generating grammar information and having rustc compile it (the latter is much slower than the former). However, previously we were a bit sloppy, not included the recovered as part of the rebuild cache.

This commit fixes that oversight, and also documents a little more clearly what expectations the rebuild cache needs to satisfy.

### DIFF
--- a/lrpar/src/lib/builder.rs
+++ b/lrpar/src/lib/builder.rs
@@ -62,6 +62,8 @@ const STABLE_FILE_EXT: &str = "stable";
 /// A `ParserBuilder` allows one to specify the criteria for building a statically generated
 /// parser.
 pub struct ParserBuilder<StorageT=u32> {
+    // Anything stored in here almost certainly needs to be included as part of the rebuild_cache
+    // function below so that, if it's changed, the grammar is rebuilt.
     recoverer: RecoveryKind,
     phantom: PhantomData<StorageT>
 }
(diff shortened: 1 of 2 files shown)


lrpar/src/lib/builder.rs
```diff
@@ -235,26 +238,33 @@ pub fn parse(lexemes: &[Lexeme<{storaget}>])
...
-    /// later.
-    fn gen_ids_map_cache(&self, grm: &YaccGrammar<StorageT>) -> String {
+    /// Generate the cache, which determines if anything's changed enough that we need to
+    /// regenerate outputs and force rustc to recompile.
+    fn rebuild_cache(&self, grm: &YaccGrammar<StorageT>) -> String {
+        // We don't need to be particularly clever here: we just need to record the various things
+        // that could change between builds.
         let mut cache = String::new();
         cache.push_str(" \n/* CACHE INFORMATION\n");
+
```
snim2: I'm not sure that you really need these extra spaces, especially when there are already comments to break up the code.

ltratt: They're to break up the different parts of the rebuild cache. I think whoever comes along in the future might be grateful for the visual clues. [Given that I got this bit wrong, and I'm the idiot who wrote it, I'm assuming it'll be harder for the people who come after me ;)]

snim2: Do you not think the comments are sufficient for that? Or could the comments be improved?

ltratt: Whitespace inside functions is like a paragraph break to me; functions are sections. I wouldn't want to read sections in papers without paragraphs!

PR #11, opened 2018-09-18
Homogenise naming

ltratt:
The names we used in grmtools were previously very inconsistent. For example, we sometimes called things tokens, sometimes terminals; sometimes rules, sometimes nonterminals, sometimes non-terminals; and so on. In other words, we were not consistent with concepts. To add insult to injury, we were then horribly inconsistent with variable names (and, to a lesser extent, with parameters and function names). For example, there was `ntidx`, `nt_idx`, `rule_i`, `r_idx` and (my personal favourite) `nidx` (a clear typo, but it happened in two functions!).

This PR first attempts to settle on simple terminology in https://github.com/softdevteam/grmtools/commit/4c36cdacf76f20e5b18456226df8986e435b6656 -- here's the important bits:

```
//! A library for manipulating Context Free Grammars (CFG). It is impractical to fully homogenise
//! all the types of grammars out there, so the aim is for different grammar types
//! to have completely separate implementations. Code that wants to be generic over more than one
//! grammar type can then use an "adapter" to homogenise the particular grammar types of interest.
//! Currently this is a little academic, since only Yacc-style grammars are supported (albeit
//! several variants of Yacc grammars).
//!
//! Unfortunately, CFG terminology is something of a mess. Some people use different terms for the
//! same concept interchangeably; some use different terms to convey subtle differences of meaning
//! (but without complete uniformity). "Token", "terminal", and "lexeme" are examples of this: they
//! are synonyms in some tools and papers, but not in others.
//!
//! In order to make this library somewhat coherent, we therefore use some basic terminology
//! guidelines for major concepts (acknowledging that this will cause clashes with some grammar
//! types).
//!
//!   * A *grammar* is an ordered sequence of *productions*.
//!   * A *production* is an ordered sequence of *symbols*.
//!   * A *rule* maps a name to one or more productions.
//!   * A *token* is the name of a syntactic element.
//!
//! For example, in the following Yacc grammar:
//!
//!   R1: "a" "b" | R2;
//!   R2: "c";
//!
//! the following statements are true:
//!
//!   * There are 3 productions. 1: ["a", "b"] 2: ["R2"] 3: ["c"]`
//!   * There are two rules: R1 and R2. The mapping to productions is {R1: {1, 2}, R2: {3}}
//!   * There are three tokens: a, b, and c.
```

The aim here isn't to pick "perfect" terminology: I doubt such a thing exists, and if it does, it's beyond me. But it's hopefully simple enough that it won't complete confuse people.

The commits after that then try to change the entire code base to respect the above terminology. Each commit aims to be a sort-of atomic unit which shifts things in that direction; to that end, I can confirm that each commit passes all tests. Despite the scary number of commits, most are trivial: those of the form "mechanically rename X to Y with [srep](https://tratt.net/laurie/src/srep/)" are mechanical search-and-replace, so it's going to be hard to spot problems (e.g. if I got the search and replace wrong, we might have variab;e shadowing): thus, a brief flick at each is probably enough. The commits "homogenise variable names in X" are all done by hand, and might be worth some checking for typos if nothing else.

This wasn't the most fun PR to do; and definitely won't be the most fun PR to review; but it starts shifting the code in a direction that will hopefully make future maintenance more palatable.

### DIFF
--- a/cfgrammar/src/lib/idxnewtype.rs
+++ b/cfgrammar/src/lib/idxnewtype.rs
@@ -66,12 +66,6 @@ macro_rules! IdxNewtype {
     }
 }
 
-// Will anyone create a grammar with more than 65535 non-terminals, productions, symbols within a
-// production, or terminals? Yes, now that I've said it out loud, they probably will. But all
-// practical grammars I know of are comfortably within these limits, so use narrow storage types
-// for now, knowing that we can transparently move the storage type from u16 to u32 in the future
-// without changing the user visible API.
-
 IdxNewtype!(
     /// A type specifically for rule indices.
     ///
--- a/cfgrammar/src/lib/idxnewtype.rs
+++ b/cfgrammar/src/lib/idxnewtype.rs
@@ -73,7 +73,7 @@ macro_rules! IdxNewtype {
 // without changing the user visible API.
 
 IdxNewtype!(
-    /// A type specifically for nonterminal indices.
+    /// A type specifically for rule indices.
     ///
     /// It is guaranteed that `RIdx` can be converted, without loss of precision, to `usize` with
     /// the idiom `RIdx::from(x_usize)`. `usize` values can be converted to `RIdx`, causing a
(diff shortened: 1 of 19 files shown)


ltratt:
A test has failed, but it seems to be a checksum in a download: I'll kick it and hope it magically goes away.

ltratt:
OK, kicking it worked.

ptersilie:
Okay, looks ok. Do you want to squash anything down or leave it as is?

ltratt:
As-is please.

PR #12, opened 2018-09-18
Remove traits

ltratt:
My original plan was to have a sort-of OO-ish superclass-ish trait called Grammar, which various grammar types (e.g. Yacc, ANTLR) could implement. This might have worked *but* the performance was going to be terrible. The basic problem is that dynamic dispatch in Rust can be horribly inefficient, and complex traits tend to hit those bad cases.

For example, at one point I had this:

```rust
trait Grammar {
  pub fn iter_pidxs<'a>(&'a self, ntidx: NTIdx<StorageT>)
                     -> Box<Iterator<Item=PIdx<StorageT>> + 'a>;
}

impl Grammar for YaccGrammar {
  pub fn iter_pidxs<'a>(&'a self, ntidx: NTIdx<StorageT>)
                     -> Box<impl Iterator<Item=PIdx<StorageT>> + 'a>
  {
      self.rules_prods[usize::from(ntidx)].iter().cloned()
  }
}
```

This looks innocuous, but it made compiling the Java grammar over 30% slower compared to:

```rust
impl YaccGrammar {
  pub fn iter_pidxs<'a>(&'a self, ntidx: NTIdx<StorageT>)
                     -> impl Iterator<Item=PIdx<StorageT>> + 'a
  {
      self.rules_prods[usize::from(ntidx)].iter().cloned()
  }
}
```

A 30% slowdown for something so simple is not exactly encouraging, and going further down this route seems only like to make the problem ever worse.

Therefore remove the Grammar trait entirely and (since it's currently our only grammar type) move stuff into YaccGrammar; do the same with the Firsts trait. In the future, my gut feeling is that the right way for an application to deal with multiple grammar types will be to have grammar "adapters" that can turn incompatible-underneath grammar types into seemingly compatible grammars. With luck, the generic guarantees that cfgrammar makes about grammars (in src/lib/mod.rs) will prove tenable for different grammar types going forward, but we can cross that bridge when we come to it.

### DIFF
--- a/cfgrammar/src/lib/mod.rs
+++ b/cfgrammar/src/lib/mod.rs
@@ -81,8 +81,6 @@ extern crate num_traits;
 #[cfg(feature="serde")] #[macro_use] extern crate serde;
 extern crate vob;
 
-use num_traits::{AsPrimitive, PrimInt, Unsigned};
-
 mod idxnewtype;
 pub mod yacc;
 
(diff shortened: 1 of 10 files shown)


ptersilie:
Looks good. Merged.

PR #13, opened 2018-09-19
Cherry pick output from rustfmt.

ltratt:
The formatting of various parts of grmtools has become quite poor, mostly because of very recent changes. Manually reformatting it is unappealing, so I hoped that [rustfmt](https://github.com/rust-lang-nursery/rustfmt) could do the job. It almost can. However, it has a bug or two which makes it generate incorrect output in a couple of cases; it struggles to format a handful of inline comments correctly (which is a bit weird); and it makes some of the tests much, much harder to read (which isn't really its fault: it just can't know that formatting things in a different way is the right thing to do).

What I've done is manually select as much of the formatting it's done as possible. I've probably selected about 95% of the non-test changes; but quite a lot of the tests I've left as-is, because they're easier to read as they are (splitting them over multiple lines, as is rustfmt's wont, makes comparing multiple `assert`s harder).

It might, in the future, be possible to let rustfmt loose over the whole codebase. Until then, this is, I think, better than nothing.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,6 @@
+indent_style="Block"
+imports_indent="Block"
+trailing_comma="Never"
+use_try_shorthand=true
+use_field_init_shorthand=true
+merge_imports=true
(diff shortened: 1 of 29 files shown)


ltratt:
As discussed offline, the new commit formats structs in a different way (and also includes a couple of reasonable-rustfmt-things that I missed in the first round). If you're OK with this, I suggest that I squash things down.

ptersilie:
Squash please!

ltratt:
Squashed.

PR #14, opened 2018-09-20
Force lrpar dependencies to be rebuilt if lrpar is recompiled.

ltratt:
Previously, if you changed the source code of lrpar, then libraries depending on it wouldn't be recompiled. This caused me some surprise and pain recently when I changed lrpar and couldn't work out why other libraries were not doing the right thing.

This commit uses vergen to embed the timestamp of the build into the lrpar cache; if lrpar is recompiled (e.g. as a result of someone changing the source code), then the cached timestamp will change, and libraries depending on lrpar will be recompiled.

### DIFF
--- a/lrpar/src/lib/builder.rs
+++ b/lrpar/src/lib/builder.rs
@@ -274,6 +274,11 @@ pub fn parse(lexemes: &[Lexeme<{storaget}>])
         let mut cache = String::new();
         cache.push_str(" \n/* CACHE INFORMATION\n");
 
+        // Record the time that this version of lrpar was built. If the source code changes and
+        // rustc forces a recompile, this will change this value, causing anything which depends on
+        // this build of lrpar to be recompiled too.
+        cache.push_str(&format!("   Build time: {:?}", env!("VERGEN_BUILD_TIMESTAMP")));
+
         // Record the recoverer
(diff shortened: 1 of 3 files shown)


ltratt:
[It would be nice if cargo gave us this information directly rather than using an external crate. As far as I know it doesn't do that.]

lrpar/src/lib/builder.rs
```diff
@@ -274,6 +274,11 @@ pub fn parse(lexemes: &[Lexeme<{storaget}>])
         let mut cache = String::new();
         cache.push_str(" \n/* CACHE INFORMATION\n");
 
+        // Record the time that this version of lrpar was built. If the source code changes and
```
ptersilie: Record the time _when_ this version of lrpar was built.

ltratt: I think either is OK, although I do marginally prefer "that". Thoughts?

ptersilie: I thought when using "that" it should be: Record the time that this version of lrpar was built __at__.

It just sounded weird to me. But you're the Englishman. ;-)

ltratt: Ha, OK, let's stick with what we have then, since it's a bit shorter.

ptersilie: Fair.

PR #15, opened 2018-09-21
Control rustfmt further

ltratt:
The basic aim of this PR is to ttop rustfmt formatting some things that it can't do a good job of, allowing us to reduce the diff between `cargo fmt --all` and the repo. In fact, we reduce that diff by 15x with this PR (from ~1500 lines to ~100).

The basic problem is that some of the tests in this repository are formatted to make reading them (and spotting problems) much easier (generally by following vertical information). rustfmt doesn't (and can't) know about it. Although this commit makes all of grmtools nightly only (at least until [tool_attributes stabilises](https://github.com/rust-lang/rust/issues/44690), it gets us much closer to "run rustfmt all the time."

https://github.com/softdevteam/grmtools/commit/686059e6f7852aef7fc379a2becbe60308a65615 is the important commit: it uses `#[rustfmt::skip]` to turn rustfmt off on functions where it can't do a job. https://github.com/softdevteam/grmtools/commit/695a65b7dd0e0884d7034ee4c7cbe3d3a418a29e then simply applies many of the remaining diffs that rustfmt produces to our repository.

### DIFF
--- a/cfgrammar/src/lib/mod.rs
+++ b/cfgrammar/src/lib/mod.rs
@@ -77,10 +77,13 @@
 
 #![feature(tool_attributes)]
 
-#[macro_use] extern crate lazy_static;
+#[macro_use]
+extern crate lazy_static;
 extern crate indexmap;
 extern crate num_traits;
-#[cfg(feature="serde")] #[macro_use] extern crate serde;
+#[cfg(feature = "serde")]
+#[macro_use]
+extern crate serde;
 extern crate vob;
 
 mod idxnewtype;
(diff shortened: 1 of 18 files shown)


ptersilie:
Looks good. I guess there isn't a way to tell rustfmt to just skip all functions starting with `test_`?

ltratt:
We could say "skip everything in the test module" (I think), but that would catch too many things we don't want, so it's just a case of selecting those `test_*` functions where we can do a better job by hand than rustfmt can manage automatically.

ptersilie:
Fair!

PR #17, opened 2018-09-26
Use packed vectors to store state tables

ptersilie:
This PR replaces the state table HashMaps (`goto` and `actions`) with packed vectors. This sacrifices memory usage for performance, as it makes table lookups O(1).

Previously, states were stored as `(stIdx, ridx) => stidx`, where `stidx` is a state id, and ridx is a rule id. Now, lookups are done by calculating an index into the vector from `stidx` and `ridx` which reveals the state, e.g. `goto[stidx * nt_len + ridx]`, where `nt_len` is the amount of nonterminals in the grammar.

We then use https://github.com/gabi-250/packed-vec to pack the vectors to reduce memory usage.

### DIFF
--- a/lrtable/src/lib/statetable.rs
+++ b/lrtable/src/lib/statetable.rs
@@ -35,7 +35,8 @@ use std::{
     error::Error,
     fmt::{self, Debug},
     hash::{Hash},
-    marker::PhantomData
+    marker::PhantomData,
+    convert::TryFrom
 };
 
 use cfgrammar::{
@@ -119,7 +120,8 @@ const ERROR: u8 = 0;
 
 impl<StorageT: 'static + Hash + PrimInt + Unsigned> StateTable<StorageT>
 where
-    usize: AsPrimitive<StorageT>
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
 {
     pub fn new(
         grm: &YaccGrammar<StorageT>,
@@ -134,7 +136,7 @@ where
         let maxa = usize::from(grm.tokens_len()) * usize::from(sg.all_states_len());
         let maxg = usize::from(grm.rules_len()) * usize::from(sg.all_states_len());
         // We only have 30 bits to store state IDs
-        assert!(usize::from(sg.all_states_len()) < (u32::max_value() - 4) as usize);
+        assert!(u32::try_from(sg.all_states_len()) < Ok(u32::max_value() - 4));
         let mut actions: Vec<u32> = Vec::with_capacity(maxa);
         actions.resize(maxa, 0);
         let mut gotos: Vec<StIdx> = Vec::with_capacity(maxg);
@@ -314,7 +316,7 @@ where
 
         match action {
             SHIFT => Action::Shift(StIdx::from(val)),
-            REDUCE => Action::Reduce(PIdx(StorageT::from(val).unwrap())),
+            REDUCE => Action::Reduce(PIdx(val.as_())),
             ACCEPT => Action::Accept,
             ERROR => Action::Error,
             _ => unreachable!()
@@ -462,7 +464,8 @@ fn resolve_shift_reduce<StorageT: 'static + Hash + PrimInt + Unsigned>(
     stidx: StIdx
 ) -> u64
 where
-    usize: AsPrimitive<StorageT>
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
 {
     let mut shift_reduce = 0;
     let tidx_prec = grm.token_precedence(tidx);
--- a/lrtable/src/lib/statetable.rs
+++ b/lrtable/src/lib/statetable.rs
@@ -87,7 +87,7 @@ pub struct StateTable<StorageT> {
     // is represented as a hashtable {0: shift 1, 2: shift 0, 3: reduce 4}.
     actions: HashMap<u32, Action<StorageT>, BuildHasherDefault<FnvHasher>>,
     state_actions: Vob,
-    gotos: HashMap<u32, StIdx, BuildHasherDefault<FnvHasher>>,
+    gotos: Vec<StIdx>,
     core_reduces: Vob,
     state_shifts: Vob,
     reduce_states: Vob,
@@ -127,7 +127,11 @@ where
                 .unwrap(),
             false
         );
-        let mut gotos = HashMap::with_hasher(BuildHasherDefault::<FnvHasher>::default());
+        let max = usize::from(grm.rules_len()) * usize::from(sg.all_states_len());
+        assert!(usize::from(sg.all_states_len()) < usize::from(StIdx::max_value()));
+        let mut gotos: Vec<StIdx> = Vec::with_capacity(max);
+        gotos.resize(max, StIdx::max_value());
+
         let mut reduce_reduce = 0; // How many automatically resolved reduce/reduces were made?
         let mut shift_reduce = 0; // How many automatically resolved shift/reduces were made?
         let mut final_state = None;
@@ -206,9 +210,9 @@ where
                 match sym {
                     Symbol::Rule(s_ridx) => {
                         // Populate gotos
-                        let off = (u32::from(stidx) * u32::from(nt_len)) + u32::from(s_ridx);
-                        debug_assert!(gotos.get(&off).is_none());
-                        gotos.insert(off, *ref_stidx);
+                        let off = ((u32::from(stidx) * u32::from(nt_len)) + u32::from(s_ridx)) as usize;
+                        debug_assert!(gotos[off] == StIdx::max_value());
+                        gotos[off] = *ref_stidx;
                     }
                     Symbol::Token(s_tidx) => {
                         // Populate shifts
@@ -364,8 +368,13 @@ where
 
     /// Return the goto state for `stidx` and `ridx`, or `None` if there isn't any.
     pub fn goto(&self, stidx: StIdx, ridx: RIdx<StorageT>) -> Option<StIdx> {
-        let off = (u32::from(stidx) * self.rules_len) + u32::from(ridx);
-        self.gotos.get(&off).and_then(|x| Some(*x))
+        let off = ((u32::from(stidx) * self.rules_len) + u32::from(ridx)) as usize;
+        if self.gotos[off] == StIdx::max_value() {
+            None
+        }
+        else {
+            Some(self.gotos[off])
+        }
     }
 }
 
@@ -551,7 +560,7 @@ mod test {
         assert_eq!(st.core_reduces(s4).collect::<HashSet<_>>(), *s4_core_reduces);
 
         // Gotos
-        assert_eq!(st.gotos.len(), 8);
+        assert_eq!(st.gotos.len(), 9*4);
         assert_eq!(st.goto(s0, grm.rule_idx("Expr").unwrap()).unwrap(), s1);
         assert_eq!(st.goto(s0, grm.rule_idx("Term").unwrap()).unwrap(), s2);
         assert_eq!(st.goto(s0, grm.rule_idx("Factor").unwrap()).unwrap(), s3);
--- a/lrtable/src/lib/statetable.rs
+++ b/lrtable/src/lib/statetable.rs
@@ -31,10 +31,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 use std::{
-    collections::hash_map::{Entry, HashMap, OccupiedEntry},
+    collections::hash_map::{HashMap},
     error::Error,
     fmt::{self, Debug},
-    hash::{BuildHasherDefault, Hash},
+    hash::{Hash},
     marker::PhantomData
 };
 
@@ -42,7 +42,6 @@ use cfgrammar::{
     yacc::{AssocKind, YaccGrammar},
     PIdx, RIdx, Symbol, TIdx
 };
-use fnv::FnvHasher;
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
 use vob::{IterSetBits, Vob};
 
@@ -85,7 +84,7 @@ pub struct StateTable<StorageT> {
     //   0  shift 1
     //   1  shift 0  reduce B
     // is represented as a hashtable {0: shift 1, 2: shift 0, 3: reduce 4}.
-    actions: HashMap<u32, Action<StorageT>, BuildHasherDefault<FnvHasher>>,
+    actions: Vec<u32>,
     state_actions: Vob,
     gotos: Vec<StIdx>,
     core_reduces: Vob,
@@ -109,8 +108,14 @@ pub enum Action<StorageT> {
     /// Reduce production X in the grammar.
     Reduce(PIdx<StorageT>),
     /// Accept this input.
-    Accept
+    Accept,
+    Error
 }
+
+const SHIFT: u8 = 1;
+const REDUCE: u8 = 2;
+const ACCEPT: u8 = 3;
+const ERROR: u8 = 0;
 
 impl<StorageT: 'static + Hash + PrimInt + Unsigned> StateTable<StorageT>
 where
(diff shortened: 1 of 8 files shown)


lrtable/src/lib/statetable.rs
```diff
@@ -127,7 +127,11 @@ where
                 .unwrap(),
             false
         );
-        let mut gotos = HashMap::with_hasher(BuildHasherDefault::<FnvHasher>::default());
+        let max = usize::from(grm.rules_len()) * usize::from(sg.all_states_len());
+        let mut gotos: Vec<StIdx> = Vec::with_capacity(max);
+        for _ in 0..max {
```
ltratt: I think you can replace the loop with `gotos.resize(max, StIdx::max_value())`.

lrtable/src/lib/statetable.rs
```diff
@@ -127,7 +127,11 @@ where
                 .unwrap(),
             false
         );
-        let mut gotos = HashMap::with_hasher(BuildHasherDefault::<FnvHasher>::default());
+        let max = usize::from(grm.rules_len()) * usize::from(sg.all_states_len());
```
ltratt: I think we need an assert along the lines of `assert!(max < StIdx::max_value())` just in case someone manages to create a grammar with `StIdx::max_value()` items.

ltratt:
Have you been able to test the performance of this on e.g. parsing a big Java file? My guess is that it doesn't make much difference either way, but it would be interesting to know!

lrtable/src/lib/statetable.rs
ptersilie: Ah, yes, good idea! That was my concern that someone creates a grammar with exactly `u16::max_value()` items in which case nothing would fail and it just silently would create weird behavior.

lrtable/src/lib/statetable.rs
ptersilie: I'll try.

ptersilie:
I tested it with a >300kb Java file, but I can't see any significant speed gain. So maybe I should remove the claim this is a performance improvement from the PR description. Which begs the question what advantage this method has over the old HashMap?

ptersilie:
No, I didn't see a slowdown either. From my brief tests, it looks like performance is exactly the same between the two versions.

I'll have a look into these compressing techniques.

lrtable/src/lib/statetable.rs
ptersilie: Okay, changed this to `max < usize::from(StIdx::max_value())`, but for some reason the assertion fails with the Java7 grammar.

ptersilie: Seems like the java7 grammar has 1148 states and 241 rules (1148*241=276668), which means we are way over the 65k limit of a `u16`.

ptersilie: Ah! The assertion is bollocks, obviously I only need to check `assert!(sg.all_states_len() < StIdx::max_value())`!

ptersilie:
Of course, I had already changed that, but didn't push it because I needed to fix the failing assertion first.

lrtable/src/lib/statetable.rs
ltratt: Aha, yes, I agree. [Sorry, I didn't engage my thinking brain in my first comment!]

ptersilie:
Added commits to implement the actions table as a vector. The commits are split up as follows:
- 940e2f5 touches only lrtable to convert the actions HashMap into a `Vec<u32>`.
- 1f301aa touches lrpar to adjust that we now have a new `Action` type `Error`. However, function `action` in `lrtable` still returns an `Option` even though this is not needed anymore. However, removing the `Option` requires adjusting a lot more of the other files. So I will do this once the lrtable changes are approved.
- d4ff391 fixes some lrtable tests (the table lenghts have changed now that we store actions in a different format), and addresses some rust warnings.

lrtable/src/lib/statetable.rs
```diff
@@ -85,7 +85,9 @@ pub struct StateTable<StorageT> {
     //   0  shift 1
     //   1  shift 0  reduce B
     // is represented as a hashtable {0: shift 1, 2: shift 0, 3: reduce 4}.
-    actions: HashMap<u32, Action<StorageT>, BuildHasherDefault<FnvHasher>>,
+    // Present as bits
+    // Action (4), Argument (max(pdix, stidx))
```
ltratt: This comment doesn't seem to match the `encode` function?

lrtable/src/lib/statetable.rs
```diff
@@ -306,10 +305,45 @@ where
...
+        }
+        else {
+            return Action::Error
+        }
+    }
+
+    fn encode(action: Action<StorageT>) -> u32 {
+        let mut enc:u32 = 0;
+        match action {
+            Action::Shift(stidx) => { enc = (1 << 8) | (u32::from(stidx)); },
```
ltratt: This looks dodgy: surely one needs to do something like `u32::from(stidx) << 8`, otherwise the tag (`1<<8`) will be or'd with `stidx`?

lrtable/src/lib/statetable.rs
```diff
@@ -306,10 +305,45 @@ where
         })
     }
 
+    fn decode(bits: u32) -> Action<StorageT> { // XXX howto find StorageT? encode in int as well?
+        //           Action   State/Prod
+        let mask = 0b00000000_11111111;
```
ltratt: An 8 bit mask for 4 values is quite wasteful: we can compress this down to 2 bits. I would expect something like this:

```
let kind = bits & 0b11;
let val = (bits & ~0b11) >> 2;
```

should work OK?

lrtable/src/lib/statetable.rs
```diff
@@ -552,7 +551,7 @@ mod test {
         let st = StateTable::new(&grm, &sg).unwrap();
 
         // Actions
-        assert_eq!(st.actions.len(), 15);
+        assert_eq!(st.actions.len(), 9*4); // XXX why is it 4 tokens and not 3 (id, *, -)?
```
ltratt: The EOF token is the 4th token.

lrtable/src/lib/statetable.rs
```diff
@@ -617,7 +616,9 @@ mod test {
           ").unwrap();
         let sg = pager_stategraph(&grm);
         let st = StateTable::new(&grm, &sg).unwrap();
-        assert_eq!(st.actions.len(), 8);
+
+        let length = usize::from(grm.tokens_len()) * usize::from(sg.all_states_len());
```
ltratt: Conventionally this is `len` (or `l`).

lrtable/src/lib/statetable.rs
ptersilie: Oh, yes, that is a left over and should be deleted.

lrtable/src/lib/statetable.rs
ptersilie: I'm not sure what you mean, `1<<8` is the first part of the `u32` encoding the actions, I then `or` this with the second part which is `stidx`.

ltratt: Ah, yes, I was wrong. Though it means you can only have an 8 bit `stidx`, so I think this code won't work with e.g. the java7 grammar.

lrtable/src/lib/statetable.rs
ptersilie: Indeed, my plan was to start simple at and get it to work first and then get some  comments from you before optimising this. I was just wondering what type we should use. Do you think `u16` is enough or still too much?

lrtable/src/lib/statetable.rs
ptersilie: Ah, of course!

lrtable/src/lib/statetable.rs
ptersilie: Ok, guess I was still in Python world, where `len` is reserved.

lrtable/src/lib/statetable.rs
ltratt: It's not trivial, because we need to represent `StorageT::max_value() << 2`: if someone uses `u16` for `StorageT`, we'll be running a risk of crashing. I would probably use a `u64` (with an appropriate `assert`... just in case!), because we'll later use an integer packing vec, at which point it doesn't really matter what type we're storing.

lrtable/src/lib/statetable.rs
ptersilie: So shall we use a `u64` and use the first 2 bits for the `Action` and the remaining 62 bits for the states? Or use a `u64` and use the first 32 bits for the action and the remaining 32 for the states, to keep the values low. I guess since we say `StorageT` is maximum `u32` (and internally even `u16`) 32 bits for the states should be enough?

lrtable/src/lib/statetable.rs
ptersilie: Isn't `let val = (bits & ~0b11) >> 2;` the same as `let val = bits >> 2;`? Also, I think your version assumes that the action type is encoded at the end of the integer, however I designed it to be at the beginning. Maybe that was a bad choice? Should I swap them around?

ltratt: Oops, you're quite right. Yes `bits >> 2` is fine. Putting the type bits starting at position 0 is definitely the right way to go, since that leaves a variable number of bits for the "data".

ptersilie: I understand now! That makes indeed more sense to have the unknown sized number at the beginning and the fixed size at the end!

ptersilie:
I updated the encoder/decoder. I still need to update the actions vector to 64bit, so we can store state values up to 32bit, right? Even tho StIdx is internally only 16bit.

ltratt:
Yes. Even though `StIdx` is a `u16`, it might change; and reductions can be done on grammars with unbounded numbers of reductions.

lrtable/src/lib/statetable.rs
```diff
@@ -303,10 +303,8 @@ where
     }
 
     fn decode(bits: u32) -> Action<StorageT> {
-        //           Action   State/Prod
-        let mask = 0b00000000_11111111;
-        let action = bits >> 8;
-        let reference = bits & mask;
+        let action = bits & 0b11;
+        let reference = bits >> 2;
```
ltratt: Let's call this `val` or `v` or something. `reference` is very long and misleading.

lrtable/src/lib/statetable.rs
```diff
@@ -303,10 +303,8 @@ where
...
 
         if action == 1 {
```
ltratt: I suggest you make an `enum` for these. It'll be something like:

```
enum EncodedActions { ACCEPT = 0; REDUCE = 1; ... }

...

match action {
  ACCEPT => Action::Accept,
  REDUCE => Action::Reduce(val)
  ...
}
```

then you can statically prove that the `else` branch can't happen (so there's no need to dynamically panic).

ptersilie: I was about to ask how I can make this more elegant! ;-)

ptersilie:
Updated to use `u64` to store the encoded actions data. However, I'm not sure about the various uses of `as u64`. Should I implement a `from_u64` trait instead?

lrtable/src/lib/statetable.rs
```diff
@@ -112,6 +112,11 @@ pub enum Action<StorageT> {
     Error
 }
 
+const SHIFT: u8 = 1;
+const REDUCE: u8 = 2;
+const ACCEPT:u8 = 3;
```
ltratt: Space after the `:`.

lrtable/src/lib/statetable.rs
```diff
@@ -303,33 +308,25 @@ where
     }
 
     fn decode(bits: u64) -> Action<StorageT> {
-        let action = bits & 0b11;
-        let reference = (bits >> 2) as u32;
+        let action = (bits & 0b11) as u8;
+        let val = (bits >> 2) as u32;
```
ltratt: Does `let val: u64 = bits >> 2` not work? Similarly for the line above.

lrtable/src/lib/statetable.rs
ptersilie: Whoops!

lrtable/src/lib/statetable.rs
ptersilie: No, because `StIdx` can't convert from `u64` unless I add a `from_u64` to it.

ltratt: OK then `let val: u32 = ...`.

ltratt:
OK, if we're going to only store `u32` (well, really `u30` as we lose bits!), there's not much point using the `u64` type. On reflection I'm OK using `u32` instead of `u64`, *but* we must put checks in to ensure that if the user gives us a value that we can't represent in 30 bits we explicitly fail rather than silently do weird things. Sound like a plan?

ptersilie:
Okay, addressed your comments. The question now is, if I should do the vector packing in this PR or a new one? If the latter, shall I rebase this PR?

ltratt:
I suggest using https://github.com/gabi-250/packed-vec in this PR if possible; and then we can think about whether that's exactly the right library going forward afterwards. [But we should think about the `u64`/`u32` thing first.]

ptersilie:
Oh, just saw your last comment. Sure, sounds like a plan. I'll change back to u32 and add a check to it as well.

ptersilie:
I just noticed we kinda already have a check for this:https://github.com/softdevteam/grmtools/pull/17/files#diff-3c84593bd2e014b4056436615cd05cdfR136

It's checking for the size of `StIdx`, which is `u32` to the outside world.

ltratt:
Yes, but we can only store a `u30`, so we need an additional check.

lrtable/src/lib/statetable.rs
```diff
@@ -302,10 +307,35 @@ where
         })
     }
 
+    fn decode(bits: u32) -> Action<StorageT> {
+        let action = (bits & 0b11) as u8;
+        let val: u32 = bits >> 2;
+
+        match action {
+            SHIFT => Action::Shift(StIdx::from(val)),
+            REDUCE => Action::Reduce(PIdx(StorageT::from(val).unwrap())),
```
ltratt: I think this can be `PIdx(val.as_())` because we know, by definition, that `val` will fit in `StorageT`. [Using the `as_` should be quite a bit faster than `from(...).unwrap()`.] We need to document *why* the `as_` call is safe though. Something like "Since val was originally stored in a StorageT, we know by definition that it can fit into a StorageT now and hence the call to as_ is safe."

ptersilie: Doesn't seem to work: `the trait num_traits::AsPrimitive<StorageT> is not implemented for u32`

lrtable/src/lib/statetable.rs
```diff
@@ -133,7 +133,8 @@ where
         );
         let maxa = usize::from(grm.tokens_len()) * usize::from(sg.all_states_len());
         let maxg = usize::from(grm.rules_len()) * usize::from(sg.all_states_len());
-        assert!(usize::from(sg.all_states_len()) < usize::from(StIdx::max_value()));
+        // We only have 30 bits to store state IDs
+        assert!(usize::from(sg.all_states_len()) < (u32::max_value() - 4) as usize);
```
ltratt: Can we do `assert!(usize::from(sg.all_states_len()) < usize::from(u32::max_value() - 4))`? I'm keen to avoid `as` whenever possible.

ptersilie: Erch, `From<u32> not implemented for usize`.

ltratt: Ah. `assert!(u32::try_from(sg.all_states_len()).unwrap() < u32::max_value() - 4);`?

ptersilie: So that would return an error if `sg.all_states_len()` is bigger than `u32`, which then also makes the `assert` fail?

ltratt: Yeah, which is fine.

ltratt:
This all looks good! Do you have any benchmarking data?

ptersilie:
I'm gonna run some benchmarks now. What are we going to do about the goto tables tho?

ltratt:
I suggest a second PR for that. If we keep this PR open too long we're going to be in merge hell (we already have quite a few merge conflicts between this branch and mine).

ptersilie:
That's alright. Maybe we should have even merged this without the packing first and make a different PR for the packing.

ptersilie:
I just did a quick performance test, using `nimbleparse` to parse all files within `javastdlib/awt/` and it looks like this PR increases performance by roughly 5%, which is actually more than I expected.

ltratt:
Honestly, at this point, if it doesn't make things slower, I'm happy. Sounds like we have a winner!

ptersilie:
Alright, shall I squash?

ptersilie:
Squashed! You might want to check the commit messages which I expanded a bit.

ltratt:
"This replaces the HashMap representation for statetable lookups with a Vector, encoding actions as a `u32` integer, where the first 30 bits describe the value of the action, and the last
2 bits the action type. For example, the action Shift(3) would be encoded as `0b1101`, and Reduce(4) would be `0b10010`." Conventionally the "first bits" are those on the RHS, so it should be "the first 2 bits are the action type and the remaining 30 bits the value" or similar.

ptersilie:
Ah yes, of course, you are right! Should I also expand on the last commit a bit?

ltratt:
Good question. I'd probably add a bit about benchmarking. Something like "This change makes parsing the Java standard library about 5% faster."

One quick question: how stable are the parsing performance numbers? In other words, if you run things a few times, is 5% "in the noise"?

ptersilie:
Updated the commits.

I ran it a couple of times (maybe 10) and it looked stable.

ltratt:
OK, looks like we're ready to merge!

PR #18, opened 2018-09-27
Break lrpar's dependency on lrlex

ltratt:
Previously, lrpar had a hard dependency on lrlex, in the sense that lrpar could only operate on the `Lexeme` type defined in lrlex. This largely precluded people using a non-lrlex lexer, which is an annoying limitation.

This PR inverts that dependency: lrlex now depends on lrpar. In essence, lrpar now defines a `Lexeme` type of its own and a `Lexer` trait which arbitrary external lexers can then conform to.

That sounds simple, but there are a couple of warts imposed upon us. First, we can no longer have an `lrpar` binary, because that would need to call lrlex (which imports lrpar), creating a circular dependency which Rust can't handle. Thus https://github.com/softdevteam/grmtools/commit/d87f967820b680fdd9a721ab4f3431c441ec2f07 breaks out the old `lrpar` binary (i.e. the command-line program which, given lex and grammar files, and an input, parses the input and prints out a parse tree) into a new binary-only project called `nimbleparse`. Second, we have to implement a mini-version of lrlex in https://github.com/softdevteam/grmtools/commit/5b372608750a0dde93e680b2ae76ceb938546a52 so that lrpar's tests don't have to import lrlex (weirdly, you can get things to compile if you use `dev-dependencies`, but then end up with two copies of lrpar... hilarity then ensues as you get messages along the lines of "expected type T, but got type T").

A future PR will improve the `Lexer` trait quite a bit, but since this PR touches quite a bit of the code base, I think it's worth looking at sooner. I'm fairly confident that subsequent PRs will have much more localised changes.


lrpar/src/lib/parser.rs
```diff
@@ -573,13 +572,56 @@ pub(crate) mod test {
...
+    // small_lexer and small_lex are our highly simplified version of lrlex (allowing us to avoid
+    // having to have lrlex as a dependency of lrpar). The format is the same as lrlex *except*:
+    //   * The initial "%%" isn't needed, and only "'" is valid as a rule name delimiter.
+    //   * "Unnamed" rules aren't allowed (e.g. you can't have a rule which discards whitespaces).
+    fn small_lexer(lexs: &str, ids_map: HashMap<String, u16>) -> Vec<(u16, Regex)> {
+        let mut rules = Vec::new();
+        for l in lexs.split("\n").map(|x| x.trim()).filter(|x| !x.is_empty()) {
+            println!("{}", l);
+            assert!(l.rfind("'") == Some(l.len() - 1));
+            let name_beg = l[..l.len() - 1].rfind("'").unwrap();
```
ptersilie: what is a `beg`?

ltratt: "beginning" (maybe not the best name ever...)

ptersilie: Maybe name it `name_start`?

ltratt: Will do.

ptersilie:
Okay, PR looks good. I just have a few general questions:

1) I assume `nimbleparse` just replaces the command line program `lrpar`? I assume it's not supposed to be used in production and that's why we don't call it `parse`?

2) How are users expected to use these libraries. I assume that they always want to import them as libraries into other projects, and never really want to use them as stand-alone binaries, right? Also, from what I understood due to lrpar and lrlex depending on each other, there will always be a one directional dependency and we never can use both of them on their own. Which is why we choose to let lrpar be stand-alone and lrlex have the dependency. Which means, however, that lrlex can never be used as a library on its own?

ltratt:
1) Yes, `nimbleparse` replaces `lrpar`'s command-line program. It's not exactly that it's not production ready, but I can't imagine why you'd want to use it in production (a program that prints out a parse tree isn't useful very often).

2) You can use lrpar without lrlex (that's what this PR is all about). lrpar requires lrtable and cfgrammar. lrtable requires cfgrammar. cfgrammar can be used on its own. Most people, though, will probably use lrpar and lrlex (and most of the remainder will use lrpar and a lexer of their choosing). To an extent this stuff is documented, but we need a "formal" document (separate from the API docs) that tells people how to do things. That's for a later date. If you want to see how to use things in the interim, `calcparse` is the way to go. It's a runnable binary so you can do `cd lrpar/examples/calcparse && cargo run` and see it do simple calculator tasks with integers.

lrpar/src/lib/parser.rs
```diff
@@ -581,9 +581,9 @@ pub(crate) mod test {
         for l in lexs.split("\n").map(|x| x.trim()).filter(|x| !x.is_empty()) {
             println!("{}", l);
             assert!(l.rfind("'") == Some(l.len() - 1));
-            let name_beg = l[..l.len() - 1].rfind("'").unwrap();
-            let name = &l[name_beg + 1 .. l.len() - 1];
-            let re = &l[..name_beg - 1].trim();
+            let name_st = l[..l.len() - 1].rfind("'").unwrap();
```
ptersilie: I know you love short variable names, but would it really hurt so much to rename this to `name_start` so that it's absolutely clear what this variable contains and readers don't have to guess what `st` means?

ltratt: In retrospect, "start" is a bit misleading, because it's the position of the input after the opening `'`. So a truthful variable name would be something like `first_character_of_name_after_quote`. It might be clearer to just give it a meaningless name like `i`?

ltratt:
If `lrpar` and `lrlex` were completely independent, it would force the user to do plumbing to tie them together, which would be frustrating. As you say, I don't think there's a meaningful use case for `lrlex` besides using it with `lrpar`.

lrpar/src/lib/parser.rs
ptersilie: Hah, ye maybe a meaningless name is better in this instance. ;-)

ptersilie:
I agree.

ltratt:
If you're OK with that last commit, I think we're ready for squashing?

ptersilie:
Yes, please squash.

ltratt:
Squashed.

PR #20, opened 2018-10-04
Make the Lexer trait be a pseudo-iterator.

ltratt:
Traditional `lex` semi-lazy in the sense that it only produces tokens when Yacc asks it to (and Yacc can advance the input such that lex implicitly skips some parts of the input). One advantage to this approach is memory usage. If you tokenize a huge file into a Vec, it can take up more memory than one might like.

This commit enables lazy lexing through the Lexer trait: the next() method simply says "give me the next token". It's up to the individual lexer whether it's actually lazy or not, but this at least gives it the option to do so. Currently the lrpar parser isn't itself lazy, but making it lazy later should be an internal change without external impact.

As part of this, it's time to finally rethink the external parsing API that users use. We now have very similar compile-time (`CTParserBuilder`) and run-time (`RTParserBuilder`) interfaces. They will start to diverge a bit over time in terms of some of the options they offer, but for the time being they're very similar.

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -31,17 +31,15 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 use std::{
-    cell::RefCell,
     collections::{HashMap, HashSet},
     hash::Hash,
-    rc::Rc,
     slice::Iter
 };
 
 use num_traits::{PrimInt, Unsigned};
 use regex::{self, Regex, RegexBuilder};
 
-use lrpar::{LexError, Lexer, Lexeme};
+use lrpar::{LexError, Lexeme, Lexer};
 
 pub struct Rule<StorageT> {
     /// If `Some`, the ID that lexemes created against this rule will be given (lrlex gives such
@@ -199,7 +197,7 @@ impl<StorageT: Copy + Eq + Hash + PrimInt + Unsigned> LexerDef<StorageT> {
 pub struct LRLexer<'a, StorageT: 'a> {
     lexerdef: &'a LexerDef<StorageT>,
     s: &'a str,
-    newlines: Rc<RefCell<Vec<usize>>>
+    newlines: Vec<usize>
 }
 
 impl<'a, StorageT: Copy + Eq> LRLexer<'a, StorageT> {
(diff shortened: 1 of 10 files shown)


lrlex/src/lib/lexer.rs
```diff
@@ -213,16 +215,13 @@ impl<'a, StorageT: Copy + Eq> LRLexer<'a, StorageT> {
...
 {
-    /// Return all this lexer's lexemes or a `LexError` if there was a problem when lexing.
-    fn lexemes(&mut self) -> Result<Vec<Lexeme<StorageT>>, LexError> {
-        let mut i = 0; // byte index into s
-        let mut lxs = Vec::new(); // lexemes
-
-        while i < self.s.len() {
+    fn next(&mut self) -> Option<Result<Lexeme<StorageT>, LexError>> {
+        while self.i < self.s.len() {
+            let old_i = self.i;
```
ptersilie: Why do we need `old_i`? It seems to me that within this context, `old_i` is always equal to `self.i`.

lrpar/src/lib/parser.rs
```diff
@@ -643,30 +608,53 @@ pub(crate) mod test {
...
     }
 
     fn check_parse_output(lexs: &str, grms: &str, input: &str, expected: &str) {
         let (grm, pt) = do_parse(RecoveryKind::MF, lexs, grms, input);
         assert_eq!(expected, pt.unwrap().pp(&grm, &input));
     }
 
-    // small_lexer and small_lex are our highly simplified version of lrlex (allowing us to avoid
-    // having to have lrlex as a dependency of lrpar). The format is the same as lrlex *except*:
+    // SmallLexer is ourhighly simplified version of lrlex (allowing us to avoid having to have
```
ptersilie: `our highly`

lrpar/src/lib/parser.rs
```diff
@@ -643,30 +608,53 @@ pub(crate) mod test {
...
     //   * "Unnamed" rules aren't allowed (e.g. you can't have a rule which discards whitespaces).
+    struct SmallLexer<StorageT> {
+        lexemes: Vec<Lexeme<StorageT>>,
+        i: usize
+    }
+
+    impl<StorageT: Hash + PrimInt + Unsigned> Lexer<StorageT> for SmallLexer<StorageT> {
+        fn next(&mut self) -> Option<Result<Lexeme<StorageT>, LexError>> {
+            let old_i = self.i;
+            if old_i < self.lexemes.len() {
```
ptersilie: Why can't we do:
```
if self.i < self.lexemes.len() {
    self.i += 1;
    return Some(Ok(self.lexemes[self.i]));
}
```

lrlex/src/lib/lexer.rs
ltratt: At line 257, we increment `self.i` but still want `old_i`. I could rewrite it not to do this though, if you think that would be clearer. [I'm neutral on this.]

lrpar/src/lib/parser.rs
ltratt: Oops.

lrpar/src/lib/parser.rs
ltratt: Because that would skip tokens. The first time you called `next` you'd get the token at position 1 (not, as you want, position 0).

lrlex/src/lib/lexer.rs
ptersilie: Ah I missed that. Should we maybe just do `let old_i = self.i` just before line 257?

lrpar/src/lib/parser.rs
ptersilie: Ah whoops, missed the `old_i` in the return again. My bad!

lrlex/src/lib/lexer.rs
ltratt: Ah, looking again, we'd also need to do this in lines 246 & 250. So it might be better off as it is.

ptersilie: You are right. In that case it's better to leave it as is.

ptersilie:
Looks good. Please squash.

ltratt:
Squashed.

PR #21, opened 2018-10-04
Tidy up the StIdx type.

ltratt:
The `StIdx` type was, previously, something of a mess. We said it stored a `u32`, even though it stored a `u16`, and we relied upon that `u32`ness in a few places. This made it quite a bit more annoying to deal with than the `PIdx`/`RIdx' types.

This PR more-or-less brings `StIdx` into line with those other types: the only guarantee we now make is that an `StIdx` can be infallibly converted into a `usize`. This has various knock-on effects which this PR deals with.

[In the long run, I think we should get rid of the size checks inside `from` methods in `StIdx`. Those, however, are for another day.]

### DIFF
--- a/lrpar/src/lib/mf.rs
+++ b/lrpar/src/lib/mf.rs
@@ -801,7 +801,7 @@ where
 mod test {
     use std::{collections::HashMap, fmt::Debug};
 
-    use num_traits::{AsPrimitive, PrimInt, ToPrimitive, Unsigned};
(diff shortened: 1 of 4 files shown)


ptersilie:
This PR causes a slight slowdown of 0.5% when parsing the Java AWT library. However, since this PR allows us to use `PackedVec` for the goto tables as well, this is likely going to improve again.

PR #27, opened 2018-10-07
Skipping the top element in the parsing stack is dangerous.

ltratt:
On the first iteration of panic mode, skipping the top element of the stack is a valid optimisation: however, if the input is advanced, then the stack is restored, and the top element of the stack *might* match. On the big corpus of Java programs, this makes a noticeable (positive) 2% change in the number of files which can be successfully recovered from.


ptersilie:
After merging the previous PR this has a conflict now, which we need to fix first.

ltratt:
Rebased against master.

PR #31, opened 2018-10-17
Use sparsevec for parsing tables

ptersilie:
This PR changes the datastructure for storing the actions and goto tables from PackedVec to SparseVec. This new data structure first compresses the tables using row displacement and then stores the result in a PackedVec.

For the Java parser, this results in a 8% smaller binary, saves 67% of memory for the actions table, and 83% for the gotos table, and gives a speed improvement of 2% when parsing the Java Standard Library.

### DIFF
--- a/cfgrammar/Cargo.toml
+++ b/cfgrammar/Cargo.toml
@@ -14,4 +14,4 @@ indexmap = "1.0"
 num-traits = "0.2"
 regex = "1.0"
 serde = { version="1.0", features=["derive"], optional=true }
-vob = "1.3"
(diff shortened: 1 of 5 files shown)


ltratt:
Can we get something about the speed-up (or not...) of this into the commit message? Otherwise this looks good to go.

ptersilie:
I added some comments about improvements to the PR text.

ltratt:
Looks good to me. Can we add that to the commit message? Then I can merge.

ptersilie:
Done!

PR #32, opened 2018-10-18
Implement actions

ptersilie:
This PR is under construction and should not be merged in its current state. The goal of this PR is to implement grammar actions similar to YACC, for example:

```
%start Expr
%%
Expr: Term 'PLUS' Expr { add($0, $2) }
    | Term { $0 }
    ;
Term: Factor 'MUL' Term { mul ($0, $2) }
    | Factor { $0 }
    ;
Factor: 'LBRACK' Expr 'RBRACK' { $1 }
      | 'INT' { $0 }
      ;
%%
type TYPE = u64;

fn add(arg1: u64, arg2: u64) -> u64 {
    arg1 + arg2
}
fn mul(arg1: u64, arg2: u64) -> u64 {
    arg1 * arg2
}
```

### DIFF
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -104,7 +104,8 @@ pub struct YaccGrammar<StorageT = u32> {
     implicit_rule: Option<RIdx<StorageT>>,
     /// User defined Rust programs which can be called within actions
     actions: Vec<Option<String>>,
-    programs: Option<String>
+    programs: Option<String>,
+    actiontype: Option<String>
 }
 
 // Internally, we assume that a grammar's start rule has a single production. Since we manually
@@ -342,13 +343,19 @@ where
             prod_precs: prod_precs.into_iter().map(|x| x.unwrap()).collect(),
             implicit_rule: implicit_rule.and_then(|x| Some(rule_map[&x])),
             actions,
-            programs: ast.programs
+            programs: ast.programs,
+            actiontype: ast.actiontype
         })
     }
 
     /// Get the programs part of the grammar
     pub fn programs(&self) -> &Option<String> {
         &self.programs
+    }
+
+    /// Get the action return type as defined by the user
+    pub fn actiontype(&self) -> &Option<String> {
+        &self.actiontype
     }
 
     /// How many productions does this grammar have?
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -101,7 +101,10 @@ pub struct YaccGrammar<StorageT = u32> {
     prod_precs: Vec<Option<Precedence>>,
     /// The index of the rule added for implicit tokens, if they were specified; otherwise
     /// `None`.
-    implicit_rule: Option<RIdx<StorageT>>
+    implicit_rule: Option<RIdx<StorageT>>,
+    /// User defined Rust programs which can be called within actions
+    actions: Vec<Option<String>>,
+    programs: Option<String>
 }
 
 // Internally, we assume that a grammar's start rule has a single production. Since we manually
@@ -224,6 +227,7 @@ where
         let mut prods = vec![None; ast.prods.len()];
         let mut prod_precs: Vec<Option<Option<Precedence>>> = vec![None; ast.prods.len()];
         let mut prods_rules = vec![None; ast.prods.len()];
+        let mut actions = vec![None; ast.prods.len()];
         for astrulename in &rule_names {
             let ridx = rule_map[astrulename];
             if astrulename == &start_rule {
@@ -245,6 +249,7 @@ where
                 prods.push(Some(start_prod));
                 prod_precs.push(Some(None));
                 prods_rules.push(Some(ridx));
+                actions.push(None);
                 continue;
             } else if implicit_start_rule
                 .as_ref()
@@ -313,6 +318,10 @@ where
                 prods[prod_idx] = Some(prod);
                 prod_precs[prod_idx] = Some(prec);
                 prods_rules[prod_idx] = Some(ridx);
+                match astprod.action {
+                    Some(ref s) => actions[prod_idx] = Some(String::from(s.clone())),
+                    None => ()
+                };
             }
         }
 
@@ -331,8 +340,15 @@ where
             prods_rules: prods_rules.into_iter().map(|x| x.unwrap()).collect(),
             prods: prods.into_iter().map(|x| x.unwrap()).collect(),
             prod_precs: prod_precs.into_iter().map(|x| x.unwrap()).collect(),
-            implicit_rule: implicit_rule.and_then(|x| Some(rule_map[&x]))
+            implicit_rule: implicit_rule.and_then(|x| Some(rule_map[&x])),
+            actions,
+            programs: ast.programs
         })
+    }
+
+    /// Get the programs part of the grammar
+    pub fn programs(&self) -> &Option<String> {
+        &self.programs
     }
 
     /// How many productions does this grammar have?
@@ -352,6 +368,11 @@ where
     /// Get the sequence of symbols for production `pidx`. Panics if `pidx` doesn't exist.
     pub fn prod(&self, pidx: PIdx<StorageT>) -> &[Symbol<StorageT>] {
         &self.prods[usize::from(pidx)]
+    }
+
+    /// Get the action for production `pidx`. Panics if `pidx` doesn't exist.
+    pub fn action(&self, pidx: PIdx<StorageT>) -> &Option<String> {
+        &self.actions[usize::from(pidx)]
     }
 
     /// How many symbols does production `pidx` have? Panics if `pidx` doesn't exist.
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -125,18 +125,7 @@ impl YaccParser {
         // every byte within the string is also a valid character).
         let mut i = self.parse_declarations(0)?;
         i = self.parse_rules(i)?;
-        // We don't currently support the programs part of a specification. One day we might...
-        match self.lookahead_is("%%", i) {
-            Some(j) => {
-                if self.parse_ws(j)? == self.src.len() {
-                    Ok(i)
-                } else {
-                    i = self.parse_programs(i)?;
-                    Ok(i)
-                }
-            }
-            None => Ok(i)
-        }
+        self.parse_programs(i)
     }
 
     pub(crate) fn ast(self) -> GrammarAST {
@@ -233,10 +222,7 @@ impl YaccParser {
 
     fn parse_rules(&mut self, mut i: usize) -> YaccResult<usize> {
         // self.parse_declarations should have left the input at '%%'
-        match self.lookahead_is("%%", i) {
-            Some(j) => i = j,
-            None => panic!("Internal error.")
-        }
+        i = self.lookahead_is("%%", i).unwrap();
         i = self.parse_ws(i)?;
         while i < self.src.len() {
             if self.lookahead_is("%%", i).is_some() {
@@ -342,12 +328,13 @@ impl YaccParser {
             let ch = self.src[j..].chars().next().unwrap();
             c += match ch {
                 '{' => 1,
+                '}' if c == 1 => {
+                    c = 0;
+                    break;
+                }
                 '}' => -1,
                 _ => 0
             };
-            if c == 0 {
-                break;
-            }
             j += ch.len_utf8();
         }
         if c > 0 {
@@ -358,13 +345,21 @@ impl YaccParser {
         }
     }
 
-    fn parse_programs(&mut self, i: usize) -> YaccResult<usize> {
-        let mut j = i + 2; // Skip %%
-        j = self.parse_ws(j)?;
-        let prog = self.src[j..].to_string();
-        j = j + prog.len();
-        self.ast.add_programs(prog);
-        Ok(j)
+    fn parse_programs(&mut self, mut i: usize) -> YaccResult<usize> {
+        if let Some(j) = self.lookahead_is("%%", i) {
+            i = self.parse_ws(j)?;
+            if i == self.src.len() {
+                Ok(i)
+            } else {
+                let prog = self.src[i..].to_string();
+                i = i + prog.len();
+                self.ast.add_programs(prog);
+                Ok(i)
+            }
+        }
+        else {
+            Ok(i)
+        }
     }
 
     fn parse_ws(&mut self, mut i: usize) -> YaccResult<usize> {
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -151,6 +151,18 @@ impl YaccParser {
                 }
                 continue;
             }
+            if let Some(j) = self.lookahead_is("%type", i) {
+                i = self.parse_ws(j)?;
+                while i < self.src.len() {
+                    if self.lookahead_is("%", i).is_some() {
+                        break;
+                    }
+                    let (j, n) = self.parse_name(i)?;
+                    self.ast.actiontype = Some(n);
+                    i = self.parse_ws(j)?;
+                }
+                continue;
+            }
             if let Some(j) = self.lookahead_is("%start", i) {
                 if self.ast.start.is_some() {
                     return Err(self.mk_error(YaccParserErrorKind::DuplicateStartDeclaration, i));
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -51,6 +51,7 @@ pub enum YaccParserErrorKind {
     IllegalString,
     IncompleteRule,
     IncompleteComment,
+    IncompleteAction,
     MissingColon,
     PrematureEnd,
     ProgramsNotSupported,
@@ -78,6 +79,7 @@ impl fmt::Display for YaccParserError {
             YaccParserErrorKind::IllegalString => "Illegal string",
             YaccParserErrorKind::IncompleteRule => "Incomplete rule",
             YaccParserErrorKind::IncompleteComment => "Incomplete comment",
+            YaccParserErrorKind::IncompleteAction => "Incomplete action",
             YaccParserErrorKind::MissingColon => "Missing colon",
             YaccParserErrorKind::PrematureEnd => "File ends prematurely",
             YaccParserErrorKind::ProgramsNotSupported => "Programs not currently supported",
@@ -129,7 +131,8 @@ impl YaccParser {
                 if self.parse_ws(j)? == self.src.len() {
                     Ok(i)
                 } else {
-                    Err(self.mk_error(YaccParserErrorKind::ProgramsNotSupported, i))
+                    i = self.parse_programs(i)?;
+                    Ok(i)
                 }
             }
             None => Ok(i)
@@ -259,16 +262,18 @@ impl YaccParser {
         }
         let mut syms = Vec::new();
         let mut prec = None;
+        let mut action = None;
         i = self.parse_ws(i)?;
         while i < self.src.len() {
             if let Some(j) = self.lookahead_is("|", i) {
-                self.ast.add_prod(rn.clone(), syms, prec);
+                self.ast.add_prod(rn.clone(), syms, prec, action);
                 syms = Vec::new();
                 prec = None;
+                action = None;
                 i = self.parse_ws(j)?;
                 continue;
             } else if let Some(j) = self.lookahead_is(";", i) {
-                self.ast.add_prod(rn.clone(), syms, prec);
+                self.ast.add_prod(rn.clone(), syms, prec, action);
                 return Ok(j);
             }
 
@@ -286,6 +291,10 @@ impl YaccParser {
                     return Err(self.mk_error(YaccParserErrorKind::PrecNotFollowedByToken, i));
                 }
                 i = k;
+            } else if self.lookahead_is("{", i).is_some() {
+                let (j, a) = self.parse_action(i)?;
+                i = j;
+                action = Some(a);
             } else {
                 let (j, sym) = self.parse_token(i)?;
                 if self.ast.tokens.contains(&sym) {
@@ -324,6 +333,38 @@ impl YaccParser {
             }
             None => Err(self.mk_error(YaccParserErrorKind::IllegalString, i))
         }
+    }
+
+    fn parse_action(&mut self, i: usize) -> YaccResult<(usize, String)> {
+        let mut j = i;
+        let mut c = 0; // Count braces
+        while j < self.src.len() {
+            let ch = self.src[j..].chars().next().unwrap();
+            c += match ch {
+                '{' => 1,
+                '}' => -1,
+                _ => 0
+            };
+            if c == 0 {
+                break;
+            }
+            j += ch.len_utf8();
+        }
+        if c > 0 {
+            Err(self.mk_error(YaccParserErrorKind::IncompleteAction, j))
+        } else {
+            let s = self.src[i + 1..j - 1].trim().to_string();
+            Ok((j + 1, s))
+        }
+    }
+
+    fn parse_programs(&mut self, i: usize) -> YaccResult<usize> {
+        let mut j = i + 2; // Skip %%
+        j = self.parse_ws(j)?;
+        let prog = self.src[j..].to_string();
+        j = j + prog.len();
+        self.ast.add_programs(prog);
+        Ok(j)
     }
 
     fn parse_ws(&mut self, mut i: usize) -> YaccResult<usize> {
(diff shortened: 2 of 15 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -104,6 +104,7 @@ lazy_static! {
     static ref RE_NAME: Regex = { Regex::new(r"^[a-zA-Z_.][a-zA-Z0-9_.]*").unwrap() };
     static ref RE_TOKEN: Regex =
         { Regex::new("^(?:(\".+?\")|('.+?')|([a-zA-Z_][a-zA-Z_0-9]*))").unwrap() };
+    static ref RE_ACTION: Regex = { Regex::new(r"^\{[^\{]*\}").unwrap() };
```
ltratt: We can't use a regex for actions, because regexs can't count brackets. For example this is a valid action block:

```rust
{
   if ... { } else { }
}
```
but the regex will (I think) only match `{ if ... { }`.

I think we're going to have to write a little hand-coded action matcher which counts open/close curlies so that it knows when to stop. It's probably only 5 or 6 LoC.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -326,6 +333,17 @@ impl YaccParser {
...
     }
 
+    fn parse_action(&mut self, i: usize) -> YaccResult<(usize, String)> {
+        match RE_ACTION.find(&self.src[i..]) {
+            Some(m) => {
+                assert!(m.start() == 0 && m.end() > 0);
+                let s = self.src[i + 1..i + m.end() - 1].trim().to_string();
+                Ok((i + m.end(), s))
+            }
+            None => Err(self.mk_error(YaccParserErrorKind::IllegalString, i))
```
ltratt: I think you'll need a new error kind here, something like `IncompleteAction` or `IncompleteAction`.

ltratt:
Two comments on that first commit, one substantial, the other minor.

cfgrammar/src/lib/yacc/parser.rs
ptersilie: Whoops, you are right. For some reason I didn't think about having to match curly braces within the actions.

ptersilie:
Fixed the way we parse actions by counting braces.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -334,13 +335,24 @@ impl YaccParser {
     }
 
     fn parse_action(&mut self, i: usize) -> YaccResult<(usize, String)> {
-        match RE_ACTION.find(&self.src[i..]) {
-            Some(m) => {
-                assert!(m.start() == 0 && m.end() > 0);
-                let s = self.src[i + 1..i + m.end() - 1].trim().to_string();
-                Ok((i + m.end(), s))
+        let mut j = i + 1; // Skip opening brace
```
ltratt: I would probably just make `i` mutable in the function signature (i.e. `fn parse_action(&mut self, mut i: usize)` and then there's no point incrementing `+ 1` here. Just `let mut c = 0` and it should handle everything.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -334,13 +335,24 @@ impl YaccParser {
...
+        let mut c = 1; // Count braces
+        while j < self.src.len() {
+            c += match self.src[j..].chars().next().unwrap() {
+                '{' => 1,
+                '}' => -1,
+                _ => 0
+            };
+            if c <= 0 {
```
ltratt: `c` can never be less than 0, so `if c == 0` is fine.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -334,13 +335,24 @@ impl YaccParser {
...
+                break;
             }
-            None => Err(self.mk_error(YaccParserErrorKind::IllegalString, i))
+            j += 1;
```
ltratt: This isn't safe in unicode: characters can be more than 1 byte long. So you need to know which character you read above and increment the byte count appropriately here.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -334,13 +335,24 @@ impl YaccParser {
...
+        }
+        if c == 0 {
```
ltratt: It might be clearer to write the exceptional case first:

```rust
if c > 0 {
    return Err(...);
}
let s = ...;
Ok(...)
```
?

cfgrammar/src/lib/yacc/parser.rs
ptersilie: You mean get rid of `j`?

ltratt: Yep.

ptersilie: I don't think I can get rid of `j` since I need both `i` and `j` to slice the result string.

ltratt: Aha, good point! Yes, agreed. Ignore what I said here.

ptersilie:
Pushed a commit that implements basic actions. The commit is a bit messy for two reasons:
1. I had to clone two functions `Parser.parse` and `RTParserBuilder.parse` (cloned as `parse2`) in order to forward the actions and the user input to the `lr` function and have the new `parse` functions return the user defined type `TYPE` instead of `Node`. I wonder if we can be smarter about this without having to clone those two parse functions.
2. `CTParserBuilder` now has an `ActionKind`, which if set to `CustomAction` generates the code needed to execute production actions and calls the aforementioned `parse2` to forwards the necessary data. Due to the order `CTParserBuilder` generates the code, this requires multiple places where we need to do `match self.actionkind` and generate different code depending on its value. I wonder if we can make this less chaotic by only having one `match` to generate the two `CTParserBuilder` versions, tho this might mean we'd have to duplicate some of the output strings.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -129,7 +131,8 @@ impl YaccParser {
                 if self.parse_ws(j)? == self.src.len() {
                     Ok(i)
                 } else {
-                    Err(self.mk_error(YaccParserErrorKind::ProgramsNotSupported, i))
+                    i = self.parse_programs(i)?;
```
ltratt: Just `Ok(self.parse_programs(i)?)`?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -326,6 +335,38 @@ impl YaccParser {
...
+        let mut j = i;
+        let mut c = 0; // Count braces
+        while j < self.src.len() {
+            let ch = self.src[j..].chars().next().unwrap();
+            c += match ch {
+                '{' => 1,
+                '}' => -1,
+                _ => 0
+            };
+            if c == 0 {
```
ltratt: Should we check this in the `}` clause above, since neither of the other two clauses can trigger this (I think)?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -326,6 +335,38 @@ impl YaccParser {
...
+        }
+        if c > 0 {
+            Err(self.mk_error(YaccParserErrorKind::IncompleteAction, j))
+        } else {
+            let s = self.src[i + 1..j - 1].trim().to_string();
+            Ok((j + 1, s))
+        }
+    }
+
+    fn parse_programs(&mut self, i: usize) -> YaccResult<usize> {
```
ltratt: I suggest that we move the `match self.lookahead_is("%%")` from the function that calls `parse_programs` into here. Then we can do something similar to:

```rust
    fn parse_rules(&mut self, mut i: usize) -> YaccResult<usize> {
        // self.parse_declarations should have left the input at '%%'
        match self.lookahead_is("%%", i) {
            Some(j) => i = j,
            None => panic!("Internal error.")
        }
```

Although the code is a bit silly and should probably be:

```rust
    fn parse_rules(&mut self, mut i: usize) -> YaccResult<usize> {
        // self.parse_declarations should have left the input at '%%'
        i = self.lookahead_is("%%").unwrap();
```

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -326,6 +335,38 @@ impl YaccParser {
...
+        let mut j = i + 2; // Skip %%
+        j = self.parse_ws(j)?;
```
ltratt: I don't think you need to introduce `j` here: you can keep overwriting `i`.

lrlex/src/lib/lexer.rs
```diff
@@ -280,6 +280,10 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
             l.start() - self.newlines[self.newlines.len() - 1] + 1
         ))
     }
+
+    fn get_input(&self) -> &str {
```
ltratt: Should this just be called `input`?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -60,13 +61,19 @@ const RUST_FILE_EXT: &str = "rs";
 const SGRAPH_FILE_EXT: &str = "sgraph";
 const STABLE_FILE_EXT: &str = "stable";
 
+pub enum ActionKind {
+    GenericParseTree,
```
ltratt: We should document these briefly with `///`. Also, should they be in alphabetical order, or sholud we keep `GenericParseTree` first because it's the default? I can be convinced either way!

lrpar/src/lib/ctbuilder.rs
```diff
@@ -235,35 +261,87 @@ pub fn parse(lexer: &mut Lexer<{storaget}>)
...
+                out_sgraph.to_str().unwrap(),
+                out_stable.to_str().unwrap()));
+
+        match self.actionkind {
+            ActionKind::CustomAction => {
+                // action function references
+                outs.push_str("\n        let mut actions: Vec<Option<&Fn(Vec<TYPE>) -> TYPE>> = Vec::new();\n");
+                for pidx in grm.iter_pidxs() {
+                    let a = grm.action(pidx);
+                    match a {
```
ltratt: This would be simpler, I think, as:

```rust
if a.is_some() {
    ...
} else {
    ...
}
```

lrpar/src/lib/ctbuilder.rs
```diff
@@ -235,35 +261,87 @@ pub fn parse(lexer: &mut Lexer<{storaget}>)
...
                     StorageT::type_name(),
                     usize::from(ridx)
                 ));
             }
         }
 
+        match self.actionkind {
+            ActionKind::CustomAction => {
+                let re: Regex = { Regex::new(r"\$([0-9]+)").unwrap() };
+                match grm.programs() {
```
ltratt: `if let Some(s) == grm.programs()` is easier.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -235,35 +261,87 @@ pub fn parse(lexer: &mut Lexer<{storaget}>)
...
+                        outs.push_str(s);
+                    }
+                    None => ()
+                };
+
+                // Convert actions to functions
+                outs.push_str("\n/* Converted actions */\n\n");
+                for pidx in grm.iter_pidxs() {
+                    let a = grm.action(pidx);
+                    match a {
```
ltratt: Same comment as above.

lrpar/src/lib/parser.rs
```diff
@@ -180,11 +219,30 @@ where
                     pstack.drain(pop_idx..);
                     let prior = *pstack.last().unwrap();
                     pstack.push(self.stable.goto(prior, ridx).unwrap());
+
+                    // Process actions
+                    match actiondata {
+                        Some((actions, ref mut astack, _, _)) => {
+                            let f = actions[usize::from(pidx)];
+                            let args: Vec<K> = astack.drain(pop_idx -1..).collect();
+                            match f {
```
ltratt: `if let Some(f) == f` (although maybe calling both things `f` isn't great?),

lrpar/src/lib/parser.rs
```diff
@@ -180,11 +219,30 @@ where
                     pstack.drain(pop_idx..);
                     let prior = *pstack.last().unwrap();
                     pstack.push(self.stable.goto(prior, ridx).unwrap());
+
+                    // Process actions
+                    match actiondata {
```
ltratt: `if let Some(actions, ref mu astack, _, _) == actiondata`

lrpar/src/lib/parser.rs
```diff
@@ -508,6 +566,29 @@ where
...
     }
+
+    /// Parse input. On success return a parse tree. On failure, return a `LexParseError`: a
+    /// `LexError` means that no parse tree was produced; a `ParseError` may (if its first element
+    /// is `Some(...)`) return a parse tree (with parts filled in by this builder's recoverer).
+    pub fn parse2<K>(
+        &self,
+        lexer: &mut Lexer<StorageT>,
+        actions: Vec<Option<&Fn(Vec<K>) -> K>>,
+        input: String,
```
ltratt: I think we want a `&str` here, so that we don't force users to care about moving the string to us.

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -353,6 +369,15 @@ where
         &self.prods[usize::from(pidx)]
     }
 
+    pub fn actions(&self) -> &Vec<Option<String>> {
```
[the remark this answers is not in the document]
ptersilie: Actually, I don't think this function is used anymore, so we can just kill it.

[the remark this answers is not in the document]
ltratt: Works for me!

cfgrammar/src/lib/yacc/parser.rs
ptersilie: Can't the `_` clause trigger this?

cfgrammar/src/lib/yacc/parser.rs
ptersilie: Don't I have to make `i` mutable then?

cfgrammar/src/lib/yacc/parser.rs
ltratt: `parse_action` is only called if the input starts with a `{`:

```rust
            } else if self.lookahead_is("{", i).is_some() {
                let (j, a) = self.parse_action(i)?;
```

so, by definition, the first time this loop executes, c will be set to 1, and thus the `_` clause can't cause can't trigger this.

cfgrammar/src/lib/yacc/parser.rs
ltratt: Yes, but that's transparent to the caller, so it's fine to do so.

cfgrammar/src/lib/yacc/parser.rs
ptersilie: But then we get a "TypeError" instead of "Internal error" if `self.lookahead_is` returns `None`. Is that ok?

ptersilie: Actually, we get an unwrap error.

cfgrammar/src/lib/yacc/parser.rs
ptersilie: You are right. Now I'm just trying to decide if this is more readable:
```
c += match ch {
    '{' => 1,
    '}' => {
        if c == 1 { break };
        -1
    }
    _ => 0
};
```

cfgrammar/src/lib/yacc/parser.rs
ptersilie: I simplified `parse_rules` and similarly `parse_programs` with the above. Though I can't move the `match self.lookahead_is("%%")` from the function that calls `parse_programs`, since programs are optional and I only want to parse them if there is `%%`. Though, I can do that check within `parse_programs` and make it return `i` if there is no `%%`.

lrpar/src/lib/ctbuilder.rs
ptersilie: I'm happy with alphabetical.

cfgrammar/src/lib/yacc/parser.rs
ltratt: I think you can do:

```rust
c += match ch {
    '{' => 1,
    '}' if c == 1 => { break; }
    '}' => -1
    _ => 0
};
```

cfgrammar/src/lib/yacc/parser.rs
ltratt: OK with me.

ptersilie:
Addressed your comments. If you are happy with them I will squash the last two commits.

ptersilie:
> It might be good to put a small example of a grammar and its output with this patch, just so that one can visualise what's going on.

Do you mean within the commit message?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -358,13 +344,22 @@ impl YaccParser {
...
-        let prog = self.src[j..].to_string();
-        j = j + prog.len();
-        self.ast.add_programs(prog);
-        Ok(j)
+    fn parse_programs(&mut self, mut i: usize) -> YaccResult<usize> {
+        if let Some(j) = self.lookahead_is("%%", i) {
+            i = self.parse_ws(j)?;
+            if i == self.src.len() {
+                Ok(i)
+            }
```
ltratt: Join these two lines?

ptersilie: Not sure I can, I still need `i` further down, so I can't just do
```
            if self.parse_ws(j)? == self.src.len() {
                Ok(i)
            }
```

ptersilie: Ah sorry, I realised you were referring to the `} else` bit. Github showed me the wrong source code extract which mislead me.

ltratt:
> Do you mean within the commit message?

Probably not a bad idea; it could be in the PR, if you want though.

cfgrammar/src/lib/yacc/parser.rs
ltratt: OK :)

ptersilie:
I guess I can just copy the grammar from the actions example (lrpar/examples/actions).

ltratt:
Works for me.

ltratt:
An alternative might be to work out how to make a test out of this?

ptersilie:
May be a bit early for a test since the API is gonna change a bit very soon, but a test is definitely on my todo list.

ltratt:
OK, please squash those last bits,

cfgrammar/src/lib/yacc/parser.rs
ltratt: Did this make it in yet?

ptersilie: I would have to write this as
```
c += match ch {
    '{' => 1,
    '}' if c == 1 => {
        c = 0;
        break;
    }
    '}' => -1
    _ => 0
};
```

So I am still uncertain if this increases readability.

ltratt: I think either of the solutions you've suggested here are OK: but they are better than what's currently in the commit, as they make clear that the condition is satisfied within the loop.

ptersilie: Okay, I've updated it.

ptersilie:
Addressed the last remaining comment. Will squash a bit once approved.

ltratt:
Please squash a bit.

ptersilie:
I just added two new commits, which are ready for review:
- e1ab296 mainly improves grammar actions by allowing the user to parse Lexemes upon their reduction. Unlike Yacc, which requires users to parse tokens during shifting, this gives users more context to decide how tokens are parsed (e.g. a number could be parsed as an INT or a FLOAT depending on which production we are in). This commit also changes the indexing within actions akin to Yacc (indexes start at 1) and prefixes generated code as to not clash with user code.
- 79c16de allows the user to define the name of the return type that will be used for the actions. Previously, this was a fixed value `TYPE`. Now the name can be customised using the grammar option `%type`.

ptersilie:
> I'm not sure `UserType` is a great name, because there could be multiple user types in different parts of the program. Maybe `ActionType` would be a better name?

Agreed, `ActionType` sounds good.

ltratt:
Please remember to only do a force push when you're explicitly asked to, because it confuses the life out of reviewers (e.g. me).

I still heartily dislike the commit message for https://github.com/softdevteam/grmtools/pull/32/commits/5900759ca68aa2c7cabd1ac902a8d66bbc993edf which gives the strong impression we're affecting parsing, when we aren't. All we're doing is changing the way actions are processed relative to Yacc. It's really important to be clear here because actions *can* influence parsing in Yacc, but that's not what this commit is doing at all.

ltratt:
Let's get the vector reuse thing in (which is a 2 minute job), then we can hopefully start squashing.

ptersilie:
Yes, sorry for the force push. I took your dislike of the commit message as implicitly allowing me to force push, as there's no other way I can edit the message. But I should have mentioned that as to not confuse you. I am going to edit it again, so do I have the explicit permission to force push when I do?

ltratt:
Let's leave that until the very end (which isn't far away, once the vector reuse thing is done), so we get all the commit message stuff done at the same time as squashing.

ptersilie:
Okay, pushed the reusable vector change. Please let me know when I can squash!

ltratt:
Looks good to me. Please squash.

ptersilie:
Squashed!

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -351,6 +353,11 @@ where
         &self.programs
     }
 
+    /// Get the action return type as defined by the user
+    pub fn usertype(&self) -> &Option<String> {
```
ltratt: Shouldn't this be called `actiontype` (and the attribute name too)?

ltratt:
OK, so we need some commit rewording. In this repo, at least, we always put full stops after the first line of a commit message. Yacc is spelt "Yacc" (not YACC); Rust is "Rust" (not rust).

https://github.com/softdevteam/grmtools/pull/32/commits/9a3334193df1166e61f0254d821dc6f421c5c536 is hard to parse: split the sentence into two maybe?

https://github.com/softdevteam/grmtools/pull/32/commits/c5610b65b168bf32572ebb821831df91a4b2af14 is still misleading (sorry to harp on about this!).  "This commit improves grammar actions by allowing the user to process Lexemes during reductions, unlike Yacc, which requires processing during shifting. This commit also changes the indexing within actions akin to Yacc (indexes start at 1) and prefixes generated code as to not clash with user code." could be something like "Allow users access to lexemes when executing actions."

There's a non-commit message comment to deal with first though.

cfgrammar/src/lib/yacc/grammar.rs
ptersilie: Oh, I missed that. Fixing.

ptersilie:
Pushed the `usertype -> actiontype` change. Will squash that in when I reword the commits. As for the Lexeme commit. For me Yacc also gives the user access to the lexemes via the `yylex` function, wherein the user needs to set `yylval` to push stuff onto the action stack. So the user can only push stuff onto the action stack during shifts. I wanted to make the distinction that lrpar allows us to push stuff onto the stack during reductions, which gives us slightly more expressiveness.

ptersilie:
How about: "Unlike Yacc, where the action stack can only be manipulated during shifting, these changes give us access to the lexemes when executing actions."

ltratt:
> I wanted to make the distinction that lrpar allows us to push stuff onto the stack during reductions, which gives us slightly more expressiveness.

I think two things are getting muddled here.

First, lex and yacc in normal parsing communicate via `yylex` (a function which returns the ID of the next token) and `yylval` (a variable which contains the token's value). This is a nasty way of doing things. grmtools doesn't have that problem, and the solution was in a PR a few weeks ago (https://github.com/softdevteam/grmtools/pull/20).

Second, yacc doesn't store string values into `yylval`; it expects the user to preemptively convert lexemes into `ActionType`s during lexing (not during shifting AFAICS). grmtools gives actions the string representing a token and allows them to do whatever they want with it.

The critical thing here, IMHO, is that "shifting" and "reducing" are core parsing actions; lexing and actions aren't. And I think the latter makes it clearer what's going on in our case.

ptersilie:
My goal was to make it clear that by giving the user access to lexemes within the actions, it gives us more expressiveness, as the user can convert tokens into different types depending on the production they are in. This is not possible in yacc, due to the fact how Yacc handles `yylval`, which is indeed is nasty. But if you believe that "Allow users access to lexemes when executing actions." is enough to make that distinction, then I'm fine with it.

ltratt:
Right, I think the thing to do is to excise any mention of shift/reduce and be clear that we're talking about lexing/actions, then we should be fine.

ptersilie:
Alright, let's give this another go. Are you happy with the last commit. Then I can squash that in as well!

ptersilie:
Squashed and reworded. Squashing of that last commit was a bit more complicated as I hoped, since it was changing multiple commits (I ended up dropping the last commit, and manually reapplying the changes do the other commits). I also noticed that I forgot to rename `UserT` to `ActionT`, which I've also done at the time.

I also rephrased the commit messages. See if they are better now.

ltratt:
Looks good to me. Merge as-is?

ptersilie:
Unless you feel like we can squash down more commits, I'm happy with merging as is.

PR #34, opened 2018-10-28
Add the ability to calculate a Yacc grammar's Follow sets.

ltratt:
This single-commit PR is all that's left of a failed attempt to implement Follow-set recovery for an LR parser. In essence, I stumbled across [this description of error recovery](https://www.cs.clemson.edu/course/cpsc827/material/LRk/LR%20Error%20Recovery.pdf) which made me think that follow set error recovery (which is widely used in LL parsing) might work in LR parsing. However, I'm pretty sure that it can't work: you can end up in loops where the error recovery transforms *(tokens, pstack)* into *(tokens, pstack')* which, when run, gets converted into *(tokens, pstack)* (ad infinitum).

However, I implemented a Follow sets calculation for Yacc grammars as part of this, and other people might find it useful, so I think it's probably worth keeping.

### DIFF
--- a/cfgrammar/src/lib/yacc/follows.rs
+++ b/cfgrammar/src/lib/yacc/follows.rs
@@ -1,0 +1,247 @@
+// Copyright (c) 2018 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any person obtaining a
+// copy of this software, associated documentation and/or data (collectively the "Software"), free
+// of charge and under any and all copyright rights in the Software, and any and all patent rights
+// owned or freely licensable by each licensor hereunder covering either (i) the unmodified
+// Software as contributed to or provided by such licensor, or (ii) the Larger Works (as defined
+// below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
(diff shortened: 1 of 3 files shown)


ptersilie:
Looks good. Though maybe we can add a sentence describing how this differs from the traditional follow set calculation and whether or not there are any benefits?

ltratt:
Does it differ? It's not meant to!

ptersilie:
Oh from the description I assumed this was a different way of doing follow sets. My memory of follow sets is a bit hazy so I thought the "doing it backwards" bit was novel. So didn't we have follow set calculation for YACC grammars before now?

ltratt:
All the descriptions I've seen of follow sets have been a specification, not an algorithm, so this is what I came up with. The test cases are ones I found online, but if you can think of a trickier test case that might stretch the algorithm, we can probably put it in.

ltratt:
Eco has a follow set calculation in there, so there are reasons why one might want it (presumably!). And it doesn't *hurt* (apart from being a bit more code to maintain).

ptersilie:
Yes, I checked Eco as well before I wrote the above. It calculates FOLLOW sets, but they are never used. I assume this is a left-over from when Eco used an LALR parser before we switched to LR(1). But you are right, it can't hurt to have this option, as long as we don't calculate them when generating a parser that doesn't need them.

ltratt:
OK, so merge or ...?

ptersilie:
Yes, I will merge.

PR #37, opened 2018-11-08
Mention nimbleparse in the README.

ltratt:
I meant to do this a while back but forgot.


ptersilie:
Looks good, is it worth mentioning that `nimbleparse` is not meant to be used for production and instead grammars should be compiled as in the calcparse example?

ltratt:
I don't think we quite want to say that, because the ability to dynamically load grammars is important for some use cases (e.g. Eco v2).

ptersilie:
Fair. I'll merge as is then.

PR #38, opened 2018-11-15
Started documentation

ptersilie:
This PR adds a documentation for lrpar. The documentation is written in markdown and can be compiled with `mdbook` (can be installed with `cargo install mdbook`).

Todo:
- [x] Quickstart Guide
- [ ] Tutorial (will be done in a separate PR)
- [x] Automatically generate docs and publish them online (maybe on github.io?)


lrpar/doc/src/quickstart.md
```diff
@@ -0,0 +1,94 @@
...
+lrlex = { path="<path to lrlex>" }
+lrpar = { path="<path to lrpar>" }
+
+[dependencies]
+cfgrammar = { path="<path to cfgrammar>" }
+lrlex = { path="<path to lrlex>" }
+lrpar = { path="<path to lrpar>" }
+```
+
+In the root of the project we create a file `build.rs`, which processes our lexer and grammar files.
```
ltratt: We should wrap lines to 80 chars, otherwise future diffs will be a nightmare.

lrpar/doc/src/quickstart.md
```diff
@@ -0,0 +1,94 @@
...
+name = "project"
+version = "0.0.1"
+authors = ["<authors>"]
+
+[[bin]]
+doc = false
+name = "project"
+
+[build-dependencies]
+lrlex = { path="<path to lrlex>" }
```
ltratt: When we make a release (which hopefully won't be long!), we have to remember to change this to `lrlex = "0.1"` and so on.

lrpar/doc/src/quickstart.md
```diff
@@ -0,0 +1,94 @@
...
+lrpar = { path="<path to lrpar>" }
+
+[dependencies]
+cfgrammar = { path="<path to cfgrammar>" }
+lrlex = { path="<path to lrlex>" }
+lrpar = { path="<path to lrpar>" }
+```
+
+In the root of the project we create a file `build.rs`, which processes our lexer and grammar files.
+Here we can also choose the integer size for the token ID's and whether we want to generate a default
```
ltratt: "'" is a possessive in this context which doesn't make sense: "IDs" is all we need.

lrpar/doc/src/quickstart.md
```diff
@@ -0,0 +1,94 @@
...
+extern crate lrpar;
+
+use lrlex::LexerBuilder;
+use lrpar::{CTParserBuilder, ActionKind};
+
+fn main() -> Result<(), Box<std::error::Error>> {
+    // We specify the integer type for token ID's as u8, which suffices for our
+    // purposes. Note, that this type *must* be big enough to fit all token
+    // IDs in.
+    let lex_rule_ids_map = CTParserBuilder::<u8>::new().action_kind(ActionKind::CustomAction).process_file_in_src("project.y")?;
```
ltratt: If you're using `CustomAction`, it doesn't matter too much (I think) what integer size you use, because you're not storing many of them a go, so I would drop this entirely (including the sentence earlier). It will default to `u32` if not specified.

ptersilie:
I cleaned this branch up a bit since the deploy testing got a bit out of hand. Unfortunately, this meant I had to squash everything down to a single commit. You've already seen `docs/*` changes, so you only need to review the `.travis.yaml` file and the second commit e3d5bd1, which needs to be squashed afterwards.

doc/src/quickstart.md
```diff
@@ -0,0 +1,91 @@
+# Quickstart Guide
+
+This is a short guide to get you started. For a more detailed explanation about
+what's going on, please refer to the [tutorial](tutorial/index.html).
+
+We begin by creating a `Cargo.toml` file and load the necessary dependencies:
```
ltratt: s/load/specify/

doc/src/quickstart.md
```diff
@@ -0,0 +1,91 @@
+# Quickstart Guide
+
+This is a short guide to get you started. For a more detailed explanation about
+what's going on, please refer to the [tutorial](tutorial/index.html).
```
ltratt: I think we need to quickly mention the main terminology: Yacc and lex files, at least.

doc/src/quickstart.md
```diff
@@ -0,0 +1,91 @@
...
+lrpar = { path="<path to lrpar>" }
+
+[dependencies]
+cfgrammar = { path="<path to cfgrammar>" }
+lrlex = { path="<path to lrlex>" }
+lrpar = { path="<path to lrpar>" }
+```
+
+In the root of the project we create a file `build.rs`, which processes our
+lexer and grammar files. Here we can choose whether we want to generate a
```
ltratt: We should tell people where their grammar and lexer files should be stored.

doc/src/quickstart.md
```diff
@@ -0,0 +1,91 @@
...
+    let lex_rule_ids_map = CTParserBuilder::new().action_kind(ActionKind::CustomAction).process_file_in_src("project.y")?;
+    LexerBuilder::new()
+        .rule_ids_map(lex_rule_ids_map)
+        .process_file_in_src("project.l")?;
+    Ok(())
+}
+```
+
+Now we can create the main function for our program which uses the parser
+generated in `build.rs`. When run this problem will take user input, lex and
```
ltratt: This example doesn't make much sense without the lex and yacc files.

doc/src/quickstart.md
```diff
@@ -0,0 +1,91 @@
...
+parse it according to the grammar, and return the result of the grammar
+actions:
+
+```rust
+extern crate lrpar;
+
+use lrpar::{LexParseError, Lexer};
+
+lrlex_mod!(project_l);
```
ltratt: This stuff needs to be commented otherwise it's just gobbledegook. `lrpar/examples/calcparse`'s `build.rs` and `src/main.rs` might be commented well enough?

doc/src/quickstart.md
ptersilie: Maybe I could do this in an Introduction section?

ltratt: I think a sentence or two in the quick start would be a good idea too. An ideal quick start section is enough to get semi-experienced users up and running on its own, so it needs to be fairly self contained.

doc/src/quickstart.md
ptersilie: I was thinking about this `build.rs` file. I wonder if you could make this is a bit simpler for the user by providing a function that takes the action type, the lexer, and grammar files, and then does the `CTBuilder::new` and `LexerBuilder::new` thing for them. Or are there reasons why they need to see these steps?

ltratt: The point is that many users won't be able to use `lrlex` because regexs aren't powerful enough to lex their language (e.g. PHP). So `lrpar` is stand-alone because of that and thus `CTBuilder` shouldn't know anything about `lrlex` (though `lrlex` does know about `lrpar`).

ptersilie: Ah, I see. That makes sense!

doc/src/quickstart.md
ptersilie: I removed the comments, because I wanted the quickstart guide to be concise and move the explanations to the tutorial. But I will put the comments back in.

doc/src/quickstart.md
```diff
@@ -1,9 +1,19 @@
 # Quickstart Guide
 
-This is a short guide to get you started. For a more detailed explanation about
-what's going on, please refer to the [tutorial](tutorial/index.html).
+This is a short guide to get you started with lrpar and shows how to generate a
+standalone parser from a given grammar. lrpar is meant to be used as a
+replacement for the parser generator
+[Yacc](https://en.wikipedia.org/wiki/Yacc), and thus uses the same syntax
+whereever possible. The main difference are the grammar actions, which unlike
```
ltratt: Move the comma to after "which".

doc/src/quickstart.md
```diff
@@ -1,9 +1,19 @@
...
+Yacc, need to be written in Rust. The input for lrpar is a lexer and a grammar
```
ltratt: lrpar takes in only a grammar. lrlex takes in the lexer.

ptersilie:
See if this is better now.

doc/src/quickstart.md
```diff
@@ -1,9 +1,19 @@
...
+file, where the former describes lexing rules in form of regular expressions
+which are used to tokenise the input. The latter describes rules to parse the
+tokenised input returned by the lexer while either generating a parse tree or
+executing the code specified within the grammar actions.
+For a more information and detailed explanations please refer to the
```
ltratt: "For more"

doc/src/quickstart.md
```diff
@@ -1,9 +1,19 @@
...
+[tutorial](tutorial/index.html).
 
-We begin by creating a `Cargo.toml` file and load the necessary dependencies:
+We begin by creating a `Cargo.toml` file and specify the necessary dependencies:
```
ltratt: Maybe "specifying"?

doc/src/quickstart.md
```diff
@@ -45,6 +55,55 @@ fn main() -> Result<(), Box<std::error::Error>> {
...
+```
+%%
+[0-9]+ "INT"
+\+ "PLUS"
+\* "MUL"
+\( "LBRACK"
+\) "RBRACK"
+[\t ]+ ;
+```
+
```
ltratt: Need something here otherwise people won't realise when the lex file stops and the grammar starts.

doc/src/quickstart.md
```diff
@@ -45,6 +55,55 @@ fn main() -> Result<(), Box<std::error::Error>> {
...
+Term: Factor 'MUL' Term { mul ($1, $3) }
+    | Factor { $1 }
+    ;
+
+Factor: 'LBRACK' Expr 'RBRACK' { $2 }
+      | 'INT' { int($1) }
+      ;
+%%
+
+type MYTYPE = u64;
```
ltratt: Need to add a comment explaining `MYTYPE` (maybe at the `%type` line?).

doc/src/quickstart.md
```diff
@@ -45,6 +55,55 @@ fn main() -> Result<(), Box<std::error::Error>> {
...
+
+fn int(s: &str) -> MYTYPE {
```
ltratt: Add a brief comment here saying something like "this function is in scope for all the grammar actions above"?

doc/src/quickstart.md
ptersilie: I meant that `parse()` takes a lexer as input, but let me see how I can rewrite this.

ltratt: Ah, yeah, this is hard to phrase. Maybe "lrpar takes as input a lexer instance (from which it consumes lexemes) and a grammar (which it parses the lexemes against)" or similar?

ptersilie:
Have another look.

ltratt:
I think we can probably get this in as a start. Squash?

ptersilie:
Squashed!

PR #42, opened 2018-11-20
Allow grammars to specify pretty-printed versions of tokens for error recovery

ltratt:
**This is a request for comments, at least initially: is this a good idea?**

Consider this following Lua fragment:

```lua
  if then end
```

If we run our parser on it we get:

```
  Error at line 1 col 4. Repairs found:
    Insert LONG_STR
    Insert SHORT_STR
    Insert DOTDOTDOT
    Insert NUMERAL
    Insert TRUE
    Insert FALSE
    Insert NIL
    Insert NAME
```

The problem is that the user might not know what those token names mean. What's the difference between LONG_STR and SHORT_STR?

This commit allows users to add %epp <token> <pp> declarations to a grammar such as:

```
  %epp LONG_STR [=[...]=]
  %epp SHORT_STR "..."
```

When an %epp declaration exists for a token that is used instead of the token name. Thus this commit makes error recovery show the following to users:

```
  Error at line 1 col 4. Repairs found:
    Insert [=[...]=]
    Insert "..."
    Insert ...
    Insert <Numeral>
    Insert true
    Insert false
    Insert nil
    Insert <Name>
```

It's a bit of a toss-up as to whether this information should go in the lexer or parser. Since error recovery is an entirely grammar-based thing, I've come down on the side of putting it in the grammar.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -51,6 +51,8 @@ pub struct GrammarAST {
     pub tokens: IndexSet<String>,
     pub precs: HashMap<String, Precedence>,
     pub implicit_tokens: Option<HashSet<String>>,
+    // Error pretty-printers
+    pub epp: HashMap<String, String>,
     pub programs: Option<String>,
     pub actiontype: Option<String>
 }
@@ -81,7 +83,8 @@ pub enum GrammarValidationErrorKind {
     InvalidStartRule,
     UnknownRuleRef,
     UnknownToken,
-    NoPrecForToken
+    NoPrecForToken,
+    UnknownEPP
 }
 
 /// `GrammarAST` validation errors return an instance of this struct.
@@ -114,7 +117,10 @@ impl fmt::Display for GrammarValidationError {
                 f,
                 "Token '{}' used in %prec has no precedence attached",
                 self.sym.as_ref().unwrap()
-            )
+            ),
+            GrammarValidationErrorKind::UnknownEPP => {
+                write!(f, "Unknown token '{}' in %epp declaration", self.sym.as_ref().unwrap())
+            }
         }
     }
 }
@@ -138,6 +144,7 @@ impl GrammarAST {
             tokens: IndexSet::new(),
             precs: HashMap::new(),
             implicit_tokens: None,
+            epp: HashMap::new(),
             programs: None,
             actiontype: None
         }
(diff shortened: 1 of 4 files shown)


ltratt:
Of course, finding good pretty-printings for some tokens is hard. Maybe LONG_STR/SHORT_STR should be:

```
%epp SHORT_STR "<String>"
%epp LONG_STR [=[<String>]=]
```

to avoid confusion with the DOTDOTDOT token?

ptersilie:
This sounds reasonable to me. Though I have to high-level comments before I look at the code:
1) What does `epp` stand for? I can't immediately think of anything. I was expecting something like `exp` for "explanation".
2) Though this may be more helpful for the developer than the user, we might want to have the error name show both the token type AND the explanatory string. Although on second thought, the developer can just put that into the string if they wanted to. Which is probably better.

ltratt:
"EPP" == "Error Pretty Print". I'm open to other names.

[I agree with your second point: the grammar developer can print out whatever they want, including the token name if they so choose.]

ptersilie:
I just thought of a third point. Maybe we extend the `%token` declaration, in that it takes an explanation as well, e.g. `%token TRUE "true"`. Then we don't need to write an entirely new block of `%epp`, since we already have a block with token declarations.

ltratt:
In our case, I think `%token` is entirely redundant (I might be wrong), so I guess we could overload it. But is that more confusing than a separate declaration? I'm not sure either way.

ltratt:
To be clear, you don't ever need to use `%token` in our set-up.

ptersilie:
You mean theoretically, or practically? Because atm if I don't define `%token` the parser complains.

ltratt:
https://github.com/softdevteam/grammars/blob/master/lua5_3/lua5_3.y for example contains no `%token` declarations.

ptersilie:
Ah yes, but only because all tokens are in quotes. You only need the `%token` declaration if you reference tokens without quotes, so the parser can separate them from grammar rules. It's just another weird Yacc thing, which we'll have to decide if we are going to keep it, or get rid of it and document that this doesn't work in lrpar.

ltratt:
We already support this aspect, so I don't see any utility in removing it. I agree it's an odd Yacc feature but, to the extent we can, it's nice if Yacc grammars "just work" without modifications.

On that basis, I think I'd prefer to stick with a new declaration and not reuse `%token`. If you take a grmtools `.y` file and move it to Yacc, you can just delete the `%epp` declarations, which is easier than manually editing the `%token` declarations (and/or working out which ones you even need). Sound like a plan?

ptersilie:
While I personally would have preferred them in a single declaration, I can see now that this would cause problems when we want to define an `%epp` for a string token `"tok", when there is also a rule `tok`. This wouldn't be possible if we only use `%token` as then the rule would be interpreted as a token.

Code looks also fine. So I will merge.

ptersilie:
Oh, you'll need to rebase first.

PR #44, opened 2018-11-20
More careful whitespace skipping

ltratt:
This commit first tightens up our skipping of whitespace (https://github.com/softdevteam/grmtools/commit/d25c68130b8d66fc71ea34be0af94a8f16914ea4) before fixing an entertaining bug I found in the "%epp" declarations in the Lua grammar (https://github.com/softdevteam/grmtools/commit/2ee56f64650ad9b891468e23fa2f451c85d1f245).

The second commit (https://github.com/softdevteam/grmtools/commit/2ee56f64650ad9b891468e23fa2f451c85d1f245) is worthy of some thought: do we want to make %epp declarations quoted? If so, we can somewhat easily do so, and we should do it sooner rather than later (though, probably, in another PR).

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -60,7 +60,8 @@ pub enum YaccParserErrorKind {
     PrecNotFollowedByToken,
     DuplicateImplicitTokensDeclaration,
     DuplicateStartDeclaration,
-    DuplicateEPP
+    DuplicateEPP,
+    ReachedEOL
 }
 
 /// Any error from the Yacc parser returns an instance of this struct.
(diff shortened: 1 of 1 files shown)


ptersilie:
Is this ready for review (since I wasn't assigned to it)?

ltratt:
That was a mistake on my part. Fixed.

ptersilie:
The last bit is a very interesting question. I do like that we don't have to use quotes and escape stuff, but I can also see how the current fix is a bit...weird. If we would allow the use of single quotes and double quotes, would that reduce the amount of escaping the user has to do?

ltratt:
Ah, I see. I'm inclined to agree. Shall we do that in a later PR then?

ptersilie:
If you believe that is a better solution than what we got now, sure! I can merge this one first then.

ltratt:
I should probably rebase first.

ptersilie:
Go ahead!

PR #46, opened 2018-11-21
Force %epp declaration values to be quoted.

ltratt:
What was:

```
  %epp A xyz
```

is now:

```
  %epp A "xyz"
```

Escaping works e.g.:

```
  %epp String "\"<String\""
```

If you want to test, the updated grammars are at https://github.com/softdevteam/grammars/commit/25d57ba890af7a54cd337d690462d94cb3665b0d.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -61,7 +61,8 @@ pub enum YaccParserErrorKind {
     DuplicateImplicitTokensDeclaration,
     DuplicateStartDeclaration,
     DuplicateEPP,
-    ReachedEOL
+    ReachedEOL,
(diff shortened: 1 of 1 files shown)


ptersilie:
I thought we agreed that one could use either double or single quotes, so that the user won't have to escape them, e.g. `%epp String '"<String"'`

ltratt:
Yes, you can use either (see the tests). But escaping works too!

ptersilie:
Nevermind, just saw the code implements this. I just assumed it doesn't because of your example in the PR desciption! Sorry!

ptersilie:
Maybe we can add a small sentence to the commit message to say that single quotes are also supported?

ltratt:
Done!

PR #48, opened 2018-11-22
Rustfmt everything

ltratt:
We previously partly rustfmt'd the codebase. Unfortunately there were a few places where rustfmt made things worse, and I fixed most, and manually excluded a few more. However, this isn't really a good long-term plan. I've now fixed the remaining cases that rustfmt mangles (https://github.com/softdevteam/grmtools/commit/51d02cbed70a2d135d11aab553d5b89d25dc4e39) so that we can run rustfmt over the entire code base (https://github.com/softdevteam/grmtools/commit/6d9a554ae0bc78736e33a6925eedd6f6159e7d96) and enforce its use in future PRs with Travis. Probably only the first and third commits really need checking in depth.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,18 @@ cache:
     - cargo
 install:
     - cargo install --list | grep mdbook || cargo install mdbook --vers "0.2.2"
+before_script:
+    - rustup component add rustfmt-preview
 script:
+    - cargo fmt --all -- --check
     - cargo test --all
     - cd $TRAVIS_BUILD_DIR/cfgrammar && cargo test --features "serde"
     - cd $TRAVIS_BUILD_DIR/lrtable && cargo test --features "serde"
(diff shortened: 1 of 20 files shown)


ptersilie:
I'm not a huge fan of those formattings, where it puts the `.something()` on the next line, e.g.
```
]
.iter()
```

But I'm not sure if this can be toggle or we even want it to. Probably not worth thinking about, but I thought I point it out.

ltratt:
rustfmt does have various settings, but they're somewhat crude overall. I'm not a total fan of rustfmt's output, but I think it's better overall than what we had before, and it removes the stylistic differences that such a syntactically rich language otherwise leads to. So I think we might have to live with this one :/

ptersilie:
Yes, it doesn't bother beyond a mild stylistic preference. Anyways, the branch needs rebasing anyways, then I can merge.

ltratt:
Rebased.

PR #51, opened 2018-11-22
Make grmtools build on stable.

ltratt:
There are a couple of compromises involved in this:

  * We have to use the external try_from crate, as TryFrom is not yet stabilised  to be honest, there isn't much sign of that changing).

  * I deleted one benchmark that requires the (unstable) test crate. The  isn't really useful any more.

  * stable doesn't have tool lints so the Clippy shutter-upper `#[allow(clippy::let_and_return)]` has to go. This can probably return when rustc 1.31 is out.

However, I suspect these are worthwhile in order that we no longer require users to use nightly. Thoughts welcome!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
 language: rust
-rust: nightly
 cache:
     - cargo
 install:
     - cargo install --list | grep mdbook || cargo install mdbook --vers "0.2.2"
 before_script:
-    - rustup component add rustfmt-preview
+    - rustup toolchain install nightly
+    - rustup component add --toolchain nightly rustfmt-preview
+    - which rustfmt || cargo install --force rustfmt-nightly
 script:
-    - cargo fmt --all -- --check
+    - cargo +nightly fmt --all -- --check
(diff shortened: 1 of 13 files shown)


ptersilie:
All sounds reasonable to me.

ltratt:
It turns out that stable rustfmt is quite different from nightly rustfmt. Hopefully that latest commit makes us use nightly rustfmt (even though we test everything else on stable). We'll find out in 10 minutes or so...

ltratt:
OK, so this is interesting. The *branch* build passes OK https://github.com/softdevteam/grmtools/pull/51/checks?check_run_id=34179189 (i.e. if we run this branch "as is"). However the *pull request build* uses the .travis.yml file from master and thus fails. At least I think that's what's going on. The Travis documentation on this is confusing https://docs.travis-ci.com/user/pull-requests/.

So I *think* we should merge this after squashing, but I'm not sure.

ptersilie:
Ok, please squash!

ltratt:
Squashed.

PR #53, opened 2018-11-23
Build a book for each tag version (including master)

ptersilie:
**PR in progress:** We should test this build script once this repo has tags.

This PR changes the book building script so that it creates and
publishes a book for each tag as well as the master branch.

### DIFF
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -1,1 +1,2 @@
 book
+release
--- a/doc/build.sh
+++ b/doc/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# build books for each tag
+for tag in $(git tag)
+do
+    echo "Building book for $tag"
+    git checkout $tag
+    mdbook build -d release/$tag/book/
+done
+
+# build book/apidoc for master
+git checkout master
+echo "Building book for master"
+mdbook build -d release/master/book
+echo "Building docs for master"
+RUSTDOCFLAGS="-Z unstable-options --enable-index-page" cargo +nightly doc --no-deps
+mv ../target/doc release/master/api
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 cache:
     - cargo
 install:
-    - cargo install --list | grep mdbook || cargo install mdbook --vers "0.2.2"
+    - cargo install --list | grep mdbook || cargo install mdbook --vers "=0.2"
 before_script:
     - rustup toolchain install nightly
     - rustup component add --toolchain nightly rustfmt-preview || cargo +nightly install --force rustfmt-nightly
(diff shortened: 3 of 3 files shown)


doc/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# build book/apidoc for each tag
+for tag in $(git tag)
+do
+    git checkout $tag
+    mdbook build -d book/$tag
+    cargo doc --target-dir book/api/$tag
```
ltratt: I think it might be better to bundle everything under the version. So we'd have something like:

```sh
    mdbook build -d $tag/book
    cargo doc --target-dir $tag/api
```

ptersilie:
There's one issue I just realised. If we build the docs for each release, then the more releases (i.e. tags) we have, the longer it will take for travis to complete. Luckily this only happens on master commits, but it will still take a long time once we have a couple of releases.

There is a way to tell travis to do incremental commits, but we'll have to see how well that works.

ptersilie:
As discussed offline I changed the PR to only generated docs for the master branch and use docs.rs for the releases. Books still get created for the master branch and each release tag and will be hosted on softdevteam.github.io/grmtools.

One problem with `cargo doc` at the moment is that it doesn't create an index page of the available crates. However this was fixed in https://github.com/rust-lang/rust/pull/54543 and will hopefully be part of a release soon.

ltratt:
Just to check: the thought here is to wait until rust 1.31 is out before merging this?

ptersilie:
Given it will be merged then, yes. The alternative is to provide our own index page.

ptersilie:
It appears the index-page feature is still unstable. I tried using it as described here (https://github.com/rust-lang/rust/blob/e53a5ffd6b1c28903b4ad9a4c433fd554802cd9c/src/doc/rustdoc/src/unstable-features.md#--index-page-provide-a-top-level-landing-page-for-docs) but couldn't get it to work at all.

Back to the drawing board, I guess.

ltratt:
I'm a bit out of thoughts on this one TBH...

ptersilie:
I managed to get the default index-page working and also adjusted the travis deploy script to pick up the new folder.

doc/.gitignore
```diff
@@ -1 +1 @@
-book
```
ltratt: I think we still need `book/` in the `.gitignore` because when we do `mdbook build` locally, it'll output it in `book/`.

ltratt:
You probably want to cherry-pick b0182a9a into this branch?

ptersilie:
Addressed your comment and merged in master. Let's wait for the tests and then we can rebase and merge.

ltratt:
If the tests succeed, please squash and merge without waiting for me to say so.

ptersilie:
Squashed and rebased!

PR #55, opened 2018-11-28
Started work on conflict API

ptersilie:
** PR in progress **

The goal of this PR is to provide a nice API for the user (i.e. the language developer) to inspect shift/reduce and reduce/reduce conflicts and handle them manually.

At the moment the API is **very** basic, but I thought I get some early thoughts on the direction before continuing. An example usage of the API can be found in nimbleparse, where it just pretty prints the conflicts to the terminal.

My main question would be, where the user would be expected to handle these errors. For example within the calc examples, would that be in `main.rs` or `build.rs`? Or am I off track again, and we want to allow the user to handle the conflict as it happens and they then can manually change how the parser generator resolves those conflicts?

### DIFF
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -491,8 +491,8 @@ where
     }
 
     /// Get the action return type as defined by the user
-    pub fn actiontype(&self) -> &Option<String> {
-        &self.actiontype
+    pub fn actiontype(&self) -> Option<&String> {
+        self.actiontype.as_ref()
     }
 
     /// Get the programs part of the grammar
@@ -555,6 +555,22 @@ where
                 return false;
             }
         }
+    }
+
+    /// Returns the string representation of a given production `pidx`.
+    pub fn pp_prod(&self, pidx: PIdx<StorageT>) -> String {
+        let mut sprod = String::new();
+        let ridx = self.prod_to_rule(pidx);
+        sprod.push_str(self.rule_name(ridx));
+        sprod.push_str(":");
+        for sym in self.prod(pidx) {
+            let s = match sym {
+                Symbol::Token(tidx) => self.token_name(*tidx).unwrap(),
+                Symbol::Rule(ridx) => self.rule_name(*ridx)
+            };
+            sprod.push_str(&format!(" \"{}\"", s));
+        }
+        sprod
     }
 
     /// Return a `SentenceGenerator` which can then generate minimal sentences for any rule
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -35,7 +35,7 @@ use std::{
     convert::AsRef,
     env::{current_dir, var},
     error::Error,
-    fmt::Debug,
+    fmt::{self, Debug},
     fs::{self, read_to_string, File},
     hash::Hash,
     io::Write,
@@ -49,7 +49,7 @@ use cfgrammar::{
     Symbol
 };
 use filetime::FileTime;
-use lrtable::{from_yacc, Minimiser, StateGraph, StateTable};
+use lrtable::{from_yacc, statetable::Conflicts, Minimiser, StateGraph, StateTable};
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
@@ -79,14 +79,72 @@ pub enum ActionKind {
     GenericParseTree
 }
 
+struct CTConflictsError<StorageT: Eq + Hash> {
+    pub grm: YaccGrammar<StorageT>,
+    pub sgraph: StateGraph<StorageT>,
+    pub stable: StateTable<StorageT>
+}
+
+impl<StorageT> fmt::Display for CTConflictsError<StorageT>
+where
+    StorageT: 'static + Debug + Hash + PrimInt + Serialize + TypeName + Unsigned,
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let conflicts = self.stable.conflicts().unwrap();
+        write!(
+            f,
+            "CTConflictsError{{{} Shift/Reduce, {} Reduce/Reduce}}",
+            conflicts.sr_len(),
+            conflicts.rr_len()
+        )
+    }
+}
+
+impl<StorageT> fmt::Debug for CTConflictsError<StorageT>
+where
+    StorageT: 'static + Debug + Hash + PrimInt + Serialize + TypeName + Unsigned,
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let conflicts = self.stable.conflicts().unwrap();
+        write!(
+            f,
+            "CTConflictsError{{{} Shift/Reduce, {} Reduce/Reduce}}",
+            conflicts.sr_len(),
+            conflicts.rr_len()
+        )
+    }
+}
+
+impl<StorageT> Error for CTConflictsError<StorageT>
+where
+    StorageT: 'static + Debug + Hash + PrimInt + Serialize + TypeName + Unsigned,
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
+{
+}
+
 /// A `CTParserBuilder` allows one to specify the criteria for building a statically generated
 /// parser.
-pub struct CTParserBuilder<StorageT = u32> {
-    // Anything stored in here almost certainly needs to be included as part of the rebuild_cache
-    // function below so that, if it's changed, the grammar is rebuilt.
+pub struct CTParserBuilder<StorageT = u32>
+where
+    StorageT: Eq + Hash
+{
+    // Anything stored in here (except `conflicts` and `error_on_conflict`) almost certainly needs
+    // to be included as part of the rebuild_cache function below so that, if it's changed, the
+    // grammar is rebuilt.
     recoverer: RecoveryKind,
     phantom: PhantomData<StorageT>,
-    actionkind: ActionKind
+    actionkind: ActionKind,
+    error_on_conflicts: bool,
+    conflicts: Option<(
+        YaccGrammar<StorageT>,
+        StateGraph<StorageT>,
+        StateTable<StorageT>
+    )>
 }
 
 impl CTParserBuilder<u32> {
@@ -132,7 +190,9 @@ where
         CTParserBuilder {
             recoverer: RecoveryKind::MF,
             phantom: PhantomData,
-            actionkind: ActionKind::GenericParseTree
+            actionkind: ActionKind::GenericParseTree,
+            error_on_conflicts: true,
+            conflicts: None
         }
     }
 
@@ -156,9 +216,9 @@ where
     /// If `StorageT` is not big enough to index the grammar's tokens, rules, or
     /// productions.
     pub fn process_file_in_src(
-        &self,
+        &mut self,
         srcp: &str
-    ) -> Result<(HashMap<String, StorageT>), Box<Error>> {
+    ) -> Result<HashMap<String, StorageT>, Box<Error>> {
         let mut inp = current_dir()?;
         inp.push("src");
         inp.push(srcp);
@@ -171,6 +231,30 @@ where
     pub fn action_kind(mut self, ak: ActionKind) -> Self {
         self.actionkind = ak;
         self
+    }
+
+    /// If set to true, `process_file_in_src` will return an error if the given grammar contains
+    /// any Shift/Reduce or Reduce/Reduce conflicts. Defaults to `true`.
+    pub fn error_on_conflicts(mut self, b: bool) -> Self {
+        self.error_on_conflicts = b;
+        self
+    }
+
+    /// If there are any conflicts in the grammar, return a tuple which allows users to inspect
+    /// and pretty print them; otherwise returns `None`. Note: The conflicts feature is currently
+    /// unstable and may change in the future.
+    pub fn conflicts(
+        &self
+    ) -> Option<(
+        &YaccGrammar<StorageT>,
+        &StateGraph<StorageT>,
+        &StateTable<StorageT>,
+        &Conflicts<StorageT>
+    )> {
+        if let Some((grm, sgraph, stable)) = &self.conflicts {
+            return Some((grm, sgraph, stable, &stable.conflicts().unwrap()));
+        }
+        None
     }
 
     /// Statically compile the Yacc file `inp` into Rust, placing the output file(s) into
@@ -194,10 +278,10 @@ where
     /// If `StorageT` is not big enough to index the grammar's tokens, rules, or
     /// productions.
     pub fn process_file<P, Q>(
-        &self,
+        &mut self,
         inp: P,
         outd: Q
-    ) -> Result<(HashMap<String, StorageT>), Box<Error>>
+    ) -> Result<HashMap<String, StorageT>, Box<Error>>
     where
         P: AsRef<Path>,
         Q: AsRef<Path>
@@ -256,6 +340,15 @@ where
         fs::remove_file(&outp_rs).ok();
 
         let (sgraph, stable) = from_yacc(&grm, Minimiser::Pager)?;
+
+        if stable.conflicts().is_some() && self.error_on_conflicts {
+            return Err(Box::new(CTConflictsError {
+                grm,
+                sgraph,
+                stable
+            }));
+        }
+
         // Because we're lazy, we don't write our own serializer. We use serde and bincode to
         // create files $out_base.grm, $out_base.sgraph, and $out_base.out_stable which contain
         // binary versions of the relevant structs, and then include those binary files into the
@@ -268,12 +361,12 @@ where
         // Header
         let mod_name = inp.as_ref().file_stem().unwrap().to_str().unwrap();
         let actiontype = match grm.actiontype() {
-            Some(t) => t,
+            Some(t) => t.clone(), // Probably unneeded once NLL is in stable
             None => {
                 match self.actionkind {
                     ActionKind::CustomAction => panic!("Action return type not defined!"),
                     ActionKind::GenericParseTree => {
-                        "" // Dummy string that will never be used
+                        String::new() // Dummy string that will never be used
                     }
                 }
             }
@@ -475,6 +568,10 @@ where
 
         let mut f = File::create(outp_rs)?;
         f.write_all(outs.as_bytes())?;
+
+        if stable.conflicts().is_some() {
+            self.conflicts = Some((grm, sgraph, stable));
+        }
         Ok(rule_ids)
     }
 
@@ -564,4 +661,70 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
     let sgraph = deserialize(sgraph_buf).unwrap();
     let stable = deserialize(stable_buf).unwrap();
     (grm, sgraph, stable)
+}
+
+#[cfg(test)]
+mod test {
+    extern crate temp_testdir;
+    use std::{fs::File, io::Write, path::PathBuf};
+
+    use self::temp_testdir::TempDir;
+    use super::{ActionKind, CTConflictsError, CTParserBuilder};
+
+    #[test]
+    fn test_conflicts() {
+        let temp = TempDir::default();
+        let mut file_path = PathBuf::from(temp.as_ref());
+        file_path.push("grm.y");
+        let mut f = File::create(&file_path).unwrap();
+        let _ = f.write_all(
+            "%start A
+%%
+A : 'a' 'b' | B 'b';
+B : 'a' | C;
+C : 'a';"
+                .as_bytes()
+        );
+
+        let mut ct = CTParserBuilder::new()
+            .error_on_conflicts(false)
+            .action_kind(ActionKind::GenericParseTree);
+        ct.process_file_in_src(file_path.to_str().unwrap()).unwrap();
+
+        match ct.conflicts() {
+            Some((_, _, _, conflicts)) => {
+                assert_eq!(conflicts.sr_len(), 1);
+                assert_eq!(conflicts.rr_len(), 1);
+            }
+            None => panic!("Expected error data")
+        }
+    }
+
+    #[test]
+    fn test_conflicts_error() {
+        let temp = TempDir::default();
+        let mut file_path = PathBuf::from(temp.as_ref());
+        file_path.push("grm.y");
+        let mut f = File::create(&file_path).unwrap();
+        let _ = f.write_all(
+            "%start A
+%%
+A : 'a' 'b' | B 'b';
+B : 'a' | C;
+C : 'a';"
+                .as_bytes()
+        );
+
+        match CTParserBuilder::new()
+            .action_kind(ActionKind::GenericParseTree)
+            .process_file_in_src(file_path.to_str().unwrap())
+        {
+            Ok(_) => panic!("Expected error"),
+            Err(e) => {
+                let cs = e.downcast_ref::<CTConflictsError<u32>>();
+                assert_eq!(cs.unwrap().stable.conflicts().unwrap().sr_len(), 1);
+                assert_eq!(cs.unwrap().stable.conflicts().unwrap().rr_len(), 1);
+            }
+        }
+    }
 }
--- a/lrtable/src/lib/statetable.rs
+++ b/lrtable/src/lib/statetable.rs
@@ -50,6 +50,68 @@ use stategraph::StateGraph;
 use StIdx;
 use StIdxStorageT;
 
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct Conflicts<StorageT> {
+    reduce_reduce: Vec<(PIdx<StorageT>, PIdx<StorageT>, StIdx)>,
+    shift_reduce: Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)>
+}
+
+impl<StorageT: 'static + Hash + PrimInt + Unsigned> Conflicts<StorageT>
+where
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
+{
+    /// Return an iterator over all shift/reduce conflicts.
+    pub fn sr_conflicts(&self) -> impl Iterator<Item = &(TIdx<StorageT>, PIdx<StorageT>, StIdx)> {
+        self.shift_reduce.iter()
+    }
+
+    /// Return an iterator over all reduce/reduce conflicts.
+    pub fn rr_conflicts(&self) -> impl Iterator<Item = &(PIdx<StorageT>, PIdx<StorageT>, StIdx)> {
+        self.reduce_reduce.iter()
+    }
+
+    /// How many shift/reduce conflicts are there?
+    pub fn sr_len(&self) -> usize {
+        self.shift_reduce.len()
+    }
+
+    /// How many reduce/reduce conflicts are there?
+    pub fn rr_len(&self) -> usize {
+        self.reduce_reduce.len()
+    }
+
+    /// Returns a pretty-printed version of the conflicts.
+    pub fn pp(&self, grm: &YaccGrammar<StorageT>) -> String {
+        let mut s = String::new();
+        if self.sr_len() > 0 {
+            s.push_str("Shift/Reduce conflicts:\n");
+            for (tidx, pidx, stidx) in self.sr_conflicts() {
+                s.push_str(&format!(
+                    "   State {:?}: Shift(\"{}\") / Reduce({})\n",
+                    usize::from(*stidx),
+                    grm.token_name(*tidx).unwrap(),
+                    grm.pp_prod(*pidx)
+                ));
+            }
+        }
+
+        if self.rr_len() > 0 {
+            s.push_str("Reduce/Reduce conflicts:\n");
+            for (pidx, r_pidx, stidx) in self.rr_conflicts() {
+                s.push_str(&format!(
+                    "   State {:?}: Reduce({}) / Reduce({})\n",
+                    usize::from(*stidx),
+                    grm.pp_prod(*pidx),
+                    grm.pp_prod(*r_pidx)
+                ));
+            }
+        }
+        s
+    }
+}
+
 /// The various different possible Yacc parser errors.
 #[derive(Debug)]
 pub enum StateTableErrorKind {
@@ -95,10 +157,7 @@ pub struct StateTable<StorageT> {
     rules_len: RIdx<StorageT>,
     prods_len: PIdx<StorageT>,
     tokens_len: TIdx<StorageT>,
-    /// The number of reduce/reduce errors encountered.
-    pub reduce_reduce: u64,
-    /// The number of shift/reduce errors encountered.
-    pub shift_reduce: u64,
+    conflicts: Option<Conflicts<StorageT>>,
     pub final_state: StIdx
 }
 
@@ -149,8 +208,9 @@ where
         assert!(usize::from(sg.all_states_len()) < (usize::from(StIdx::max_value()) - 1));
         gotos.resize(maxg, 0);
 
-        let mut reduce_reduce = 0; // How many automatically resolved reduce/reduces were made?
-        let mut shift_reduce = 0; // How many automatically resolved shift/reduces were made?
+        // Store automatically resolved conflicts, so we can print them out later
+        let mut reduce_reduce = Vec::new();
+        let mut shift_reduce = Vec::new();
         let mut final_state = None;
 
         for (stidx, state) in sg
@@ -186,10 +246,10 @@ where
                             // By default, Yacc resolves reduce/reduce conflicts in favour
                             // of the earlier production in the grammar.
                             if pidx < r_pidx {
-                                reduce_reduce += 1;
+                                reduce_reduce.push((pidx, r_pidx, stidx));
                                 actions[off as usize] = StateTable::encode(Action::Reduce(pidx));
                             } else if pidx > r_pidx {
-                                reduce_reduce += 1;
+                                reduce_reduce.push((r_pidx, pidx, stidx));
                             }
                         }
                         Action::Accept => {
@@ -230,13 +290,15 @@ where
                         match StateTable::decode(actions[off as usize]) {
                             Action::Shift(x) => assert_eq!(*ref_stidx, x),
                             Action::Reduce(r_pidx) => {
-                                shift_reduce += resolve_shift_reduce(
+                                resolve_shift_reduce(
                                     grm,
                                     &mut actions,
                                     off as usize,
                                     s_tidx,
                                     r_pidx,
-                                    *ref_stidx
+                                    *ref_stidx,
+                                    &mut shift_reduce,
+                                    stidx
                                 );
                             }
                             Action::Accept => panic!("Internal error"),
@@ -307,6 +369,14 @@ where
         let actions_sv = SparseVec::<usize>::from(&actions, 0, usize::from(grm.tokens_len()));
         let gotos_sv = SparseVec::<usize>::from(&gotos, 0, usize::from(grm.rules_len()));
 
+        let mut conflicts = None;
+        if !(reduce_reduce.is_empty() && shift_reduce.is_empty()) {
+            conflicts = Some(Conflicts {
+                reduce_reduce,
+                shift_reduce
+            });
+        }
+
         Ok(StateTable {
             actions: actions_sv,
             state_actions,
@@ -317,8 +387,7 @@ where
             rules_len: grm.rules_len(),
             prods_len: grm.prods_len(),
             tokens_len: grm.tokens_len(),
-            reduce_reduce,
-            shift_reduce,
+            conflicts,
             final_state: final_state.unwrap()
         })
     }
@@ -424,6 +493,11 @@ where
             None => unreachable!()
         }
     }
+
+    /// Return a struct containing all conflicts or `None` if there aren't any.
+    pub fn conflicts(&self) -> Option<&Conflicts<StorageT>> {
+        self.conflicts.as_ref()
+    }
 }
 
 fn actions_offset<StorageT: PrimInt + Unsigned>(
@@ -478,13 +552,13 @@ fn resolve_shift_reduce<StorageT: 'static + Hash + PrimInt + Unsigned>(
     off: usize,
     tidx: TIdx<StorageT>,
     pidx: PIdx<StorageT>,
-    stidx: StIdx
-) -> u64
-where
+    stidx: StIdx, // State we want to shift to
+    shift_reduce: &mut Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)>,
+    conflict_stidx: StIdx // State in which the conflict occured
+) where
     usize: AsPrimitive<StorageT>,
     u32: AsPrimitive<StorageT>
 {
-    let mut shift_reduce = 0;
     let tidx_prec = grm.token_precedence(tidx);
     let pidx_prec = grm.prod_precedence(pidx);
     match (tidx_prec, pidx_prec) {
@@ -492,7 +566,7 @@ where
             // If the token and production don't both have precedences, we use Yacc's default
             // resolution, which is in favour of the shift.
             actions[off] = StateTable::encode(Action::Shift(stidx));
-            shift_reduce += 1;
+            shift_reduce.push((tidx, pidx, conflict_stidx));
         }
         (Some(token_prec), Some(prod_prec)) => {
             if token_prec.level == prod_prec.level {
@@ -524,7 +598,6 @@ where
             // the reduce as-is.
         }
     }
-    shift_reduce
 }
 
 #[cfg(test)]
@@ -868,6 +941,42 @@ mod test {
                    Action::Shift(s6));
         assert_eq!(st.action(s10, grm.eof_token_idx()),
                    Action::Reduce(grm.rule_to_prods(grm.rule_idx("Expr").unwrap())[0]));
+    }
+
+    #[test]
+    fn conflicts() {
+        let grm = YaccGrammar::new(
+            YaccKind::Original,
+            &"
+%start A
+%%
+A : 'a' 'b' | B 'b';
+B : 'a' | C;
+C : 'a';
+          "
+        )
+        .unwrap();
+        let sg = pager_stategraph(&grm);
+        let st = StateTable::new(&grm, &sg).unwrap();
+        let conflicts = st.conflicts().unwrap();
+        assert_eq!(conflicts.sr_len(), 1);
+        assert_eq!(conflicts.rr_len(), 1);
+        assert_eq!(
+            conflicts.sr_conflicts().next().unwrap(),
+            &(
+                grm.token_idx("b").unwrap(),
+                grm.rule_to_prods(grm.rule_idx("B").unwrap())[0],
+                StIdx::from(2)
+            )
+        );
+        assert_eq!(
+            conflicts.rr_conflicts().next().unwrap(),
+            &(
+                grm.rule_to_prods(grm.rule_idx("B").unwrap())[0],
+                grm.rule_to_prods(grm.rule_idx("C").unwrap())[0],
+                StIdx::from(2)
+            )
+        );
     }
 
     #[test]
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -159,6 +159,11 @@ fn main() {
         }
     };
 
+    if let Some(c) = stable.conflicts() {
+        println!("{}", c.pp(&grm));
+        println!("Stategraph:\n{}\n", sgraph.pp_core_states(&grm));
+    }
+
     {
         let rule_ids = grm
             .tokens_map()
--- a/lrpar/Cargo.toml
+++ b/lrpar/Cargo.toml
@@ -29,3 +29,4 @@ vob = "2.0"
 regex = "1.0"
 
 [dev-dependencies]
+temp_testdir = "0.2"


ltratt:
>My main question would be, where the user would be expected to handle these errors. For example within the calc examples, would that be in main.rs or build.rs?

The user can handle these wherever they want, which means whenever they have access to a statetable. At the moment, unfortunately, there's an asymmetry between `RTParserBuilder` (where the user explicitly builds statetables themselves and can thus do what you've done with nimbleparse) and `CTParserBuilder` (where the statetable is built for them and where they can't get access to the conflicts). We need to think what to do about the latter.

My initial suggestion is the following. First we add a `bool` to `CTParserBuilder` called something like `error_on_conflicts` which defaults to true. If set to true, the user loads a grammar in with conflicts, `process_file_in_src` will return an `Err` of some sort. If set to false, `process_file_in_src` will set (new) attributes `sr_conflicts` and `rr_conflicts` in `CTParserBuilder` to whatever `stable` returned, and the user can access them (or ignore them) as they wish.

>Or am I off track again, and we want to allow the user to handle the conflict as it happens and they then can manually change how the parser generator resolves those conflicts?

I don't think we need to worry about that sort of thing yet (maybe not ever!).

ptersilie:
> My initial suggestion is the following. First we add a bool to CTParserBuilder called something like error_on_conflicts which defaults to true. If set to true, the user loads a grammar in with conflicts, process_file_in_src will return an Err of some sort. If set to false, process_file_in_src will set (new) attributes sr_conflicts and rr_conflicts in CTParserBuilder to whatever stable returned, and the user can access them (or ignore them) as they wish.

Yes, that sounds sensible. I had something similar in mind, minus the `error_on_conflicts` bool.

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -555,6 +555,26 @@ where
         }
     }
 
+    /// Returns the name of a given Symbol `sym`.
```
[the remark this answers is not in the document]
ptersilie: I moved the function into the place where it's used. Rather than documenting why it panics, maybe we should just pretty print EOF as `"EOF"`?

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -555,6 +555,26 @@ where
...
+            Symbol::Token(tidx) => self.token_name(*tidx).unwrap(),
+            Symbol::Rule(ridx) => self.rule_name(*ridx),
+        }
+    }
+
+    /// Returns the string representation of a given production `pidx`.
+    pub fn prod_to_string(&self, pidx: PIdx<StorageT>) -> String {
+        let mut sprod = String::new();
+        let ridx = self.prod_to_rule(pidx);
+        sprod.push_str(self.rule_name(ridx));
```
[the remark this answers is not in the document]
ptersilie: I added the rule name as otherwise the conflict information is not as helpful. I also checked that your above example still works in `cfgrammar`.

cfgrammar/src/lib/yacc/grammar.rs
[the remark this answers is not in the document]
ltratt: No, we can't give it any name, because that might clash with a user defined name for another token (and then confusion will ensue!).

cfgrammar/src/lib/yacc/grammar.rs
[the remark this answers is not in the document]
ltratt: Perfect!

lrtable/src/lib/statetable.rs
```diff
@@ -96,9 +96,9 @@ pub struct StateTable<StorageT> {
     prods_len: PIdx<StorageT>,
     tokens_len: TIdx<StorageT>,
     /// The number of reduce/reduce errors encountered.
-    pub reduce_reduce: u64,
+    pub reduce_reduce: Vec<(PIdx<StorageT>, PIdx<StorageT>)>,
```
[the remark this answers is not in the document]
ptersilie: Then it's probably not worth the hassle. The production + rule name should be enough to find the location within the grammar manually. I'll just add the state for now.

cfgrammar/src/lib/yacc/grammar.rs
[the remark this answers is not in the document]
ptersilie: That is true.

lrtable/src/lib/statetable.rs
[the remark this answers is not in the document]
ltratt: OK.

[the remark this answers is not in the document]
ltratt: BTW, there is a `pp_core_states` function on `StateGraph` which might be useful for you?

cfgrammar/src/lib/yacc/grammar.rs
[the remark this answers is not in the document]
ptersilie: Anyways, now that I moved this into `prod_to_string`, this can't happen, as EOF can't be a symbol within a production, right? So I can add an `unreachable!()` for that case.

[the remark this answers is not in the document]
ltratt: Correct.

lrtable/src/lib/statetable.rs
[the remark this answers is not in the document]
ptersilie: Yes I've seen that. Maybe we should allow users to save that to a file if they want to look deeper into a certain conflict?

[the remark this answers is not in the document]
ltratt: I don't like the idea of saving to a file, but we might want to expose the string to the user in some way or another.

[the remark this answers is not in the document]
ptersilie: Ok, I guess if we give them the string, they can do whatever they want with it.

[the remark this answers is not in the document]
ltratt: Exactly.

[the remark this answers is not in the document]
ltratt: Perhaps what we need to do is return some sort of "conflicts struct", which, amongst other things, has the stategraph?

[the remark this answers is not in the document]
ltratt: One method called `conflicts` which returns `Option<Conflicts>` would probably be nicer.

ptersilie:
The last commit takes a shot at making grammar conflicts available from within `CTParserBuilder`. This API needs some more work and there's a few things I'm not super happy with, but wanted to discuss them here first.

As discussed this adds `error_on_conflicts` to `CTParserBuilder` which, if set, errors if the grammar contains any conflicts. I wanted to give the user the option here to get some useful information out of the error, instead of just `Error: Grammar contains conflicts`. So the error contains grm, sgraph, and stable, and functions to pretty print the results. The debug print of the error is something like `Error: CTConflictsError(1 Shift/Reduce, 5 Reduce/Reduce)`, though we may want to pretty print the actual conflicts along with the graph here. Suggestions?

If `error_on_conflicts` isn't set, then the `CTParserBuilder` receives references to grm, sgraph, and stable, which the user can then manually query for conflicts and do whatever they want with them, e.g.

```
    let mut ctb = CTParserBuilder::<u8>::new()
        .error_on_conflicts(false)
        .action_kind(ActionKind::CustomAction);
    let lex_rule_ids_map = ctb
        .process_file_in_src("calc.y")?;
    panic!("{}", ctb.pp_conflicts());
```

would print out

```
Found shift/reduce conflicts:
   State 5: Shift("d") / Reduce(E: "c")
Found reduce/reduce conflicts:
   State 10: Reduce(A: "b") / Reduce(B: "a" "b")
```

The change in ctbuilder.rs:358 is also not ideal as we need to clone a string here to be able to move `grm` into `CTParserBuilder`. The alternative would be to wrap most of this function into a block so `actiontype` goes out of focus before we set `self.grm` below.

lrtable/src/lib/statetable.rs
```diff
@@ -428,14 +464,9 @@ where
...
-    /// Return a list of shift/reduce conflicts.
-    pub fn sr_conflicts(&self) -> &Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)> {
-        &self.shift_reduce
-    }
-
-    /// Return a list of reduce/reduce conflicts.
-    pub fn rr_conflicts(&self) -> &Vec<(PIdx<StorageT>, PIdx<StorageT>, StIdx)> {
-        &self.reduce_reduce
+    /// Return a struct containing all conflicts or None if there aren't any.
+    pub fn conflicts(&self) -> &Option<Conflicts<StorageT>> {
```
ltratt: It might be better to make this `Option<&Conflicts<...>>` to give yourself some flexibility in the future. You'd need to change the body to `self.conflicts.as_ref()`.

nimbleparse/src/main.rs
```diff
@@ -162,26 +162,26 @@ fn main() {
...
-    if sr_conflicts.len() > 0 {
-        println!("Found shift/reduce conflicts:");
-        for (tidx, pidx, stidx) in sr_conflicts{
-            println!("   State {:?}: Shift(\"{}\") / Reduce({})",
-                usize::from(*stidx),
-                grm.token_name(*tidx).unwrap(),
-                grm.pp_prod(*pidx));
+    if let Some(c) = stable.conflicts() {
+        if c.sr_len() > 0 {
+            println!("Found shift/reduce conflicts:");
```
ltratt: I'm not sure the "found" bit is useful. Maybe just `"Shift/reduce conflicts:"?

nimbleparse/src/main.rs
```diff
@@ -162,26 +162,26 @@ fn main() {
...
-    if rr_conflicts.len() > 0 {
-        println!("Found reduce/reduce conflicts:");
-        for (pidx, r_pidx, stidx) in rr_conflicts{
-            println!("   State {:?}: Reduce({}) / Reduce({})",
-                usize::from(*stidx),
-                grm.pp_prod(*pidx),
-                grm.pp_prod(*r_pidx));
+        // Report reduce/reduce conflicts
+        if c.rr_len() > 0 {
+            println!("Found reduce/reduce conflicts:");
```
ltratt: Same "found" comment as above.

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -362,8 +362,13 @@ where
     }
 
     /// Get the action return type as defined by the user
-    pub fn actiontype(&self) -> &Option<String> {
-        &self.actiontype
+    pub fn actiontype(&self) -> Option<&String> {
+        if let Some(ref a) = self.actiontype {
```
ltratt: I think this can be `self.actiontype.as_ref()`.

lrpar/examples/calc_actions/build.rs
```diff
@@ -44,6 +44,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
     // enough to fit all IDs in) as well as the input file (which must end in ".y" for lrpar, and
     // ".l" for lrlex).
     let lex_rule_ids_map = CTParserBuilder::<u8>::new()
+        .error_on_conflicts(false)
```
ltratt: There are no conflicts here (I think!), so we don't want to set this to `false` do we?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -73,14 +74,58 @@ pub enum ActionKind {
...
+
+impl<StorageT> fmt::Display for CTConflictsError<StorageT>
+where
+    StorageT: 'static + Debug + Hash + PrimInt + Serialize + TypeName + Unsigned,
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let conflicts = self.stable.conflicts().unwrap();
+        write!(f, "{}", conflicts.pp(&self.grm)) //
```
ltratt: Stray `//`.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -73,14 +74,58 @@ pub enum ActionKind {
...
+}
+
+impl<StorageT> fmt::Display for CTConflictsError<StorageT>
+where
+    StorageT: 'static + Debug + Hash + PrimInt + Serialize + TypeName + Unsigned,
+    usize: AsPrimitive<StorageT>,
+    u32: AsPrimitive<StorageT>
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let conflicts = self.stable.conflicts().unwrap();
```
ltratt: This `unwrap` seems a bit dodgy: if there are no conflicts it'll cause a panic I think?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -73,14 +74,58 @@ pub enum ActionKind {
...
+        write!(f, "{}", conflicts.pp(&self.grm)) //
+    }
+}
+
+impl<StorageT> fmt::Debug for CTConflictsError<StorageT>
```
ltratt: Can we not `#[derive(Debug)]` something that's good enough for debug support?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -150,6 +197,69 @@ where
...
+        if let Some((g, _, _)) = &self.error_data {
+            Some(&g)
+        }
+        else {
+            None
+        }
+    }
+
+    pub fn sgraph(&self) -> Option<&StateGraph<StorageT>> {
+        if let Some((_, sg, _)) = &self.error_data {
```
ltratt: `self.error_data.as_ref()[1]` or similar might work here.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -150,6 +197,69 @@ where
...
+        if let Some((_, _, st)) = &self.error_data {
+            Some(&st)
+        }
+        else {
+            None
+        }
+    }
+
+    /// Returns a pretty-printed version of the graph.
+    pub fn pp_sgraph(&self) -> String {
```
ltratt: Can't someone do `sgraph().pp_core_states(&grm)`? I'm not sure we need this function.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -150,6 +197,69 @@ where
...
+        if let Some((ref grm, ref sgraph, _)) = self.error_data {
+            sgraph.pp_core_states(&grm)
+        }
+        else {
+            "".to_string()
+        }
+    }
+
+    /// Returns a pretty-printed version of the conflicts.
+    pub fn pp_conflicts(&self) -> String {
```
ltratt: Similarly, this seems redundant.

ltratt:
A few high-level comments:

* `as_ref` is a wonderful function on `Option` :)
* We need tests for this stuff.
* I'm not convinced that we want so many methods on `CTBuilder` exposing stuff to the outside: the fewer we have, the lower the chance of API breakage in the future IMHO.

With regards to that last point, I think I'd prefer to only expose a `conflicts()` function with a signature along the lines of:

```rust
fn conflicts(self) -> ConflictsStruct
```

so that it consumes the `CTBuilder` (otherwise what happens if you process two files? Do the old conflicts get thrown away?). Then the `ConflictsStruct` would, minimally, have functions like `pp_sgraph` (but probably not `sgraph`).

Alternatively we could even have the `ConflictsStruct` returned from `process_file`, along the lines of:

```rust
    pub fn process_file<P, Q>(
        &self,
        inp: P,
        outd: Q
    ) -> Result<(HashMap<String, StorageT>, Option<ConflictsStruct>), Box<Error>>
    where
        P: AsRef<Path>,
        Q: AsRef<Path>
```

making it clear that the conflicts only relate to one file.

lrtable/src/lib/statetable.rs
```diff
@@ -465,8 +491,13 @@ where
     }
 
     /// Return a struct containing all conflicts or None if there aren't any.
-    pub fn conflicts(&self) -> &Option<Conflicts<StorageT>> {
-        &self.conflicts
+    pub fn conflicts(&self) -> Option<&Conflicts<StorageT>> {
+        if let Some(c) = &self.conflicts {
```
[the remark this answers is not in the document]
ptersilie: I didn't know about `as_ref` but it's exactly what I needed in this case and I was thinking to myself that there must be a better way.

lrpar/src/lib/ctbuilder.rs
ptersilie: They can, but then they need to retrieve `grm` first from the `Conflicts` struct as it's needed as an argument. I thought this was more user friendly.

lrpar/src/lib/ctbuilder.rs
ptersilie: `CTConflictsError` only gets created if there are actual any errors, so `stable.conflicts` can not be `None` here.

lrpar/examples/calc_actions/build.rs
ptersilie: Yes, I thought I'll leave it in as an example, but I'm happy to remove it.

ptersilie:
> as_ref is a wonderful function on Option :)
I didn't know about it, but it's definitely what I wanted.

> I'm not convinced that we want so many methods on CTBuilder exposing stuff to the outside: the fewer we have, the lower the chance of API breakage in the future IMHO.
Agreed, I also think the fewer methods the better, but wasn't sure what needed to be exposed.

> Alternatively we could even have the ConflictsStruct returned from process_file, along the lines of
I think I like this solution the most, as then we won't have functions that return `None` or empty strings, if there weren't any errors.

lrpar/src/lib/ctbuilder.rs
ltratt: Aha!

lrpar/examples/calc_actions/build.rs
ltratt: I want to encourage people not to do `error_on_conflicts(false)` unless they know they have conflicts (e.g. the Lua grammar), so I think we should remove this.

lrpar/src/lib/ctbuilder.rs
ptersilie: Then we would also have to add `#[derive(Debug)]` to `YaccGrammar` and `StateTable`, and I'm not sure what the output would then be.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -238,12 +355,12 @@ where
         // Header
         let mod_name = inp.as_ref().file_stem().unwrap().to_str().unwrap();
         let actiontype = match grm.actiontype() {
-            Some(t) => t,
+            Some(t) => t.clone(), // We need to clone here so we can move `grm` at the end of the function
```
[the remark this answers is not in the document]
ptersilie: Shall I leave the comment in then, so we know to remove this once this is the case?

lrpar/src/lib/ctbuilder.rs
ltratt: We have to implement `Debug` because of `Error` I guess?

lrpar/src/lib/ctbuilder.rs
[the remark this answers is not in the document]
ltratt: You could maybe change it to something like "Probably unneeded once NLL is in stable"?

lrpar/src/lib/ctbuilder.rs
ptersilie: I'm not sure what you mean. How I see it we have 3 options:
1) We use `#[derive(Debug)]`, which means we also have to add to `YaccGrammar` and `StateTable`, and when this error is returned it prints out the entire grammar and statetable. Which IMO is a bit too much information.

2) We leave this is at is, printing out something like `CTConflictsError(1 Shift/Reduce, 5 Reduce/Reduce)` as a debug message.

3) We update this to print out the stategraph + the actual errors encountered, resulting in something like:
```
0:  [^ -> . Expr, {'$'}]
    Term -> 2
    Factor -> 4
    Expr -> 3
    'INT' -> 5
    'LBRACK' -> 1
1:  [Factor -> 'LBRACK' . Expr 'RBRACK', {'PLUS', 'MUL', 'RBRACK', '$'}]
    Expr -> 6
    'LBRACK' -> 1
    'INT' -> 5
    Factor -> 4
    Term -> 2
...

Shift/Reduce conflicts:
   State 2: Shift("PLUS") / Reduce(Term: "Term")
Reduce/Reduce conflicts:
   State 2: Reduce(Expr: "Term") / Reduce(Term: "Term")
   State 2: Reduce(Expr: "Term") / Reduce(Term: "Term")
   State 11: Reduce(Term: "Factor" "MUL" "Term") / Reduce(Term: "Term")
   State 11: Reduce(Term: "Factor" "MUL" "Term") / Reduce(Term: "Term")
   State 11: Reduce(Term: "Factor" "MUL" "Term") / Reduce(Term: "Term")

```

I'm probably tending to the last one, since then when the user sets `error_on_conflicts(true)`, they immediately get all the info they need to debug this error. Otherwise they have to first set `error_on_conflicts(false)` and then retrieve the `ConflictsStruct` from the result, and then pretty print it using a `panic!`.

ltratt: I guess what I'm asking is: do we need to implement `Debug` at all?

ptersilie: It seems it's needed by `Error`:
```
impl<StorageT> Error for CTConflictsError<StorageT>                                                                                                                                     
    |                ^^^^^ `ctbuilder::CTConflictsError<StorageT>` cannot be formatted using `{:?}` 
```

ltratt: Right, OK, that was my guess. Then I'm more-or-less OK with this as-is. Should it be `CTConflictsError{{{} Shift/Reduce, {} Reduce/Reduce}}` to make it look more like other debug output?

ptersilie: Can do. What's your thoughts on my suggestion above to output the graph and more detailed conflicts instead, so the user doesn't need to jump through so many hoops to get to this information.

ltratt: I think `Debug` output should me minimal, so I'd prefer not to output the graph etc.

ptersilie: Yes, I agree `Debug` should be minimal. I've moved the pretty printed stuff into the `Display` method, though I'm not sure if this is the correct place for it either.

ptersilie:
See if that new commit makes things better.

lrpar/src/lib/ctbuilder.rs
ltratt: I must admit, I generally make `Debug` and `Display` the same; and if someone wants the full pretty-printed output I'd make them call `pp`. But this is something over which reasonable people differ.

lrpar/examples/calc_actions/build.rs
```diff
@@ -43,8 +43,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
     // Note that we specify the integer type (u8) we'll use for token IDs (this type *must* be big
     // enough to fit all IDs in) as well as the input file (which must end in ".y" for lrpar, and
     // ".l" for lrlex).
-    let lex_rule_ids_map = CTParserBuilder::<u8>::new()
-        .error_on_conflicts(false)
+    let (lex_rule_ids_map, _cs) = CTParserBuilder::<u8>::new()
```
ltratt: Just `_` is enough.

lrpar/examples/calc_parsetree/build.rs
```diff
@@ -43,7 +43,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
     // Note that we specify the integer type (u8) we'll use for token IDs (this type *must* be big
     // enough to fit all IDs in) as well as the input file (which must end in ".y" for lrpar, and
     // ".l" for lrlex).
-    let lex_rule_ids_map = CTParserBuilder::<u8>::new().process_file_in_src("calc.y")?;
+    let (lex_rule_ids_map, _cs) = CTParserBuilder::<u8>::new().process_file_in_src("calc.y")?;
```
ltratt: `_`

lrpar/src/lib/ctbuilder.rs
```diff
@@ -78,21 +78,40 @@ pub enum ActionKind {
...
 
-struct CTConflictsError<StorageT: Eq + Hash> {
+
+pub struct ConflictsStruct<StorageT: Eq + Hash> {
     pub grm: YaccGrammar<StorageT>,
     pub sgraph: StateGraph<StorageT>,
     pub stable: StateTable<StorageT>
 }
 
+impl<StorageT> ConflictsStruct<StorageT>
```
ltratt: In retrospect, this is a weird name. Shouldn't it just be `Conflicts`?

ltratt:
Some minor comments, but I think we're heading in the right direction.

lrpar/src/lib/ctbuilder.rs
ptersilie: Yes this was kind of a placeholder, I forgot to ask what we should name this. `Conflicts` is already the name of the `Conflicts` struct within `StateTable`. Which is why I wasn't sure how to name this.

lrpar/src/lib/ctbuilder.rs
ptersilie: Okay, looks go with your way then to be consistent.

ptersilie: If they want a prettyprinted version for the error, they can just call `error.cs.pp()`.

lrpar/src/lib/ctbuilder.rs
ltratt: That's another file and we don't import that struct in here I think, so reusing the name isn't a total disaster.

lrpar/src/lib/ctbuilder.rs
ltratt: Works for me.

lrpar/src/lib/ctbuilder.rs
ptersilie: Won't that be confusing from a documentation perspective. We'd have a struct `Conflicts`, which has a `StateTable`, which has a `Conflicts`. But the two `Conflicts` are not the same.

ptersilie: Maybe I name the one in StateTable `GrammarConflicts`

ltratt: They'll never appear on the same page...

I'm not saying it's ideal, but we can't add `Struct` as a suffix and I can't think of a better name...

ptersilie: See my above suggestion to rename the `StateTable` one to `GrammarConflicts`.

ltratt: They're all grammar conflicts in one way...

If you really want to separate the names, maybe make `StateTable`'s struct be `STConflicts`?

ptersilie: Alternatively we can name the CTParserBuilder one `CTConflicts`?

ptersilie:
I pushed a commit that addresses the comments, and one that adds a `Conflicts` test to `StateTable`. I'll look into adding `CTParserBuilder` tests for conflicts tomorrow.

lrtable/src/lib/statetable.rs
```diff
@@ -947,6 +947,44 @@ mod test {
...
+        assert_eq!(
+            itersr.next().unwrap(),
+            &(
+                grm.token_idx("b").unwrap(),
+                grm.rule_to_prods(grm.rule_idx("B").unwrap())[0],
+                StIdx::from(2)
+            )
+        );
+        let mut iterrr = conflicts.rr_conflicts();
+        assert_eq!(
```
ltratt: Would it be easier to say `assert_eq!(conflicts.rr_conflicts().next().unwrap(), ...);`?

ptersilie:
I added tests for `CTParserBuilder`. Creating those tests made me realise that it'sa bit tedious to retrieve the data from `CTConflictsError` since it gets returned as a `Box<Error>`. Do we need to use a boxed type here? If so, do you have an idea how we can make it easier for the user to get the data out of the error?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -618,3 +618,67 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
...
+    use self::temp_testdir::TempDir;
+    use super::{CTParserBuilder, ActionKind, CTConflictsError};
+
+    #[test]
+    fn test_conflicts() {
+
+        let temp = TempDir::default();
+        let mut file_path = PathBuf::from(temp.as_ref());
+        file_path.push("grm.y");
+        let mut f = File::create(file_path.clone()).unwrap();
```
ltratt: I think you can do `File::create(&file_path)` since the argument is `AsRef`.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -618,3 +618,67 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
...
+        let _ = f.write_all("%start A
+%%
+A : 'a' 'b' | B 'b';
+B : 'a' | C;
+C : 'a';".as_bytes());
+
+        match CTParserBuilder::<u8>::new()
```
ltratt: Once you merge in master you can (and probably will have to) remove the `<u8>` annotation here.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -618,3 +618,67 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
...
+            .error_on_conflicts(false)
+            .action_kind(ActionKind::GenericParseTree)
+            .process_file_in_src(file_path.to_str().unwrap()) {
+            Ok((_, conflicts)) => {
+                assert_eq!(conflicts.as_ref().unwrap().stable.conflicts().unwrap().sr_len(), 1);
```
ltratt: Addressing https://github.com/softdevteam/grmtools/pull/55#pullrequestreview-181277687 (assuming it's possible) will remove at least one `as_ref`.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -618,3 +618,67 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
...
+        let _ = f.write_all("%start A
+%%
+A : 'a' 'b' | B 'b';
+B : 'a' | C;
+C : 'a';".as_bytes());
+
+        match CTParserBuilder::<u8>::new()
+            .action_kind(ActionKind::GenericParseTree)
+            .process_file_in_src(file_path.to_str().unwrap()) {
+            Ok((_, _)) => {
```
ltratt: `Ok(_)` should be OK here.

ltratt:
You only need to return `Box<Error>` if you can return multiple different error types. In this case, since you can only return conflicts, I doubt you need the `Box`.

ptersilie:
Shall we get rid of the `Box<Error>` then?

ltratt:
If the complexity isn't needed... off with its head!

lrpar/src/lib/ctbuilder.rs
ptersilie: I'm not sure what you mean. Your link is related to statetable, here we don't have statetable but `process_in_src` returns an `Option<Conflicts>` which is moved (we can't return a reference, since the value will be out of scope once we return from the function).

ptersilie:
Hmm, looks like `Box` is needed: `he size for values of type `(dyn std::error::Error + 'static)` cannot be known at compilation time `

lrpar/src/lib/ctbuilder.rs
ltratt: Mea culpa!

However, this API is borderline unusable: `as_ref().unwrap().stable.conflicts().unwrap().sr_len()` is almost impossible to remember. Can we do *something* to make things easier? One thing might be to not expose `stable`, but instead expose things like `sr_len`? But that might cause us the same duplication-of-methods problem we had earlier in the PR...

ptersilie: I can give the `CTConflictsError` struct some functions, at the very least `fn conflicts() -> Conflicts`.

ltratt: That would certainly help!

ltratt:
Right, so when I looked, `process_file_in_src` can currently return multiple different types of errors. But I don't think anyone in their right mind will do:

```rust
        match CTParserBuilder::<u8>::new()
            .action_kind(ActionKind::GenericParseTree)
            .process_file_in_src(file_path.to_str().unwrap()) {
            Ok(_) => {
                panic!("Expected error")
            }
            Err(e) => {
                let cs = e.downcast_ref::<CTConflictsError<u8>>();
                assert_eq!(cs.unwrap().cs.stable.conflicts().unwrap().sr_len(), 1);
                assert_eq!(cs.unwrap().cs.stable.conflicts().unwrap().rr_len(), 1);
            }
        }
```

so we don't have to worry too much. Anyone with half a brain will do `error_on_conflicts(false)` and access the conflicts via the `Ok` case.

ptersilie:
Right. I pushed some more commits addressing your comments (935d10d) and merging in master and applying the necessary changes (1f35d6f).

lrpar/src/lib/ctbuilder.rs
```diff
@@ -97,6 +98,10 @@ where
         s.push_str(&self.stable.conflicts().unwrap().pp(&self.grm));
         s
     }
+
+    pub fn conflicts(&self) -> &STConflicts<StorageT> {
```
ltratt: Should this be `stconflicts`?

ltratt: If it's confusing, it might be worth trying a different configuration of methods to find something better. I generally find my second attempt at an API is better than the first. And I've recently sometimes had to have three attempts to find something good...

ptersilie: Okay, I've tried to improve it a little bit.

ltratt:
Dumb question: what's in `Conflicts` and what's in `ConflictsData`? Maybe there's a way to squeeze them down into one struct (e.g. perhaps there are fields like `grm: Option<&YaccGrammar>` in one of the structs?

Warning: I'm not saying this is a good idea, but I'm just trying to think of alternatives.

ptersilie:
`ConflictsData` just groups the graph, statetable, and grammar into one struct to return it in `process_in_src`. I guess we could also just return a tuple, but I'm not sure that makes it better. The advantage of `ConflictsData` is that it gives us some functions like `pp`.

ltratt:
I know what the two structs contain :) My point was more that the names don't mean anything to users and that's why I was wondering if we could combine them e.g. somethinglike:

```rust
struct Conflicts {
    reduce_reduce: Vec<(PIdx<StorageT>, PIdx<StorageT>, StIdx)>,
    shift_reduce: Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)>
    grm: Option<YaccGrammar<StorageT>>,
    sgraph: Option<StateGraph<StorageT>>,
    stable: Option<StateTable<StorageT>>
}
```

BTW, in general we probably don't want `pub` elements on structs: in general, we should provide a method to access things (it gives us a bit more flexibility in the future).

ptersilie:
The problem with that is, that `Conflicts` lives on `StateTable`, so having `StateTable` within the `Conflicts` struct wouldn't make sense.

ltratt:
In that case, `stable` would be `None`? I'm not saying that that struct is necessarily a good idea, but can we do *something* to improve things?

ptersilie:
Alright, I've tried to simplify the API a bit. See if this goes in the right direction!

ltratt:
Can you show me an example of the new API in use? It's too hard for me to guess just from the diff.

ptersilie:
The tests (which also have been updated in the commit) should show it's used.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -691,16 +670,17 @@ C : 'a';"
...
+        let mut ct = CTParserBuilder::new()
             .error_on_conflicts(false)
-            .action_kind(ActionKind::GenericParseTree)
-            .process_file_in_src(file_path.to_str().unwrap())
-        {
-            Ok((_, conflicts)) => {
-                assert_eq!(conflicts.as_ref().unwrap().conflicts().sr_len(), 1);
-                assert_eq!(conflicts.as_ref().unwrap().conflicts().rr_len(), 1);
+            .action_kind(ActionKind::GenericParseTree);
+        let _ = ct.process_file_in_src(file_path.to_str().unwrap());
```
ltratt: `let _ = ` can be removed.

ptersilie: Not without creating a warning tho.

ltratt:
I can only see the `CTParserBuilder` API being tested?

lrpar/src/lib/ctbuilder.rs
ltratt: You can fix it in nicer ways than this ;)

ptersilie:
What other API do you mean?

ltratt:
Mea culpa! I meant: how does it look from `StateTable`?

ptersilie:
The `StateTable` API hasn't changed since the last time. Though I did find another `as_ref` opportunity that I just fixed and pushed.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -691,16 +670,17 @@ C : 'a';"
...
+
+        match ct.conflicts() {
+            Some((_, _, _, conflicts)) => {
```
ltratt: The `_`s are grm, sgraph, stable respectively?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -256,6 +235,13 @@ where
         self
     }
 
+    pub fn conflicts(&self) -> Option<(&YaccGrammar<StorageT>, &StateGraph<StorageT>, &StateTable<StorageT>, &Conflicts<StorageT>)> {
+        if let Some((ref grm, ref sgraph, ref stable)) = self.cdata {
```
ltratt: Can you simplify this to `if let Some((grm, sgraph, stable)) = &self.cdata`? I'm not sure.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -256,6 +235,13 @@ where
...
+            return Some((&grm, &sgraph, &stable, &stable.conflicts().unwrap()));
```
ltratt: I'm not sure you need the `&` on `grm` and `sgraph`.

lrpar/src/lib/ctbuilder.rs
ptersilie: Yes, I was thinking about putting `_grm`, etc, but last time I did this you told me `_` is enough.

lrpar/src/lib/ctbuilder.rs
ptersilie: Looks like I can! :-)

lrpar/src/lib/ctbuilder.rs
ptersilie: And you are right again! :-)

lrpar/src/lib/ctbuilder.rs
ltratt: I didn't mean you should change them: I just wanted to check.

ptersilie:
Addressed some more comments. Hope it's getting there.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -173,6 +232,27 @@ where
...
     }
 
+    /// If set to true, `process_file_in_src` will return an error if the given grammar contains
+    /// any Shift/Reduce or Reduce/Reduce conflicts. Defaults to `true`.
+    pub fn error_on_conflicts(mut self, b: bool) -> Self {
+        self.error_on_conflicts = b;
+        self
+    }
+
+    pub fn conflicts(
```
ltratt: I'm inclined to think that this API should be explicitly marked as unstable until we're sure it's in the final form we want. There isn't an off-the-shelf way to do this, but we could add an `unstable` feature to the crate (as suggested by https://github.com/rust-lang/rfcs/issues/1491#issuecomment-345455136). Thoughts?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -565,3 +656,70 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
     let stable = deserialize(stable_buf).unwrap();
     (grm, sgraph, stable)
 }
+
+#[cfg(test)]
+mod test {
+
```
ltratt: Extraneous blank line.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -79,14 +79,71 @@ pub enum ActionKind {
...
+    StorageT: Eq + Hash
+{
     // Anything stored in here almost certainly needs to be included as part of the rebuild_cache
     // function below so that, if it's changed, the grammar is rebuilt.
     recoverer: RecoveryKind,
     phantom: PhantomData<StorageT>,
-    actionkind: ActionKind
+    actionkind: ActionKind,
+    error_on_conflicts: bool,
+    cdata: Option<(
```
ltratt: `cdata` is kind of a weird name: it certainly needs to be explicitly documented what it's for, and pointed out that it's not part of the cache. You'll need to update that comment above too slightly.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -475,6 +564,8 @@ where
 
         let mut f = File::create(outp_rs)?;
         f.write_all(outs.as_bytes())?;
+
+        self.cdata = Some((grm, sgraph, stable));
```
ltratt: Maybe it would make more sense to have something like:

```rust
struct CTParserBuilder {
 ...
 conflicts: Option<(YaccGrammar, StateGraph, StateTable, Conflicts)>
```

and this line would be:

```rust
if let Some(c) = stable.conflicts() {
    self.conflicts = Some((grm, sgraph, stable, c));
}
```

and then `fn conflicts` would just  be `self.conflicts`.

lrtable/src/lib/statetable.rs
```diff
@@ -307,6 +369,14 @@ where
         let actions_sv = SparseVec::<usize>::from(&actions, 0, usize::from(grm.tokens_len()));
         let gotos_sv = SparseVec::<usize>::from(&gotos, 0, usize::from(grm.rules_len()));
 
+        let mut conflicts = None;
+        if reduce_reduce.len() > 0 || shift_reduce.len() > 0 {
```
ltratt: You want to use `is_empty` otherwise Clippy will complain...

lrtable/src/lib/statetable.rs
```diff
@@ -478,21 +552,21 @@ fn resolve_shift_reduce<StorageT: 'static + Hash + PrimInt + Unsigned>(
     off: usize,
     tidx: TIdx<StorageT>,
     pidx: PIdx<StorageT>,
-    stidx: StIdx
-) -> u64
-where
+    stidx: StIdx, // State we want to shift to
+    shift_reduce: &mut Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)>,
+    state: StIdx // State in which the conflict occured
```
ltratt: Ouch, this is a terrible name given `stidx` also exists! We have to come up with something that's not so much of a duplicate. Maybe `shift_stidx` and `conflict_stidx` or something like that?

lrpar/src/lib/ctbuilder.rs
ptersilie: Do you have an idea for a better name? Maybe just `data`?

ltratt: In a later comment, I sort-of suggest renaming this to `conflicts` (for other reasons).

lrpar/src/lib/ctbuilder.rs
ptersilie: Since `c` is a reference, I would have to add lifetimes, which isn't really worth it for a private variable that can only be accessed via `fn conflicts` anyways. So I just renamed `cdata` to `conflicts`.

lrtable/src/lib/statetable.rs
ptersilie: I changed it. But now I'm wondering if this is a similar problem and needs addressing: https://github.com/softdevteam/grmtools/blob/f8e5a1d9d8e58faf1a05dd370064ebb653548498/lrtable/src/lib/statetable.rs#L216

ltratt: That's less bad, because `stidx` is `STIdx` and `state` is an entirely different type.

lrpar/src/lib/ctbuilder.rs
ltratt: Fair enough.

ptersilie:
See if this fixes things.

lrpar/src/lib/ctbuilder.rs
ptersilie: I have come to the opinion that this may be too confusing, and I rather just leave a comment in the documentation that this function is unstable.

ltratt: Fair enough.

lrtable/src/lib/statetable.rs
```diff
@@ -370,7 +370,7 @@ where
         let gotos_sv = SparseVec::<usize>::from(&gotos, 0, usize::from(grm.rules_len()));
 
         let mut conflicts = None;
-        if reduce_reduce.len() > 0 || shift_reduce.len() > 0 {
+        if reduce_reduce.is_empty() || shift_reduce.is_empty() {
```
ltratt: Should be `!is_empty` right?

lrtable/src/lib/statetable.rs
```diff
@@ -554,7 +554,7 @@ fn resolve_shift_reduce<StorageT: 'static + Hash + PrimInt + Unsigned>(
     pidx: PIdx<StorageT>,
     stidx: StIdx, // State we want to shift to
     shift_reduce: &mut Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)>,
-    state: StIdx // State in which the conflict occured
+    shift_stidx: StIdx // State in which the conflict occured
```
ltratt: This variable name seems odd, given the comment on `stidx`. Can we come up with something better?

lrtable/src/lib/statetable.rs
ltratt: Ah, this was fixed in a later commit.

ptersilie: Yes, see my `whoops` commit. ;-)

lrtable/src/lib/statetable.rs
ptersilie: I was just following your suggestion, though I do agree this can be somewhat confusing. I guess your other suggestion `conflict_stidx` it is then.

lrtable/src/lib/statetable.rs
```diff
@@ -554,7 +554,7 @@ fn resolve_shift_reduce<StorageT: 'static + Hash + PrimInt + Unsigned>(
     pidx: PIdx<StorageT>,
     stidx: StIdx, // State we want to shift to
     shift_reduce: &mut Vec<(TIdx<StorageT>, PIdx<StorageT>, StIdx)>,
-    shift_stidx: StIdx // State in which the conflict occured
+    conflicts_stidx: StIdx // State in which the conflict occured
```
ltratt: Is it `conflicts_idx` plural or just `conflict_idx`?

ptersilie: Yes, should probably be singular. But I think it should still be `conflict_stidx` as it is the id of the state the conflict occurs in.

ltratt: `conflict_stidx` works for me.

ptersilie:
Renamed `conflicts_stidx`. This should be all now I think.

ltratt:
Now that we've hit 150 comments, I think it's time to squash :)

ptersilie:
I think so too. I'm getting quite a few conflicts when I squash everything down, so this is gonna take me a bit.

ptersilie:
Rebase and squashed. When the tests pass, I think we are ready to go!

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -557,6 +557,28 @@ where
...
+    /// Returns the string representation of a given production `pidx`.
+    pub fn pp_prod(&self, pidx: PIdx<StorageT>) -> String {
+        let mut sprod = String::new();
+        let ridx = self.prod_to_rule(pidx);
+        sprod.push_str(self.rule_name(ridx));
+        sprod.push_str(":");
+        for sym in self.prod(pidx) {
+            let s = match sym {
+                Symbol::Token(tidx) => {
+                    if *tidx == self.eof_token_idx() {
```
ltratt: The `if` statement is unnecessary: just `self.token_name(*tidx).unwrap()` is needed (it'll `panic` if it hits `eof_token_idx`).

lrpar/src/lib/ctbuilder.rs
```diff
@@ -173,6 +233,29 @@ where
...
     }
 
+    /// If set to true, `process_file_in_src` will return an error if the given grammar contains
+    /// any Shift/Reduce or Reduce/Reduce conflicts. Defaults to `true`.
+    pub fn error_on_conflicts(mut self, b: bool) -> Self {
+        self.error_on_conflicts = b;
+        self
+    }
+
+    /// Returns the grammar conflicts found during generation of the parser, or `None` if there
```
ltratt: Should this be something like "If the grammar had conflicts, return a tuple which allows users to see what they were; or `None` if there were no conflicts"?

lrtable/src/lib/statetable.rs
```diff
@@ -424,6 +493,11 @@ where
             None => unreachable!()
         }
     }
+
+    /// Return a struct containing all conflicts or None if there aren't any.
```
ltratt: `None` needs backticking.

nimbleparse/src/main.rs
```diff
@@ -159,6 +159,11 @@ fn main() {
         }
     };
 
+    if let Some(c) = stable.conflicts() {
+        println!("Stategraph:\n{}\n", sgraph.pp_core_states(&grm));
```
ltratt: Should we say something like "Conflicts found: {} shift/reduce {} reduce/reduce\n  {}" here? Otherwise the user might not realise what the stategraph is for?

ltratt:
Doing a final read through I spotted a few minor things. Won't take more than a minute or two to fix.

Once that's done and checked, we should expand the main commit's message: it would be nice to have a brief example in the commit of the API in use (for `StateTable` and `CTParserBuilder`), because it's a bit hard to find it in the diff.

nimbleparse/src/main.rs
ptersilie: Hmm, I think the next line already says something like
```
Shift/Reduce conflicts:
   State: 1: Shift(2) / Reduce(E: T)
```

So we'd mention "Shift/Reduce conflicts" twice. Are you still fine with that?

ptersilie: Or maybe we just swap the lines, and put the conflicts before the graph?

ltratt: Swapping the lines works for me!

ptersilie:
Addressed your comments. See if that works for you.

ltratt:
Works for me! Please squash (and also please edit the commit message along the lines suggested a bit earlier). Almost there!

ptersilie:
Squashed and also edited the commit message. See what you think!

PR #56, opened 2018-11-29
Fix actions and parse errors

ltratt:
I wanted to do more in this PR, but it's already getting quite big, and the next stage is going to take me a little while. So it's probably best to get this in now as it already makes things somewhat better.

After a few minor commits, the meat of this PR is in https://github.com/softdevteam/grmtools/commit/9f2961e7939dc237dad725869d934f1b147038fd. This means that we no longer get hard crashes when error recovery occurs on grammars with actions, and it moves us towards (but doesn't get all the way) to a more powerful actions API. The commit message goes into a lot more detail. Finally, https://github.com/softdevteam/grmtools/commit/c1e352dd0ba8318772e79083971a6b5a6005dfa7 is a small but useful commit which gives us more flexibility going forwards.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -301,7 +301,7 @@ where
         match self.actionkind {
             ActionKind::CustomAction => {
                 // action function references
-                outs.push_str(&format!("\n        let mut actions: Vec<Option<&Fn(RIdx<{storaget}>, &str, vec::Drain<AStackType<{actiont}, {storaget}>>) -> {actiont}>> = Vec::new();\n",
+                outs.push_str(&format!("\n        let mut actions: Vec<Option<&Fn(RIdx<{storaget}>, &Lexer<{storaget}>, vec::Drain<AStackType<{actiont}, {storaget}>>) -> {actiont}>> = Vec::new();\n",
                     storaget=StorageT::type_name(),
                     actiont=actiontype)
                 );
@@ -318,10 +318,9 @@ where
                 }
                 outs.push_str(&format!(
                     "
-        let s = lexer.input().to_string();
         RTParserBuilder::new(&grm, &sgraph, &stable)
             .recoverer(RecoveryKind::{})
-            .parse_actions(lexer, &actions, &s)\n",
+            .parse_actions(lexer, &actions)\n",
                     recoverer,
                 ));
             }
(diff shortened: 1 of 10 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -363,24 +365,19 @@ where
...
-                            let num = match &s[m.start() + 1..m.end()].parse::<usize>() {
-                                Ok(val) => val - 1,
-                                Err(_) => unreachable!()
-                            };
+                        for i in 0..grm.prod(pidx).len() {
                             outs.push_str(&format!(
-                                "    let {prefix}arg_{} = match {prefix}args[{}] {{",
-                                num + 1,
-                                num,
+                                "    let {prefix}arg_{} = match {prefix}args.next().unwrap() {{",
```
ptersilie: I like this! The drain is nicer than the reusable Vector.

ptersilie: One question though. Doesn't that generate variables that might not be used, creating `unused variable` warnings when compiling things?

lrpar/src/lib/parser.rs
```diff
@@ -231,35 +252,30 @@ where
...
                     // Process actions
-                    if let Some((actions, ref mut astack, input)) = actiondata {
-                        action_vec.clear();
-                        action_vec.extend(astack.drain(pop_idx - 1..));
-                        if let Some(f) = actions[usize::from(pidx)] {
-                            astack.push(AStackType::ActionType(f(input, &action_vec)));
-                        }
+                    if let Some(f) = self.actions[usize::from(pidx)] {
+                        let v = AStackType::ActionType(f(ridx, input, astack.drain(pop_idx - 1..)));
+                        astack.push(v);
```
ptersilie: Should we squash these two lines into one?

ltratt: Normally I'd agree, but the next commit in the PR reorders this, so it's kind of moot in this case.

lrpar/src/lib/parser.rs
```diff
@@ -331,25 +347,29 @@ where
...
-                    if let Some(ref mut tstack_uw) = *tstack {
-                        let nodes = tstack_uw
-                            .drain(pop_idx - 1..)
-                            .collect::<Vec<Node<StorageT>>>();
-                        tstack_uw.push(Node::Nonterm { ridx, nodes });
+                    if let Some(ref mut astack_uw) = *astack {
+                        if let Some(f) = self.actions[usize::from(pidx)] {
+                            let v = AStackType::ActionType(f(
+                                ridx,
+                                "", // XXX,
```
ptersilie: What does the `XXX` stand for?

ptersilie:
The commit message of 9f2961e is missing stuff after `That means that a grammar rule such as:`.

lrpar/src/lib/parser.rs
ltratt: Good question. I should have explained this a bit better in the commit message (it's hinted at in the last paragraph): these are points where the API wanted us to pass around the entire user input, which wasn't practical. These XXXs disappear in https://github.com/softdevteam/grmtools/pull/56/commits/c1e352dd0ba8318772e79083971a6b5a6005dfa7

ltratt:
> The commit message of 9f2961e is missing stuff

Oops! When you've finished reviewing things I'll fix that.

lrpar/src/lib/parser.rs
ptersilie: Fair.

lrpar/src/lib/parser.rs
ptersilie: I've just read the final commit where it became clear these were placeholders. Probably not worth removing them then.

ptersilie:
Rest looks fine. Looks like only the commit message needs fixing.

ltratt:
Shall I force push a fix?

ptersilie:
Yes please!

ltratt:
Updated.

ptersilie:
Need to squash anything?

ltratt:
Nope, merge as-is please!

PR #58, opened 2018-12-03
Allow errors to percolate upwards in the actions API

ltratt:
This PR contains two major things.

First it provides a way to differentiate between lexemes of zero length (e.g. `DEDENT` tokens in Python) and lexemes created by error recovery (which have no content) in https://github.com/softdevteam/grmtools/commit/3db2be06e5c0980d272739dee322360007195c86.

Second it changes the actions API in two related ways in https://github.com/softdevteam/grmtools/commit/faee83c26bea71c0040ccbf47194bb88cec523fc: it now passes `Lexeme`s to actions (instead of strings); and it passes `Lexeme`s in an a `Result<Lexeme, Lexeme>` type, allowing users to easily tell if a lexeme is the result of user input (the `Ok` case) or error recovery (the `Err` case).

When combined with the ability to use arbitrary types for `%type` https://github.com/softdevteam/grmtools/commit/46c4265ebdf20d41dc0cd068105dbad4d34175ab, this allows users a lot of flexibility to write flexible actions, but also makes it relatively painless to do so. For example, here's the new calculator grammar:

```
start Expr
%type Result<u64, ()>
%%
Expr: Term 'PLUS' Expr { Ok($1? + $3?) }
    | Term { $1 }
    ;

Term: Factor 'MUL' Term { Ok($1? * $3?) }
    | Factor { $1 }
    ;

Factor: 'LBRACK' Expr 'RBRACK' { $2 }
      | 'INT' {
            let l = $1.map_err(|_| ())?;
            match $lexer.lexeme_str(&l).parse::<u64>() {
                Ok(v) => Ok(v),
                Err(_) => {
                    let (_, col) = $lexer.line_and_col(&l);
                    eprintln!("Error at column {}: '{}' cannot be represented as a u64",
                              col,
                              $lexer.lexeme_str(&l));
                    Err(())
                }
            }
        }
      ;
```

and here's a session using it:

```
>>> 1+2
Result: 3
>>> 1++2
Parsing error at column 3.
Result: 3
>>> 1+18446744073709551616
Error at column 3: '18446744073709551616' cannot be represented as a u64
Unable to evaluate
>>> 1+18446744073709551616+18446744073709551616
Error at column 3: '18446744073709551616' cannot be represented as a u64
Error at column 24: '18446744073709551616' cannot be represented as a u64
Unable to evaluate
```

In essence, we give users the choice for how to handle the semantic consequences of syntactic errors. Interestingly, my first attempt at this embedded the `Resulte<u64, ()>` idiom into `lrpar`, before I realised that we can let the user do it themselves just as easily.

There is one part of `lrpar`'s API which this PR makes even worse: the return type of `parse_action` is a mess, requiring the user to deal with too many cases. Fixing that is a medium sized job and best left to a subsequent PR IMHO. [It might also be a good idea to provide easy functions for the user to pretty print things like repair sequences, but that might turn out to be bloat. I'll experiment.]

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -256,13 +256,14 @@ where
         outs.push_str(&format!("mod {}_y {{\n", mod_name));
         outs.push_str(
             "    use lrpar::{{Lexer, LexParseError, RecoveryKind, RTParserBuilder}};
-    use lrpar::ctbuilder::_reconstitute;"
+    use lrpar::ctbuilder::_reconstitute;
+"
         );
 
         match self.actionkind {
             ActionKind::CustomAction => {
                 outs.push_str(&format!(
-                    "use lrpar::parser::AStackType;
+                    "use lrpar::{{Lexeme, parser::AStackType}};
     use cfgrammar::RIdx;
     use std::vec;
 
@@ -378,28 +379,34 @@ where
                         actiont = actiontype
                     ));
                     for i in 0..grm.prod(pidx).len() {
-                        outs.push_str(&format!(
-                            "    let {prefix}arg_{} = match {prefix}args.next().unwrap() {{",
-                            i + 1,
-                            prefix = ACTION_PREFIX
-                        ));
                         match grm.prod(pidx)[i] {
-                            Symbol::Rule(_) => outs.push_str(
+                            Symbol::Rule(_) => outs.push_str(&format!(
                                 "
+    let {prefix}arg_{} = match {prefix}args.next().unwrap() {{
         AStackType::ActionType(v) => v,
         AStackType::Lexeme(_) => unreachable!()
-    };
-"
-                            ),
-                            Symbol::Token(_) => outs.push_str(&format!(
-                                "
-        AStackType::ActionType(_) => unreachable!(),
-        AStackType::Lexeme(ref l) => {prefix}lexer.lexeme_str(l)
     }};
 ",
+                                i + 1,
+                                prefix = ACTION_PREFIX
+                            )),
+                            Symbol::Token(_) => outs.push_str(&format!(
+                                "
+    let {prefix}arg_{}: Result<Lexeme<_>, Lexeme<_>> = match {prefix}args.next().unwrap() {{
+        AStackType::ActionType(_) => unreachable!(),
+        AStackType::Lexeme(l) => {{
+            if l.len().is_some() {{
+                Ok(l)
+            }} else {{
+                Err(l)
+            }}
+        }}
+    }};
+",
+                                i + 1,
                                 prefix = ACTION_PREFIX
                             ))
-                        };
+                        }
                     }
                     if let Some(s) = grm.action(pidx) {
                         // Replace $1 ... $n with the correct local variable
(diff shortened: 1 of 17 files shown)


lrpar/src/lib/lex.rs
```diff
@@ -69,18 +83,23 @@ impl<StorageT: Copy> Lexeme<StorageT> {
...
+        } else {
+            None
+        }
     }
 
-    pub fn is_empty(&self) -> bool {
-        self.len == 0
+    /// Length in bytes of the lexeme, or `None` if this lexeme is the result of error recovery.
+    pub fn len(&self) -> Option<usize> {
+        if self.len == u32::max_value() {
```
ptersilie: Can this restriction have any negative consequences to the user, e.g. when they have a lexeme of exactly u32 length? Should we document this somewhere?

ltratt: Yes, it definitely is restrictive: they can only have lexemes of (4 gigabytes - 1 byte). I must admit that I don't think we'll hit that case for a while and, if/when we do, we can transparently change the storage of `Lexeme` to store more bits. It's one of those "I should fix this, but it's low priority" things.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -398,11 +402,19 @@ where
                         };
                     }
                     if let Some(s) = grm.action(pidx) {
-                        let ns = re.replace_all(
-                            s,
-                            format!("{prefix}arg_$1", prefix = ACTION_PREFIX).as_str()
+                        // Replace $1 ... $n with the correct local variable
```
ptersilie: This will also overwrite any `$X`s inside of strings which can restrict the user from some reasonable actions. Probably something we'll have to address in the future.

ltratt: Agreed, but fixing this perfectly would be incredibly hard -- close to impossible (we'd have to parse arbitrary Rust fragments, which is hard, given that Rust doesn't have a formal grammar). I think we might just have to document this as "gotcha" somewhere?

ptersilie:
Two comments, rest looks good.

lrpar/src/lib/lex.rs
ptersilie: I agree, it's unlikely someone will run into this, but you never know. Isn't the fix just changing `self.len` to `Option` as well? Or am I missing something?

ltratt: Changing it to `Option` will add a whole extra machine word to this struct, which is pretty punishing for the case of generic parse trees. The good news is that, if someone does create an input with a single lexeme of 4GiB, it will at least crash gracefully at the first possible opportunity.

lrpar/src/lib/ctbuilder.rs
ptersilie: I quickly checked, and Yacc ignores `$X` within strings, so we should at least document this somewhere. My first thought at a somewhat ok fix for this would be to count unescaped quotes and only replace `$X` if the count is even. It's not enough, but it might be better than nothing?

lrpar/src/lib/lex.rs
ptersilie: Ah! I hadn't considered the memory implication of that. I agree that in this case it's not worth it then.

lrpar/src/lib/ctbuilder.rs
ltratt: Agreed, we could do that. Do you want to file an issue and tackle it in another PR? I don't really want to make this PR any more complex than it already is!

ptersilie: Sounds like a plan!

ptersilie:
Merge as is?

ltratt:
Yes please!

PR #70, opened 2018-12-13
Point at the new location for the github.io page.

ltratt:
[We need to wait until the github.io pages are pushed. If the URLs in here are correct, then this PR can be merged. If not, I need to fix things!]


ptersilie:
You forgot to update the error recovery link. Should we also link `lrpar`?

ltratt:
See if that commit fixes things for you.

I don't think it's worth making every lrpar/lrlex etc. link clickable (particularly when someone's already in the lrpar section of the book).

ptersilie:
Looking good. If you squash I will merge as soon as the tests pass.

ltratt:
Squashed.

PR #71, opened 2018-12-17
Quoted epp strings

ltratt:
This PR was originally intended to fix a bug which meant that `%epp` statements like:

```
%epp '"string"'
```

caused a compile-time error (https://github.com/softdevteam/grmtools/commit/87a16197f54d9bdba130991d4ba2bba5333105f9). But I then realised that we can make the testing of compile-time code generation more robust so that we don't have to rely on bunging tests in the `calc_actions` example. So this PR separates out a new way of running full compile-time code generation in tests. The new file `tests/ct.rs` is a bit hard to read, but it does the job. The tests in there aren't run normally but are run if you do:

```
$ cargo test -- --ignored
```

though note that doesn't run the normal tests! So if you want to run *all* the tests you need to do:

```
$ cargo test
$ cargo test --ignored
```

which is what is now in `.travis.yml`.

### DIFF
--- a/lrpar/tests/ct.rs
+++ b/lrpar/tests/ct.rs
@@ -71,7 +71,9 @@ where
 
     // If we're the last thread to be waiting on DUMMY_TESTDIR, then drop it.
     if DUMMY_WAITING.fetch_sub(1, Ordering::SeqCst) == 1 {
-        unsafe { DUMMY_TESTDIR = None; }
+        unsafe {
+            DUMMY_TESTDIR = None;
+        }
     }
     DUMMY_SPINLOCK.store(false, Ordering::SeqCst);
     match r {


lrpar/tests/ct.rs
```diff
@@ -7,19 +7,73 @@
...
+    sync::atomic::{AtomicBool, Ordering},
+    thread,
+    time::Duration
 };
 
 use self::tempfile::TempDir;
 
-fn create_dummy() -> TempDir {
-    let tdir = TempDir::new().unwrap();
+// This is ugly, but effective: compiling a create dummy project is slow because we have to
```
ptersilie: `compiling a dummy project`?

ltratt: Oops!

lrpar/tests/ct.rs
```diff
@@ -7,19 +7,73 @@
...
+// compile dependecies. We thus cheat and only create one dummy project, which we continually
```
ptersilie: s/dependecies/dependencies

ltratt: Oops!

ptersilie:
Only some typos. Rest looks good.

ltratt:
I've fixed the typos in one commit and, in a second commit, added a named constant `SPIN_WAIT` (I meant to do this earlier but forgot). If you're OK with those two commits, I can squash.

lrpar/tests/ct.rs
```diff
@@ -24,11 +24,11 @@ use std::{
 
 use self::tempfile::TempDir;
 
-// This is ugly, but effective: compiling a create dummy project is slow because we have to
-// compile dependecies. We thus cheat and only create one dummy project, which we continually
-// refresh, leaving only its target/ directory behind. In order to do that, we use the
-// `DUMMY_RUNNING` bool as a simple spin-lock to prevent more than one test trying to make use of
-// the dummy test directory at once.
+// This is ugly, but effective: compiling a dummy project is slow because we have to compile
+// dependecies. We thus cheat and only create one dummy project, which we continually refresh,
```
ptersilie: You forgot to fix `dependencies`.

ptersilie: Ah never mind. You fixed it in the other commit.

ptersilie:
Looks good. Please squash!

ltratt:
Squashed but... I've just noticed that `lazy_static` doesn't call `Drop`, so our test directory is not being deleted after tests are run :/ I think I should try and fix that before we try merging this PR if you're OK with that.

ptersilie:
Sure, I can wait. :-)

ltratt:
OK, that was a bit more fiddly than I expected: we can't use `lazy_static`, but I hacked together an equivalent that allows us to GC `DUMMY_TESTDIR` whenever there are no tests queued which want to use it.

ptersilie:
Looks good, please squash!

PR #78, opened 2018-12-29
Faster deduplication of repair sequences.

ltratt:
I found a real world (if, admittedly, completely bonkers) example of broken Java input which leads to hundreds of thousands of repair sequences being found at one point. The previous deduplicator spent over 15 minutes (maybe longer... I gave up!) doing its business. This commit changes things to use a HashSet which gets the time down sufficiently that everything (parsing, printing the tree, etc. etc.) is about 8 seconds. The key change is https://github.com/softdevteam/grmtools/pull/78/files#diff-6bb37983b3729c36a792027b9c0b1dbaL627.

[I did briefly try sorting then deduping, but that would require us to implement Ord on ParseRepair which felt more artificial than implementing Hash. Not that it makes a huge difference in the long run I suppose.]

### DIFF
--- a/lrpar/src/lib/cpctplus.rs
+++ b/lrpar/src/lib/cpctplus.rs
@@ -447,7 +447,7 @@ fn ends_with_parse_at_least_shifts<StorageT: PrimInt + Unsigned>(
 
 #[cfg(test)]
 mod test {
-    use std::fmt::Debug;
+    use std::{fmt::Debug, hash::Hash};
 
     use cfgrammar::yacc::YaccGrammar;
     use num_traits::{AsPrimitive, PrimInt, ToPrimitive, Unsigned};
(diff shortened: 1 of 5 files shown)


ptersilie:
Well, that was my only comment anyways, so rest looks good. Will merge once the tests pass.

ptersilie:
I also just merged the other open PR, so you'll need to do a rebase before we can merge this.

ltratt:
I was a very naughty boy with the 5000ms thing -- sorry!

ltratt:
I was thoroughly wrong about the cause of the test failure. It's actually something far more interesting. https://github.com/softdevteam/grmtools/pull/78/commits/f61c47d224ebb57ab4c65ada018e8ecf145b10c2 has more details.

ltratt:
Squashed.

PR #79, opened 2018-12-31
Add a %avoid_insert directive.

ltratt:
**THIS IS A REQUEST FOR COMMENTS: PLEASE DON'T MERGE YET**

With the calc grammar and the input "2++3" there are two possible repair sequences:

```
  Insert Int
  Delete +
```

If you're building a parse tree, both are equally reasonable. However, if you're executing actions, "Insert Int" isn't great, because there is no value we can insert, so we stop execution. In contrast, "Delete" allows us to continue execution as normal. Since our current ranking is non-deterministic, that means that we end up with this surprising outcome (from a real calc_actions session):

```
  >>> 2++3
  Parsing error at line 1 column 3. Repair sequences found:
     1: Insert INT
     2: Delete +
  Unable to evaluate expression.
  >>> 2++3
  Parsing error at line 1 column 3. Repair sequences found:
     1: Delete +
     2: Insert INT
  Result: 5
```

The problem is that, in general, we probably prefer to Insert new content than delete content, on the basis that there is more value in what the user entered than what they didn't enter. However, as the above example shows, tokens whose value we care about are not ideal when recovering from errors. Unfortunately, we have no idea which tokens have important values or not.

This commit adds a new grammar-level directive `%avoid_insert` which takes one or more token names. When ranking repairs, repair sequences which contain an Insert that references a token in `%avoid_insert` are penalised. If we add `%avoid_insert 'INT'` to the calc grammar, then it is guaranteed that the repair sequences above are always returned in the following order:

```
  Delete +
  Insert Int
```

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -35,6 +35,7 @@ pub enum YaccParserErrorKind {
     UnknownDeclaration,
     DuplicatePrecedence,
     PrecNotFollowedByToken,
+    DuplicateAvoidInsertDeclaration,
     DuplicateImplicitTokensDeclaration,
     DuplicateStartDeclaration,
     DuplicateTypeDeclaration,
@@ -67,6 +68,9 @@ impl fmt::Display for YaccParserError {
             YaccParserErrorKind::UnknownDeclaration => "Unknown declaration",
             YaccParserErrorKind::DuplicatePrecedence => "Token already has a precedence",
             YaccParserErrorKind::PrecNotFollowedByToken => "%prec not followed by token name",
+            YaccParserErrorKind::DuplicateAvoidInsertDeclaration => {
+                "Duplicate %avoid_insert declaration"
+            }
             YaccParserErrorKind::DuplicateImplicitTokensDeclaration => {
                 "Duplicate %implicit_tokens declaration"
             }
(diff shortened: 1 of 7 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -67,6 +68,9 @@ impl fmt::Display for YaccParserError {
             YaccParserErrorKind::UnknownDeclaration => "Unknown declaration",
             YaccParserErrorKind::DuplicatePrecedence => "Token already has a precedence",
             YaccParserErrorKind::PrecNotFollowedByToken => "%prec not followed by token name",
+            YaccParserErrorKind::DuplicateAvoidInsertDeclaration => {
+                "Duplicate %avoid_insert declaration"
```
ptersilie: Do we need to disallow multiple `%avoid_insert`s? Given it's a HashSet anyways, we could just merge them all together, similar to the `%token` declarations. This gives users more options to format their grammars (e.g. if they have loads of tokens they want to avoid, they don't need to be all in a single line), and has as far as I can tell no disadvantage.

ltratt: I sort-of agree, but `%implicit_tokens` has the same restriction. Perhaps we should loosen both?

ptersilie: I don't see a reason why not. But we can do this in separate PR.

ltratt: Agreed, that sounds like a plan. Let's do both in one go?

ptersilie:
This all sounds perfectly reasonable to me, so I don't have anything to say that speaks against these changes.

ltratt:
Then I guess we can merge!

ptersilie:
Do you need to squash anything?

ltratt:
Good point. Squashed!

PR #80, opened 2019-01-02
Allow duplicate %implicit_tokens and %avoid_insert definitions

ptersilie:
This commit loosens the definition of the %implicit_token and %avoid_insert directives. Previouly, using the directives multiple times would result in a duplication error. However, there is no disadvantage in letting the user define those directives multiple times and allowing it instead gives the user more options to format the grammar by grouping things togther. Similary to the %token directive, multiple definitions are just merged together.


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -189,15 +176,18 @@ impl YaccParser {
                     avoid_insert.insert(n);
                     i = self.parse_ws(j, true)?;
                 }
-                self.ast.avoid_insert = Some(avoid_insert);
+
+                if let Some(ref mut ai) = self.ast.avoid_insert {
```
ltratt: Might be simpler as something like:

```rust
if self.ast.avoid_insert.is_none() {
    self.ast.avoid_insert = Some(HashSet::new())
}
self.ast.avoid_insert.extend(avoid_insert);
```
?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -207,7 +197,13 @@ impl YaccParser {
                         implicit_tokens.insert(n);
                         i = self.parse_ws(j, true)?;
                     }
-                    self.ast.implicit_tokens = Some(implicit_tokens);
+                    if let Some(ref mut it) = self.ast.implicit_tokens {
```
ltratt: Same comment as above.

ltratt:
We should decide what we want to happen with:

```
%avoid_insert T T
```

i.e. duplicates. Ignore? Error? I suspect the latter is safer; if so, we'll also need a corresponding test.

cfgrammar/src/lib/yacc/parser.rs
ptersilie: I did try that but was getting:
```
error[E0507]: cannot move out of borrowed content                                                                                                                                                                                                             
   --> cfgrammar/src/lib/yacc/parser.rs:181:21                                                                                                                                                                                                                
    |                                                                                                                                                                                                                                                         
181 |                     self.ast.avoid_insert.unwrap().extend(avoid_insert);                                                                                                                                                                                
    |                     ^^^^ cannot move out of borrowed content 
```

ltratt: That error suggests you need `self.ast.avoid_insert.as_ref().unwrap().extend(...)`.

ptersilie: But I can replace the for loop. I actually didn't see HashSet had an extend method and tried to use `union` instead, which didn't work.

ptersilie:
I've just added duplicate detection as per your comment above. See if that addresses your comments.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -170,40 +178,52 @@ impl YaccParser {
                 let mut avoid_insert = HashSet::new();
                 i = self.parse_ws(j, false)?;
                 let num_newlines = self.newlines.len();
+                if self.ast.avoid_insert.is_none() {
+                    self.ast.avoid_insert = Some(HashSet::new());
+                }
                 while j < self.src.len() && self.newlines.len() == num_newlines {
                     let (j, n) = self.parse_token(i)?;
                     self.ast.tokens.insert(n.clone());
+                    if avoid_insert.contains(&n)
```
ltratt: I reckon we should remove the `avoid_insert` variable and just stuff things in `self.ast.avoid_insert` directly, which will simplify this `if`.

ltratt:
One simplification suggestion, but otherwise, I think this is heading in the right direction.

cfgrammar/src/lib/yacc/parser.rs
ptersilie: Good idea!

ptersilie:
Changes pushed!

ltratt:
Please squash.

ptersilie:
Squashed!

ptersilie:
Now that #81 is merged I will have to rebase this. Is it okay if I do this now?

PR #81, opened 2019-01-02
Stop lrlex from creating identifiers from tokens that are invalid in Rust

ltratt:
A simple lex rule such as:

```
  = "="
```

would lead to the following Rust code being generated:

```
  pub const T_= = <some int>;
```

which, of course, is completely invalid.

This commit doesn't generate `T_` variables for token names that can't be Rust identifiers. Many users won't use these variables so will never notice the difference: and those who do want to use this feature will have to use Rust identifier-friendly token names!

### DIFF
--- a/lrlex/src/lib/builder.rs
+++ b/lrlex/src/lib/builder.rs
@@ -20,6 +20,7 @@ use std::{
 };
 
 use num_traits::{PrimInt, Unsigned};
+use regex::Regex;
 use try_from::TryFrom;
 use typename::TypeName;
 
@@ -29,6 +30,10 @@ use parser::parse_lex;
 const LEX_SUFFIX: &str = "_l";
 const LEX_FILE_EXT: &str = "l";
 const RUST_FILE_EXT: &str = "rs";
+
+lazy_static! {
+    static ref RE_TOKEN_ID: Regex = Regex::new(r"^[a-zA-Z_][a-zA-Z_0-9]*$").unwrap();
+}
 
 /// A `LexerBuilder` allows one to specify the criteria for building a statically generated
 /// lexer.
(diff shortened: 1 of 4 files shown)


lrlex/src/lib/builder.rs
```diff
@@ -169,12 +174,14 @@ where
         // Token IDs
         if let Some(ref rim) = self.rule_ids_map {
             for (n, id) in rim {
-                outs.push_str(&format!(
-                    "#[allow(dead_code)]\npub const T_{}: {} = {:?};\n",
-                    n.to_ascii_uppercase(),
-                    StorageT::type_name(),
-                    *id
-                ));
+                if RE_TOKEN_ID.is_match(n) {
```
ptersilie: Will this just silently ignore my tokens if I used weird names? In that case we might want to throw an error message or something so people are not confused if something doesn't work as expected.

ltratt: It does silently ignore them, but I'm not really sure what else to sensibly do: unless you're using generic parse trees, you pretty much don't care about these. And if you use generic parse trees, you'll very quickly realise why there isn't a variable called `T=`.

ptersilie: Can you remind me again what these `T_` variables are used for? So it won't happen that the program compiles but behaves weirdly because we ignore variables with weird names?

ltratt: They tell you the `TIdx` of a token name. If you have a generic parse tree and want to (say) check if a terminal is an ID, you need to use these. Otherwise you can ignore them.

ptersilie: Ah I see. In that case I agree that there is probably no need to raise an error. I will merge as is then.

Issue #75, opened 2018-12-18
Should CTParserBuilder::action_kind be CTParserBuilder::actionkind?


ltratt:
The underscore in there seems a bit of an odd choice, in retrospect, especially given the struct is `ActionKind`. Any thoughts @ptersilie?

ptersilie:
Don't really have a preference. Happy to change it to `actionkind` though.

PR #82, opened 2019-01-03
Tidyup YaccKinds and co.

ltratt:
This is very much a "PR that will make later PRs easier", but it's probably big enough to make it worth considering individually.

The main aim of the PR is to solve the weird dichotomy we had between YaccKind (Original or Eco?) and ActionKind (parse tree, user action, no action?). This felt like some sort of weird feature matrix, except sometimes you could specify YaccKind and not ActionKind (or vice versa), making it simply confusing. https://github.com/softdevteam/grmtools/commit/6238a48deb1ac918595c38440e8a9c3961a285b6 is thus the most important commit, though https://github.com/softdevteam/grmtools/commit/6ec897526bfd8fe137b2ec1db247592e65acad4d is a useful follow-up (I kept these two commits separate to make reviewing easier: both commits pass all tests in their own right). This indirectly fixes https://github.com/softdevteam/grmtools/issues/75 by virtue of removing `action_kind` entirely! [I should add that I'm not sure `YaccOriginalActionKind` is a great name due to its verbosity. However, it does have the virtue of being accurate. Thoughts welcome!]

There are then a couple of secondary aspects to the PR.

https://github.com/softdevteam/grmtools/commit/1b6fe62214cbedd9e0588a09cf41d933f5ad5603 does a proper job of a hack in Yacc ASTs which will prove useful to us soon.

Finally, https://github.com/softdevteam/grmtools/commit/c34676293a47800fb3e8042c8ccc8d259a854a3d makes the compile-time generation code slightly less unreadable.

### DIFF
--- a/cfgrammar/src/lib/yacc/firsts.rs
+++ b/cfgrammar/src/lib/yacc/firsts.rs
@@ -20,7 +20,7 @@ use TIdx;
 /// `Firsts` stores all the first sets for a given grammar. For example, given this code and
 /// grammar:
 /// ```text
-///   let grm = YaccGrammar::new(YaccKind::Original, "
+///   let grm = YaccGrammar::new(YaccKind::Original(YaccOriginalActionKind::GenericParseTree), "
 ///     S: A 'b';
 ///     A: 'a'
 ///      | ;").unwrap();
@@ -156,7 +156,7 @@ where
 mod test {
     use super::YaccFirsts;
     use num_traits::{AsPrimitive, PrimInt, Unsigned};
-    use yacc::{YaccGrammar, YaccKind};
+    use yacc::{YaccGrammar, YaccKind, YaccOriginalActionKind};
 
     fn has<StorageT: 'static + PrimInt + Unsigned>(
         grm: &YaccGrammar<StorageT>,
@@ -193,7 +193,7 @@ mod test {
     #[test]
     fn test_first() {
         let grm = YaccGrammar::new(
-            YaccKind::Original,
+            YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
             &"
           %start C
           %token c d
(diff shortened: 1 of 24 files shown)


ltratt:
I should have thought of this earlier: https://github.com/softdevteam/grmtools/pull/82/commits/181a264c5c8429879d3015580478ab8e6e448560 renames `%type` to `%actiontype`. I think this better matches the `ActionT` naming inside grmtools, and it makes it clearer what effect the type in question has on affairs. That said, maybe this is one change too many: if you don't like it, we can back it out.

cfgrammar/src/lib/yacc/firsts.rs
```diff
@@ -20,7 +20,7 @@ use TIdx;
 /// `Firsts` stores all the first sets for a given grammar. For example, given this code and
```
ptersilie: Typo in the commit message: `requires a a %actiontype declaration.`

cfgrammar/src/lib/yacc/mod.rs
```diff
@@ -23,8 +23,19 @@ pub use self::{
...
-    Original,
+    /// [Johnson](http://dinosaur.compilertools.net/yacc/index.html),
+    Original(YaccOriginalActionKind),
     /// The variant used in the [Eco language composition editor](http://soft-dev.org/src/eco/)
     Eco
 }
+
+#[derive(Clone, Copy)]
+pub enum YaccOriginalActionKind {
+    /// Execute user-specified actions attached to each production; also requires a a %actiontype
```
ptersilie: Same typo as commit message.

ltratt: Oops!

ltratt:
The commit message typo will need a rebase to fix, so I'll wait for you to OK that before doing so.

ptersilie:
The rest looks good. Ready for rebase and squash!

ltratt:
Squashed and reworded.

PR #83, opened 2019-01-06
Allow different rules to have different action types.

ltratt:
This falls under the "why didn't this occur to me earlier?" heading.

Our previous %actiontype solution is, more-or-less, useless in practise. It works in C because the union type that's created allows C programmers to treat chunks of memory as dynamically typed (happily seg-faulting if the programmer fished the wrong thing out of the union). Consider this (highly contrived) grammar:

```
  %start S
  %%
  S: A | S A;
  A: 'a';
```

i.e. match one or more "a" characters. Let's assume that, for each "a" we match we want to produce some value. How should we write this? Our only option is "%actiontype Vec<...>" because "S" must return a Vec, even though "A" only sensibly returns a singular value. In this case, this feels inelegant, but it works.

However, what happens if you've got a real programming language building up an AST? Consider a simple language which allow assignments and expressions:

```
  Assignments: Assignment | Assignments Assignment;
  Assignment: "ID" "=" Expr
  Expr: "INT" "+" Expr | "INT"
```

"Assignments" should return "Vec<Assign>" and "Expr" should return an "Expr". What should our %actiontype be now? Well, we can make an enum:

```rust
  enum AssignsOrExprs {
    Assigns(Vec<Assign>),
    Expr(Expr)
  }
```

and then we can have "%actiontype AssignsOrExprs". But we now have to scale this up to every AST type in our entire program which is horrible, and also turns our seemingly statically typed Rust into dynamically typed code: the number of "match" statements doesn't bear thinking about!

How do other grammar generators deal with this? Well they do the obvious (in retrospect) thing of allowing each different rule's actions to return different types (note: each production in a rule must return the same type!). That's what this commit does. Interestingly, this requires a simple tweak to cfgrammar; a big change to ctbuilder.rs; and not much else.

First we add `YaccKind::Grmtools`, a new variant on Yacc syntax. `%actiontype` is not valid in this grammar type. Instead, each rule in this grammar type *must* have a type after its name:

```
  %start S
  %%
  S::Vec<A>: A { vec![$1] }
    | S A {
        $1.push($2);
        $1
    }
    ;
  A::A: 'a' { A } ;
  %%
  pub struct A;
```

How does this work? Before we translated each production's action into a rule which (for the first production in the grammar above) looked roughly like this:

```rust
  fn action_prod0(..., args: Drain<...>) -> Vec<A> {
     let arg1 = match args.next().unwrap() {
        AStackType::ActionType(x) => x,
        _ => unreachable!()
     };
     vec![arg1]
  }
```

The crucial detail now is that we know the types of all rules in advance. So we generate an enum of all action types, and translate each action into a wrapper and an action:

```rust
  enum ActionsKind {
    AKS(Vec<A>),
    AKA(A)
  }

  fn wrapper_prod0(..., args: Drain<...>) -> Vec<A> {
    let arg1 = match args.next().unwrap() {
      AStackType::ActionType(x) => x,
      _ => unreachable!()
    };
    action_prod0(arg1)
  }

  fn action_prod0(..., mut arg1: A) -> Vec<A> {
    vec![arg1]
  }

  fn wrapper_prod1(..., args: Drain<...>) -> Vec<A> {
    let arg1 = match args.next().unwrap() {
      AStackType::ActionType(x) => x,
      _ => unreachable!()
    };
    let arg2 = match args.next().unwrap() {
      AStackType::ActionType(x) => x,
      _ => unreachable!()
    };
    action_prod0(arg1, arg2)
  }

  fn action_prod1(..., mut arg1: Vec<A>, mut arg2: A) -> Vec<A> {
    arg1.push(arg2);
    arg1
  }

  fn wrapper_prod2(..., args: Drain<...>) -> Vec<A> {
    action_prod2()
  }

  fn action_prod2(...) -> A {
    A
  }
```

The cunning thing about this is that we don't have to change the way the parser works: it receives "action" functions that are wrappers, all of which have the same function signature (i.e. from the parser's perspective it's a bit like we wrote `%actiontype ActionsKind`). The wrapper functions then unpack the Drain, and call the "actual" action functions which contain user code. The actual action functions have their function's arguments and return types statically typed, so Rust statically guarantees that you can't, for example, mix your Classes and Imports. This does incur a mild additional overhead because of the ActionsKind enum (one extra machine word, at worst, per ActionType we're holding), but not enough to hugely worry us. And now we can write grammars which
generate ASTs with ease!

As a happy bonus, I realised that we can make arguments to action functions be mutable (hence the `mut arg1: Vec<A>` above), which makes doing thing like flattening lists a lot more ergonomic.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -31,14 +31,14 @@ pub struct GrammarAST {
     pub implicit_tokens: Option<HashSet<String>>,
     // Error pretty-printers
     pub epp: HashMap<String, String>,
-    pub programs: Option<String>,
-    pub actiontype: Option<String>
+    pub programs: Option<String>
 }
 
 #[derive(Debug)]
 pub struct Rule {
     pub name: String,
-    pub pidxs: Vec<usize> // index into GrammarAST.prod
+    pub pidxs: Vec<usize>, // index into GrammarAST.prod
+    pub actiont: Option<String>
 }
 
 #[derive(Debug, Eq, PartialEq)]
@@ -126,17 +126,17 @@ impl GrammarAST {
             avoid_insert: None,
             implicit_tokens: None,
             epp: HashMap::new(),
-            programs: None,
-            actiontype: None
+            programs: None
         }
     }
 
-    pub fn add_rule(&mut self, name: String) {
+    pub fn add_rule(&mut self, name: String, actiont: Option<String>) {
         self.rules.insert(
             name.clone(),
             Rule {
                 name,
-                pidxs: Vec::new()
+                pidxs: Vec::new(),
+                actiont
             }
         );
     }
@@ -279,7 +279,7 @@ mod test {
     fn test_invalid_start_rule() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("B".to_string());
+        grm.add_rule("B".to_string(), None);
         grm.add_prod("B".to_string(), vec![], None, None);
         match grm.complete_and_validate() {
             Err(GrammarValidationError {
@@ -294,7 +294,7 @@ mod test {
     fn test_valid_start_rule() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![], None, None);
         assert!(grm.complete_and_validate().is_ok());
     }
@@ -303,8 +303,8 @@ mod test {
     fn test_valid_rule_ref() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
-        grm.add_rule("B".to_string());
+        grm.add_rule("A".to_string(), None);
+        grm.add_rule("B".to_string(), None);
         grm.add_prod("A".to_string(), vec![rule("B")], None, None);
         grm.add_prod("B".to_string(), vec![], None, None);
         assert!(grm.complete_and_validate().is_ok());
@@ -314,7 +314,7 @@ mod test {
     fn test_invalid_rule_ref() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![rule("B")], None, None);
         match grm.complete_and_validate() {
             Err(GrammarValidationError {
@@ -330,7 +330,7 @@ mod test {
         let mut grm = GrammarAST::new();
         grm.tokens.insert("b".to_string());
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![token("b")], None, None);
         assert!(grm.complete_and_validate().is_ok());
     }
@@ -342,7 +342,7 @@ mod test {
         let mut grm = GrammarAST::new();
         grm.tokens.insert("b".to_string());
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![rule("b")], None, None);
         assert!(grm.complete_and_validate().is_err());
     }
@@ -351,7 +351,7 @@ mod test {
     fn test_invalid_token_ref() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![token("b")], None, None);
         match grm.complete_and_validate() {
             Err(GrammarValidationError {
@@ -366,7 +366,7 @@ mod test {
     fn test_invalid_rule_forgotten_token() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![rule("b"), token("b")], None, None);
         match grm.complete_and_validate() {
             Err(GrammarValidationError {
@@ -381,7 +381,7 @@ mod test {
     fn test_invalid_epp() {
         let mut grm = GrammarAST::new();
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod("A".to_string(), vec![], None, None);
         grm.epp.insert("k".to_owned(), "v".to_owned());
         match grm.complete_and_validate() {
@@ -405,7 +405,7 @@ mod test {
         );
         grm.start = Some("A".to_string());
         grm.tokens.insert("b".to_string());
-        grm.add_rule("A".to_string());
+        grm.add_rule("A".to_string(), None);
         grm.add_prod(
             "A".to_string(),
             vec![token("b")],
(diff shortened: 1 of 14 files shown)


ptersilie:
Two high level comments before I review the code:

1. I'm trying to make sense of the code example in the PR description. I don't understand how `wrapper_prod2` can return `Vec<A>`, if it returns `action_prod2`, which only returns `A`. Also, where does ActionKind come into play here? I assumed that all the wrapper functions would have to return `ActionKind` to be able to have the same signature.
2. The new grammar style is hard to read: `A::A:` is very confusing. Unfortunately, I don't immediately have a better idea, except allowing the user to alternatively (we can allow both) use something like `%ruletype S Vec<A>`, or similar?

ltratt:
1) Oops, that's a typo. It should be `A`. And the example should use `ActionsKind` as follows:

```rust
  enum ActionsKind {
    AKS(Vec<A>),
    AKA(A)
  }

  fn wrapper_prod0(..., args: Drain<...>) -> Vec<A> {
    let arg1 = match args.next().unwrap() {
      AStackType::ActionType(x) => x,
      _ => unreachable!()
    };
    action_prod0(arg1)
  }

  fn action_prod0(..., mut arg1: A) -> Vec<A> {
    vec![arg1]
  }

  fn wrapper_prod1(..., args: Drain<...>) -> Vec<A> {
    let arg1 = match args.next().unwrap() {
      AStackType::ActionType(x) => x,
      _ => unreachable!()
    };
    let arg2 = match args.next().unwrap() {
      AStackType::ActionType(x) => x,
      _ => unreachable!()
    };
    action_prod0(arg1, arg2)
  }

  fn action_prod1(..., mut arg1: Vec<A>, mut arg2: A) -> Vec<A> {
    arg1.push(arg2);
    arg1
  }

  fn wrapper_prod2(..., args: Drain<...>) -> Vec<A> {
    action_prod2()
  }

  fn action_prod2(...) -> A {
    A
  }
```

Sorry about that!

2) I'm open to suggestions on the syntax, though I think moving the `%ruletype` declaration to the prelude would be even worse, as you'd end up scrolling up and down in the file to see what a rule's return type is.

ptersilie:
1. I'm still confused. The new code you posted is identical to the one in the description!
2. Well I guess you can use a split window to have the `%ruletype`s in scope, but I get you point. It's just the excessive use of `:` that I find distracting. Maybe we can use `A<A>: 'a'`?

ltratt:
1. You might need to reload the page. I made a boo-boo but updated it within maybe 10s, but you might have seen the old version.

2. We can't use `A<T>` because, at some point, we might want to allow type parameters and then they'll need to live in the angled brackets. How about removing one colon so you'd have `A:T: ...`?

ptersilie:
1. Aha! That's more what I was expecting! :-)
2. Why are type parameters a problem? You can have `A<Vec<A>>: 'a'`, similar to how it is in Rust. The single colon doesn't improve things much in my opinion. I think we need to separate the type parameter more clearly. I'm also happy with different brackets, e.g. `A[Vec<A>]: 'a'` or `A(Vec<A>): 'a'`. It's not idealy, but even less so is the double colon. But this may just be a personal preference problem.

ltratt:
2. If you have a function, type parameters, and the function's return type are separated, and we need to know which is which. `fn f<T>() -> U` is a simple example. Maybe we could even borrow Rust's syntax and have `A -> T: ...`? Then if we add in type parameters later it would be `A<T> -> U:`.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -31,14 +31,14 @@ pub struct GrammarAST {
     pub implicit_tokens: Option<HashSet<String>>,
```
ptersilie: Reminder to update the code in the commit message before we merge.

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -89,8 +89,8 @@ pub struct YaccGrammar<StorageT = u32> {
     actions: Vec<Option<String>>,
     /// The programs section of a grammar, if specified; otherwise `None`.
     programs: Option<String>,
-    /// The %actiontype declaration value, if specified; otherwise `None`.
-    actiontype: Option<String>,
+    /// The actiontypes of rules (one per rule).
+    actiontypes: Vec<Option<String>>,
```
ptersilie: Why is `String` optional? Don't all rules need to have a type if we are using actions?

ptersilie: Ah, or is this backward compatibility for the original YaccKind.

ltratt: Yes, but some grammar types (e.g. parse trees) don't need any actions. It turns out to be easier to deal with `Vec<Option<String>>` than `Option<Vec<String>>`.

doc/src/errorrecovery.md
```diff
@@ -299,6 +289,47 @@ Result: 2
...
+
+Why does the same input sometimes produce a result and sometimes fail to produce
+a result? The problem is that `2 + + 3` has two repair sequences `Delete +` and
+`Insert Int`. As things stand, both are equally good, and so one is chosen
+non-deterministically. If `Insert Int` is chosen, we hit the `Err` case from
+earlier, and fail to produce a result; if the `Delete` case is chosen, we can
+produce a result.
+
+There is thus a simple mechanism to bias error recovery away from inserting
+lexemes that cause grammar actions problem. `%avoid_insert L` causes grmtools to
```
ptersilie: `that cause the grammar actions problem` or `that cause grammar actions problems`.

cfgrammar/src/lib/yacc/grammar.rs
ptersilie: Ok.

ptersilie:
Just a few typo comments.

ptersilie:
2. After having seen the action type used in action (no pun intended), I don't mind the colons the as much anymore as in the `A::A:` example. My only alternative idea was to use Rust's `->`, which I just saw you already suggested above. So maybe we stick with that then?

doc/src/cfgrammar.md
```diff
@@ -1,5 +1,12 @@
 # `cfgrammar`
 
 `cfgrammar` reads in grammar files, processes them, and provides a convenient
-API for operating with them. It may be of interest to those manipulating
-grammars directly, or who wish to use custom types of parsers.
+API for operating with them. Most users need care about `cfgrammar` only to the
```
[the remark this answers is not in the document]
ltratt: The "to" isn't necessary here (don't ask me why...).

doc/src/errorrecovery.md
ltratt: Hmm, this sentence is a bit awkward isn't it? Let me try rephrasing.

ltratt:
Do you mean move to the `->` notation?

doc/src/cfgrammar.md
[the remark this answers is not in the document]
ptersilie: But it won't hurt. And it sounds bad without it, in my opinion.

ptersilie:
Yes, I meant moving to the `->` notation. I can live with the `::` if you insist on them, but the `->` is probably better.

ltratt:
OK, let me have a go at updating things.

ltratt:
See if that new commit convinces you of `->` one way or the other. [I think I like it.]

ltratt:
Actually, the further doc update would make sense as a separate PR, so let me address your comment above, then we can think about squashing with luck.

doc/src/cfgrammar.md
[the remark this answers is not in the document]
ltratt: OK.

ltratt:
See if those two new commits work for you. If so, I think we're ready to squash?

ptersilie:
It looks like ` doc/src/errorrecovery.md` still has some `::` that need replacing?

ltratt:
Oops!

ltratt:
OK, try that newest commit.

ptersilie:
Looks good. Ready for squashing (and updating the commit message while doing so).

ltratt:
Squashed.

ltratt:
Oops, fixed!

PR #89, opened 2019-01-16
Make a 0.2.1 release.

ltratt:
In one way this is a bit annoying, but since our documentation is sufficiently wrong that it makes using things hard, probably the right thing to do.


ptersilie:
Agreed.

PR #90, opened 2019-01-16
Improve ctbuilder testing

ptersilie:
This is the first attempt at improving the ctbuilder tests. All tests have been moved from ct.rs into a new test suite. The main advantage is that tests don't need to be recompiled each time `cargo test` is run. Defining test grammars is also made easier by using a Yaml file, which avoids escaping as grammars don't need to be defined in strings anymore.

New test grammars can now be defined as *.test-files within `lrpar/test/src/` using Yaml syntax. Tests for these grammars can then be written within `lrpar/tests/src/lib.rs` after importing the compiled lexer and grammar. See `lrpar/tests/src/lib.rs` for examples.


ltratt:
You might want to add `#[allow(dead_code)]` to the code generator(s) to deal with such cases (see `ctbuilder.rs` for examples).

ptersilie:
I'm not really generating any code myself, though (I'm just using `CTParserBuilder` and `LexerBuilder`).

ltratt:
Yes, but in order to remove those warnings (and we can't have warnings in a merged PR except in highly unusual circumstances), you'd have to change `ctbuilder.rs` and the like.

ptersilie:
Ah, I didn't know you were implying I should change `ctbuilder.rs`. I thought you meant I could fix this from within `lrpar/tests`.

ptersilie:
Fixed the warnings and moved the tests around. Have another look.

lrpar/tests/build.rs
```diff
@@ -0,0 +1,57 @@
...
+fn main() -> Result<(), Box<std::error::Error>> {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    for entry in glob("src/*.test")? {
+        match entry {
+            Ok(path) => {
+                if path.is_file() {
+                    // Parse test file
+                    let mut file = fs::File::open(&path)?;
+                    let mut s = String::new();
+                    file.read_to_string(&mut s)?;
```
ltratt: You might prefer:

```rust
let s = fs::read_to_string(&path);
```

lrpar/tests/build.rs
```diff
@@ -0,0 +1,57 @@
...
+                    let grm = &docs[0]["grammar"].as_str().unwrap();
+                    let lex = &docs[0]["lexer"].as_str().unwrap();
+                    let yacckind = match docs[0]["yacckind"].as_str().unwrap() {
+                        "NoAction" => YaccKind::Original(YaccOriginalActionKind::NoAction),
+                        "UserAction" => YaccKind::Original(YaccOriginalActionKind::UserAction),
+                        "Grmtools" => YaccKind::Grmtools,
+                        "GenericParseTree" => {
+                            YaccKind::Original(YaccOriginalActionKind::GenericParseTree)
+                        }
+                        s => panic!("Yacckind '{}' not supported", s)
```
ltratt: `YaccKind` with a capital "K".

lrpar/tests/build.rs
```diff
@@ -0,0 +1,57 @@
...
+            Ok(path) => {
+                if path.is_file() {
+                    // Parse test file
+                    let mut file = fs::File::open(&path)?;
+                    let mut s = String::new();
+                    file.read_to_string(&mut s)?;
+                    let docs = YamlLoader::load_from_str(&s).unwrap();
+                    let grm = &docs[0]["grammar"].as_str().unwrap();
+                    let lex = &docs[0]["lexer"].as_str().unwrap();
+                    let yacckind = match docs[0]["yacckind"].as_str().unwrap() {
```
ltratt: I wonder if we should make the user type out the full `YaccKind` e.g. `Original(YaccOriginalActionKind::NoAction)` and so on? Then it's one less thing to learn the rules about mapping the two.

lrpar/tests/build.rs
```diff
@@ -0,0 +1,57 @@
...
+                    pg.push(format!("{}.y", base));
+                    fs::write(&pg, &grm).unwrap();
+                    let mut pl = PathBuf::from(&out_dir);
+                    pl.push(format!("{}.l", base));
+                    fs::write(&pl, &lex).unwrap();
+
+                    // Build parser and lexer
+                    let lex_rule_ids_map = CTParserBuilder::new()
+                        .yacckind(yacckind)
+                        .process_file(pg.to_str().unwrap(), &out_dir)?;
```
ltratt: Deosn't `process_file` take a file (not a directory) as its second arg?

lrpar/tests/build.rs
```diff
@@ -0,0 +1,57 @@
...
+use glob::glob;
+use lrlex::LexerBuilder;
+use lrpar::CTParserBuilder;
+use std::{env, fs, io::Read, path::PathBuf};
+use yaml_rust::YamlLoader;
+
+fn main() -> Result<(), Box<std::error::Error>> {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    for entry in glob("src/*.test")? {
+        match entry {
```
ltratt: Since the `Err` path is `panic`, I think you can remove the `match` and just say `let path = entry.unwrap()`?

lrpar/tests/src/goal
```diff
@@ -0,0 +1,37 @@
+// build.rs, requires grm/lex in files
```
ltratt: I'm not sure what this file is doing. Any hints?

lrpar/tests/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+use lrlex::lrlex_mod;
+use lrpar::lrpar_mod;
+
+lrlex_mod!(multitypes_l);
```
ltratt: Let's sort these things alphabetically (calc, calcactions, multitypes).

lrpar/tests/src/calc.test
```diff
@@ -0,0 +1,39 @@
+name: Test NoAction using the calculator grammar
```
ltratt: We should give this a better name -- maybe `calc_noactions`?

lrpar/tests/src/calcactions.test
```diff
@@ -0,0 +1,39 @@
+name: Test basic user actions using the calculator grammar
```
ltratt: Maybe call this one `calc_actiontype`?

ltratt:
I have some minor comments, but nothing substantial. I think this looks good! Can I suggest we add a new test in which basically incorporates what's currently `lrpar/examples/calc_actions` in?

lrpar/tests/build.rs
ptersilie: Only `LexerBuilder` apparently. `CTParserBuilder` takes a directory and generates the `_y.rs` filename automatically, it appears.

lrpar/tests/src/goal
ptersilie: Oh, I dunno how this snug in all of a sudden. I was being careful not to add it. This can obviously  go!

lrpar/tests/build.rs
ltratt: Hmm, we should make those two consistent (one way or another -- I don't feel strongly). Can you maybe make a PR after this one to do that?

ptersilie: Will do.

ptersilie:
I addressed your comments and added the calc_action example as a new test.

lrpar/tests/src/calc_action_example.test
```diff
@@ -0,0 +1,44 @@
+name: Test basic user actions using the calculator grammar
```
ltratt: Based on the other files, should this be called `calc_multitypes`?

lrpar/tests/src/calc_action_example.test
```diff
@@ -0,0 +1,44 @@
...
+yacckind: Grmtools
+grammar: |
```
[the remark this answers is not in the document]
ptersilie: Yes, it tells Yaml that the following is a multiline string. Otherwise it'll only parse the first line.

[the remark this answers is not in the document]
ltratt: I did not know that!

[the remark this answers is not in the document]
ptersilie: Me neither. I had to look it up when it failed to parse the test files.

ptersilie:
Addressed your last comment.

ltratt:
Please squash.

ptersilie:
Squashed!

ltratt:
I've just realised: we need to document the YAML format used somewhere, even if only roughly, and even if only in a comment in Rust file. Otherwise I'll guarantee we'll baffle ourselves at some point in the future!

ptersilie:
Documented the syntax of the test files. See what you think.

lrpar/tests/build.rs
```diff
@@ -5,6 +5,11 @@ use lrpar::CTParserBuilder;
 use std::{env, fs, path::PathBuf};
 use yaml_rust::YamlLoader;
 
+// Compiles the `*.test` files within `src`. Test files are written in Yaml syntax and have 4
+// sections: name (describing what the test does), yacckind (defining the grammar type used),
```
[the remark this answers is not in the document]
ptersilie: Fixed.

ltratt:
Please squash.

ptersilie:
Squashed!

PR #91, opened 2019-01-21
Make process_file consistent between LexerBuilder and CTParserBuilder

ptersilie:
Previously, `process_file` in LexerBuilder expected a filename, while
CTParserBuilder's `process_file` expects a directory. This commit adjusts
LexerBuilder to generate the filename from a given directory similar to
CTParserBuilder, which makes the function a bit easier to use as can be
seen in lrpar/tests/build.rs where we the change saves a few lines of code.

### DIFF
--- a/lrlex/src/lib/builder.rs
+++ b/lrlex/src/lib/builder.rs
@@ -113,8 +113,8 @@ where
         self.process_file(inp, outp)
     }
 
-    /// Statically compile the `.l` file `inp` into Rust, placing the output into `outp`. The
-    /// latter defines a module with a function `lexerdef()`, which returns a
+    /// Statically compile the `.l` file `inp` into Rust, placing the output into the file `outp`.
+    /// The latter defines a module with a function `lexerdef()`, which returns a
     /// [`LexerDef`](struct.LexerDef.html) that can then be used as normal.
     pub fn process_file<P, Q>(
         self,
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -194,9 +194,13 @@ where
         let mut inp = current_dir()?;
         inp.push("src");
         inp.push(srcp);
-        let mut outd = PathBuf::new();
-        outd.push(var("OUT_DIR").unwrap());
-        self.process_file(inp, outd)
+        let mut leaf = inp.file_stem().unwrap().to_str().unwrap().to_owned();
+        leaf.push_str(&YACC_SUFFIX);
+        let mut outp = PathBuf::new();
+        outp.push(var("OUT_DIR").unwrap());
+        outp.push(leaf);
+        outp.set_extension(RUST_FILE_EXT);
+        self.process_file(inp, outp)
     }
 
     /// Set the `YaccKind` for this parser to `ak`.
@@ -229,8 +233,10 @@ where
         None
     }
 
-    /// Statically compile the Yacc file `inp` into Rust, placing the output file(s) into
-    /// the directory `outd`. The latter defines a module with the following functions:
+    /// Statically compile the Yacc file `inp` into Rust, placing the output into the file `outp`.
+    /// Note that three additional files will be created with the same name as `outp` but with the
+    /// extensions `grm`, `sgraph`, and `stable`, overwriting any existing files with those names.
+    /// `outp` defines a module with the following functions:
     ///
     /// ```text
     ///    fn parse(lexemes: &Vec<Lexeme<StorageT>>)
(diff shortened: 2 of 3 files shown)


ltratt:
I must admit, I mildly prefer that the user can specify the output file, in case they want to do something weird with the filename. Any thoughts?

ptersilie:
I think a user almost always wants to use `process_file_in_source` anyways, which is why I decided to change `LexerBuilder` instead (I already had changed `CTParserBuilder` but then realised that this leads to more code, while changing `LexerBuilder` gets rid of some code, which is why I changed it back).

ltratt:
The key word is "almost" I think! Is it "almost always" or "always"?

ltratt:
I'm imagining someone writing some weird tool where they choose to compile `_l.rs` and `_y.rs` files in a way we can't currently foresee. Maybe they have some filename clashes they want to deal with or somesuch.

ptersilie:
Sigh, alright, I'll change it back again.

ptersilie:
Alright, I changed it back. I'm still unsure why anyone would change the file extension to anything other than `.rs` (does Rust even build files with different extensions?), but who knows what users want to do.

ltratt:
I just realised what the killer use case is: if you have a grammar / lexer file that doesn't end in `.y` or `.l` (and there might be good reasons for that), we wouldn't be able to do the suffix switching trick in `process_file`. So I think we've done the right thing!

Please squash.

ltratt:
I forgot: weren't we going to also update the doc comments to make clear that `inp` and `outp` are file names?

ptersilie:
You are right. I did that in the initial commit but when I undid everything to change CTParserBuilder instead I forgot to update the doc there.

ptersilie:
I updated the doc to match that of `LexerBuilder`. The problem with that is that it doesn't clearly say if it expects a directory or a filename. Changing the description to say it takes an output file is a bit misleading, since `CTParserBuilder::process_file` generates additional files (e.g. grammar, graph, stable), which are put into the same base directory as `outp`. Suggestions?

ltratt:
Hmmm, that is a really good point. Admittedly, those files are a hack that one day I hope will go away. Let me check on this for a little bit.

ltratt:
https://github.com/softdevteam/grmtools/pull/91/commits/06524f92e29ba8d31f7b575601829d07f854c0a8 doesn't make sense because it references `outd`.

That aside, can we not just say something like "Statically compile the Yacc file `inp` into Rust, writing the output to the file `outp`."? It might be worth adding something like "Note that three additional files will be created `outp.grm`, `outp.sgraph`, and `outp.stable`, overwriting any existing files with those names."? One day, hopefully, we can remove that second sentence...

ptersilie:
Whoops, that should have been `outp`. Anyways, I like your suggestion and will update the doc-string accordingly.

ptersilie:
Doc-strings updated.

lrlex/src/lib/builder.rs
```diff
@@ -113,8 +113,8 @@ where
         self.process_file(inp, outp)
     }
 
-    /// Statically compile the `.l` file `inp` into Rust, placing the output into `outp`. The
-    /// latter defines a module with a function `lexerdef()`, which returns a
+    /// Statically compile the `.l` file `inp` into Rust, placing the output to the file `outp`.
```
ltratt: Needs to be "into" the file.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -233,8 +233,10 @@ where
         None
     }
 
-    /// Statically compile the Yacc file `inp` into Rust, placing the output into `outd`. The
-    /// latter defines a module with the following functions:
+    /// Statically compile the Yacc file `inp` into Rust, placing the output to the file `outp`.
+    /// Note that three additional files will be created with the same name as `outp` but with the
+    /// extensions `grm`, `sgraph`, and `stable`, overwriting any existing files with those names.
+    /// The latter defines a module with the following functions:
```
ltratt: What does "the latter" refer to? Probably just say "`outp`"?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -233,8 +233,10 @@ where
         None
     }
 
-    /// Statically compile the Yacc file `inp` into Rust, placing the output into `outd`. The
-    /// latter defines a module with the following functions:
+    /// Statically compile the Yacc file `inp` into Rust, placing the output to the file `outp`.
```
ltratt: Same comment as above.

ptersilie:
Addressed your comments.

lrlex/src/lib/builder.rs
ltratt: This one is yet to be fixed?

ptersilie: I'm an idiot. I changed it but didn't save the file before comitting.

ptersilie: Fixed.

ptersilie:
Squashed!

PR #92, opened 2019-03-07
Add surrounding_line_str functions.

ltratt:
When e.g. printing out error messages, it's often useful to get hold of the entire line that a lexeme is contained on.

The tests are probably the easiest way of seeing what this does.

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -259,6 +259,30 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
         )
     }
 
+    fn surrounding_line_str(&self, i: usize) -> &str {
+        fn surrounding_line_off<StorageT>(lexer: &LRLexer<StorageT>, i: usize) -> (usize, usize) {
+            if i > lexer.s.len() {
+                panic!("Offset {} exceeds known input length {}", i, lexer.s.len());
+            }
+
+            if lexer.newlines.is_empty() {
+                return (0, lexer.s.len());
+            } else if i < lexer.newlines[0] {
+                return (0, lexer.newlines[0] - 1);
+            }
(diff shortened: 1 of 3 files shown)


lrlex/src/lib/lexer.rs
```diff
@@ -259,6 +259,30 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
         )
     }
 
+    fn surrounding_line_str(&self, i: usize) -> &str {
+        fn surrounding_line_off<StorageT>(lexer: &LRLexer<StorageT>, i: usize) -> (usize, usize) {
```
ptersilie: Why is this a function? Can't be just inline the code, since all the outer `surrounding_line_str` does is call the inner function?

ltratt: If you look at the logic inside, it's quite tricky to make it part of the outer function. [I actually did a version which inlined everything, but I found it much harder to follow.]

ltratt:
Sausages, Travis's version of rustfmt and mine are out of sync. I'll manually change that rustfmt error.

lrlex/src/lib/lexer.rs
ptersilie: Ah, I see what you mean now.

ptersilie:
Looks good. Please squash!

ltratt:
I'll just wait for the tests to pass to make sure I haven't done anything silly.

ltratt:
Squashed.

ltratt:
A thought before we merge: should this actually be called `surrounding_line` and not `surrounding_line_str`?

ptersilie:
I don't mind either way. Given it does return a string "surrounding_line_str" is ok IMO.

ltratt:
OK, let's keep it as-is.

PR #94, opened 2019-03-07
Having offset_line_col return the byte offset of columns is annoying to use.

ltratt:
Byte offsets are efficient ways of indexing UTF-8 strings, but are confusing to users, since bytes and characters aren't always the same thing. In other words, in the face of multi-byte characters, offset_line_col gave confusing results.

This commit fixes this, returning the character offset for the column which is nearly always what users want. To make this clear, this commit also renames `offset_line_col` to `line_col`. This is a breaking change, but probably the right thing to do.

[I'm not sure why this didn't occur to me earlier -- probably I made the classic mistake of forgetting that byte and character offsets aren't the same thing in UTF-8. Besides, before the `surrounding_line` function was added, there wasn't really an easy way to perform the necessary calculation.]

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -369,6 +369,34 @@ if 'IF'
     }
 
     #[test]
+    fn test_multibyte() {
+        let src = "%%
+[a❤]+ 'ID'
+[ ] ;"
+            .to_string();
+        let mut lexerdef = parse_lex(&src).unwrap();
+        let mut map = HashMap::new();
+        map.insert("ID", 0u8);
+        assert_eq!(lexerdef.set_rule_ids(&map), (None, None));
+
+        let mut lexer = lexerdef.lexer("a ❤ a");
+        let lexemes = lexer.all_lexemes().unwrap();
+        assert_eq!(lexemes.len(), 3);
+        let lex1 = lexemes[0];
+        assert_eq!(lex1.start(), 0);
+        assert_eq!(lex1.len(), Some(1));
+        assert_eq!(lexer.lexeme_str(&lex1), "a");
+        let lex2 = lexemes[1];
+        assert_eq!(lex2.start(), 2);
+        assert_eq!(lex2.len(), Some(3));
+        assert_eq!(lexer.lexeme_str(&lex2), "❤");
+        let lex3 = lexemes[2];
+        assert_eq!(lex3.start(), 6);
+        assert_eq!(lex3.len(), Some(1));
+        assert_eq!(lexer.lexeme_str(&lex3), "a");
+    }
+
+    #[test]
     fn test_line_col() {
         let src = "%%
 [a-z]+ 'ID'
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -239,24 +239,30 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
         None
     }
 
-    fn offset_line_col(&self, i: usize) -> (usize, usize) {
+    fn line_col(&self, i: usize) -> (usize, usize) {
         if i > self.s.len() {
             panic!("Offset {} exceeds known input length {}", i, self.s.len());
         }
 
-        if self.newlines.is_empty() || i < self.newlines[0] {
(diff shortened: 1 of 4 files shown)


lrlex/src/lib/lexer.rs
```diff
@@ -376,31 +382,53 @@ if 'IF'
...
+
+        let mut lexer = lexerdef.lexer(" a\n❤ b");
+        let lexemes = lexer.all_lexemes().unwrap();
+        assert_eq!(lexemes.len(), 3);
+        assert_eq!(lexer.line_col(lexemes[0].start()), (1, 2));
+        assert_eq!(lexer.line_col(lexemes[1].start()), (2, 1));
+        assert_eq!(lexer.line_col(lexemes[2].start()), (2, 3));
+        assert_eq!(lexer.surrounding_line_str(lexemes[0].start()), " a");
+        assert_eq!(lexer.surrounding_line_str(lexemes[1].start()), "❤ b");
+        assert_eq!(lexer.surrounding_line_str(lexemes[1].start()), "❤ b");
```
ptersilie: I think this last assert was meant to be `assert_eq!(lexer.surrounding_line_str(lexemes[2].start()), "b");`

ltratt: Oops, deliberate mistake! [I might be lying.] I'll fix it.

ltratt:
Fix pushed.

ptersilie:
Rest looks good. Please squash.

lrlex/src/lib/lexer.rs
```diff
@@ -423,7 +423,7 @@ if 'IF'
         assert_eq!(lexer.line_col(lexemes[2].start()), (2, 3));
         assert_eq!(lexer.surrounding_line_str(lexemes[0].start()), " a");
         assert_eq!(lexer.surrounding_line_str(lexemes[1].start()), "❤ b");
-        assert_eq!(lexer.surrounding_line_str(lexemes[1].start()), "❤ b");
+        assert_eq!(lexer.surrounding_line_str(lexemes[2].start()), "❤ b");
```
ptersilie: Wait, should the last string only be `"b"`?

ltratt: No, because it's the whole line that that lexeme is contained in.

ptersilie: Either I'm missing something, or the tests shouldn't have passed.

ptersilie: Ah, wait, this is testing `surrounding_line_str`. Nevermind.

ptersilie: Wouldn't it make more sense to test here if the lexemes are correct? e.g.
```
assert_eq!(lexemes[1], "❤");
assert_eq!(lexemes[2], "b");
```

ltratt: A unicode test for the pure lexer is a good idea -- I'll add that in.

ltratt:
I've added a multibyte test for the plain lexer. See what you think.

lrlex/src/lib/lexer.rs
```diff
@@ -368,6 +368,31 @@ if 'IF'
...
+
+        let mut lexer = lexerdef.lexer("a ❤ a");
+        let lexemes = lexer.all_lexemes().unwrap();
+        assert_eq!(lexemes.len(), 3);
+        let lex1 = lexemes[0];
+        assert_eq!(lex1.start(), 0);
+        assert_eq!(lex1.len(), Some(1));
+        let lex2 = lexemes[1];
+        assert_eq!(lex2.start(), 2);
+        assert_eq!(lex2.len(), Some(3));
```
ptersilie: Would it be difficult to add another check here, that checks the content of the lexeme, just so we also test, that the lexer doesn't lex something like ` \xe2\x9d` as the second lexeme (e.g. space + the first two characters of ❤)?

ltratt: Done -- try the newest commit.

ptersilie: Perfect!

ptersilie:
Looking good now. Please squash!

ltratt:
Squashed (note that I kept the testing-the-lexer commit separate, because it's really stand alone).

PR #97, opened 2019-03-25
Packaging changes

ltratt:


ptersilie:
Looks alright. Though we can probably squash the two commits, since the second removes the changed line from the first anyways.

ltratt:
Agreed. Squashed.

PR #100, opened 2019-04-09
Add various rustdoc comments for lrpar.

ltratt:
Before this, the docs.rs output was Spartan, to say the least, and probably off-putting due to the lack of documentation.

### DIFF
--- a/lrpar/src/lib/mod.rs
+++ b/lrpar/src/lib/mod.rs
@@ -7,19 +7,190 @@
 // at your option. This file may not be copied, modified, or distributed except according to those
 // terms.
 
+//! `lrpar` provides a Yacc-compatible parser (where grammars can be generated at
+//! compile-time or run-time). It can take in traditional `.y` files and convert
+//! them into an idiomatic Rust parser. More details can be found in the [grmtools
+//! book](https://softdevteam.github.io/grmtools/master/book); the
+//! [quickstart guide](https://softdevteam.github.io/grmtools/master/book/quickstart.html)
+//! is a good place to start.
+//!
+//!
+//! ## Example
+//!
+//! Let's assume we want to statically generate a parser for a simple calculator
+//! language (and let's also assume we are able to use
+//! [`lrlex`](https://softdevteam.github.io/grmtools/master/book/lrlex.html) for the
+//! lexer). We need to add a `build.rs` file to our project which tells `lrpar` to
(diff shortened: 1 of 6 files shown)


ltratt:
OK, that was more annoying than I expected: the rustdoc documentation seems incorrect. `ignore` still compiles Rust code! Anyway, this eventually fixes things, and is now ready for review.

ltratt:
[I think this might be fixed in Rust nightly, because on b8 I couldn't trigger the same error. But I'm not sure.]

lrpar/src/lib/mod.rs
```diff
@@ -7,20 +7,192 @@
...
+//!         Ok(val) => Ok(val),
+//!         Err(_) => {
+//!             eprintln!("{} cannot be represented as a u64", s);
+//!             Err(())
+//!         }
+//!     }
+//! }
+//! ```
+//!
+//! Because we specified that our Yacc file is in `Grmtools` format, each Rule has a
```
ptersilie: Is Rule referring to a class and thus capitalised on purpose?

lrpar/src/lib/mod.rs
```diff
@@ -7,20 +7,192 @@
...
+//!                     _ => eprintln!("Unable to evaluate expression.")
+//!                 }
+//!             }
+//!             _ => break
+//!         }
+//!     }
+//! }
+//! ```
+//!
+//! We can now cargo run our project and evaluate simple expressions:
```
ptersilie: cargo run in backticks?

ptersilie:
Some very minor comments.

ptersilie:
If it turns out this is fixed in latest Rust, should we then put back the `rust,ignore` flag. Because current with `text` it wouldn't syntax highlight these parts in the documentation, right?

ltratt:
> If it turns out this is fixed in latest Rust, should we then put back the rust,ignore flag. Because current with text it wouldn't syntax highlight these parts in the documentation, right?

I think we can wait to see what happens in the next stable release. I think this is a long-standing bug / flaw / feature, so I'm not really sure what's going on.

ltratt:
I've addressed comments in the latest commit. I've also fixed the `README` (where some of this material is inevitably duplicated from).

ptersilie:
Excellent! Looks good. Please squash!

ltratt:
Squashed.

PR #104, opened 2019-04-22
Tentatively add a $span pseudo-variable

ltratt:
This allows you to tell how much input the current production has matched, which can be useful for giving better debugging information to users. Its type is `(usize, usize)` where these represent (start, end) offsets in the input.

For example if you have a rule:

```
  Expr: Term '+' Expr { println!("{:?}", $span); } ;
```

and input along the lines of "2+3", you will get the output "(0, 3)".

Interestingly, users can mostly calculate this same information themselves (by inspecting tokens start/end positions), *except* for epsilon rules where there's no way of knowing where in the input we are. So this production can't be made to work sensibly except with $span:

```
  R: { println!("{:?}", $span); }
```

This is a bit tentative, because I haven't used this enough yet to know if it's the right design: feedback is welcome! The major commit is https://github.com/softdevteam/grmtools/commit/2b64530eb381c3f009d78aaf18b49c739ac43bff with documentation in https://github.com/softdevteam/grmtools/commit/b81b095a0901b6953a4ebe03637926cd3b11825a; the other commits are mostly shuffling a few things around.


ltratt:
That last commit, if nothing else, will need to be squashed away.

lrpar/cttests/src/lib.rs
```diff
@@ -88,3 +91,51 @@ fn test_calc_multitypes() {
...
+        }
+        _ => unreachable!()
+    }
+
+    let mut lexer = lexerdef.lexer("2+3*4");
+    match span_y::parse(&mut lexer) {
+        (Some(ref spans), _)
+            if spans
+                == &vec![
+                    (0, 1),
```
ptersilie: It's scary how it didn't even confuse me why some of these appear multiple times, because my head had already parsed the above input and made the `P->T` reductions.

ltratt: Ha! It took me a few times to get my head around it!

lrpar/src/lib/mf.rs
```diff
@@ -236,13 +237,16 @@ where
             return (in_laidx, vec![]);
         }
         simplify_repairs(parser, &mut rnk_rprs);
+        dbg!(&(&in_pstack, &span));
```
ptersilie: Debug statement?

ltratt: Oops!

lrpar/src/lib/parser.rs
```diff
@@ -289,9 +300,15 @@ where
                     let prior = *pstack.last().unwrap();
                     pstack.push(self.stable.goto(prior, ridx).unwrap());
 
+                    let tail = span[span.len() - 1];
```
ptersilie: What is the purpose if this? I can't really tell just from the code.

doc/src/yaccvariants.md
```diff
@@ -4,19 +4,28 @@ grmtools can deal with several different variants of Yacc input.
 
 ## Grmtools
 
-`YaccKind::Grmtools` is grmtools own variant of Yacc syntax. The sole difference
-is that rules have a Rust type to which all their production's actions must adhere
-to. Consider the following snippet:
+`YaccKind::Grmtools` is grmtools own variant of Yacc syntax, and the one that
```
ptersilie: `grmtools'`?

ltratt: Good point.

doc/src/actioncode.md
```diff
@@ -0,0 +1,19 @@
+# Action code
+
+Action code is normal Rust code with the addition of the following special variables:
+
+ * `$1` ... `$n` refer to the respective symbol in the production, numbered
+   from 1 (i.e. `$1` refers to the first symbol in the production). If the
+   symbol references a rule `R` then an instance of `R`'s type will be stored
+   in the `$x` variable; if the symbol references a lexeme then an
```
ptersilie: Maybe use `$i` instead of `$x` as that may make it clearer we are talking about a variable from `$1...$n`.

ptersilie:
How easy would it be for the user to use the span information to extract the part from the input that has successfully been parsed? I don't currently see the span on its own being useful for debugging. Also, doesn't the occurrence of the first error already tell us how far the parser managed to parse the input?

lrpar/src/lib/parser.rs
ltratt: This code is actually where most of the drama when creating this PR happened! Here's an easier rewrite:

```rust
                    let sp = (span[pop_idx - 1], span[span.len() - 1]);
                    if pop_idx < span.len() - 1 {
                        span.drain(pop_idx..span.len() - 1);
                    }
```

Basically we want to delete from `pop_idx..span.len() - 1` *but* sometimes `popidx == span.len() - 1` and then `drain` complains about that (why... I don't know). Unfortunately there isn't a `remove_range` function, hence the use of `drain`.

ptersilie: Ah! That makes sense.

doc/src/actioncode.md
ltratt: Agreed.

ltratt:
>How easy would it be for the user to use the span information to extract the part from the input that has successfully been parsed?

Not hugely easy, but it is easy to turn the numbers into (line, col) pairs. I'm basically thinking that when we build an AST we sometimes want to say "this AST spans this part of the input (start, end)" and that's the use case for `$span`.

ptersilie:
> we sometimes want to say "this AST spans this part of the input (start, end)" and that's the use case for `$span`.

Fair enough. Maybe this could also come in handy when we have a go at incremental error recovery.

Please squash.

ltratt:
I'm going to have a quick peek at that bit of code that confused you: I think I can make it faster (and since it's in a critical loop, that's probably worth doing).

ltratt:
I can speed things up, but I can't make them easier to read. I don't know if that latest commit makes things worse/same/better. Thoughts?

ptersilie:
Maybe just add a comment that says, `this deletes pop_idx..span.len() - 1`? When I first read it I couldn't understand why we were jumping through so many hoops, e.g. get the tail, remove stuff, and attach the tail.

ltratt:
That new commit should fix things.

However, I am starting to think we should put `$span` behind a feature gate of some sort, because it will definitely have a performance hit, and if you don't use it, you probably don't want to pay for it. Thoughts?

ptersilie:
Looks good. Please squash.

ptersilie:
> That new commit should fix things.
> 
> However, I am starting to think we should put `$span` behind a feature gate of some sort, because it will definitely have a performance hit, and if you don't use it, you probably don't want to pay for it. Thoughts?

Yes, I agree with that. Should it be a setting at the top of the grammar, maybe?

ltratt:
I think it has to be a flag in `CTBuilder`: conceptually that could remove the run-time cost entirely. [It won't make any performance difference at the moment, but that's because we don't generate a proper compile-time version of the parser: but we could in the future.]

ptersilie:
Fair enough. Do you want to put this in the current PR or create a new one?

ltratt:
I've just done another commit: I think it makes sense to do it here. See what you think.

ptersilie:
Looks good. So we just need to remember to remove the span variable all together in a future PR to get a performance gain from this, right?

ltratt:
> So we just need to remember, in a future PR, to remove the span variable completely when it is disabled to get a performance gain from this, right?

Sort of, but maybe not exactly. The new commit means that you the `$span` variable doesn't exist if you don't call `span_var(true)` *but* the run-time parser still creates the `span` stack (etc.). Turning those calculations off is what we'd need to do.

ptersilie:
> Sort of, but maybe not exactly. The new commit means that you the `$span` variable doesn't exist if you don't call `span_var(true)` _but_ the run-time parser still creates the `span` stack (etc.). Turning those calculations off is what we'd need to do.

Sorry, that's exactly what I meant. I assume we'll do this in a separate PR though?

ltratt:
> I assume we'll do this in a separate PR though?

Yes, and probably not for a while either.

ptersilie:
Okay, time to squash then. Should we leave a note somewhere that this needs to be done someday? Because currently we are slowing things down for debugging info that is likely not being used.

ptersilie:
But until then this PR slows down parsing for the sake of having this debug information?

ltratt:
I wouldn't call it debugging info, as such -- it's info that people are probably most often going to use for helping debugging, but I'm sure someone will find some other interesting use case for it.

ptersilie:
Fair enough. Let's squash this then and get it merged.

ltratt:
Squashed (that was a really hard one to squash down!).

PR #112, opened 2019-06-10
Add the dyn keyword where appropriate

ltratt:
The second commit is key here, because it stops generated grammars causing tens of warnings along the lines of:

```
    warning: trait objects without an explicit `dyn` are deprecated
       --> /home/ltratt/scratch/softdev/yksom/target/debug/build/yksom-b0187fd6ceceb296/out/som_y.rs:475:36
        |
    475 |                       __gt_lexer: &Lexer<u32>,
        |                                    ^^^^^^^^^^ help: use `dyn`: `dyn Lexer<u32>`
```

when compiled with a new(ish) rustc.


ptersilie:
I feel like we can squash those commits, as they both do the same thing: fix the deprecation warnings. It doesn't really matter if one was fixed manually and the other by `cargo fix`.

ltratt:
Fair enough. Shall I do so now?

ptersilie:
Please do.

Issue #59, opened 2018-12-03
Ignoring `$X` within strings for actions


ptersilie:
When translating action code we currently replace all `$X` within it to reference the correct value from the actions stack. However, since we are using a simple regex to do this, we also replace any occurrence of `$[0-9]+` within comments and strings. While we can live with altering comments, changing strings may prevent users from generating reasonable programs.

Since parsing rust is too big a task, a quick solution to this problem would be to count unescaped quotes, and only replace occurrences of `$X`, if the count is equal. This should at least get rid of replacements within user strings.

ltratt:
Fixed in #144.

PR #114, opened 2019-06-22
Parse $ variables in actions in a better fashion.

ltratt:
This is more efficient than the old mechanism, and also adds an escape mechanism for "$": "$$" in an action is equivalent to "$" in normal Rust code. The passthrough test hopefully makes this clear.

In a sense, this is an alternative to the fix suggested in https://github.com/softdevteam/grmtools/issues/59, which I think we can never really make work well. The "$$" escaping thing is relatively well known in other contexts, so hopefully it isn't too surprising, though note that Yacc uses `$$` to mean the return value of an action (a convention which we do not use). So perhaps it's not ideal in that regard, but it's hard to imagine anyone thinking that Yacc's special return pseudo-variable is a great idea!

Comments welcome: this isn't a "merge and forget" sort of PR. Maybe this isn't the best solution!

### DIFF
--- a/doc/src/actioncode.md
+++ b/doc/src/actioncode.md
@@ -18,3 +18,7 @@ Action code is normal Rust code with the addition of the following special varia
    captures how much of the user's input the current production matched. This
    can be useful when storing debugging information. Note that this variable is
    not enabled by default: use `CTParserBuilder::span_var(true)` to enable it.
+
+ * `$$` is equivalent to `$` in normal Rust code.
+
+Any other variables beginning with `$` are treated as errors.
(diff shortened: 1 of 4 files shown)


ltratt:
I found a new way to mess this up: I ran `cargo fmt` in the wrong repository. The new commit should fix things.

ptersilie:
I can't really find a flaw with this. It's unfortunate that we didn't think about `$` already being reserved in Rust, otherwise we might have picked another character. Although, I couldn't even think of an alternative.

ltratt:
I think any character we pick will have the same problem, and at least Rust doesn't use `$` very much itself (apart from macros).

ptersilie:
Yes, that's what I meant. So I think this looks merge-able to me after squashing.

ltratt:
Squashed.

PR #116, opened 2019-09-30
Update stale version numbers in the quickstart guide.

ltratt:


ltratt:
Updated the Travis check so that the ones-that-can't-succeed don't show up. This should be OK to merge now.

PR #117, opened 2019-11-19
Make CPCT+ the default error recovery algorithm.

ltratt:
On reflection, I think MF is too complex for the additional gains we get from it over CPCT+. On that basis, this commit makes CPCT+ the default, and hides any mention of MF from the outside world. At least for the time being, however, we'll maintain all the code internally, in case we change our minds on this.

### DIFF
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -33,7 +33,7 @@ fn usage(prog: &str, msg: &str) -> ! {
     }
     writeln!(
         &mut stderr(),
-        "Usage: {} [-r <cpctplus|mf|none>] [-y <eco|grmtools|original>] [-q] <lexer.l> <parser.y> <input file>",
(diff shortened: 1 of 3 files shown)


nimbleparse/src/main.rs
```diff
@@ -62,8 +62,8 @@ fn main() {
         .optopt(
             "r",
             "recoverer",
-            "Recoverer to be used (default: mf)",
-            "cpctplus|mf|none"
+            "Recoverer to be used (default: cpctplus)",
+            "mf|none"
```
ptersilie: I think  you wanted to remove `mf` here and keep `cpctplus`.

ltratt: Good spot!

ltratt: Fixed in https://github.com/softdevteam/grmtools/pull/117/commits/7316e6fe583d0b642501920920625cefead6eafb.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #120, opened 2019-12-02
Zero copy input

ltratt:
This doesn't look like a big PR, but it is a painful one. The ultimate goal is to allow people to return / store substrings from the user's input in a grammar without having to copy memory. In other words, before this PR, if you wanted to extract a substring you had two choices:

  * Return a `Lexeme<StorageT>` -- this is often useful because it stores span information, but it's not always ergonomic.
  * Call `to_owned` on `lexer.lexeme_str(...)` -- which is wasteful.

In essence, this PR adds a lifetime `'input` which allows users to return / store `&'input str` substrings with zero copying. Perhaps the easiest place to start is the `calc_ast` grammar which makes use of this facility (https://github.com/softdevteam/grmtools/commit/13c23b32cc4a477a75698137f41a07ba8103d752#diff-a9e315df25061f8a6a0e4317eb4f3ad9).

Unfortunately this is a more complex -- though not particularly big -- change than I first expected. In essence, when I wrote much of grmtools I didn't understand the finer points of lifetimes. Some of the commits in this PR (https://github.com/softdevteam/grmtools/commit/8559653a26983e4dd4c328edaf0d1acdef980959, https://github.com/softdevteam/grmtools/commit/33f8581b97a85d40fa5be8ad140ed449a295c263, https://github.com/softdevteam/grmtools/commit/b2e1c118340949ca2835dde711ff20f9d33aa50f) fix minor mistakes that I made in my ignorance.

In a sense, https://github.com/softdevteam/grmtools/commit/b9f2885a73e6514bef254a40259b68a08ba19119 is the most important commit in that it changes the API (in a backwards compatible way) so that the lifetimes of grammars and input are separated. In other words, the lexer can live a different length of time than the grammar. That fairly directly opens the way for the seemingly most important commit https://github.com/softdevteam/grmtools/commit/0d5d70c8ef3adcf383ddf59624633b86d2053896 that allows grammar files to actually reference the `'input` lifetime.

Unfortunately that showed that the current design of having `mut` lexers is totally unergonomic: if your grammar returned any `&'input str`s you couldn't do things like print out syntax errors! https://github.com/softdevteam/grmtools/commit/52351a708f41150079754b611f601e038bbeda67 is thus an annoying commit which removes the mutability requirement on lexers, but probably means that all of them (`lrlex` included) have to resort to interior mutability. As unfortunate as this is, I don't know of a better trade off.


lrlex/examples/calclex/src/main.rs
```diff
@@ -17,7 +17,7 @@ fn main() {
             Some(Ok(ref l)) => {
```
ptersilie: Typo in commit message: `This commit thus change the API from:`

ltratt: Agreed.

lrlex/src/lib/lexer.rs
```diff
@@ -201,7 +202,7 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
                 }
             }
             if longest > 0 {
-                self.newlines.extend(
+                unsafe { &mut *self.newlines.get() }.extend(
```
ptersilie: Why is this unsafe?

ltratt: It's an `UnsafeCell`.

ptersilie: Why does `newlines` need to be in an `UnsafeCell`?

ltratt: The commit message https://github.com/softdevteam/grmtools/pull/120/commits/52351a708f41150079754b611f601e038bbeda67 hopefully helps. Basically removing `&mut self` means we have to move to interior mutability and for something like a `Vec` that means `UnsafeCell` (with all its attendant horrors) is our only option.

ptersilie: I think I understand now.

ptersilie:
Just one typo, rest looks good.

ltratt:
Update to commit message pushed.

ptersilie:
Merge as is?

ltratt:
Yes please.

PR #122, opened 2019-12-06
Panic if timeout in debug

ltratt:
Error recovery has a timeout, but that is not very useful in test mode which is slow enough that it's not hard to hit the limit. And, when that happens, much confusion results. This PR changes things so that in test mode the timeout is a) increased b) if triggered causes a `panic`. This should stop us being confused again in the future...

### DIFF
--- a/lrpar/src/lib/astar.rs
+++ b/lrpar/src/lib/astar.rs
@@ -62,6 +62,9 @@ where
         }
 
         if !neighbours(true, &n, &mut next) {
+            #[cfg(test)]
+            panic!("Timeout occurred.");
+            #[cfg(not(test))]
             return Vec::new();
         }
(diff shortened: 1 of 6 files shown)


lrpar/src/lib/astar.rs
```diff
@@ -62,6 +62,9 @@ where
         }
 
         if !neighbours(true, &n, &mut next) {
```
jacob-hughes: Does this always mean "no neighbours were found within our `RECOVERY_TIME_BUDGET`"? Could this accidentally panic in the case that the search was exhausted (but within the allotted time) ?

ltratt: At the moment it's exclusively used to mean "timeout".

jacob-hughes: Fair enough

PR #152, opened 2020-02-07
Move from "point" indexes to spans

ltratt:
Previously we identified lexemes (and points of errors) as happening at a single byte offset in the index. This is sometimes sufficient (e.g. for lexing errors), but parse errors are more naturally expressed as happening across a span of the input.

This PR thus introduces the concept of a "span" [*]. A span tells one a start and end byte offset in input. This is a breaking API change, and has several knock on effects, though much of the PR is inevitably minor churn. One advantage of spans can be seen in the [`test_multiline_lexeme`](https://github.com/softdevteam/grmtools/commit/8620ac5eeadf2ab6bf235e31801db0b378a2575f#diff-1a9216bfbedf1a82401b1d47edc4a375R527) test.

[*] The terminology is from rustc, though the concept is as old as the hills -- Converge did this properly, and I have no good justification for why I didn't start things properly off in grmtools. Oh well.

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -232,6 +232,7 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
     }
 
     fn line_col(&self, span: Span) -> ((usize, usize), (usize, usize)) {
+        debug_assert!(span.end() >= span.start());
         if span.end() > self.s.len() {
             panic!(
                 "Span {:?} exceeds known input length {}",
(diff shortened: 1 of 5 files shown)


lrlex/src/lib/lexer.rs
```diff
@@ -252,9 +256,16 @@ impl<'a, StorageT: Copy + Eq + Hash + PrimInt + Unsigned> Lexer<StorageT>
             )
         }
 
-        let (line_idx, col_byte) = line_col_off(self, i);
-        let line = self.surrounding_line_str(i);
-        (line_idx, line[..col_byte].chars().count() + 1)
+        fn line_col<StorageT: Copy + Eq + Hash + PrimInt + Unsigned>(
```
ptersilie: Does this mean we are defining a function `line_col` within another function also called `line_col`? If so that's a little bit confusing.

ltratt: It's a local function so the two can't be confused. Maybe I could call the inner functions `lc`?

ptersilie:
> but parse errors are more naturally expressed as happening across a span of the input.

I'm not sure I understand what this means. Parse errors are expressed in tokens, e.g. `Parse error at token X in file:line:col`. How does the introduction of `Span` affect/improve our parse errors?

lrlex/src/lib/lexer.rs
ltratt: Changed in https://github.com/softdevteam/grmtools/pull/152/commits/23b18e6324a0649930b7481dce1197d53633d733. See if you think it's better or not!

ptersilie: I noticed it's in inner function but I still found it confusing that it had the same name. Renaming it is a good idea IMO.

ltratt:
> How does the introduction of Span affect/improve our parse errors?

At the moment not *much*, but this PR enables using `Span`s across production rules (i.e. more than the span of a lexeme, which was all you could do before this PR) (e.g. you could report an error across a whole `if` statement or a `fn` etc.). That is something I used to do in Converge, and it can give much better error messages, because you can now highlight multi-lexeme blocks in the user's input.

ptersilie:
I see. Okay, please squash.

ltratt:
Squashed.

PR #154, opened 2020-02-10
span_str

ltratt:
The main aim of this PR is to change `lexeme_str(...: &Lexeme)` into `span_str(...: Span)` (https://github.com/softdevteam/grmtools/commit/6b283f0311725d9badde6b427ee89d7982162129 adds `span_str` and https://github.com/softdevteam/grmtools/commit/6d35508f7271c1ed0e0cf3af425714ac9c7aae41 removes `lexer_str`).

That then enables us to write some nicer code. I expanded the `calc_ast` example -- not in terms of lines of code very much, but in terms of functionality -- substantially to take advantage of all of this. https://github.com/softdevteam/grmtools/commit/359cfa72a36fd515cc1890e797b0baf41c463dad contains those changes and https://github.com/softdevteam/grmtools/commit/c55cd58728e8167c7e58a6621399150039fa15f1 documents them.


lrpar/src/lib/lex.rs
```diff
@@ -62,6 +61,14 @@ pub trait Lexer<StorageT: Hash + PrimInt + Unsigned> {
     ///
     /// If the span exceeds the known input.
     fn span_lines_str(&self, span: Span) -> &str;
+
+    /// Return `((start line, start column char), (end line, end column char))` for `span`. Note
```
ptersilie: What is `start column char`. I thought this would just return the line and column, e.g. `4, 5`. Does this have to do with the fact that we are returning chars and not bytes? Either way `start column char` is a bit confusing as it sounds like this returns a character instead of an index. Maybe `start column char index` would be better, if a bit long though.

doc/src/actioncode.md
```diff
@@ -10,11 +10,12 @@ Action code is normal Rust code with the addition of the following special varia
 
  * `$lexer` allows access to the lexer and its [various
    functions](https://softdevteam.github.io/grmtools/master/api/lrpar/trait.Lexer.html).
-   The most commonly used of these is the `lexeme_str` function, which allows
-   us to turn `Lexeme`s into `&'input str`s representing the corresponding
-   portion of the user's input. As this may suggest, actions may also reference the
-   special lifetime `'input` (without any `$` prefix), which allows strings to
-   be returned / stored by the grammar without copying memory.
+   The most commonly used of these is the `span_str` function, which allows us
```
ptersilie: Better: The most common use for this is...

doc/src/quickstart.md
```diff
@@ -156,11 +156,12 @@ the respective symbol in the production, numbered from 1 (i.e. `$1` refers to th
 the production). If the symbol references a rule `R` then an instance of
 `R`'s type will be stored in the `$x` variable; if the symbol references a lexeme then an `Option<Lexeme>`
 instance is returned. A special `$lexer` variable allows access to the lexer.
-This allows us to turn `Lexeme`s into strings with the `lexeme_str` function,
-which given a `Lexeme` returns a `&'input str` representing the corresponding portion of
-the user's input. As this may suggest, actions may also reference the
-special lifetime `'input` which allows strings to be returned / stored
-by the grammar without copying memory.
+The most commonly used function that `$lexer` exposes is the `span_str`
```
ptersilie: Same here.

CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+   * `Lexer::span_str` replaces `Lexer::lexeme_str` function. Roughly speaking this:
+       ```rust
+       let s = lexer.lexeme_str(&lexeme);
+       ```
+     becomes:
+       ```rust
+       let s = lexer.span_str(lexeme.span());
+       ```
+   * `Lexer::line_col` now takes a `Span` rather than a `usize` and, since a
+     `Span` can be over multiple lines returns `((start line, start column),
```
ptersilie: missing `,` after `lines`.

CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+     (end line, end column))`.
+   * `Lexer::surrounding_line_str` is removed in favour of `span_lines_str`
+     which takes a `Span` and returns a (possibly multi-line) `&str` of the
+     lines containing that `Span`.
+   * The `$span` special variable now returns a `Span` rather than `(usize,
```
ptersilie: Did we have already the special variable `$span` before this version?

lrpar/src/lib/lex.rs
ltratt: Think of it as `(start, end)` where both `start` and `end` have `(line, column char)`. So, yes, it's because we return chars not bytes. I suppose we could say `((start line, start column char off)), ...)` or similar?

doc/src/actioncode.md
ltratt: It's referring to the lexer's functions, so I think the current phrasing is OK?

CHANGES.md
ltratt: Fixed in https://github.com/softdevteam/grmtools/pull/154/commits/4469eb9a9a2d77f6fe6ded208c9952b478139c09.

CHANGES.md
ltratt: Yes -- although I admit I'd forgotten about it too! It's currently not enabled by default, which I think we should change, but that'll be for another PR.

ltratt:
I've just realised that we can simplify the AST evaluator even further in https://github.com/softdevteam/grmtools/pull/154/commits/04e58cb306d4c21ebefe06f2a059743824f2abaf. Adding it now is a bit naughty, but I think it makes the documentation sufficiently simpler to read that it's probably worth it -- let me know if you think it's too much for this PR or not.

doc/src/actioncode.md
ptersilie: Ah, I see it now. Ok that's fine then.

lrpar/src/lib/lex.rs
ptersilie: I think just shortening it to `start line, start column` is probably fine, especially with the note further down explaining that we count chars not bytes.

ptersilie:
Fine by me. If it makes stuff easier we should add it now.

lrpar/src/lib/lex.rs
ltratt: Fixed in https://github.com/softdevteam/grmtools/pull/154/commits/0b522f3a06dedcaf711cb8d7a644294ea6e04cef.

ltratt:

ptersilie:
Yes, please squash!

ltratt:
Squashed.

ptersilie:
So are we doing the AST evaluator simplification in this PR or a separate one?

ltratt:
I left it in as I thought you were OK with that?

ltratt:
It's in https://github.com/softdevteam/grmtools/pull/154/commits/b69f04e2b95339950a693a1dfc562212d1b26aa6

ptersilie:
Ah that was it already! For some reason I expected this change to be much bigger since you were so careful about adding it at this point in time.

Everything's well then!

bors r+

Issue #133, opened 2019-12-15
Increase visibility of this project


pablosichert:
Hey there,

I just ported over a grammar written in Bison/Flex to grmtools – and I was super pleased with how the process went. (Sorry for bombarding you with a lot of small issues, those were just things I noted while implementing.)

Considering how mature this library seems to me, I'm wondering if you deliberatively chose not to advertise it more (-> more maintenance effort)?

I spent a lot of time researching and trying parser generators for Rust, but none of the existing ones ticked all the boxes that I were looking for, until I found this one.

In case you're interested in this project having more visibility (which in my opinion would be useful to a lot of people) I think the Readme could be a bit more elaborate and the repository could be tagged with some topics. Submitting it to relevant aggregators (/r/rust, hackernews, ...) after having a representative landing page would certainly help as well.

Thank you for the huge effort and developing this project in public!

ltratt:
I'm glad you've found this useful, and thanks for pointing out limitations -- it helps make things better! There's been no particular intent to keep this quiet, nor any intent so far to publicise it. I view the libraries herein as semi-mature, in the sense that I'm still prepared to make breaking API changes (e.g. https://github.com/softdevteam/grmtools/pull/125) if it usefully improves things. That said, I've been using these libraries in other projects fairly happily, and I know others have too, which is a decent sign that things are probably ready for a somewhat wider audience. I take your point about the `README`'s brevity, though I hope that the [quickstart](https://softdevteam.github.io/grmtools/master/book/quickstart.html) goes some way to rectifying that. I will beef up the `README`. We'll probably have a `0.5.0` release at some point this week or so and maybe that will help find these libraries a few more users!

ptersilie:
> I take your point about the README's brevity, though I hope that the quickstart goes some way to rectifying that.

We could probably just copy over parts of the (or the entire) quickstart guide into the README, which should save some people with particularly short attention spans an extra click.

ptersilie:
I will have a go at the README then.

ltratt:
Thanks! As well as doing this, it might be worth checking that the quickstart guide is still 100% accurate and as simple as it could be. It wouldn't surprise me if it could be improved.

rljacobson:
[...]

PR #158, opened 2020-02-22
Briefly document major differences relative to Lex/Yacc.

ltratt:
This is intended to start addressing https://github.com/softdevteam/grmtools/issues/133#issuecomment-589827654 -- we could, and probably should, do more, but this is a start at least.


doc/src/lexcompatibility.md
```diff
@@ -0,0 +1,27 @@
+# Lex compatibility
+
+grmtools currently supports one common use of Lex, which is to produce a
+sequence of tokens. All Lex files require at least some porting to grmtools,
+though in many cases this is fairly trivial. Nevertheless aspects such as
```
ptersilie: `,` after `nevertheless`?

doc/src/lexcompatibility.md
```diff
@@ -0,0 +1,27 @@
...
+
+There are several major differences between Lex and grmtools:
+
+ * Lex has its own regular expression language whereas grmtools uses the well
+   known Rust [regex crate](https://crates.io/crates/regex) for regular
+   expressions. These two regular expression languages are very similar, but
+   complex regular expressions might not be supported under one or the other.
+
+ * Lex files consist of a sequence of regular expressions and an action for each.
+   grmtools lex files consists of a sequence of regular expressions and simple
```
ptersilie: s/simple/simply/

doc/src/lexcompatibility.md
```diff
@@ -0,0 +1,27 @@
...
+   a token name. Actions are not currently supported (and, by extension, nor
+   are special action expressions such as `ECHO` and `REJECT`).
+
+ * Start conditions, character sets, and changes to internal array sizes are
```
ptersilie: With `Start conditions`, do you mean context sensitivity?

ltratt: It's a Lex feature: I've never thought much about it to be honest, so I don't know it in detail.

ltratt:
Fixed in https://github.com/softdevteam/grmtools/pull/158/commits/239d85c398f724627000a6fa6915c218d90231e2.

doc/src/lexcompatibility.md
ptersilie: I think we should also mention that we do not support context sensitivity, which Lex does:

> Sometimes it is desirable to have several sets of lexical rules to be applied at different times in the input. For example, a compiler preprocessor might distinguish preprocessor statements and analyze them differently from ordinary statements.

ptersilie: Actually, nevermind. "Start conditions" is used for context-sensitivity (see Chapter 10 in the Lex doc).

ltratt:
OK to squash?

ptersilie:
Please squash.

ltratt:
Squashed.

PR #162, opened 2020-02-25
Remove stray 'dbg'.

ltratt:
This shouldn't have ever squeezed in!


ltratt:
@ptersilie I'm not sure why this died: buildbot seemed to work OK. Can we retry?

PR #163, opened 2020-02-25
More book bits

ltratt:
More improvements to the grmtools book.


doc/src/errorrecovery.md
```diff
@@ -329,6 +329,63 @@ With this, the `Delete +` repair sequence is consistently favoured over `Insert
...
+. "UNKNOWN"
+```
+
+Any single character which is not matched by any other lex rule will now lead
+to a token of type `UNKNOWN`. Note that it is vital that this is the last rule
+in your `.l` file, and that only a single character is matched, otherwise you
+will incorrectly lex correct input as `UNKNOWN`!
+
+We then need to add a dummy rule to your `.y` file, simply so that `lrpar`
+knows about `UNKNOWN` tokens. This dummy rule won't be referenced by other rule
```
ptersilie: s/rule/rules

ltratt: Fixed in https://github.com/softdevteam/grmtools/pull/163/commits/82e2b256527d8c53550433fae86e3a87b8ac84b7.

ptersilie:
Looks good. Please squash.

ltratt:
Squashed.

PR #166, opened 2020-02-28
Update example in README

TennyZhuang:
Use explicit dyn trait to disable warning by rustc.

```
warning: trait objects without an explicit `dyn` are deprecated
 --> src/main.rs:5:29
  |
5 | fn main() -> Result<(), Box<std::error::Error>> {
  |                             ^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn std::error::Error`
  |
  = note: `#[warn(bare_trait_objects)]` on by default
```


ltratt:
Good spot -- thanks!

PR #169, opened 2020-03-05
Use "unmatched" rather than "unknown".

ltratt:
Following Wagner's precedent, I think "unmatched" is a better name than "unknown" (even though, ironically, we now match unmatched content...).


ltratt:
`cargo deny` doesn't build at the moment :/

ltratt:
@ptersilie I think we can try-re-r+-ing this now.

PR #174, opened 2020-03-28
decouple lexer & text lifetimes

valarauca:
Happy to make/remake any changes that are requested. Great project BTW!

I think I found a small error when working with the `lrpar` library.

This is the rough example: [original source code (sorry for the mess)](https://github.com/valarauca/trollparse/blob/a2a170ad534546c881bea89ba32fc4a3e62b7f5e/src/lib.rs#L31-#L36), the [super::error_handling](https://github.com/valarauca/trollparse/blob/a2a170ad534546c881bea89ba32fc4a3e62b7f5e/src/lib.rs#L39-#L56) is just something to give me syntax highlighting on errors.

```rust

    /// handles parsing of the input and attempts to return an AST
    pub fn parser<'a>(input: &'a str) -> Result<Source<'a>,String> {
        let l = lexerdef();
        let lexer_def = l.lexer(input);
        super::error_handling(input, parse, &lexer_def)
    }
 ````

Gives me the following error

```
error[E0515]: cannot return value referencing local variable `lexer_def`
  --> src/lib.rs:35:9
   |
35 |         super::error_handling(input, parse, &lexer_def)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^----------^
   |         |                                   |
   |         |                                   `lexer_def` is borrowed here
   |         returns a value referencing data owned by the current function

error[E0515]: cannot return value referencing local variable `l`
  --> src/lib.rs:35:9
   |
34 |         let lexer_def = l.lexer(input);
   |                         - `l` is borrowed here
35 |         super::error_handling(input, parse, &lexer_def)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

This confuses me because [`Source<'a>`](https://github.com/valarauca/trollparse/blob/master/src/parser2.y#L102-#L109) has references to the underlying source code, but not fields/elements owned by the lexer.

I think the problem is with [`impl Lexer<StorageT>` is created](https://docs.rs/lrlex/0.6.2/lrlex/struct.LexerDef.html#method.lexer) 

```
    pub fn lexer<'a>(&'a self, s: &'a str) -> impl Lexer<StorageT> + 'a
```

The `LexerDef` equates its own `&self` lifetime to that of the input buffer. Which means `impl Lexer<StorageT>` will live as long as our input buffer, but our input buffer also cannot outlive `impl Lexer<StorageT>`. 

I am _attempting_ to decouple this, because while we do what `impl Lexer` to keep the input buffer alive (if necessary) we also want to ensure that the input buffer can outlive `impl Lexer`.

This is why I'm suggesting the introduction of the `'a: 'b` to a few places, _but I'm not sure I hit them all_. The basic idea is stating `'input: 'lexer` or "_the input may outlive the lexer, but the lexer may not outlive the input_". 

**TL;DR** 

Right now I see a generated function of the signature:

````
pub fn parse<'input>(
    lexer: &'input (dyn Lexer<u32> + 'input)
) -> (Option<Result<Source<'input>, Lexeme<u32>>>, Vec<LexParseError<u32>>)
````

And I think it should be

````
pub fn parse<'lexer, 'lexer: 'input>(
    lexer: &'lexer (dyn Lexer<u32> + 'input)
) -> (Option<Result<Source<'input>, Lexeme<u32>>>, Vec<LexParseError<u32>>)
````

But I'm not sure where to make this change.

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -1,7 +1,8 @@
 use std::{
     collections::{HashMap, HashSet},
     hash::Hash,
-    slice::Iter
+    slice::Iter,
+    marker::PhantomData,
 };
 
 use num_traits::{PrimInt, Unsigned};
@@ -56,13 +57,13 @@ impl<StorageT: Copy + Eq + Hash + PrimInt + Unsigned> LexerDef<StorageT> {
     }
 
     /// Get the `Rule` at index `idx`.
-    pub fn get_rule(&self, idx: usize) -> Option<&Rule<StorageT>> {
+    pub fn get_rule<'a>(&'a self, idx: usize) -> Option<&'a Rule<StorageT>> {
         self.rules.get(idx)
     }
 
     /// Get the `Rule` instance associated with a particular lexeme ID. Panics if no such rule
     /// exists.
-    pub fn get_rule_by_id(&self, tok_id: StorageT) -> &Rule<StorageT> {
+    pub fn get_rule_by_id<'a>(&'a self, tok_id: StorageT) -> &'a Rule<StorageT> {
         &self
             .rules
             .iter()
@@ -71,7 +72,7 @@ impl<StorageT: Copy + Eq + Hash + PrimInt + Unsigned> LexerDef<StorageT> {
     }
 
     /// Get the `Rule` instance associated with a particular name.
-    pub fn get_rule_by_name(&self, n: &str) -> Option<&Rule<StorageT>> {
+    pub fn get_rule_by_name<'a>(&'a self, n: &str) -> Option<&'a Rule<StorageT>> {
         self.rules
             .iter()
             .find(|r| r.name.as_ref().map(String::as_str) == Some(n))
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by ltratt** [comment]
Thanks for this. I'm currently under the weather, so I might be a little slower reviewing this than normal. Sorry!

**Comment by valarauca** [comment]
No worries! Stay healthy, especially now. 

The change is a little incomplete, I think the `Lexer` trait might need to gain lifetimes, or the generated `parse` function gain arguments. I'm not sure which is _better_.

**Comment by ltratt** [comment]
> Right now I see a generated function of the signature:

This is in `ctbuilder.rs` at line 464 (you'd also want to fix the two variants below it in the same way). It's in the code generator i.e. it's basically a `print` statement of Rust code. This part of the code is fairly awful!

To understand what's going on, I made that change myself, which opens up the next round of problems, and soon enough I found myself a deep hole. I think I have exceeded my Rust borrow-fu!

It would be really interesting to see if this problem can be solved, as I think you have put your finger on a real problem.

**Comment by valarauca** [comment]
Thanks for the guidance I'll take a dive in later this afternoon.

**Comment by ltratt** [comment]
Best of luck! I think in most cases you'll end up dealing with explicitly named lifetimes `<'lexer, 'input: lexer'>` which are nice and clear. Can I suggest that when you have just `a`, `b`, you order them, so you end up with `<'a, 'b: 'a>`? I found that a bit easier to pick apart than putting `b` first. This may just be my alphabetical OCD going haywire though!

**Comment by valarauca** [comment]
Yeah that's no problem.

**Comment by valarauca** [comment]
> I made that change myself, which opens up the next round of problems, and soon enough I found myself a deep hole.

You weren't joking! I feel like I stepped right into it. 

I actually managed to get pretty much all of the `grmtools` tests passing. **BUT** any example that compiles generated code (and my own project, using generated code) was throwing lifetime errors within the generated functions.

---

This might be a few days.

**Comment by ltratt** [comment]
You have my sympathies! If there's anything I can do to help, please shout!

**Comment by valarauca** [comment]
Passed tests locally, and my project using this crate worked with just a little lifetime twisting in the error handler function. 

---

Let me know if you need anything changed (like the git history cleaned up).

**Comment by ltratt** [comment]
This looks very promising! To understand a bit more about what's going on, I made a minimal test case:

```
$ cat src/t.l
%%
[a-z] "ID"
$ cat src/main.rs
use lrlex::lrlex_mod;
use lrpar::Lexer;

lrlex_mod!("t.l");

fn main() {
    let lexerdef = t_l::lexerdef();
    let input = "a";
    let t = {
        let lexer = lexerdef.lexer(&input);
        let lx = lexer.iter().next().unwrap().unwrap();
        lexer.span_str(lx.span())
    };
    dbg!(t);
}
```

On `master` this fails to compile:

```
error[E0597]: `lexer` does not live long enough
  --> src/main.rs:12:9
   |
9  |     let t = {
   |         - borrow later stored here
...
12 |         lexer.span_str(lx.span())
   |         ^^^^^ borrowed value does not live long enough
13 |     };
   |     - `lexer` dropped here while still borrowed

```

With this PR, it does compile, so that's a very clear indication that the current API is not adequate. So my next question was then, is this PR the "minimal" (for a very loose definition of "minimal"!) that enables this to compile. I wondered, in particular, if the `'lexer` lifetime on `Lexer` was useful, so I removed that from the whole of grmtools, and the little program above still compiles.

That makes me wonder if we can simplify this PR somewhat, or whether there are additional use cases beyond the simple example above that this PR can handle, but which I hadn't considered. Any thoughts?

valarauca:
Dug through trying to slim it down. The `cpctplus` and `mf` changes appear to be required. Omitting lifetimes from `Parse` causes errors, trying to merge (repeat use of the same lifetime) causes some very weird errors. 

I was able to remove a few spurious changes.

ltratt:
Again, I have to apologise that poor health is delaying me looking in detail at this. Sorry!

valarauca:
No worries! Just focus on getting better.

ltratt:
@valarauca I'm sorry it's taken me so long to get to this PR, but if it's any consolation you provided me with a lot to think about when I was ill! I'm absolutely in awe of your work here, and in the last couple of days I've tried to get clear in my head what you've done here, because I just wouldn't have believed this PR was possible before you showed that it is possible! That led me down several false routes while I tried to get my head around both grmtools use of lifetimes, and also the deeper recesses of the borrow checker.

In the end I've ended up with an approach that is **heavily** influenced by your approach, but which also seems in some places to be a bit simpler. That makes me suspicious that I might have simplified *too* much and that my version (in https://github.com/softdevteam/grmtools/pull/177) doesn't solve the problem that motivated this PR. Might you be able to take a look at https://github.com/softdevteam/grmtools/pull/177 and see if it solves your problem or not? If it doesn't, then clearly I have not understood the problem adequately yet and need to do some more thinking.

valarauca:
Sure i'll take a look.

PR #177, opened 2020-04-26
More flexible input lifetime

ltratt:
This is based on https://github.com/softdevteam/grmtools/pull/174: it tries to decouple the lifetime of a lexer from its input. That PR is a work of near genius: I simply would not have imagined that the outcome it achieves is possible without the PR as a proof-of-existence. The only minor problem was that I couldn't work out how it achieved its effect. I therefore tried to simplify the PR, but didn't get very far. I then tried reimplementing it, and didn't get very far with that either.

This PR is the result of me taking a different approach. First I simplified and unified the existing lifetimes in grmtools, because there were several inconsistencies, which I thought might be responsible for some of the pain in #174. Once that was done I could then add an explicit `'input` lifetime in https://github.com/softdevteam/grmtools/commit/91dab3fd5f21347e87441bb7c1b34493d519ac58 which I *think* achieves the same effect as #174. Certainly it is enough to allow this program to now compile:

```rust
      fn main() {
          let lexerdef = t_l::lexerdef();
          let input = "a";
          let t = {
              let lexer = lexerdef.lexer(&input);
              let lx = lexer.iter().next().unwrap().unwrap();
              lexer.span_str(lx.span())
          };
      }
```

where previously rustc would complain:

```
error[E0597]: `lexer` does not live long enough
  --> src/main.rs:12:9
   |
9  |     let t = {
   |         - borrow later stored here
...
12 |         lexer.span_str(lx.span())
   |         ^^^^^ borrowed value does not live long enough
13 |     };
   |     - `lexer` dropped here while still borrowed

```

However, I am not sure if this PR is able to handle all the same cases as #174. I'm hoping that @valarauca will be able to let me know if this solves the problem that led him to create #174.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,31 @@
+# grmtools 0.7.0 (2020-XX-XX)
+
+## Breaking change
+
+* `Lexer` now takes a lifetime `'input` which allows the input to last longer
+  than the `Lexer` itself. `Lexer::span_str` and `Lexer::span_lines_str` have
+  changed from:
+    ```rustc
+    fn span_str(&self, span: Span) -> &str;
+    fn span_lines_str(&self, span: Span) -> &str;
+    ```
+  to:
+    ```rustc
+    fn span_str(&self, span: Span) -> &'input str;
+    fn span_lines_str(&self, span: Span) -> &'input str;
+    ```
+  This change allows users to throw away the `Lexer` but still keep around
+  structures (e.g. ASTs) which reference the user's input.
+
+  rustc infers the `'input` lifetime in some situations but not others,
+  so if you get an error:
+    ```
+    error[E0106]: missing lifetime specifier
+    ```
+  then it is likely that you need to change a type from `Lexer` to
+  `Lexer<'input>`.
+
+
 # grmtools 0.6.2 (2020-03-22)
 
 * Fix two Clippy warnings and suppress two others.
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     hash::Hash,
+    marker::PhantomData,
     slice::Iter
 };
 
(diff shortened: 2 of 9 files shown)


valarauca:
It does not.

The minimal test case you want to use is something like:

```rust

lrlex_mod!("my_lexer_module");
lrpar_mod!("my_parser_module");

pub fn parse_my_data<'a>(input: &'a str) -> Option<crate::my_parser_module::ParseResult<'a>> {
    let lexer_def = crate::my_lexer_module::lexerdef();
    let l = lexer_def.lexer(input);
    match crate::my_parser_module::parser(&l) {
        (Option::Some(Ok(x)),_) => Some(x),
        _ => None
    }
}
```

I roughly implemented [that here](https://github.com/valarauca/trollparse/blob/cody/testing_flexible_lifetimes/src/lib.rs#L35-#L43). It gives me: `error[E0597]`. That's same case I used for testing the change I pushed to this project.

The critical issue I believe is [here](https://github.com/ltratt/grmtools/blob/more_flexible_input_lifetime/lrpar/src/lib/lex.rs#L39). The resulting trait needs to track 2 separate lifetimes, the [input source lexer itself](https://github.com/valarauca/grmtools/blob/master/lrpar/src/lib/lex.rs#L39) need to have separate lifetimes.


This is why the lifetimes tend to "_explode_" when you get to internal types as you're borrowing 2 lifetimes explicitly, [so you need a 3rd lifetime](https://github.com/valarauca/grmtools/blob/master/lrpar/src/lib/ctbuilder.rs#L464). Then later in `Parser` you borrow the same 2 lifetimes, soyou still need a 3rd. But since everything doesn't [live the same amount of time (borrowing in multiple places)](https://github.com/valarauca/grmtools/blob/master/lrpar/src/lib/parser.rs#L741), you [need a 4th lifetime](https://github.com/valarauca/grmtools/blob/master/lrpar/src/lib/parser.rs#L89) to ensure the compiler is aware each borrow in `Parser` isn't identical.

ltratt:
OK, I can see what your example is doing now. Clearly I haven't yet fully understood #174 yet! I'll keep at it -- this is fascinating stuff!

ltratt:
@valarauca https://github.com/softdevteam/grmtools/pull/177/commits/c8155248730e5f7553b2e8b1fa84eeeea7412a07 now allows your example to compile. I did engage in one cheat, changing `lrlex::LexerDef` from:

```rustc
    pub fn lexer(&self, s: &str) -> impl Lexer<StorageT> {
        LRLexer::new(self, s)
    }
```

to:

```
    pub fn lexer<'lexer, 'input: 'lexer>(
        &'lexer self,
        s: &'input str
    ) -> LRLexer<'lexer, 'input, StorageT> {
        LRLexer::new(self, s)
```

AFAICS the only reason to do this is so that `Lexer` only needs an `'input` lifetime (and not a `'lexer` lifetime as well) -- I don't *think* it saves us from adding extra lifetimes internally. Warning: I might be wrong.

Do you think this commit is flexible enough or are there other things that it doesn't allow to compile?

valarauca:
Sure I'll give it a shot

valarauca:
I depended on `4bc7ee3` and it worked :D

ltratt:
@valarauca  Cool, thanks so much for trying this -- I really can't emphasise enough how much of an eye-opener your PR has been for me!

I hate doing this (because I would dislike it being done to me), but I think this PR, as it's simpler, is probably the better candidate for merger relative to #174. Are you OK with that? And do you have any comments on any of the code? If we do end up merging this PR, I'd like to make sure it's as good as it can be.

valarauca:
> Are you OK with that? 

Of course. I can close #174 it is nbd. 

> And do you have any comments on any of the code?

I took a look across it; LGTM.

Thanks for your hard work and help getting this ready to go 👍

ltratt:
Thanks for being a good sport @valarauca ! I'm going to ask @ptersilie to do a lightweight review of the PR, but since you're happy with it, I expect that will be a simple task for him.

@ptersilie I will need to squash at least one commit away before merging.

ptersilie:
Typo in commit 8d4138246e7628e0e00fdae0b1d13e103528d3b2? `probably be easier to change it [in] one go later.`

ltratt:
Agreed.

ptersilie:
Just one minor typo which you can fix during squashing. Another thought I had was, since this PR is based on @valarauca's work, maybe we could add him as a co-author to some of the commits where that makes sense? Thoughts?

ltratt:
Excellent idea! Let me squash and do that at the same time.

ptersilie:
Ok, please go ahead.

ltratt:
OK, done: I've added @valarauca to the first two commits as those are clearly based off his work (I feel I should shoulder the blame for the latter commits!) and squashed.

ltratt:
Oops! That last commit will need squashing.

bors try

ptersilie:
Please squash.

ltratt:
Squashed.

PR #180, opened 2020-04-30
Various Clippy suggestions.

ltratt:


ltratt:
Hmm, that looks like a bors problem and not a buildbot issue. Probably r plusing it again will do the trick.

Issue #185, opened 2020-05-17
Non-Relative Paths given to `include_bytes!`


valarauca:
I'm working with a `build.rs` that is roughly:

```rust
let rules_id = CTParserBuilder::new()
    .yacckind(YaccKind::Grmtools)
    .recoverer(RecoveryKind::None)
    .error_on_conflicts(true)
    .process_file(
        "src/internals/parser/generated/parser.y",
        "src/internals/parser/generated/parser.rs")?;
```

Now the `build.rs` works. My parser gets generated. But as the project get's compiled I see:

```
error: couldn't read src/internals/parser/generated/src/internals/parser/generated/parser.grm: No such file or directory (os error 2)
error: couldn't read src/internals/parser/generated/src/internals/parser/generated/parser.sgraph: No such file or directory (os error 2)
error: couldn't read src/internals/parser/generated/src/internals/parser/generated/parser.stable: No such file or directory (os error 2)
```

Within my generated `parser.rs` I see:

```
::lrpar::ctbuilder::_reconstitute( include_bytes!("src/internals/parser/generated/parser.grm"),
include_bytes!("src/internals/parser/generated/parser.sgraph"),
include_bytes!("src/internals/parser/generated/parser.stable"));
```

The problem is [`include_bytes!`](https://doc.rust-lang.org/std/macro.include_bytes.html) specifies

> The file is located relative to the current file

ltratt:
Fixed by https://github.com/softdevteam/grmtools/pull/186.

valarauca:
You might want to increment the version number. Cargo is still fetching the broken version.

ltratt:
0.8.0 is now released!

PR #186, opened 2020-05-17
attempt to fix relative pathing error

valarauca:
I threw this together in an attempt to address: https://github.com/softdevteam/grmtools/issues/185

I'm not sure if you want to fix it in house, or use an external crate.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -6,10 +6,10 @@ use std::{
     convert::AsRef,
     env::{current_dir, var},
     error::Error,
-    fmt::{self, Debug},
+    fmt::{self, Debug, Write as fmtWrite},
     fs::{self, create_dir_all, read_to_string, File},
     hash::Hash,
-    io::Write,
+    io::{self, Write},
     marker::PhantomData,
     path::{Path, PathBuf}
 };
@@ -24,7 +24,7 @@ use lazy_static::lazy_static;
 use lrtable::{from_yacc, statetable::Conflicts, Minimiser, StateGraph, StateTable};
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
 use regex::Regex;
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Serialize};
 
 use crate::RecoveryKind;
 
@@ -34,10 +34,11 @@ const ACTIONS_KIND: &str = "__GTActionsKind";
 const ACTIONS_KIND_PREFIX: &str = "AK";
 const ACTIONS_KIND_HIDDEN: &str = "__GTActionsKindHidden";
 
-const GRM_FILE_EXT: &str = "grm";
 const RUST_FILE_EXT: &str = "rs";
-const SGRAPH_FILE_EXT: &str = "sgraph";
-const STABLE_FILE_EXT: &str = "stable";
+
+const GRM_CONST_NAME: &str = "__GRM_DATA";
+const SGRAPH_CONST_NAME: &str = "__SGRAPH_DATA";
+const STABLE_CONST_NAME: &str = "__STABLE_DATA";
 
 lazy_static! {
     static ref RE_DOL_NUM: Regex = Regex::new(r"\$([0-9]+)").unwrap();
@@ -293,11 +294,6 @@ where
             .collect::<HashMap<_, _>>();
         let cache = self.rebuild_cache(&grm);
 
-        // out_base is the base filename for the output (e.g. /path/to/target/out/grm_y) to which
-        // we will write filenames with various extensions below.
-        let mut outp_base = outp.as_ref().to_path_buf();
-        outp_base.set_extension("");
-
         // We don't need to go through the full rigmarole of generating an output file if all of
         // the following are true: the output file exists; it is newer than the input file; and the
         // cache hasn't changed. The last of these might be surprising, but it's vital: we don't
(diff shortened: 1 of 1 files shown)


ltratt:
I agree that #185 is a problem that needs fixing, so thanks for looking into this! I don't know whether we should fix this in-house or use a crate. A quick look suggested that something like https://docs.rs/include_dir/0.6.0/include_dir/ might solve things for us? Then my question would probably be whether we think our solution is stable going forward (at which point not dragging in another dependency is probably a good thing) or whether we want to allow some other poor soul to keep things working in the future (at which point a dependency is probably worth the cost). Any thoughts?

valarauca:
> Any thoughts?

The whole situation could also be avoided by just serializing into a `const NAME: &'static [u8]` instead of other files.

Generated sources almost always lead to "_ugly diffs_" in git. I doubt this would surprise/annoy down stream users. As an example [`grpc`/`protobuf`](https://github.com/grpc/grpc-go/blob/master/examples/route_guide/routeguide/route_guide.pb.go#L315) include the base source used to generate the file, so it can communicate metadata.

ltratt:
> The whole situation could also be avoided by just serializing into a const NAME: &'static [u8] instead of other files.

I'm not sure I understand -- can you expand a little?

ltratt:
> The whole situation could also be avoided by just serializing into a const NAME: &'static [u8] instead of other files.

Oh, I have just realised what you mean!

Are there any downsides to doing this? If not, maybe that's the most stable long-term fix? I guess we would need to check that it doesn't tank compile times though.

ltratt:
[The particular mechanism we use is intentionally hidden from the user as the idea "one day" (in the mystical perfect future) is that we generate much better code for the statetable. So provided none of our changes leak out to the user, I don't have a strong opinion on how we do things -- `include_bytes` was just the first thing that came to my mind when I implemented this way back when.]

valarauca:
Instead of serializing the data to disk, as a file, and loading that file via `include_bytes!()`.

The data could just be serialized to a Rust-Array within the generated source code directly.

ltratt:
> The data could just be serialized to a Rust-Array within the generated source code directly.

Yes, I was slow to realise what you meant.

If this doesn't tank compile times, I think it might be the neatest solution? It's certainly the least code to maintain.

ltratt:
Hello @valarauca, I realise my last message probably wasn't very clear. Basically I'm open-minded about which solution you think is best (the one you've initially done in the PR or the "embed everything in the string"), and would welcome your opinion about which you think is the best long-term fix. Any thoughts?

valarauca:
I think the best long term solution would be the constant.

ltratt:
Would you like to do that for this PR, or shall we merge this then do the constant?

valarauca:
I'll take a crack at this afternoon.

valarauca:
2 new commits.

`inline_constants` is the bulk of the work. It does indeed convert the values into inline constants.

The second commit modifies the formatting. I was originally using hexadecimal for the generated byte arrays, but decimal seems to save a fair amount of space.

On a local project, with a fairly trivial grammar. Pre-Patch the generated file was `19KiB`, Post-Patch it is `30KiB`. Using Hexadecimal the generated file size was `40KiB`, that is why I patched in decimal support. 

---

I haven't seen a change in compile times.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -492,22 +492,13 @@ where
...
         };
 
-        // Because we're lazy, we don't write our own serializer. We use serde and bincode to
-        // create files $out_base.grm, $out_base.sgraph, and $out_base.out_stable which contain
-        // binary versions of the relevant structs, and then include those binary files into the
-        // Rust binary using "include_bytes" (see below).
-        let out_grm = self.bin_output(&outp_base, GRM_FILE_EXT, &grm)?;
-        let out_sgraph = self.bin_output(&outp_base, SGRAPH_FILE_EXT, &sgraph)?;
-        let out_stable = self.bin_output(&outp_base, STABLE_FILE_EXT, &stable)?;
+        // bincode constants are deserialized within
```
ltratt, line 495: This comment seems a bit unfinished?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -892,57 +869,57 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
...
+        Self {
+            buffer: format!(r#"#[allow(dead_code)] const {}: &[u8] = &["#, name),
         }
-        Option::None => {
-            // nothing in the paths is shared.
-            PathBuf::from(other)
+    }
+
+    /// complete the array, and return the finished string
+    fn finish(self) -> String {
```
ltratt, line 897: I *think* (but I might be wrong) we could mildly simplify this to:

```rust
fn finish (mut self) {
    buffer.push_str(...);
    buffer
}
```

?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -892,57 +869,57 @@ pub fn _reconstitute<'a, StorageT: Deserialize<'a> + Hash + PrimInt + Unsigned>(
...
+        let mut buffer = self.buffer;
+        buffer.push_str("];\n");
+        buffer
+    }
+}
+impl Write for ArrayWriter {
+    #[allow(dead_code)]
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        use std::fmt::Write;
```
ltratt, line 906: This should probably go with the other `use`s at the top of the file?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -906,7 +906,7 @@ impl Write for ArrayWriter {
         use std::fmt::Write;
         for b in buf {
             // specifying hexidecimal format.
```
ltratt, line 908: I think we can remove this comment?

valarauca:
Good feedback. I'll get them addressed ASAP.

lrpar/src/lib/ctbuilder.rs
valarauca: I think we'll still need the `self.` prefix. But I understand what you're saying.

ltratt: Mea culpa! Yes, that was silly of me.

lrpar/src/lib/ctbuilder.rs
valarauca: I can just remove it.

ltratt: Works for me.

lrpar/src/lib/ctbuilder.rs
valarauca: I don't think so. We already have `std::io::Write` within the namespace, so bringing in `std::fmt::Write` generates a collision.

ltratt: That's a good point -- I hadn't encountered that particular clash before. I *think* you can use `as` to rename the two things out of the way of each other. Whether that's worth doing or not is another thing.

valarauca: Hmm the `as` alias would work.

valarauca: Updated the PR to use it.

ltratt:
This looks good to me! Can you `cargo fmt` (alas, using nightly `rustfmt` :/ )? Then you can squash the commits down as you see fit, and we can merge.

ltratt:
Thanks!

bors r+

PR #196, opened 2020-05-30
change generated module scope to public

valarauca:
The contents are not accessible by default.


ltratt:
Is there a particular use case for this?

ltratt:
Ah, I see what the use case is, because you can't do `pub lrpar_mod!(...)` and so on. However, I don't think we can make things `pub` by default. I think `CTParserBuilder` would need to gain something like a `visibility(v: Visibility)` function where `Visibility` is an `enum` {Private, Pub, PubCrate} and so on (there are actually [more visibility types](https://doc.rust-lang.org/reference/visibility-and-privacy.html) than I realised), defaulting to `Private`.

valarauca:
I'm not sure I understand why that would be necessary?

The current behavior is the [module is private](https://github.com/valarauca/foxhole/blob/706cb286f688969a50631923a3ceb873dde14640/src/internals/parser/generated/parser.rs#L1), while the only "_interesting_" public items are [`token_epp`](https://github.com/valarauca/foxhole/blob/706cb286f688969a50631923a3ceb873dde14640/src/internals/parser/generated/parser.rs#L456) & [`parser`](https://github.com/valarauca/foxhole/blob/706cb286f688969a50631923a3ceb873dde14640/src/internals/parser/generated/parser.rs#L376). I guess [some constants are publicly visible](https://github.com/valarauca/foxhole/blob/706cb286f688969a50631923a3ceb873dde14640/src/internals/parser/generated/parser.rs#L422-#L437) but I don't see how that could be abused/depended upon?

ltratt:
> I'm not sure I understand why that would be necessary?

I don't think we should place something public in a file without "permission" (so to speak). Users might not realise that `lrpar_mod!` makes something public by default, and that they're thus exporting that as part of their interface.

valarauca:
Ah I see. 

I can add a builder pattern flag for that.

ltratt:
Cool :)

valarauca:
I updated the builder pattern, and my local project.

Let me know if there are any review comment and/or squashing, styling, etc.

ltratt:
Thanks! I think we really do need to have an `enum` though with the various possibilities listed in https://doc.rust-lang.org/reference/visibility-and-privacy.html (perhaps along the lines sketched out in https://github.com/softdevteam/grmtools/pull/196#issuecomment-636431563) because there's more than just public/private (I was surprised that I didn't know all the variants!). So for your particular use case you'd probably end up doing something like `CTBuilder::new()...visibility(Visibility::Public)`.

BTW, it's easier for me to review if you push new commits without force pushing / squashing. When we've gone through the whole review process, I'll give you a chance to squash as much as you want before merging. We've found this works well across a number of projects (well, we stole this pattern of working from Rust, so we can't claim credit for it!).

valarauca:
Sounds good. I'll see what I can do. 


The force push was me really screwing things up locally. I didn't mean for that to be public. I assumed a final squash would remove it. Sorry about that.

valarauca:
Visibility enum added. 

Let me know if you like the `fmt::Display` route I went with, or would prefer a different serialize option.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -92,6 +92,59 @@ where
 {
 }
 
+/// The scope of public visibility.
+#[derive(Clone,PartialEq,Eq,Debug,Hash)]
+pub enum Scope {
+    /// Represents `pub(crate)`
+    Crate,
+    /// Represents `pub(self)`
+    _Self,
```
ltratt, line 101: A `_` prefix means something in Rust ("unused"), so `Self_` is probably preferable.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -92,6 +92,59 @@ where
...
+    /// `pub(in String)`
+    In(String),
+}
+
+/// Visibility controls how public generated code is.
+#[derive(Clone,PartialEq,Eq,Debug,Hash)]
+pub enum Visibility {
+    Private,
+    Public(Option<Scope>),
+}
```
ltratt, line 113: My OCD can't help but ask for 1 blank line between `enum`/`impl`/etc.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -92,6 +92,59 @@ where
...
+    Super,
+    /// `pub(in String)`
+    In(String),
+}
+
+/// Visibility controls how public generated code is.
+#[derive(Clone,PartialEq,Eq,Debug,Hash)]
+pub enum Visibility {
+    Private,
+    Public(Option<Scope>),
```
ltratt, line 112: I wonder if we're being a bit too clever here? I think it's fine for us to have a single `enum` with `Public/Private/Crate/Self_/In(String)` which would be easier to use. That would also get rid of the `Option` and `Scope`.

ltratt:
I like the general idea, but I do wonder if we've gone a bit too complex here: I think we can simplify the two `enum`s into one, and drop the constructor (I think people can manage `Visibility::Public` and so on). I also wondered if we're abusing `Display` here, although I could be persuaded otherwise.

lrpar/src/lib/ctbuilder.rs
valarauca: Sure.

valarauca:
Sounds good. I can remove the `fmt::Display` it was kind of a hack, and glob the 2 enums together.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -92,67 +93,43 @@ where
...
-    /// Will generate the module with the visibility `pub(in {arg})`.
-    pub fn public_in<T: AsRef<str>>(s: T) -> Self {
-        Self::Public(Some(Scope::In(s.as_ref().to_string())))
-    }
-}
-impl fmt::Display for Visibility {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+impl Visibility {
+    /// returns a string to serialize the result
+    pub fn get_str(&self) -> Cow<'static, str> {
```
ltratt, line 125: Does having the `Cow` help with the `format!` later? I ask that very naively!

valarauca: Not at all. I just just thinking of saving the extra allocation, it really doesn't matter. I can have it return a `String` if you prefer.

ltratt: If it saves an allocation, I'm all for it: I just couldn't work it out if it did or not. Does it definitely save an allocation on our common path?

valarauca: In the common path? I don't think so. The builder is normally invoked in the `build.rs`.

ltratt: Having thought about it, I see why `Cow` works here. I'm fine with it.

lrlex/src/lib/builder.rs
```diff
@@ -84,6 +87,13 @@ where
         self
     }
 
+    /// Set if the generated module should be public. This is only necessary if
```
ltratt, line 90: I'm not sure I understand the "this is only necessary" sentence: surely it's useful only *if* you use `lrlex_mod!`?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -171,6 +215,13 @@ where
         self
     }
 
+    /// Set if the generated module should be public. This is only necessary if
```
ltratt, line 218: Save comment about the second sentence as above.

valarauca: I was trying to capture your initial assertion about this not being necessary if the macro was used. Could you suggest alternative verbiage?

ltratt: I don't think I said (or, at least, I didn't mean!) that it's not necessary, but that it's not possible. You can't write `pub lrlex_mod!` for example.

I think a doc string such as "Set the visibility of the generated module to `vis` (e.g. public). Defaults to `Visibility::Private`." should be OK?

valarauca:
Updated the code with the suggested comment.

ltratt:
Please squash as you see fit (you don't have to squash to one commit, unless you think that's the right thing to do).

PR #201, opened 2020-06-11
Add a README for nimbleparse

ltratt:
nimbleparse is, I think, just about worthy of a public release which means it needs documenting. This PR does that, and then makes a couple of small consistency changes (e.g. making sure the grmtools book reflects the README).

### DIFF
--- a/doc/src/nimbleparse.md
+++ b/doc/src/nimbleparse.md
@@ -1,49 +1,177 @@
 # nimbleparse
 
-If you have a [`lrlex`](lrlex.html) compatible `.l` file and an
-[`lrpar`](lrpar.html) compatible `.y` file, you can use `nimbleparse` as a quick
-way of testing inputs and exploring the resulting parse tree:
+`nimbleparse` is a simple grammar debugging aid. It takes as input a Lex
+specification, a Yacc specification, and an input file and prints any warnings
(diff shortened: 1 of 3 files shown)


doc/src/nimbleparse.md
```diff
@@ -1,49 +1,177 @@
...
+errors and, when possible fixes. If parsing is successful, `nimbleparse` exits
+with 0; if an error is detected it exits with 1.
+
+The full command-line specification is as follows:
 
 ```
-$ cargo build --release
-$ target/release/nimbleparse -h
-Usage: nimbleparse [-r <cpctplus|mf|panic|none>] [-y <eco|original>] <lexer.l> <parser.y> <input file>
+nimbleparse [-r <cpctplus|none>] [-y <eco|grmtools|original>] [-q] <lexer.l> <parser.y> <input file>
```
vext01, line 13: `mf` isn't actually gone yet, right? So we are just not documenting it so that people don't use it?

What about `panic`?

vext01:
Just one comment.

doc/src/nimbleparse.md
ltratt: > mf isn't actually gone yet, right? So we are just not documenting it so that people don't use it?

Correct.

> What about panic?

That *might* disappear at some point. I certainly don't want people to rely on its existence at the moment.

PR #202, opened 2020-06-11
Prepare a 0.8.1 release.

ltratt:
Note that the intention is to make a release of nimbleparse on crates.io for the first time, so I've tentatively added the fields that I think crates.io will require. Guessing these in advance is always a bit random for me, so I might have got them wrong.


vext01:
lgtm.

bors r+

PR #205, opened 2020-07-02
Remove MF and panic mode

ltratt:
This PR first removes two recoverers which we have long since removed from the documentation (MF and Panic mode) and which we don't want to maintain (https://github.com/softdevteam/grmtools/commit/d1e599e1b240e2c1c5af2422a8b52d9950073afa).

Since MF required access to the stategraph, this then gives us the opportunity to remove the serialised stategraph in compile-time mode which shaves a useful amount off built binaries (https://github.com/softdevteam/grmtools/commit/a1ec4a004e6fff6b30400f156f6d02f17e21278a). At some point we should consider doing the same with the serialised grammar because we only need a subset of that at runtime (I think the gains from doing this to the statetable would, in general, be much less, as we use most of that struct at runtime), but that's for a future PR.


ptersilie:
Looks good. I've checked out the changes and `cargo test` ran fine. So I think we can merge this.

PR #206, opened 2020-07-02
Fix factor/Term names

ltratt:
Copy and paste, and a lack of thought, are dangerous bedfellows. As soon as I thought about what "factor" and "term" mean, I realised that we've got them the wrong way around! ["Factor" are the things either side of "*"; "Term" is "terminal".] Mea culpa!


ptersilie:
Looks good. Testing now. Anything to squash?

ltratt:
Squashed.

ptersilie:
Tests run. So this is good to merge.

Issue #212, opened 2020-09-02
Declaring extra parameters to be passed to parse with %parse-param


ratmice:
I couldn't find a way to do this in lrpar, but in bison there is a `%parse-param` directive which allows you to declare additional parameters to pass to the generated parse function,

I would guess there would also need to be something for lifetimes in addition to that.

ltratt:
Although it's not *exactly* what this issue is asking about, https://github.com/softdevteam/grmtools/pull/254 allows the user to define their own lexeme struct to which they can attach any data they want. I think that captures some of the use-cases that traditionally required `%parse-param` but definitely not all of them.

PR #214, opened 2020-09-20
WIP: %parse-param

ratmice:
This merely adds some basic scaffolding for %parse-param, issue #212,
It contains some ugly hacks/commented out code to get tests compiling with the unimplemented feature.
Mostly wanted to ask if this looks like an ok syntax for lifetimes.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -714,7 +714,7 @@ impl YaccParser {
 
 #[cfg(test)]
 mod test {
-    use std::{ collections::HashSet, iter::FromIterator };
+    use std::{collections::HashSet, iter::FromIterator};
 
     use super::{
         super::{
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -462,7 +462,7 @@ impl YaccParser {
 
         // Next, the '(' pattern : type, ... ')'
         i = self.parse_ws(i, false)?;
-        if let Some(_) = self.lookahead_is("(", i) {
+        if self.lookahead_is("(", i).is_some() {
             let mut j = i;
             let mut bindings: Vec<(String, String)> = Vec::new();
             while j < self.src.len() {
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -462,19 +462,23 @@ impl YaccParser {
 
         // Next, the '(' pattern : type, ... ')'
         i = self.parse_ws(i, false)?;
-        if let Some(mut j) = self.lookahead_is("(", i) {
+        if let Some(_) = self.lookahead_is("(", i) {
+            let mut j = i;
             let mut bindings: Vec<(String, String)> = Vec::new();
-
             while j < self.src.len() {
+                let c = self.src[j..].chars().next().unwrap();
+                j += c.len_utf8();
+
+                if c == ')' {
+                    break
+                }
+
                 // Some binding name, or pattern.
                 let k = self.parse_ws(j, false)?;
                 let (k, binding) = self.parse_to_single_colon(k)?;
-                let (k, typ, done) = self.parse_param_rust_type(k + ':'.len_utf8())?;
+                let (k, typ) = self.parse_param_rust_type(k + ':'.len_utf8())?;
                 j = k;
                 bindings.push((binding.trim_end().to_string(), typ));
-                if done == true {
-                    break;
-                }
             }
             if !bindings.is_empty() {
                 self.ast.parse_param_bindings = Some(bindings);
@@ -484,7 +488,10 @@ impl YaccParser {
         Ok(self.parse_ws(i, true)?)
     }
 
-    fn parse_param_rust_type(&mut self, i: usize) -> YaccResult<(usize, String, bool)> {
+    // Parse a rust type, followed by either a ',' character or an unbalanced ')'
+    // Return the char indice of the trailing character,
+    fn parse_param_rust_type(&mut self, i: usize) -> YaccResult<(usize, String)> {
+        let i = self.parse_ws(i, false)?;
         let mut j = i;
         let mut brace_count = 0;
 
@@ -494,9 +501,8 @@ impl YaccParser {
                 '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, j)),
                 ')' | ',' if brace_count == 0 => {
                     return Ok((
-                        j + c.len_utf8(),
-                        self.src[i..j].trim_start().trim_end().to_string(),
-                        c == ')',
+                        j,
+                        self.src[i..j].trim_end().to_string(),
                     ));
                 }
                 '(' | '{' | '[' | '<' => {
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -23,6 +23,7 @@ pub enum YaccParserErrorKind {
     IncompleteAction,
     MissingColon,
     MissingRightArrow,
+    MismatchedBrace,
     PrematureEnd,
     ProgramsNotSupported,
     UnknownDeclaration,
@@ -58,6 +59,7 @@ impl fmt::Display for YaccParserError {
             YaccParserErrorKind::IncompleteAction => "Incomplete action",
             YaccParserErrorKind::MissingColon => "Missing ':'",
             YaccParserErrorKind::MissingRightArrow => "Missing '->'",
+            YaccParserErrorKind::MismatchedBrace => "Mismatched brace",
             YaccParserErrorKind::PrematureEnd => "File ends prematurely",
             YaccParserErrorKind::ProgramsNotSupported => "Programs not currently supported",
             YaccParserErrorKind::UnknownDeclaration => "Unknown declaration",
@@ -434,11 +436,10 @@ impl YaccParser {
         if let Some(mut j) = self.lookahead_is("<", i) {
             let mut k = j;
             let mut lifetimes = HashSet::new();
-            let mut add_lifetime = |k, c: char| {
-                let s = self.src[j..k].trim_start().trim_end().to_string();
+            let mut add_lifetime = |j, k, c: char| {
+                let s = self.src[j..k].trim().to_string();
                 lifetimes.insert(s);
-                j = k + c.len_utf8();
-                j
+                k + c.len_utf8()
             };
 
             while k < self.src.len() {
@@ -446,11 +447,11 @@ impl YaccParser {
                 match c {
                     '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, k)),
                     ',' => {
-                        k = add_lifetime(k, ',');
-                        continue;
+                        k = add_lifetime(j, k, ',');
+                        j = k;
                     }
                     '>' => {
-                        k = add_lifetime(k, '>');
+                        k = add_lifetime(j, k, '>');
                         break;
                     }
                     _ => k += c.len_utf8(),
@@ -470,8 +471,8 @@ impl YaccParser {
                 j += c.len_utf8();
 
                 // Some binding name, or pattern.
-                let k = self.parse_ws(j, false)?;
-                let (k, binding) = self.parse_to_single_colon(k)?;
+                j = self.parse_ws(j, false)?;
+                let (k, binding) = self.parse_to_single_colon(j)?;
                 let (k, typ) = self.parse_param_rust_type(k + ':'.len_utf8())?;
                 j = k;
                 bindings.push((binding.trim_end().to_string(), typ));
@@ -503,14 +504,17 @@ impl YaccParser {
                     brace_count += 1;
                     j += c.len_utf8();
                 }
-                ')' | '}' | '>' | ']' if brace_count > 0 => {
+                ')' | '}' | '>' | ']' => {
+                    if brace_count == 0 {
+                        return Err(self.mk_error(YaccParserErrorKind::MismatchedBrace, j));
+                    }
                     brace_count -= 1;
                     j += c.len_utf8();
                 }
                 c => j += c.len_utf8(),
             }
         }
-        return Err(self.mk_error(YaccParserErrorKind::PrematureEnd, j));
+        Err(self.mk_error(YaccParserErrorKind::PrematureEnd, j))
     }
 
     /// Parse up to (but do not include) the end of line (or, if it comes sooner, the end of file).
@@ -710,6 +714,8 @@ impl YaccParser {
 
 #[cfg(test)]
 mod test {
+    use std::{ collections::HashSet, iter::FromIterator };
+
     use super::{
         super::{
             ast::{GrammarAST, Production, Symbol},
@@ -1797,10 +1803,9 @@ x"
             Err(e) => panic!("Incorrect error returned {}", e),
         }
     }
+
     #[test]
     fn test_parse_param() {
-        use super::HashSet;
-        use std::iter::FromIterator;
         let src = "
           %parse_param <'a, 'b> (x: &'a (), (y, z) : (Result<((), ()), ((), ())>, ((u32, u32), &'b ())))
           %%
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -426,32 +426,18 @@ impl YaccParser {
         }
     }
 
-    // This parses `%parse_param "<"lifetimes">" pat ":" type
-    // Eventually we need to bind type to a known variable name.
-    // And generate actions using pat.
-    //
-    // The single variable case of this is `%parse_param x : int`,
-    // while in the multi-variable case is `%parse_param (x, y) : (int, int)`
-    //
-    // To bind this to a known variable name we drop the pattern,
-    // `known_variable_name : (int, int)`.
-    //
-    // Then for user actions we recombine the pattern and the type.
-    // This all works out because patterns don't have single colons in them.
-    //
-    // This way we can void avoid knowing parameter names,
-    // and dealing commas and parenthesis in the pattern which is difficult
-    // since rust types can contain them.
+    // Handle parse_param declarations of the form:
+    // %parse_param <'a>(x: u32, y : (u32, u32))
     fn parse_param(&mut self, mut i: usize) -> YaccResult<usize> {
         i = self.parse_ws(i, false)?;
         // First gobble up all of the '<' lifetime ',' ... '>
         if let Some(mut j) = self.lookahead_is("<", i) {
             let mut k = j;
             let mut lifetimes = HashSet::new();
-            let mut add_lifetime = |k| {
+            let mut add_lifetime = |k, c: char| {
                 let s = self.src[j..k].trim_start().trim_end().to_string();
                 lifetimes.insert(s);
-                j = k + 1;
+                j = k + c.len_utf8();
                 j
             };
 
@@ -460,11 +446,11 @@ impl YaccParser {
                 match c {
                     '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, k)),
                     ',' => {
-                        k = add_lifetime(k);
+                        k = add_lifetime(k, ',');
                         continue;
                     }
                     '>' => {
-                        k = add_lifetime(k);
+                        k = add_lifetime(k, '>');
                         break;
                     }
                     _ => k += c.len_utf8(),
@@ -474,15 +460,57 @@ impl YaccParser {
             i = k;
         }
 
+        // Next, the '(' pattern : type, ... ')'
         i = self.parse_ws(i, false)?;
-        let (i, pat) = self.parse_to_single_colon(i)?;
-        let (i, typ) = self.parse_to_eol(i + ':'.len_utf8())?;
-        let pat = pat.trim_start().trim_end().to_string();
-        let typ = typ.trim_start().trim_end().to_string();
-        if !(pat.is_empty() || typ.is_empty()) {
-            self.ast.parse_param_bindings = Some((pat, typ));
+        if let Some(mut j) = self.lookahead_is("(", i) {
+            let mut bindings: Vec<(String, String)> = Vec::new();
+
+            while j < self.src.len() {
+                // Some binding name, or pattern.
+                let k = self.parse_ws(j, false)?;
+                let (k, binding) = self.parse_to_single_colon(k)?;
+                let (k, typ, done) = self.parse_param_rust_type(k + ':'.len_utf8())?;
+                j = k;
+                bindings.push((binding.trim_end().to_string(), typ));
+                if done == true {
+                    break;
+                }
+            }
+            if !bindings.is_empty() {
+                self.ast.parse_param_bindings = Some(bindings);
+            }
+            i = j;
         }
         Ok(self.parse_ws(i, true)?)
+    }
+
+    fn parse_param_rust_type(&mut self, i: usize) -> YaccResult<(usize, String, bool)> {
+        let mut j = i;
+        let mut brace_count = 0;
+
+        while j < self.src.len() {
+            let c = self.src[j..].chars().next().unwrap();
+            match c {
+                '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, j)),
+                ')' | ',' if brace_count == 0 => {
+                    return Ok((
+                        j + c.len_utf8(),
+                        self.src[i..j].trim_start().trim_end().to_string(),
+                        c == ')',
+                    ));
+                }
+                '(' | '{' | '[' | '<' => {
+                    brace_count += 1;
+                    j += c.len_utf8();
+                }
+                ')' | '}' | '>' | ']' if brace_count > 0 => {
+                    brace_count -= 1;
+                    j += c.len_utf8();
+                }
+                c => j += c.len_utf8(),
+            }
+        }
+        return Err(self.mk_error(YaccParserErrorKind::PrematureEnd, j));
     }
 
     /// Parse up to (but do not include) the end of line (or, if it comes sooner, the end of file).
@@ -1774,7 +1802,7 @@ x"
         use super::HashSet;
         use std::iter::FromIterator;
         let src = "
-          %parse_param <'a, 'b> (x, y) : (&'a (), &'b ())
+          %parse_param <'a, 'b> (x: &'a (), (y, z) : (Result<((), ()), ((), ())>, ((u32, u32), &'b ())))
           %%
           A: 'a';
          ";
@@ -1785,8 +1813,16 @@ x"
         .unwrap();
 
         let expect_lifetimes = HashSet::from_iter([&"'a", &"'b"].iter().map(|s| s.to_string()));
-        let expect_bindings = ("(x, y)".to_string(), "(&'a (), &'b ())".to_string());
-
+        let expect_bindings = [
+            ("x", "&'a ()"),
+            (
+                "(y, z)",
+                "(Result<((), ()), ((), ())>, ((u32, u32), &'b ()))",
+            ),
+        ]
+        .iter()
+        .map(|(v, t)| (v.to_string(), t.to_string()))
+        .collect::<Vec<(String, String)>>();
         assert_eq!(grm.parse_param_lifetimes, Some(expect_lifetimes));
         assert_eq!(grm.parse_param_bindings, Some(expect_bindings));
     }
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -497,10 +497,7 @@ impl YaccParser {
             match c {
                 '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, j)),
                 ')' | ',' if brace_count == 0 => {
-                    return Ok((
-                        j,
-                        self.src[i..j].trim_end().to_string(),
-                    ));
+                    return Ok((j, self.src[i..j].trim_end().to_string()));
                 }
                 '(' | '{' | '[' | '<' => {
                     brace_count += 1;
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -196,6 +196,12 @@ impl YaccParser {
                 }
                 continue;
             }
+            if let Some(j) = self.lookahead_is("%parse-param", i) {
+                i = self.parse_ws(j, false)?;
+                let (j, n) = self.parse_to_eol(i)?;
+                i = self.parse_ws(j, true)?;
+                continue;
+            }
             if let YaccKind::Eco = self.yacc_kind {
                 if let Some(j) = self.lookahead_is("%implicit_tokens", i) {
                     i = self.parse_ws(j, false)?;
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -196,10 +196,8 @@ impl YaccParser {
                 }
                 continue;
             }
-            if let Some(j) = self.lookahead_is("%parse-param", i) {
-                i = self.parse_ws(j, false)?;
-                let (j, n) = self.parse_to_eol(i)?;
-                i = self.parse_ws(j, true)?;
+            if let Some(j) = self.lookahead_is("%parse_param", i) {
+                i = self.parse_param(j)?;
                 continue;
             }
             if let YaccKind::Eco = self.yacc_kind {
@@ -426,6 +424,65 @@ impl YaccParser {
         } else {
             Ok(i)
         }
+    }
+
+    // This parses `%parse_param "<"lifetimes">" pat ":" type
+    // Eventually we need to bind type to a known variable name.
+    // And generate actions using pat.
+    //
+    // The single variable case of this is `%parse_param x : int`,
+    // while in the multi-variable case is `%parse_param (x, y) : (int, int)`
+    //
+    // To bind this to a known variable name we drop the pattern,
+    // `known_variable_name : (int, int)`.
+    //
+    // Then for user actions we recombine the pattern and the type.
+    // This all works out because patterns don't have single colons in them.
+    //
+    // This way we can void avoid knowing parameter names,
+    // and dealing commas and parenthesis in the pattern which is difficult
+    // since rust types can contain them.
+    fn parse_param(&mut self, mut i: usize) -> YaccResult<usize> {
+        i = self.parse_ws(i, false)?;
+        // First gobble up all of the '<' lifetime ',' ... '>
+        if let Some(mut j) = self.lookahead_is("<", i) {
+            let mut k = j;
+            let mut lifetimes = HashSet::new();
+            let mut add_lifetime = |k| {
+                let s = self.src[j..k].trim_start().trim_end().to_string();
+                lifetimes.insert(s);
+                j = k + 1;
+                j
+            };
+
+            while k < self.src.len() {
+                let c = self.src[k..].chars().next().unwrap();
+                match c {
+                    '\n' | '\r' => break,
+                    ',' => {
+                        k = add_lifetime(k);
+                        continue;
+                    }
+                    '>' => {
+                        k = add_lifetime(k);
+                        break;
+                    }
+                    _ => k += c.len_utf8(),
+                }
+            }
+            self.ast.parse_param_lifetimes = Some(lifetimes);
+            i = k;
+        }
+
+        i = self.parse_ws(i, false)?;
+        let (i, pat) = self.parse_to_single_colon(i)?;
+        let (i, typ) = self.parse_to_eol(i + ':'.len_utf8())?;
+        let pat = pat.trim_start().trim_end().to_string();
+        let typ = typ.trim_start().trim_end().to_string();
+        if !(pat.is_empty() || typ.is_empty()) {
+            self.ast.parse_param_bindings = Some((pat, typ));
+        }
+        Ok(self.parse_ws(i, true)?)
     }
 
     /// Parse up to (but do not include) the end of line (or, if it comes sooner, the end of file).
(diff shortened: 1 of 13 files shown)


lrpar/cttests/src/calc_parse_param.test
```diff
@@ -0,0 +1,49 @@
+name: Test %parse-param using the calculator grammar
+yacckind: Original(YaccOriginalActionKind::UserAction)
+grammar: |
+    %start Expr
+    %actiontype Result<u64, ()>
+    %avoid_insert 'INT'
+    %parse-param <'a> {x: &'a mut int}
```
ltratt, line 7: I don't have a deep opinion here, but my gut reaction is that a) we should call it `%parse_param` (to match the `_` in `%avoid_insert) b) I think the `<...>` lifetime syntax makes sense c) the `{...}` parameter syntax makes a bit less sense (at first I thought it was action code until I read Bison's documentation). I wondered if either changing `{}` -> `()` would make sense, or maybe just drop the brackets altogether? i.e. either `%parse_param <'a> (x: &'a mut int)` or `%parse_param <'a> x: &'a mut int`.

ratmice: Agree, I'd only done that because thats the way bison did it -- matching rust syntax is probably a better option and less work.
thank you for the feedback.

ltratt: My pleasure -- I think this is a useful feature so I'm happy to work with you to get it in!

ratmice: I went with the latter, there is a (probably excessive) comment which gives a rationale for the choice. That wins out over the `(x: &'a mut int)` because rust types may contain `(int, int)` making it a somewhat painful exercise.

ltratt: I found the single variable case in https://github.com/softdevteam/grmtools/pull/214/commits/1eb94e0218344ecb394d3f8dfb4f572c315e6a69 to be clear, but the multi-variable case confusing. I reckon it probably needs to be a bit closer to Rust's syntax, so something like `%parse_param <'a> (x: int, y: int)` (or `%parse_namar <'a> x: int, y: int`) perhaps?

ratmice: I agree it isn't as nice, it strangely *is* using Rust syntax, but Rust's syntax isn't uniform between parameters and tuples, and in this case it's very much easier to use the tuple syntax.

Both of the above options are unfortunately going to force us to parse a good amount of rust arrays, tuples, and `<T1, T2>`  because we need to pull out `x`, and `y` to pass them down to the user actions from parse, so we'll need to know the variable names.

Now, because `fn parse<'a>(s:&'a str, (x, y) : (u32, u32)) {}` is valid syntax,
but `fn parse<'a>(s:&'a str, (x: u32, y: u32)){}` is not.

We'd have to either parse out x and y either to get the variable names, or to translate them into tuple decl syntax.

And when it comes to  `fn parse<'a>(s:&'a str, x: Foo<(u32, u32), u32>, y: u32) {}`
We'll have to do all the bracket matching to figure out which commas separate variable decls, and which are part of the type.

I'd really like to avoid that if possible, It requires a rather deep dive into parsing rust types, and is potentially fragile if new brackets ever get added to types (unlikely?).

Perhaps i'm missing some way we can avoid parsing these out though.
rust macro?

ltratt: Point taken: I hadn't considered all these cases! I do think that we don't have to worry too much about invalid syntax, provided that we can parse valid syntax :)

I wonder if we can do something fairly simple and get away with it. Let's take your example `fn parse<'a>(s:&'a str, (x, y) : (u32, u32)) {}`. If I parse everything after the lifetime, I think I can get away with the following algorithm:

1) Everything to the left of a `:` must be a name (or some brackets etc. which we could even ignore).
2) Everything to the right of a `:` must be types, upto the next `,` or `)`.
3) We have to count brackets (`<{(` etc.), but only in a simple way.

I don't think we necessarily have to check that the user is using valid Rust syntax. For example I'd be fine with allowing this to squeeze through `<(x>)` i.e. getting the closing brackets "the wrong way around" at the moment.

Put another way: I'd rather we get a nice syntax that we can live with long-term and maybe have a parser which is a little sloppy (in the sense of accepting some invalid things) since we can improve that over time. But once we've chosen a syntax, it becomes painful for users to change in the future.

ratmice: Yeah, it doesn't seem like that would be too painful, i'll try and work on it in the morning.

ltratt: Thanks :)

ratmice: Alright, so that is done.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -474,17 +460,59 @@ impl YaccParser {
...
+            let mut bindings: Vec<(String, String)> = Vec::new();
+
+            while j < self.src.len() {
+                // Some binding name, or pattern.
+                let k = self.parse_ws(j, false)?;
+                let (k, binding) = self.parse_to_single_colon(k)?;
+                let (k, typ, done) = self.parse_param_rust_type(k + ':'.len_utf8())?;
+                j = k;
+                bindings.push((binding.trim_end().to_string(), typ));
+                if done == true {
```
ltratt, line 475: Just `if done`?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -474,17 +460,59 @@ impl YaccParser {
...
+                    ));
+                }
+                '(' | '{' | '[' | '<' => {
+                    brace_count += 1;
+                    j += c.len_utf8();
+                }
+                ')' | '}' | '>' | ']' if brace_count > 0 => {
+                    brace_count -= 1;
+                    j += c.len_utf8();
+                }
```
ltratt, line 509: Do we need to handle whitespace explicitly here? Or perhaps it's just OK to pass it through to `rustc`?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -474,17 +460,59 @@ impl YaccParser {
...
-        let (i, pat) = self.parse_to_single_colon(i)?;
-        let (i, typ) = self.parse_to_eol(i + ':'.len_utf8())?;
-        let pat = pat.trim_start().trim_end().to_string();
-        let typ = typ.trim_start().trim_end().to_string();
-        if !(pat.is_empty() || typ.is_empty()) {
-            self.ast.parse_param_bindings = Some((pat, typ));
+        if let Some(mut j) = self.lookahead_is("(", i) {
+            let mut bindings: Vec<(String, String)> = Vec::new();
+
+            while j < self.src.len() {
```
ltratt, line 468: If we make this something like:

```rust
while j < self.src.len() && self.src[j..].chars().next().unwrap() != ')'
```

then we can get rid of the slightly magical-seeming `done` variable I think?

ltratt:
I have a few small comments, but otherwise this is looking good!

cfgrammar/src/lib/yacc/parser.rs
ratmice: IIUC that this is about the whitespace trimming,
It is fine to remove it and pass it through to rust, I was mostly trimming it so the tests don't run into it.

ltratt: I wondered if we needed to manually trim it; but if what's here works as-is, that's fine with me!

cfgrammar/src/lib/yacc/parser.rs
ratmice: Its a bit difficult to do, because `j` is the character *after* ')'.
I did an update, it used a mutable to put it at the beginning of the loop, but between the break and the mutation I thought it was a toss-up, returning the char we can at least give it a better name.

cfgrammar/src/lib/yacc/parser.rs
ratmice: I managed to get rid of the `trim_start()`, but the trim_end() isn't so easy.

cfgrammar/src/lib/yacc/parser.rs
ratmice: So, I think the last thing I pushed is the closest we can come to that.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -462,19 +462,23 @@ impl YaccParser {
 
         // Next, the '(' pattern : type, ... ')'
         i = self.parse_ws(i, false)?;
-        if let Some(mut j) = self.lookahead_is("(", i) {
+        if let Some(_) = self.lookahead_is("(", i) {
```
ltratt, line 465: Can be `if self.lookahead_is(...).is_some()`?

cfgrammar/src/lib/yacc/parser.rs
ltratt: I wonder if:

```rust
while j < self.src.len() && !self.lookahead_is(")", j) {
   ...
}
```

works? I must admit I haven't thought about this part of the code in ages!

cfgrammar/src/lib/yacc/parser.rs
ratmice: good catch!

cfgrammar/src/lib/yacc/parser.rs
ratmice: It's a bit awkward, i.e. we still need to increment in the loop, and outside the loop with `parse_to_eol()`, but I made it so!

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -485,6 +481,7 @@ impl YaccParser {
             }
             i = j;
         }
+        let (i, _) = self.parse_to_eol(i)?;
```
ltratt, line 484: Can we squish this line and the below to `Ok(self.parse_ws(self.parse_to_eol(i).0, true)?)`?

ratmice: I was wondering about this myself, but hadn't checked since I hadn't seen the pattern used anywhere.  Apparently no (borrow checker, both take `&mut self`)

ltratt: Ah! Good point.

ltratt:
OK, I think we're done with the syntax phase :) Sorry for the minor comments! Do you have a good sense of what part of the code you need to prod next?

ratmice:
No problem, sorry it took so much back and forth,  but no I don't i'll have to do some reading.

ratmice:
So, looking past the syntax phase, Overall the `GrammarAST -> YaccGrammar`, and the `CTParserBuilder` aspects are pretty straightforward (but, and of course there's a but).
I'm somewhat at a loss for how to interact between the generated code, and the (non-generated) `parser::ActionFn`, and library code parameterized by it.

I think at least, %parse_param probably shouldn't accept multiple lifetimes, and I need to combine all the separate parameters into one (e.g. generate a tuple from the parameters in the `GrammarAst -> YaccGrammar` phase, so ActionFn is parameterized by a single parameter type).

Theres a lot of moving parts, So I think the right approach is to play around with it in a cut-down simplified playground which kind of reflects the overall `Parser` struct with less moving parts.

ratmice:
Here are some experiments that seem like they are leading towards something that can be generated (Although I would not call them pretty), i'll try and see what happens when I try and use something like that `Perhaps` type.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=884c27cb4ed267b4f700c217a3df4a6d

ltratt:
> I'm somewhat at a loss for how to interact between the generated code, and the (non-generated) parser::ActionFn, and library code parameterized by it.

Hmm, yes, this might be tricky, particularly with lifetimes.

That makes me wonder if what we really need is a more radical change: rather than having the core parser be a library, we could entirely generate it. That has always been my (very very very) long term plan anyway, on the assumption that it will allow the compiler to further optimise the parser. If we generate the parser, then we can make the function lifetimes / types be whatever we want. I admit that this is not a small job, but it's probably not quite as big as it looks either.

ratmice:
I think I'm still going to try the tricky type/single lifetime approach (It's probably going to be ugly, so I totally understand if it becomes unmergable!),

The main reason is, I'm not actually sure if I can use this in my project yet, it depends on if I can get the parser parse in the same order, I need to call functions on the builder parameter I'm passing in.
As such, I don't want to commit to that large of a project while uncertain I can actually use it!

So, I'll at least give that a go first and then see where things stand?

ltratt:
It's certainly worth trying, and then we can see what we end up with.

ratmice:
So, this last commit is kind of a doozy, Its hard to see any way to break it up into smaller chunks,

There are some extraneous changes to the testsuite (where i pass in (), and actually return the parameter).
I think that these should probably be reverted, and just not return the parameter.
But it was easier to implement as a first pass if I just didn't worry about it and do the whole thing the end-to-end.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -692,29 +742,34 @@ where
...
                             ridx = usize::from(ridx),
                             pidx = usize::from(pidx),
-                            args = args.join(", ")));
+                            args = args.join(", "),
+                            param_pat = self.param_pattern(grm),
+                        ));
                     }
                     _ => {
-                        outs.push_str(&format!("\n        {actionskind}::{actionskindprefix}{ridx}({prefix}action_{pidx}({prefix}ridx, {prefix}lexer, {prefix}span, {args}))",
+                        outs.push_str(&format!("\n        ({actionskind}::{actionskindprefix}{ridx}({prefix}action_{pidx}({prefix}ridx, {prefix}lexer, {prefix}span, {args})), {param_pat})",
```
ratmice, line 764: So, I need to look at this more carefully, as it seems like a very tricky part.

`{args}`, now contains param_args, from the chain added above, but then went out of scope in the action call.
It seems quite possible that this could lead to a dramatic amount of clean up in figuring out when/where we can stop threading this parameter through.  But I haven't done that detailed of a review of things yet.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -422,6 +426,93 @@ impl YaccParser {
...
+    // Handle parse_param declarations of the form:
+    // %parse_param <'a>(x: u32, y : (u32, u32))
+    fn parse_param(&mut self, mut i: usize) -> YaccResult<usize> {
+        i = self.parse_ws(i, false)?;
+        // First gobble up all of the '<' lifetime ',' ... '>
+        if let Some(mut j) = self.lookahead_is("<", i) {
+            let mut k = j;
+            let mut lifetimes = HashSet::new();
+            let mut add_lifetime = |k, c: char| {
+                let s = self.src[j..k].trim_start().trim_end().to_string();
```
ltratt, line 438: I think we can replace the two `trim`s with just `trim`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -422,6 +426,93 @@ impl YaccParser {
...
+                lifetimes.insert(s);
+                j = k + c.len_utf8();
```
ltratt, line 440: Can replace this line and the one below with just `k + c.len_utf8()`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -422,6 +426,93 @@ impl YaccParser {
...
+                j
+            };
+
+            while k < self.src.len() {
+                let c = self.src[k..].chars().next().unwrap();
+                match c {
+                    '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, k)),
+                    ',' => {
+                        k = add_lifetime(k, ',');
+                        continue;
```
ltratt, line 450: I think the `continue` is superfluous here.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -422,6 +426,93 @@ impl YaccParser {
...
+        i = self.parse_ws(i, false)?;
+        if self.lookahead_is("(", i).is_some() {
+            let mut j = i;
+            let mut bindings: Vec<(String, String)> = Vec::new();
+            while j < self.src.len() && self.lookahead_is(")", j).is_none() {
+                let c = self.src[j..].chars().next().unwrap();
+                j += c.len_utf8();
+
+                // Some binding name, or pattern.
+                let k = self.parse_ws(j, false)?;
```
ltratt, line 473: Since we don't reference `j` afterwards, I think this can be just `j = self.parse_ws(...)`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -422,6 +426,93 @@ impl YaccParser {
...
+            match c {
+                '\n' | '\r' => return Err(self.mk_error(YaccParserErrorKind::ReachedEOL, j)),
+                ')' | ',' if brace_count == 0 => {
+                    return Ok((j, self.src[i..j].trim_end().to_string()));
+                }
+                '(' | '{' | '[' | '<' => {
+                    brace_count += 1;
+                    j += c.len_utf8();
+                }
+                ')' | '}' | '>' | ']' if brace_count > 0 => {
```
ltratt, line 506: I think we can easily catch one source of mistake here by changing this to something like:

```rust
')' | '}' | '>' | ']' => {
    if brace_count == 0 { <error> }
    brace_count -= 1;
    j += c.len_utf8();
}
```

otherwise if someone does close too many brackets, it'll be a bit difficult to work out what happened.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -422,6 +426,93 @@ impl YaccParser {
...
+                    brace_count -= 1;
+                    j += c.len_utf8();
+                }
+                c => j += c.len_utf8(),
+            }
+        }
+        return Err(self.mk_error(YaccParserErrorKind::PrematureEnd, j));
```
ltratt, line 513: Can just be `Err(...)` without the `return`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1706,4 +1797,33 @@ x"
             Err(e) => panic!("Incorrect error returned {}", e),
         }
     }
+    #[test]
```
ltratt, line 1800: Missing blank line before `#[test]`?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1706,4 +1797,33 @@ x"
...
+    fn test_parse_param() {
+        use super::HashSet;
```
ltratt, line 1802: I think we should add these two `use`s to the beginning of the `test` module (and prefer `std::collections::HashSet` over `super::HashSet`).

lrpar/cttests/src/calc_parse_param.test
```diff
@@ -0,0 +1,45 @@
+name: Test %parse-param using the calculator grammar
+yacckind: Original(YaccOriginalActionKind::UserAction)
+grammar: |
+    %start Expr
+    %actiontype Result<u64, ()>
+    %avoid_insert 'INT'
+    %parse_param <'a>(x : &'a mut u64)
+    %%
+    Expr: Expr '+' Term {
+                          *x = $1? + $3?;
```
ltratt, line 10: I think the indentation would match the other `.test` files if it was more like:

```rust
Expr: Expr '+' Term {
        *x = $1? + $3?;
        ...
    }
```

I think the `Factor` rule's indent precedent is probably worth reusing?

lrpar/cttests/src/lib.rs
```diff
@@ -231,8 +234,25 @@ fn test_span() {
...
+#[test]
+fn test_parse_param() {
+    let lexerdef = calc_parse_param_l::lexerdef();
+    let lexer = lexerdef.lexer("1 + 1");
+    let mut x = 0;
+    match calc_parse_param_y::parse(&lexer, &mut x) {
+        (Some(Ok(y)), ref errs, &mut _x) if errs.is_empty() => {
+            let _x = &_x;
+            assert_eq!(&x, _x);
+            //std::ptr::eq(&x, _x));
```
ltratt, line 252: Debug comment?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -541,48 +580,54 @@ where
...
                     ridx = usize::from(self.user_start_ridx(grm)),
                     recoverer = recoverer,
+                    param_pat = self.param_pattern(grm),
                 ));
             }
             YaccKind::Original(YaccOriginalActionKind::GenericParseTree) => {
                 outs.push_str(&format!(
                     "
-        ::lrpar::RTParserBuilder::new(&grm, &stable)
+        {{ let (__x, __y, _) = ::lrpar::RTParserBuilder::new(&grm, &stable)
```
ltratt, line 616: I'm about 98% certain that these variables are generated in a function which doesn't contain any user code. [I think that function has "bare" variables like `action` generated by `ctbuilder`.] If so, I think we can drop the `__` prefix.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -802,8 +863,9 @@ where
             };
             outs.push_str(&format!(
                 "    // {rulename}
-    #[allow(clippy::too_many_arguments)]
-    fn {prefix}action_{}<'lexer, 'input: 'lexer>({prefix}ridx: ::cfgrammar::RIdx<{storaget}>,
+    #[allow(clippy::too_many_arguments, unused_variables)]
+    fn {prefix}action_{}<'lexer, 'input: 'lexer, {param_lt}>(
```
ltratt, line 867: Why allow `unused_variables`?

lrpar/src/lib/parser.rs
```diff
@@ -91,11 +101,18 @@ pub struct Parser<'a, 'b: 'a, 'input: 'b, StorageT: 'static + Eq + Hash, ActionT
     // In the long term, we should remove the `lexemes` field entirely, as the `NonStreamingLexer` API is
     // powerful enough to allow us to incrementally obtain lexemes and buffer them when necessary.
     pub(crate) lexemes: Vec<Lexeme<StorageT>>,
-    actions: &'a [ActionFn<'a, 'b, 'input, StorageT, ActionT>],
+    actions: &'a [ActionFn<'a, 'b, 'input, 'arg, StorageT, ActionT, ParamT>],
```
ltratt, line 104: This might be a silly question but I've just realised that I'm not sure why an action can *return* a `ParamT`. Is this something to do with ownership problems?

ltratt:
I've done a semi-thorough review though most of my comments are just minor things.

Overall, I'm impressed! I think that you've done a lot of the hard work, and I can now see a fairly clear path to getting this PR merged. Although we still have a bit of work to do, I think the worst is behind us!

I have two over-arching comments:

1) I think we need to make sure the "common `parse` case" is unchanged, so we can't add an extra argument to every generated `parse` function. *However*, I'm open to changing the semi-internal interface. So what I would suggest is that, if `parse_param` isn't specified, we generate a `parse` function without an extra parameter. However, I think having `parse_actions` take an extra parameter (or, maybe even better because it would be entirely backwards compatible, adding a new function to `RTParserBuilder` called something like `parse_param`?) would be acceptable.

2) I haven't worked out why `ParamT` is also a return type of actions. I wondered if there's some sort of lifetime issue, but I wasn't sure.

lrpar/src/lib/parser.rs
ratmice: Not a silly question at all, it strikes quite to the heart of it, If we look at the loop in `parser::Parser::lr`, and `lr_upto`, where we call the action inside parser (because param isn't `Copy` bounded, it moves the param field, so we have to return it there, in order to regain ownership.

I'll add some comments at the places, but I'm not sure how to do so in this thread.

lrpar/src/lib/parser.rs
```diff
@@ -396,12 +437,15 @@ where
...
-                            let v = AStackType::ActionType(self.actions[usize::from(pidx)](
+                            let (x, param_tmp) = self.actions[usize::from(pidx)](
                                 ridx,
                                 self.lexer,
                                 span,
                                 astack_uw.drain(pop_idx - 1..),
-                            ));
+                                param,
+                            );
+                            param = param_tmp;
```
ratmice, line 447: So here is one of the places we regain ownership via returning Param from action.

lrpar/src/lib/parser.rs
```diff
@@ -279,12 +315,15 @@ where
...
+                    let (x, param_tmp) = self.actions[usize::from(pidx)](
                         ridx,
                         self.lexer,
                         span,
                         astack.drain(pop_idx - 1..),
-                    ));
+                        param,
+                    );
+                    let v = AStackType::ActionType(x);
+                    param = param_tmp;
```
ratmice, line 326: Here is another where we're dealing with a param moved through an action.

lrpar/src/lib/ctbuilder.rs
ratmice: I had tried to not do this but hadn't figured out a way, this is because the param is being passed through to user action code, and so user actions like `{ $1 }` will trigger this lint.

I think one way to disable this in user-action code is force the user-code to `{ #[allow(unused_variables) { $1 } }`, but that didn't occur to me until just now.

lrpar/src/lib/parser.rs
ltratt: If we're doing this sort of moving, it suggests to me that we really intend to pass the param as a reference? Or is this because we want to allow mutation?

ratmice: Yeah its because of mutation, we have to move the mutable reference around to enforce single-ownership.

But it does bring up a point that I haven't tested this with a non-reference param.

ltratt: Maybe passing it as `&mut` would do the trick?

ratmice: Not quite certain, I can give it a try though, I had difficulty trying to use the fancy types (higher ranked lifetime bounds) here, which allows us to show that the `&mut` isn't captured by the `Fn` closure.  So this was kind of a last resort, but perhaps starting from a place where it now works I can figure that out.

ltratt: Fingers crossed!

cfgrammar/src/lib/yacc/parser.rs
ratmice: This was actually updating j, in the loop from where `add_lifetime` is called, 
I think its clearer with the above change, and mutation moved out.

Sorry it has been some years since i've done much imperative programming, and it tells.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yeah, what I wouldn't give for `(k, let foo) = ...`.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yep, it won't catch the ')' case though.

cfgrammar/src/lib/yacc/parser.rs
ltratt: Agreed -- it does lead to some ugly code :/

lrpar/cttests/src/calc_parse_param.test
ratmice: I believe I got this matching the existing tests.

cfgrammar/src/lib/yacc/parser.rs
ltratt: Totally agreed: I'm only inclined to catch the one I suggested because it's so easy it seems almost silly not to do it. There are various other cases (e.g. mismatched brackets) which this code doesn't deal with -- and I'm fine with it not doing so!

lrpar/cttests/src/calc_parse_param.test
ltratt: I think the indentation later in the same file is different, so it's probably worth syncing it, at least within that single file.

lrpar/src/lib/ctbuilder.rs
ratmice: Yep, this highlights a missing test, as it wouldn't have compiled as is when running with `%parse_param` set in the yacc.

lrpar/cttests/src/calc_parse_param.test
ratmice: I think now I get what you mean.

lrpar/src/lib/parser.rs
ratmice: Well, i'm sorry to say that I don't think it is possible, the problem is that `ActionFn` takes an owned `vec::Drain`, the HRTB works via saying `for <'c>  Fn(...)`, essentially saying the parameters can live either less than or greater than the lifetime of the closure (Essentially a `Fn` equivalent of a `fn`, but the closure itself owns the `vec::Drain`, and so must have a self which captures it.  There isn't any fine-grained per parameter way to type that I believe.

ltratt: I think I sort-of understand. I guess the challenge with passing in `ParamT` and receiving back a `ParamT` is that we've semi-destroyed the relationship between the two from the type system's perspective. This might be a very dumb thing to say, but can we cheat by wrapping the parameter in something (maybe, at its heart, an `UnsafeCell`)? I might be clutching at straws here...

ratmice: I'm really not familiar with `unsafe` in general, but I *guess* you can do anything, it would add some difficulty in getting stuff out of the unsafe cell in the way the user-code is expecting the environment.

One thing I was wondering, is whether it is possible to use a thread_local for either the `Drain` or the `ParamT`, If we can just get one of those out of the `Fn` signature while somehow avoiding moving `ParamT`.

ltratt: We definitely can't do that for the `Drain`. *But* it might be possible to convert the `Drain` into a slice which would be passed by reference -- would that make any difference?

ratmice: I think that it would be possible to get a reference to a slice to work with the `for<'c>` quantifier yes.  I'll try and work on a more reflective example in the playground again.

ratmice:
So I messed with something on the playground, that seemed like it would work/tried it out, the tests and ctbuilder.rs changes still need work, and it forces `&mut` on the caller (Not sure how to avoid that).

But otherwise it did manage to get rid of the `ParamT` return from action, so that seems like a success.  I'll try and clean it up some tomorrow.

ltratt:
I think this is an improvement, and with a bit of tidying up, we might have a viable solution. [The main thing is that we don't want to force things like `multitypes_y::parse(&lexer, &mut ());` for grammars that don't use `%parse_param`. Apart from that, I think the internal changes to grmtools are probably close to acceptable!]

ratmice:
So, I think i need to write some tests for multiple parameters, I don't think they'll survive the tuplization that is going on, and the `GenericParseTree` with `%parse_params`.

ltratt:
The new commits look good! I agree, though, that more tests would be good to check that we've covered all the main things!

ratmice:
So, the code gen is a bit idiosyncratic, I couldn't figure out any nice way to organize it.
On the test front, I couldn't figure out a way to ignore unused variables, and mutation (the `#[allows()]` don't work in this case.  So just used '_var', names.

lrpar/cttests/src/multiparam.test
```diff
@@ -0,0 +1,27 @@
+name: Test multiple types
+yacckind: Grmtools
+grammar: |
+    %parse_param <'a>(_a_count: u64, _b_count: u64)
```
ltratt, line 4: I *think* I might know the answer to this question but why the initial `_` on the variable names? Is it because of unused variable warnings in the generated code for actions that don't make use of them?

lrpar/cttests/src/multiparam.test
```diff
@@ -0,0 +1,27 @@
...
+    | AorB B { Ok($2?) }
+    | AorB A { Ok($2?) };
+
+    A  -> Result<(), ()>
+    : 'a'    { Ok(*_a_count.deref_mut() += 1) };
+
+    B  -> Result<(), ()>
+    : 'b'    { Ok(*_b_count.deref_mut() += 1) };
+    %%
+    // We get &mut &mut u64's, use deref_mut to coerce them to &mut u64
```
ltratt, line 21: This slightly surprised me: naively, I specified `u64` to %parse_param, so I would have expected `u64`, not `&mut u64` and almost certainly not `&mut &mut u64`. Clearly I've misunderstood something somewhere, but I'm not quite sure where :)

ratmice: *nod*, this part that bothers me a lot too, I really should have brought more attention to it earlier. The problem is we have to pass the tuple by reference, while avoiding moving the tuple.

Here is a mildly reflective playground link, the only difference is `(x, y)` vs  `&mut (x, y)`, of how we destructure the pattern. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cec88192e245564e61d7b9b378ef43e1)

Normally we'd expect something like the `move` variants, where we'd destructure the tuple via `&mut (x, y)` giving us an `x`, and `y` of `&u64`, what works here though is the `nomove`, which ends up adding the tuples own reference onto the existing reference so we end up with the double indirection.

With the mechanism where we return the ParamT from the actions, we can move the tuple, because wherever we move it to should give us back the place that they moved it.

It seems like a place where we are cornered between two fairly ugly choices. :face_with_head_bandage:

ratmice: The second part that bothers me about this, is we have to make a choice for the user between `&mut`, and `&` ref, because there is no mutation polymorphic reference, with unique ownership.

ratmice: In conclusion I *think* perhaps the best thing to do here, is just only accept a single parameter, meaning the parse caller has to tuplize multiple params themself.

ltratt: OK, I'm starting to see the problem. This works:

```rust
fn f(t: &mut (&mut u64, &mut u64)) {
    *t.0 = 4;
}

fn main() {
    let mut x = 2;
    let mut y = 3;
    f(&mut (&mut x, &mut y));
}
```

but this doesn't:

```rust
fn f(t: &(&mut u64, &mut u64)) {
    *t.0 = 4;
}

fn main() {
    let mut x = 2;
    let mut y = 3;
    f(&mut (&mut x, &mut y));
}
```

However, this does work:

```rust
fn f(t: (&mut u64, &mut u64)) {
    *t.0 = 4;
}

fn main() {
    let mut x = 2;
    let mut y = 3;
    f((&mut x, &mut y));
    println!("{}", x);
}
```

I'm not sure I have any deep thoughts on this yet, but at least I'm starting to see what the design issue is, and I agree that it's not an ideal one. However, I'm an optimist -- I'm sure we'll find a good solution one way or another!

ratmice: Indeed it does, i'm pretty bad at traits, but it *seems* like we might be able to do something with those having the action functions implement trait which works for some fixed number of tuples, and as well as refs.

ltratt: I could definitely live with the old "works for tuples up to size N" hack -- it isn't pretty, but it is effective (and that trick is used elsewhere in Rust, so it does have precedent)!

ratmice: Yeah seems like it might work, not sure how much of e.g. `Parser` would need to be parameterized across the various tuples though.

unfortunately, it looks as though due to some hardware issues, it might be some time before I can do any further work here, I'll try and get back to it once I have a new development machine up and running.

ltratt: > Yeah seems like it might work, not sure how much of e.g. Parser would need to be parameterized across the various tuples though.

That's a good point. The more I think about it, the more I think that if we take the "traditional" approach to parsers and generate all the code we'd avoid all of these problems, but that isn't a 5 minute job.

One possibility is that we could merge this work more-or-less as-is and either-or a) not document the feature b) explicitly say that it's going to change (in a more ergonomic way) in the future. Any thoughts?

If it would help to have access to a shell machine for development, please drop me an email, and I can sort something out!

ltratt:
As discussed off-GitHub, we're currently considering rethinking how this PR approaches things. However, I think there is merit in getting at least *some* of the code into `master`. @ratmice one thing we can do is squash this down so that we keep the cfgrammar parser code, merge this PR, then you can salvage whatever else you want from this PR for the "bigger" PR we discussed. If you have an alternative suggestion for what to do, I'm very happy to do so!

ltratt:
Hello @ratmice -- is there still interest at your end in furthering this work? I'd be very happy to see it carry on!

ratmice:
Yes @ltratt -- totally still interested in doing this, I just got distracted and haven't circled back to language projects yet I'll try and give it another go in the next few days.

ltratt:
Excellent :)

ratmice:
@ltratt So, I haven't exactly thought it through but before going and actually trying to figure out the extent that `const generics` might help in this series (unclear to me). Given the relatively new compiler required I wanted to ask beforehand whether that aspect would be a deal breaker.

ltratt:
I'm absolutely fine with requiring the latest stable (particularly as 0.10.0 was a breaking API change, so people using the current version are much more likely to have an upto date rustc).

ltratt:
This PR is starting to gradually pile up conflicts with the master branch (and #254 will probably make that worse). Any thoughts on what we should do here?

ratmice:
Sorry about that, I'd tried working on it some, but not to any success yet; I think it is fine to close it and if I get it working make a new PR I'll leave my fork/branch up for historical purposes too or in case anyone else might have better luck than I in getting it working.

ratmice:
I'll have to think about it, I can't think of anything at the moment because it's all pretty well under the hood in generated code/lifetimes rather than any visible API surface. One thing I recall is error recovery, If it was possible feature gated out error recovery, I think it'd be easier to play around/experiment with the parse tables/code-gen, and afterwords get it working with the fully featured build.  I'd toyed around with making a mini/toy version of the parser with fewer features

 --then again sometimes featuritis just make things even worse for things which are not experiments.

ltratt:
I wonder if it's error recovery or the user-runnable actions that causes the lifetime fun? Certainly, that's the bit that I remember being a pain to integrate in...

ratmice:
It's definitely user-runnable actions which are the major problem, but Error recovery also adds a little bit of fun at each iteration over getting it working.

ltratt:
Is there still a problem if you set `RecoveryKind::None` (https://docs.rs/lrpar/0.10.2/lrpar/enum.RecoveryKind.html)? I'd like to think it makes things easier but perhaps it doesn't?!

ratmice:
Specifically the thing I was recalling is that the recoverer function is generic over the ActionT kind,
```
pub(super) fn recoverer<'a, StorageT: 'static + Debug + Hash + PrimInt + Unsigned, ActionT: 'a>(
    parser: &'a Parser<StorageT, ActionT>,
) -> Box<dyn Recoverer<StorageT, ActionT> + 'a>
```

So from a compile-time perspective, it still adds a bunch of work to get it to compile, in this specific situation with tricky ownership of actions.

ltratt:
Ah, I think I see. I guess commenting out the recovery aspects, at least for experimentation, isn't too difficult: the error recoverers form a sort-of-disconnected part of the graph (they point back to the parser in various ways, but the parser only points to them in a single well-defined way). A PR which just hacked that stuff out would still be very interesting, and might give us the right direction to put it back in.

PR #216, opened 2020-10-11
cfgrammar: initial parsing/syntax for the %parse_param.

ratmice:
Here is the rebased branch from #214 which contains only the option parsing, and none of the implementation or tests which rely on the functitoning implementation.
As such it parses the `%parse_param` directive in the yacc grammar, but doesn't function to implement it.

The rebased versions of the implementation are [here](https://github.com/ratmice/grmtools/tree/parse_param), which builds off the commits in this PR, but still need work.


ltratt:
Thanks!

bors r+

PR #221, opened 2020-11-27
Caseless tokenization

timothy-martin:
Added a case insensitive option for LexerBuilder. Setting this to true will panic if LexerKind::Flex is not specified.


ltratt:
Sorry it's taken me a while to look at this! This is a good beginning, and I think we can quickly whip this into shape for inclusion. Part of the problem is that I forgot some of the details of lrlex and suggested things that aren't quite the best thing to do -- mea culpa!

On that basis, I think we can adjust things as follows:

1. I forgot that lrlex has grown its own not-quite-lex-not-quite-flex `LexKind` called `LRNonStreamingLexer`. We can thus add whatever `%option` things we want to that without worrying about backwards compatibility. In other words, we don't need to mess with `enum LexerKind`.

2. The `case_insensitive` stuff is good and can stay as-is.

3. The tests should either a) be a (new) `lrlex/cttests` directory (using `lrpar/cttests` as a precedent) or b) a new test in `lrpar/cttests`. The latter is almost certainly easier and acceptable. It does involve some mucking around in `lrpar/cttests/lib.rs` but it should be doable.

You can either do that as part of this PR or as part of a new PR. It *might* be easier to close this one and open a new PR, but that's up to you -- feel free to do whichever you're most comfortable with.

ltratt:
@timothy-martin ping.

ltratt:
Closing -- this PR can be reopened if there's additional progress.

PR #229, opened 2021-01-07
Make line_col faster.

ltratt:
`line_col` was previously a linear search that called another function that does a linear search. This PR changes the first linear search into a binary search (https://github.com/softdevteam/grmtools/commit/b62cab260a9b0f4a69a9b50d30eb57b9a43348b6) and removes the second linear search entirely (https://github.com/softdevteam/grmtools/commit/2d61edb7b4e882ef4007b431facacb031c0065b4).

This speeds up the easy bits of `line_col` significantly, moving us from `O(4n + x)` into `O(log_n + x)` where `n` is the number of lines in the file and `x` is whatever the cost to calculate the character offset from a line containing a string of bytes. Alas, this second thing is probably impossible to speed-up given that we're dealing with UTF-8. Still, this PR does at least address the most obviously egregious inefficiencies in `line_col` and it does so while decreasing the lines of code (allowing for the fact that I've added three lines of comments)!

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -212,6 +212,8 @@ impl<StorageT: Copy + Eq + Hash + PrimInt + TryFrom<usize> + Unsigned>
 pub struct LRNonStreamingLexer<'lexer, 'input: 'lexer, StorageT> {
     s: &'input str,
     lexemes: Vec<Result<Lexeme<StorageT>, LexError>>,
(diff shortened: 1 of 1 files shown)


ltratt:
https://github.com/softdevteam/grmtools/pull/229/commits/96f2ebc15b5b927438bf513a3d633a6d1cbfcebd should fix the bors error and will need squashing.

bors try

ptersilie:
Please squash.

ltratt:
Squashed.

PR #237, opened 2021-04-22
Clippy fixes

ltratt:


ptersilie:
Do you want to undo the first commit, now that the last commit fixes Clippy complaining about that? Or is that not what the last commit does?

ltratt:
They're slightly different: https://github.com/softdevteam/grmtools/pull/237/commits/23d4bd270fcdca2448ce54a1a15354392721aaf3 turns off warnings *within* grmtools but https://github.com/softdevteam/grmtools/pull/237/commits/fe8107f78bf8cd07acc30c80a5b044585f3ddcfd makes sure that when we generate code the user doesn't get camel case warnings in *their* code. I think they're best kept separate.

ptersilie:
I see.

bors r+

Issue #234, opened 2021-03-26
Add support for %expect


ltratt:
Bison has `%expect` declarations which tell the parsing system "I know there are *n* conflicts, please don't warn me if you find *n* too". Although undocumented in http://dinosaur.compilertools.net/yacc/index.html, Berkeley Yacc seems to support `%expect` too, so I think grmtools should add that. See section 3.7.9 of https://www.gnu.org/software/bison/manual/bison.html for more details.

Bison also support a `%expect-rr` option too although the manual suggests it should only apply to GLR parsers. I don't see a good reason why we couldn't accept that for normal LR grammars as well though perhaps only if the user specifies `YaccKind::Grmtools`?

PR #238, opened 2021-05-08
Add support for the %expect declaration.

ltratt:
This allows a grammar author to tell Yacc how many shift/reduce conflicts it expects. In our context, this means that compile-time grammar generation and nimbleparse don't need to report errors iff the number of shift/reduce conflicts
exactly matches that specified.

Fixes https://github.com/softdevteam/grmtools/issues/234.

### DIFF
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -124,7 +124,11 @@ fn main() {
 
     if !quiet {
         if let Some(c) = stable.conflicts() {
-            println!("{}", c.pp(&grm));
+            match grm.expect() {
+                Some(i) if i == c.sr_len() => (),
+                _ => println!("{}", c.pp_sr(&grm)),
+            }
+            println!("{}", c.pp_rr(&grm));
             println!("Stategraph:\n{}\n", sgraph.pp_core_states(&grm));
         }
     }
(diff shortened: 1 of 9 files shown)


ltratt:
I've also added support for `%expect-rr` in https://github.com/softdevteam/grmtools/pull/238/commits/25ce441002b1ad8c9f1b6ffdb31d125cff351d29.

nimbleparse/src/main.rs
```diff
@@ -124,12 +124,25 @@ fn main() {
 
     if !quiet {
         if let Some(c) = stable.conflicts() {
-            match grm.expect() {
-                Some(i) if i == c.sr_len() => (),
-                _ => println!("{}", c.pp_sr(&grm)),
+            let pp_rr = if let Some(i) = grm.expectrr() {
```
ptersilie, line 127: Any reason we don't initialise `expect_sr` and `expect_rr` to `0`? Then we don't have to do this dance here.

ptersilie:
Just one question.

nimbleparse/src/main.rs
ptersilie: Fair enough.

ptersilie:
Merge as is?

ltratt:
Yes please!

PR #247, opened 2021-09-13
Deprecate the `process_file_in_src` and `process_file` functions.

ltratt:
The `CTBuilder` API has always been a bit unsatisfactory, and some changes I'd like to make to it in the near future have made clear to me that we need some bigger changes.

https://github.com/softdevteam/grmtools/commit/4d6625c9b8210e9ac77aa6b905ec23e4bd12aa81 solves the problem that the `process_*` functions take arguments that really should have been options on the builder itself. In practise this means that `CTBuilder::process_file_in_src` is deprecated in favour of `CTBuilder::grammar_path_in_src` and `CTBuilder::process`. In most cases this means changing a `build.rs` from:

```rust
let lex_rule_ids_map = CTParserBuilder::new()
.process_file_in_src("grm.y")?;
```

to:

```rust
let lex_rule_ids_map = CTParserBuilder::new()
.grammar_path_in_src("grm.y")?
.process()?;
```

The less commonly used `process_file` function is similarly deprecated in favour of the `grammar_path`, `output_path`, and `process` functions. To my slight horror, some of the cttests used `process_file` in a way that only really worked by accident: by explicitly using `grammar_path` we lessen that chance. https://github.com/softdevteam/grmtools/commit/11981b6ee000eeb23a1d01b142b3ac397988254a ensures that.

While I'm here, I've tidied up some of the documentation that surrounds these functions: there's still more that could, and probably should, be done but that's best left to later commits.

This is a backwards compatible change in the sense that you can use the old functions but rustc will warn you about them. That should help people migrate their code more easily than a "hard" API change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,23 @@
+# grmtools 0.10.3 (XXXX-YY-ZZ)
+
+* `CTParserBuilder::process_file_in_src` is deprecated in favour of
+  `CTParserBuilder::grammar_in_src_dir` and `CTParserBuilder::process`. In most
+  cases you can change your `build.rs` from:
+  ```rust
+    let lex_rule_ids_map = CTParserBuilder::new()
+        .process_file_in_src("grm.y")?;
+  ```
+  to:
+  ```rust
+    let lex_rule_ids_map = CTParserBuilder::new()
+        .grammar_in_src_dir("grm.y")?
+        .process()?;
+  ```
+
+  The less commonly used `process_file` function is similarly deprecated in
+  favour of the `grammar_path`, `output_path`, and `process` functions.
+
+
 # grmtools 0.10.2 (2021-08-09)
 
 * Optimise `NonStreamingLexer::span_lines_str` from *O(n)* to *O(log n)*.
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -126,9 +126,11 @@ pub struct CTParserBuilder<'a, StorageT = u32>
 where
     StorageT: Eq + Hash,
 {
-    // Anything stored in here (except `conflicts` and `error_on_conflict`) almost certainly needs
-    // to be included as part of the rebuild_cache function below so that, if it's changed, the
-    // grammar is rebuilt.
+    // Anything stored in here (except `output_path`, `conflicts`, and `error_on_conflict`) almost
+    // certainly needs to be included as part of the rebuild_cache function below so that, if it's
+    // changed, the grammar is rebuilt.
+    grammar_path: Option<PathBuf>,
+    output_path: Option<PathBuf>,
     mod_name: Option<&'a str>,
     recoverer: RecoveryKind,
     yacckind: Option<YaccKind>,
@@ -149,8 +151,8 @@ impl<'a> CTParserBuilder<'a, u32> {
     ///
     /// ```text
     /// CTParserBuilder::new()
-    ///     .process_file_in_src("grm.y")
-    ///     .unwrap();
+    ///     .grammar_in_src_dir("grm.y")?
+    ///     .process()?;
     /// ```
     pub fn new() -> Self {
         CTParserBuilder::<u32>::new_with_storaget()
@@ -178,11 +180,13 @@ where
     ///
     /// ```text
     /// CTParserBuilder::<u8>::new_with_storaget()
-    ///     .process_file_in_src("grm.y")
-    ///     .unwrap();
+    ///     .grammar_in_src_dir("grm.y")?
+    ///     .process()?;
     /// ```
     pub fn new_with_storaget() -> Self {
         CTParserBuilder {
+            grammar_path: None,
+            output_path: None,
             mod_name: None,
             recoverer: RecoveryKind::CPCTPlus,
             yacckind: None,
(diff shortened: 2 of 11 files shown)


ltratt:
bors try

ltratt:
bors ping

ltratt:
bors ping

ltratt:
bors try

ltratt:
That last commit (at least) will need squashing away.

bors try

ltratt:
I'd like to pretend I was deliberately testing that https://github.com/ltratt/grmtools/commit/a125a70b2d3852ed1cf12d14608a9ddcc413b47b works, but https://github.com/ltratt/grmtools/commit/cb23764e5cd1cee1f11346e8f65f2df07c7fd67a is just the result of my idiocy.

bors try

ltratt:
@ptersilie This is ready for review (but will definitely need squashing!).

CHANGES.md
```diff
@@ -1,3 +1,20 @@
+# grmtools 0.10.3 (XXXX-YY-ZZ)
+
+* `CTBuilder::process_file_in_src` is deprecated in favour of
+  `CTBuilder::grammar_path_in_src` and `process`. In most cases you can change
```
ptersilie: I'm unsure about the name `grammar_path_in_src`. To me it suggest we give it a source which contains the grammar path. Maybe it should just be called `grammar_src`, since we are passing a grammar file, not a path.

ltratt: I'm also not sure what the best name is, and I was curious what you would think! I think `grammar_src` isn't good because that can be confused with `grammar_path`: I think we need to get the notion of "this file is relative to `src/`" across somehow (but without being too wordy). Maybe something like `grammar_file_in_src` or even `find_grammar_in_src`?

ptersilie: After reading the rest of the commit I realised that `src` was referencing the project's `src` directory. However, that is not always correct. What is actually used is the current working directory which isn't necessarily the project root. Maybe `grammar_file_in_cwd` would me more accurate, though it is a bit of a mouthful.

What does `grammar_path` take as an argument? I think you called it `path` because it takes a rust `Path`, but is that clear enough to the end user? When I read `path` I assume it takes a path and not for example a file.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -197,9 +197,9 @@ where
         }
     }
 
-    /// Set the input grammar path to a file in this project's `src` directory. This will also set
-    /// the output path (i.e. you do not need to call [CTParserBuilder::output_path]) and module
-    /// name (i.e. you do not need to call [CTParserBuilder::mod_name]).
+    /// Set the input grammar path to a file relative to this project's `src` directory. This will
```
ptersilie, line 200: This is what I was referring to in my earlier comment. The doc string suggests this is always relative to the project root, but that's not true if the binary is run from somewhere else, e.g. `/home/lukas/grmtools/bin/grmtools grm.y`.

CHANGES.md
ltratt: > After reading the rest of the commit I realised that src was referencing the project's src directory. However, that is not always correct. What is actually used is the current working directory which isn't necessarily the project root.

Are there cases when the CWD isn't the project root? If so, is there a way to genuinely use the project root?

> What does grammar_path take as an argument?

`AsRef<Path>`. It's a common idiom in the Rust libraries, but a bit hard to describe succinctly, because it can auto-transform strings to `Path`s.

lrpar/src/lib/ctbuilder.rs
ltratt: There is no `grmtools` binary so I'm not sure this can really happen.

CHANGES.md
ptersilie: > `AsRef<Path>`. It's a common idiom in the Rust libraries, but a bit hard to describe succinctly, because it can auto-transform strings to `Path`s.

I meant, what does the user pass in, not the type, but the destination. Do they pass a file or really just a path. I think it's the former, that's why I pointed out that it might be confusing to the end user, that the function is called `grammar_path` when actually they are passing in a file.

lrpar/src/lib/ctbuilder.rs
ptersilie: This was just an example. There will be some binary that the user runs, and since they can run it from anywhere, I feel like claiming this is always relative to the project root is misleading. Unless you mean it's only possible to run this from within the project's root due to other dependencies?

CHANGES.md
ltratt: Technically they pass in a relative path.

lrpar/src/lib/ctbuilder.rs
ltratt: I'm sure someone can work out how to run it in a weird way, but it's only documented as being runnable from a project's `build.rs` and I think there that CWD == project root (but I might be wrong).

ptersilie: Ah, it depends on `build.rs` which takes the grammar path and the compiles the parser. In that case I agree that it's unlikely someone will run it from anywhere but the project root.

CHANGES.md
ptersilie: To a file or directory?

ltratt: Can I suggest we address the questions to the methods' doc strings? `CHANGES.md` only includes a summary, not the complete specification.

ptersilie: What questions? I didn't see any in the doc strings. Anyway, I was just trying to help find clearer names for those methods. If you don't want to change them, that's fine.

ltratt: I mean: if you find the doc strings for these functions (they're all in this PR, as they've been changed), you might find that it helps give you ideas for the names. e.g. https://github.com/softdevteam/grmtools/blob/9cae583549012f11fe534a6f5f0ba608279ef535/lrpar/src/lib/ctbuilder.rs#L200

ptersilie: The only ideas I got left is maybe `local_grammar_file`, `relative_grammar_file`. My mind is a bit blank at the moment.

ltratt: I think `grammar_in_src_dir` might be the best bet?

ptersilie: I can live with that.

ltratt: See if https://github.com/softdevteam/grmtools/pull/247/commits/da8d4490ba6869892d189e4bb0b29ed7bec9a0bd works for you.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #248, opened 2021-09-14
Rework lrlex to have an equivalent API to lrpar.

ltratt:
`LexerBuilder::process_file_in_src` is deprecated in favour of `LexerBuilder::lexer_in_src_dir` and `LexerBuilder::process` in order that lrlex's compile-time API mirrors that of lrpar.


ltratt:
@ptersilie https://github.com/ltratt/grmtools/commit/f8d4b8955a06005893e6ff332926a6a723013a3f includes some changes that should probably have been in #247, so is perhaps best left as a separate commit.

PR #253, opened 2021-09-20
Faulty lexemes

ltratt:
Previously we supported the idea of "inserted" lexemes -- that is, lexemes inserted by error recovery, which were expected to have a type, a starting position, but no length. This PR generalises this to the concept of "faulty" lexemes which might result from, for example, lexing errors. Faulty lexemes can have a non-zero length.

I *think* this is the right thing to do, but I welcome thoughts @ptersilie!

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -254,7 +254,7 @@ impl<'lexer, 'input: 'lexer, StorageT: Copy + Eq + Hash + PrimInt + TryFrom<usiz
                 if r.name.is_some() {
                     match r.tok_id {
                         Some(tok_id) => {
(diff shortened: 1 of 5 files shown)


ptersilie:
In the example in the first commit. Does this change then produce a faulty lexeme `"blah`? This sounds similar to the way Wagner handled lexing errors, but creating error tokens (and to an extend what we did in Eco, though our error tokens where always a single char).

What we currently doing with something like `"blah`?

ltratt:
> What we currently doing with something like "blah?

We just produce a lexing error and stop.

ptersilie:
If I understood what this does correctly, then I don't see a problem with it.

PR #254, opened 2021-09-27
Make `Lexeme` a trait and separate lrpar and lrlex

ltratt:
This PR does several things (so it's not going to be fun to review -- sorry!), but everything stems from two desires:

1. `Lexeme` should be a trait not a struct (see https://github.com/softdevteam/grmtools/commit/d586d36f7a98a591a9efe8fd198567aee61e4f94 for the motivation),
2. Having done that, we can then remove the unfortunate coupling between lrlex and lrpar where the latter defined a concrete `Lexeme` struct that lrlex (or, indeed any other lexer) had to use (most of the remaining PR commits).

In the end this PR does more than I originally expected: it makes the split between lexer and parser clearer and should make it much easier for people to plug-in other lexers. But because I didn't know exactly where the PR was going to go, some of the early commits have things in (e.g. the name `StandardLexeme`) that subsequent commits change. Also, the fact that `Lexeme` is now a trait makes instantiating lexers and parsers more tedious (Rust's default type parameters aren't much help), with some inevitable changes to the building API. I have added a new `lrpar_config` option to `lrlex` which makes the standard "use lrlex and lrpar" case very easy (see the updated quickstart guide https://github.com/softdevteam/grmtools/blob/196a549c391462475f89d8e54a495a7498859c47/doc/src/quickstart.md), meaning that only "power users" have to face the full effects of the `Lexeme` trait.


ltratt:
bors try

ptersilie:
This looks, as far as I can tell, good to me. Just some typos in the commits:

26e8ceb: Having lrpar mandate use of a particular lexeme struct never ~~fault~~ felt
196a549: need to know about `StandardLexeme` and ~~so~~ so on

Please squash as needed.

ltratt:
Force pushed updated commits. Unfortunately buildbot is almost certainly out of action while our servers are currently behind a network blockage but hopefully we'll able to bors this at some point today!

ltratt:
bors try

ltratt:
bors try-

ltratt:
bors try

ltratt:
bors ping

ltratt:
bors try-

ltratt:
bors try

ltratt:
bors try

ltratt:
@ptersilie That last commit will need squashing away, assuming you're OK with it.

ptersilie:
Yeah, please squash.

ltratt:
Squashed.

PR #257, opened 2021-10-02
Implement a basic version of `%parse-param`.

ltratt:
We previously could parse `%parse_param` but not do anything with it. This commit adds support for `%parse-param` (using Bison's spelling) which, if defined, allows the user to pass an extra parameter to the parser. See `lrpar/cttests/src/parseparam.test` for an example:

```
%start S
%parse-param p: &'input u64
%%
S -> u64:
    'INT' { *p + $lexer.span_str($1.unwrap().span()).parse::<u64>().unwrap() }
;
%%
```

Now the generated parser can be called with two parameters `parse(&lexer, &1234)`.

Since Rust has tuples, `parse-param` only needs to take a single paramater (whereas Bison allows multiple extra parameters), which is fortunate, because Rust doesn't have varargs, which would make dealing with multiple parameters extremely difficult.

This commit is a bit simple in some ways: the type of the extra parameter must implement `Copy` and, if it's a reference, it must be tied to the `'input` lifetime. Perhaps these restrictions could be relaxed one day, but this is a start, and better than nothing.

Fixes https://github.com/softdevteam/grmtools/issues/212.


ltratt:
CC @ratmice -- this might be of interest to you, though it's not as ambitious as your original PR!

ratmice:
Thanks, will definitely have a look and see if what I was doing fits within the lifetime constraints!

ltratt:
Any luck, @ratmice?

ptersilie:
Just a couple of typos.

PR #259, opened 2021-10-09
Silence warnings about unused %parse-param variables.

ltratt:
If a user defines a `%parse-param` variable but doesn't use it in a particular rule, then rustc complains about an unused variable. We partly silenced this before with the `let _ = <var>;` trick -- but it wasn't quite in the right place to silence all the warnings. This PR fixes that in https://github.com/softdevteam/grmtools/pull/259/commits/3418f4231606cc009931cb95cfb1dc2520962f1b (I also fixed some whitespace ugliness that I noticed at the same time in https://github.com/softdevteam/grmtools/pull/259/commits/73dc83c49bd60e834283c423affdc2fef911bad4).

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -969,7 +969,7 @@ where
                 if actiont == "()" {
                     "".to_owned()
                 } else {
-                    format!("\n->                  {}", actiont)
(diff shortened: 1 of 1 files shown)


ltratt:
@ptersilie I'll need to squash that last commit away if you're OK with it (my fault!).

ptersilie:
Please squash.

PR #260, opened 2021-10-09
We are no longer tied to `'input`.

ltratt:
I'm not sure what I changed or when, but I no longer get compile-time errors if `%parse-param` references are not tied to `'input`. Whatever I did, I'm happy to loosen this restriction!


ltratt:
bors try

PR #268, opened 2021-11-18
Remove unneeded mutability.

ltratt:
The initial set was suggested by Clippy, which then made it possible for the compiler to suggest removing more.


vext01:
Good stuff.

bors r+

PR #281, opened 2022-03-20
Edition2021

ratmice:
When I saw the edition 2021 changes that allow 
```
let foo = "foo";
format!("{foo}");
```

style of code to work, I remembered some places in grmtools which used: `format!("{foo}", foo = foo);`
I thought perhaps this might clean some things up, I'm a bit wishy-washy on how it turned out in practice,
as now things are mixed between referring to the environment, and parameters.

i.e. `let foo = "foo"; format!("{foo}{bar}", bar = ...);`
Let me know if I should just drop that patch of the series.

It seemed worth a try however; I didn't really do a thorough examination of the code base for format strings so this likely misses some...

Leading up to that I fixed a lot of clippy lints that seemed to appear when switching to 2021 edition.
This confuses me, because I get different clippy output (nothing) for command line, and all the things fixed here from rust-analyzer when check-on-save set to clippy.  I don't know how to get rust-analyzer to tell me e.g. which compiler or command line arguments it uses.

However there are at least a few cases there which seemed like a definite win in readability and a few which were less clear
whether I would consider the new code better.

I didn't really consider compatibility while doing this, so feel free to close or postpone due to conflicts with pending changes.


ltratt:
This will need a `rustfmt`. If you're able to, please feel free to squash that into the appropriate commit(s) and force push an update. Otherwise, it can be left as a separate commit.

ltratt:
Is this ready for rereview?

ratmice:
Yeah, I believe that it should be I upgraded the compiler my command line sees reran `cargo fmt`, squashed it as you asked.
I'm quite confused why rustfmt didn't work last time, either a rustfmt version issue or I forgot to run it after the last 2 commits even though I distinctly remember doing so before push!

ratmice:
Well, I haven't investigated what is causing that test failure,
But since I've figured out that this PR update is needlessly entangled with fixes for a bunch of nightly/unstable clippy lints.
Most of these lints seem to have been moved out of running by default, So I'm having a bit of trouble now tracking down the lint that caused each change.

I'm going to close this, and open a new one with the minimal 2021 edition change, later we can investigate these lints on a case-by-case basis, and figure out what is causing that test failure, because it isn't obvious to me...

ratmice:
FWIW, I finally tracked down the thing rust-analyzer was doing differently from running `cargo clippy` on the command line.
Because of another setting, it would pass `cargo clippy --all-targets` presumably that allows the lint to catch more things.

PR #285, opened 2022-04-06
appease clippy --all-targets

ratmice:
Now that i've figured out how to reproduce all those clippy lints, I was getting in my initial 2021 edition bump.
Here is 2 patches one with one from `cargo clippy --all-targets`, and a follow up with `cargo +nightly clippy --all-targets`.

There is some subsequent cleanup I think that could be done, in particular calls to `state_exists`, pass the parameters as a reference sometimes, other times the reference is in the type of the variable passed as a parameter.

This could be made more uniform by declaring the reference in the let binding rather than call site perhaps.
I didn't do it here because it isn't required to appease clippy, and its already pretty big patch.


ltratt:
Thanks!

bors r+

PR #289, opened 2022-05-07
Add a name_span field to Rule.

ratmice:
This adds a name_span field to lexerdef's rule, I assume it is okay to depend on span here from lrpar in lrlex since the depdency is already there.

Notes about the impl:
The span refers to inside the quotes, of the quoted name, and in the case of an anonymous rule with no name,
It points to the empty string at the semicolon.

This is why name is an option but span is not, this still needs to be documented and tested, 
But I wanted to post this up for comments before writing docs/tests in case of feedback on the impl causes different behavior.
So for now I merely checked it just with `iter_rules()` manually.

It *doesn't* add spans for re_str (`re_span`?), which is private and I'm not sure it would be alright to have a public span for that private field, and i'm not sure if I would actually need it -- I still need to play around with adding this info to diagnostics to get a feel for that.

### DIFF
--- a/cfgrammar/src/lib/mod.rs
+++ b/cfgrammar/src/lib/mod.rs
@@ -53,7 +53,10 @@
 use serde::{Deserialize, Serialize};
 
 mod idxnewtype;
+pub mod span;
 pub mod yacc;
+
+pub use span::Span;
 
 /// A type specifically for rule indices.
 pub use crate::idxnewtype::{PIdx, RIdx, SIdx, TIdx};
--- a/cfgrammar/src/lib/span.rs
+++ b/cfgrammar/src/lib/span.rs
@@ -1,0 +1,45 @@
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// A `Span` records what portion of the user's input something (e.g. a lexeme or production)
+/// references (i.e. the `Span` doesn't hold a reference / copy of the actual input).
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Span {
+    start: usize,
+    end: usize,
+}
+
+impl Span {
+    /// Create a new span starting at byte `start` and ending at byte `end`.
+    ///
+    /// # Panics
+    ///
+    /// If `end` is less than `start`.
+    pub fn new(start: usize, end: usize) -> Self {
+        if end < start {
+            panic!("Span starts ({}) after it ends ({})!", start, end);
+        }
+        Span { start, end }
+    }
+
+    /// Byte offset of the start of the span.
+    pub fn start(&self) -> usize {
+        self.start
+    }
+
+    /// Byte offset of the end of the span.
+    pub fn end(&self) -> usize {
+        self.end
+    }
+
+    /// Length in bytes of the span.
+    pub fn len(&self) -> usize {
+        self.end - self.start
+    }
+
+    /// Returns `true` if this `Span` covers 0 bytes, or `false` otherwise.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -8,6 +8,8 @@ use indexmap::{IndexMap, IndexSet};
 
 use super::Precedence;
 
+use crate::Span;
+
 /// An AST representing a grammar. This is built up gradually: when it is finished, the
 /// `complete_and_validate` must be called exactly once in order to finish the set-up. At that
 /// point, any further mutations made to the struct lead to undefined behaviour.
@@ -17,6 +19,7 @@ pub struct GrammarAST {
     pub rules: IndexMap<String, Rule>,
     pub prods: Vec<Production>,
     pub tokens: IndexSet<String>,
+    pub spans: Vec<Span>,
     pub precs: HashMap<String, Precedence>,
     pub avoid_insert: Option<HashSet<String>>,
     pub implicit_tokens: Option<HashSet<String>>,
@@ -115,6 +118,7 @@ impl GrammarAST {
             rules: IndexMap::new(), // Using an IndexMap means that we retain the order
             // of rules as they're found in the input file.
             prods: Vec::new(),
+            spans: Vec::new(),
             tokens: IndexSet::new(),
             precs: HashMap::new(),
             avoid_insert: None,
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -12,7 +12,7 @@ use super::{
     parser::{YaccParser, YaccParserError},
     YaccKind,
 };
-use crate::{PIdx, RIdx, SIdx, Symbol, TIdx};
+use crate::{PIdx, RIdx, SIdx, Span, Symbol, TIdx};
 
 const START_RULE: &str = "^";
 const IMPLICIT_RULE: &str = "~";
@@ -42,9 +42,9 @@ pub struct YaccGrammar<StorageT = u32> {
     rules_len: RIdx<StorageT>,
     /// A mapping from `RIdx` -> `String`.
     rule_names: Vec<String>,
-    /// A mapping from `TIdx` -> `Option<String>`. Every user-specified token will have a name,
+    /// A mapping from `TIdx` -> `Option<(Span, String)>`. Every user-specified token will have a name,
     /// but tokens inserted by cfgrammar (e.g. the EOF token) won't.
-    token_names: Vec<Option<String>>,
+    token_names: Vec<Option<(Span, String)>>,
     /// A mapping from `TIdx` -> `Option<Precedence>`
     token_precs: Vec<Option<Precedence>>,
     /// A mapping from `TIdx` -> `Option<String>` for the %epp declaration, giving pretty-printed
@@ -188,11 +188,11 @@ where
             rule_map.insert(v.clone(), RIdx(i.as_()));
         }
 
-        let mut token_names: Vec<Option<String>> = Vec::with_capacity(ast.tokens.len() + 1);
+        let mut token_names: Vec<Option<(Span, String)>> = Vec::with_capacity(ast.tokens.len() + 1);
         let mut token_precs: Vec<Option<Precedence>> = Vec::with_capacity(ast.tokens.len() + 1);
         let mut token_epp: Vec<Option<String>> = Vec::with_capacity(ast.tokens.len() + 1);
-        for k in &ast.tokens {
-            token_names.push(Some(k.clone()));
+        for (i, k) in ast.tokens.iter().enumerate() {
+            token_names.push(Some((ast.spans[i], k.clone())));
             token_precs.push(ast.precs.get(k).cloned());
             token_epp.push(Some(ast.epp.get(k).unwrap_or(k).clone()));
         }
@@ -202,7 +202,7 @@ where
         token_epp.push(None);
         let mut token_map = HashMap::<String, TIdx<StorageT>>::new();
         for (i, v) in token_names.iter().enumerate() {
-            if let Some(n) = v.as_ref() {
+            if let Some((_, n)) = v.as_ref() {
                 token_map.insert(n.clone(), TIdx(i.as_()));
             }
         }
@@ -460,7 +460,9 @@ where
     /// Return the name of token `tidx` (where `None` indicates "the rule has no name"). Panics if
     /// `tidx` doesn't exist.
     pub fn token_name(&self, tidx: TIdx<StorageT>) -> Option<&str> {
-        self.token_names[usize::from(tidx)].as_deref()
+        self.token_names[usize::from(tidx)]
+            .as_ref()
+            .map(|x| x.1.as_str())
     }
 
     /// Return the precedence of token `tidx` (where `None` indicates "no precedence specified").
@@ -473,6 +475,11 @@ where
     /// pretty-printed value"). Panics if `tidx` doesn't exist.
     pub fn token_epp(&self, tidx: TIdx<StorageT>) -> Option<&str> {
         self.token_epp[usize::from(tidx)].as_deref()
+    }
+    pub fn token_span(&self, tidx: TIdx<StorageT>) -> Option<&Span> {
+        self.token_names[usize::from(tidx)]
+            .as_ref()
+            .map(|(span, _)| span)
     }
 
     /// Get the action for production `pidx`. Panics if `pidx` doesn't exist.
@@ -498,7 +505,7 @@ where
     pub fn tokens_map(&self) -> HashMap<&str, TIdx<StorageT>> {
         let mut m = HashMap::with_capacity(usize::from(self.tokens_len) - 1);
         for tidx in self.iter_tidxs() {
-            if let Some(n) = self.token_names[usize::from(tidx)].as_ref() {
+            if let Some((_, n)) = self.token_names[usize::from(tidx)].as_ref() {
                 m.insert(&**n, tidx);
             }
         }
@@ -509,7 +516,7 @@ where
     pub fn token_idx(&self, n: &str) -> Option<TIdx<StorageT>> {
         self.token_names
             .iter()
-            .position(|x| x.as_ref().map_or(false, |x| x == n))
+            .position(|x| x.as_ref().map_or(false, |(_, x)| x == n))
             // The call to as_() is safe because token_names is guaranteed to be small
             // enough to fit into StorageT
             .map(|x| TIdx(x.as_()))
@@ -1032,7 +1039,7 @@ mod test {
         super::{AssocKind, Precedence, YaccGrammar, YaccKind, YaccOriginalActionKind},
         rule_max_costs, rule_min_costs, IMPLICIT_RULE, IMPLICIT_START_RULE,
     };
-    use crate::{PIdx, RIdx, Symbol, TIdx};
+    use crate::{PIdx, RIdx, Span, Symbol, TIdx};
     use std::collections::HashMap;
 
     #[test]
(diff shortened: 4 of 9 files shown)


ratmice:
I converted to a draft, because I would like to think this through a bit more,
The process of going from `let (missing_from_lexer, missing_from_parser) = lexer.set_rule_ids(...)` to a span is a bit painful.

Because `set_rule_ids` takes lexer as mutable, and the return values borrow from it.
This makes `lexer.get_rule_by_name().name_span()` a little more difficult than perhaps it could be, so I'd like to play around with that. At the same time it'd be nice if I figured out getting parser spans for `missing_from_lexer` too.

ltratt:
I'm fine with lrlex's `Rule` having a `name_span` though it makes me wonder if we should convert lrlex to not store `name` directly (but that might cause problems elsewhere).

ratmice:
Yeah, this is one of the things I was wondering about,
there looks to be a few uses that are hard to remove, but 

1. checking for duplicate names
2. the mapping from parser StorageT to lexer token id in set_rule_id's is done through name equality.
3. checking if a name is empty which could be replaced by `span.length() == 0?`

the first 2 are hard to get remove, I'm not certain these are all the cases, but they are the ones that stand out to me. What I kind of notice is that at least after `set_rule_ids`
 I'm not seeing anything yet which would require them to still be around.  That is what i've been looking into.

ltratt:
> the first 2 are hard to get remove

At least the first is easy *if* you've still got a reference to the lex file (then it's just comparing two string slices).

ratmice:
true, anyhow i'm going to play around with:
```
pub struct LRNonStreamingLexerDef<LexemeT, StorageT> {
    names: HashMap<Option<Cow<'static, str>>, Vec<(lrpar::Span, usize)>>,
    rules: Vec<Rule<StorageT>>,
    phantom: PhantomData<LexemeT>,
}
```

which has a less binomial behavior on the happy path, but i'm a bit skeptical because then `rules` is no longer just valid by construction, but it would allow you to continue after hitting the first duplicate, and report all errors.

ltratt:
What's the use of `Cow` here?

If we want to use `Span`s I'd be more inclined to do something like:

```rust
pub struct LRNonStreamingLexerDef<'a, LexemeT, StorageT> {
    lexerdef_str: &'a str,
    rules: Vec<Rule<StorageT>>,
    phantom: PhantomData<LexemeT>,
}
```

There's some ickiness because now `rule.name` would a) have to become a function b) have to take a reference to the `&'a str` one way or another. So it might not be the best option?

ratmice:
I was mostly using it as a go between between `&str` and `String`, which allows you to do a lookup with an `&str` without doing a full `to_string()` in I believe it was `get_rule_by_name`.

But I would be happy to look at the above representation too.

ltratt:
My experience of these matters is that I have to try about 7 representations before I find one I don't dislike *too* much :)

ratmice:
FWIW, I was debating on whether I should even post this, I figured its worthwhile if only in that it shows all the usage sites.
The Cow stuff definitely does not seem to be carrying it's weight, I was only able to avoid conversion to string in a few cases.

It gives me some thoughts for the next pass, like the table for duplicate checking could probably be kept isolated to
`parser.rs` in which case we can drop the `Option` from the key which causes the coercion hassle which lead me to use `Cow` in the first place.
At the very least it seems worthwhile to keep any potential duplicate table to a separate from name lookup.

> There's some ickiness because now rule.name would a) have to become a function 

One of the main insights for me was the usage in lexer/src/main.rs, it already ends up doing a linear scan lookup by `tok_id`. Before it can get to the rule. I think that is the hottest path here, so i'm not sure if we really benefit from it being a field.

Anyhow this probably represents the maximum field movement (moving both name and span out of rule), replacing it with a bool to indicate the presence or lack of a name.  I should probably spend some time measuring what affect this has (if any) on actual parsing though absent the name lookups to see if it is actually worth doing.

https://github.com/softdevteam/grmtools/compare/master...ratmice:rule_name_experiment1?expand=1

ltratt:
I might be wrong (it's been a while...) but I *think* the performance impact will be near-zero, as the names are only used when joining the lexer + parser up (the names aren't used during parsing itself when everything is numeric indices).

[PS: the link at the end of your message tries to have me open a PR! I'm not sure if it was supposed to show a diff?]

ratmice:
Update the link, I figured it would show a comparison, given the url was something like /compare/... but guess not.

ltratt:
Quick comments:

* `get_info_for_rule_name`: I would probably just make this `span_for_rule_name` and not try returning a string and `usize` simultaneously (off-hand I can't even remember what the `usize` represents!).
* I agree with you about `Cow` probably not being useful enough to justify the complexity.

ratmice:
the `usize` is the index into the `rules` vector, so e.g. the same as the argument for `get_rule`, I don't see though why you would have a name but `no tok_id`, or `index: usize`, so I think you are right that it could probably be dropped.

ratmice:
I did some testing, and you are absolutely correct that it had zero impact.

So I guess I had a realization of sorts, that I hadn't previously picked up on previously.  Which I think makes it more difficult to just move name out of `Rule` and into `LRNonStreamingLexerDef`, which is the `from_rules` constructor.

Experiment 2 is basically keeping what we have but trying to make set_rule_ids return `Rc<str>`,
which cleans up lifetimes for callers and and the impl of set_rule_ids.
https://github.com/ratmice/grmtools/commit/ce7b7c8ab03c69c6100dd96b00bc4ee8ea790511

Experiment 3 was very similar to the representation you had suggested except that it had a `ByteString` instead of `&str` for the `lexerdef_str`, and returned `ByteString` slices out of `set_rule_ids`, it looked like:

```
pub struct LRNonStreamingLexerDef<'a, LexemeT, StorageT> {
    lexerdef_source: ByteString,
    rules: Vec<Rule<StorageT>>,
    phantom: PhantomData<LexemeT>,
}
fn set_rule_ids(
        &mut self,
        rule_ids_map: &HashMap<&str, StorageT>,
) -> (Option<HashSet<ByteString>>, Option<HashSet<ByteString>>);
```

Which is where I noticed the problem with `lexerdef_source`, or `lexerdef_str` in the precense of of `from_rules` is very gordian my first experiment  (I believe) just returned `None` for name related functions if they were called on a `from_rules` constructed lexerdef like is built from `CTBuilder`.

Since there really is virtually no impact, I don't see much benefit from trying to move `name` out. I was mistaken that I thought `CTBuilder` *had* a `Vec<Rules>`, but instead it is actually *constructing` a `LRNonStreamingLexerDef` *from* a `Vec<Rules>`.

Curious your thoughts about this owned set_rule_ids return value though/whether you think that is worthwhile.

ltratt:
If it's any consolation, it took me a number of attempts to get a half-sensible split between `LexerDef` and `Lexer` -- even when I had the general idea I kept putting too much in `Lexer` and not enough in `LexerDef`. I'm sure the design could be improved, perhaps fundamentally, or perhaps just in small details.

My immediate reaction is that I'm not sure that `Rc<str>` is worth the API change though. We win something in terms of memory used when we convert `String` to `Rc<str>` but we lose something when we convert `&str` to `Rc<str>` -- but the difference in either case is probably immeasurably small and `Rc<str>` is such a weird type that I had to stop and think about where the `&` had gone :)

Maybe I should now ask the high-level question I should have asked earlier: is there something in the current design that's causing you problems in your other project or was this PR more motivated by seeing something that looked a bit sub-optimal? If the latter, we might just have to chalk this up to "worth trying but time to move on" but if the former it might be worth thinking more fundamentally about what a better design could be.

ratmice:
Kind of 3 distinct motivations I guess, 

The primary motivation for introducing `Rc` can kind of be seen in the diff.
https://github.com/ratmice/grmtools/commit/ce7b7c8ab03c69c6100dd96b00bc4ee8ea790511#diff-0e2eb9d52e952250f05f36f38a48804906daea69e8dc0c52f4221294a0f70f16L314-L322

e.g. the need to call set_rule_ids in a block, do a deep copy on the results, and bubble that up so I can then can call `get_rule_by_name`.  By removing that lifetime I can avoid introducing a block since it takes lexerdef mutably.
```
let (missing_from_lexer, missing_from_parser) = {
   let (ml, mp) = lexerdef.set_rule_ids(rule_id_map); 
  (deep_copy!(ml), deep_copy!(mp))
}
for missing in missing_from_lexer {
  lexerdef.get_rule_by_name(...)
}
```

and just do 
```
  let (missing_from_lexer, missing_from_parser) = lexerdef.set_rule_ids(rule_id_map);
  for missing in missing_from_lexer {
    ...
  }
```

I think the only difference being my program is at that point already at an uncomfortable indentation level.

Other aspects I was looking into (trying to work on duplication) were aimed at 
a) emitting better diagnostics, currently when we get the error we would just get the location of the second of the duplicate found.
b) speeding up the compile-time, there seemed like a lot of linear scans that could potentially be made more efficient.

In particular I don't think the edit->save->compile loop requires fast lexer/parser *generation*, since you only compile once in that loop. But when you are building the lexer after every keystroke for instance it becomes more important, I haven't really tried with any large number of tokens yet to see at what point it currently becomes infeasible.

ltratt:
Maybe a quick experiment to see how fast it is is worthwhile. My intuition is that, because lexer generation is pretty simple it probably already will cope with compiling big lex files multiple times per second. I might be very wrong though -- lrlex isn't optimised for speed at all (yet)!

ratmice:
One more question about `setup_rule_ids` thing which would also work removing lifetimes is an alternative function,
like (I haven't thought of a good name), curious if that kind of thing would that be preferrable to trying to change `set_rule_id`
My feeling is it's a bit awkward being a part of a trait -- anyhow it isn't important I guess.

```
/// An alternative function to set_rule_id where the return type isn't bound to the lifetime of self
fn set_rule_ids_alternative(&mut self,
        rule_ids_map: &HashMap<&str, StorageT>,
) -> (Option<HashSet<StorageT>>, Option<HashSet<usize>>);
```

ltratt:
I might have misunderstood but is the lifetime aspect causing actual problems or is it just that lifetimes are always a bit frightening and we wouldn't mind them gone?

ratmice:
It doesn't really cause problems, because I can convert it to an owned `Option<HashSet<String>>`, it's just a bit more work to do so and slower, e.g. `usize -> Rule` via `get_rule` is O(1) while `get_rule_by_name` requires a `rules.iter().find` on top of the entire conversion to an owned value.

I think that is probably what bothering me about it -- that it is a minor inconvenience but ends up being more work.
Anyhow I think i'm probably being difficult here and need to learn to stop worrying and love the clone.

ratmice:
I think for now it's good to just try and add the span without any of the larger refactoring around name/other tangentally related stuff, So I went ahead and marked ready for review.

ratmice:
So, I forgot I had intended to implement spans for parser before I got distracted by shiny things.

This last implements spans for parser tokens, the patch probably needs some work, in particular because `Span` lives in `lrpar` which is downstream of `cfgrammar`.  I ended up just using `(usize, usize)` for now just to initially get it working.

ratmice:
So I think one thing to consider regarding this series is dealing with `Results` like `LexBuildError` `StateTableError` and `YaccGrammarError`.  

In these we don't have a lexerdef or grm respectively to look up spans.  Some like `LexBuildError` currently have privae `line`, `col` fields.  `YaccGrammarError::GrammarValidationError` has a symbol but without the `grm` no way to turn it into a span.

I Imagine if we're going to export span information via these, it's probably going to need to be done as a field in each error, but I think its worth thinking about if only how to make them more uniform.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -17,6 +17,7 @@ pub struct GrammarAST {
     pub rules: IndexMap<String, Rule>,
     pub prods: Vec<Production>,
     pub tokens: IndexSet<String>,
+    pub spans: Vec<(usize, usize)>,
```
ltratt, line 20: This `(usize, usize`) type makes me wonder if we should put `Span` in `cfgrammar`, so that `lrpar` can access it, but `cfgrammar` can also use it. [Subtext: I think `Span` is clearer in intent than `(usize, usize)`.] Any thoughts?

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -45,6 +45,12 @@ pub struct YaccGrammar<StorageT = u32> {
     /// A mapping from `TIdx` -> `Option<String>`. Every user-specified token will have a name,
     /// but tokens inserted by cfgrammar (e.g. the EOF token) won't.
     token_names: Vec<Option<String>>,
+    /// A mapping from `TIdx` -> `Option<(usize, usize)>`.
+    /// The first element is the absolute byte offset of the first element of the name
+    /// in the yacc file this grammar was produced from (if any).
+    /// The second is the absolute byte offset of the end of the name.
+    /// Elements not represented by a yacc source file, or inserted by cfgrammar return None.
```
ltratt, line 52: Should we just have `token_names: Vec<Option<(Span, String)>>` rather than two separate, but highly related, fields?

cfgrammar/src/lib/yacc/ast.rs
ltratt: [Though we could do this in a second PR. It's not strictly related to this PR.]

ratmice: Sounds good to me.

cfgrammar/src/lib/yacc/grammar.rs
ratmice: Will give that a go!

cfgrammar/src/lib/yacc/ast.rs
ratmice: I think this was easy enough that it was worth getting done before inclusion.

I wasn't exactly sure where in cfgrammar, I ended up putting it in a `utils` module, because yacc didn't seem right if we ever get other formats besides yacc.  But open to thoughts here if you have other ideas...

cfgrammar/src/lib/yacc/grammar.rs
ratmice: I haven't had much luck with this, It's easy enough to build the `Vec<Option<(Span, String)>>`,
but all of the `name_option.as_ref()` have given me a bit of trouble, (I think it is tuple not implementing as_ref).
And `x.map(|(_, x)| x).as_ref()` as giving me some inference troubles,

But i'll try it again now that I'm not trying to change 2 things at once.

In theory I agree `Vec<Option<(Span, String)>>` is nicer, just not sure implementation wise yet.

ratmice: Not so bad if you don't accidentally mix up `(Span, String)` with ``(String, Span)`.

ltratt:
Let's move `Span` from `utils` to either `cfgrammar/span.rs` or just in `cfgrammar/mod.rs` (I'm allergic to "utils" as a name unless I'm stumped for anything better!). I think this PR should then be ready!

ratmice:
Alright, I will probably want to follow up with `Span` for `Ridx` I think so we might want to hold off on a release until then, or we could wait until that is done to merge, but not picky in this regard.

cfgrammar/src/lib/mod.rs
```diff
@@ -53,7 +53,7 @@
 use serde::{Deserialize, Serialize};
 
 mod idxnewtype;
-pub mod utils;
+pub mod span;
```
ltratt, line 56: On reflection, we should also `use span::Span` so most people can just do `use cfgrammar::Span` without worrying about the module it's defined in. [This also makes the `crate::span::Span` dance simpler as just `crate::Span`.]

ltratt:
> Alright, I will probably want to follow up with Span for Ridx

Not sure I understand?

ratmice:
I meant the span returning equivalent, of the [rule_name](https://docs.rs/cfgrammar/latest/cfgrammar/yacc/grammar/struct.YaccGrammar.html#method.rule_name) on YaccGrammar, (Primarily i'm thinking for jumping to the place the rule is defined).

ltratt:
> I meant the span returning equivalent, of the [rule_name](https://docs.rs/cfgrammar/latest/cfgrammar/yacc/grammar/struct.YaccGrammar.html#method.rule_name) on YaccGrammar

Ah you mean to return a `Span`? I'm neutral on that.

Please squash. [In case the terminology isn't clear, let's assume we're following exactly the same process documented at https://ykjit.github.io/ykdocs/contributing/prs.html]

ratmice:
> Ah you mean to return a Span? I'm neutral on that.

Cool, it might take me a bit before i'm prepared to use it.
> Please squash. (In case the terminology isn't clear, let's assume we're following exactly the same process documented at
> https://ykjit.github.io/ykdocs/contributing/prs.html)

I felt it at 2 commits, because the first is self contained, but given that there is no overlap probably could have gone with one.

ltratt:
Could we add a bit of motivation to the commit message body of https://github.com/softdevteam/grmtools/pull/289/commits/bbc91bf7451503cdfcd1362cca06faffc3c9c65d ? So, not just what we've done but why. I find that super helpful when looking through git history! Please feel free to force push an update.

cfgrammar/src/lib/span.rs
```diff
@@ -0,0 +1,41 @@
+/// A `Span` records what portion of the user's input something (e.g. a lexeme or production)
+/// references (i.e. the `Span` doesn't hold a reference / copy of the actual input).
+#[derive(Clone, Copy, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
```
ratmice, line 3: Just noticed an issue with this, I believe they need feature guards here for serde.

ratmice: Fixed, this only showed up when doing `cargo build` under `cfgrammar/`, because if I recall I believe one of the top-level crates uses the feature.

ltratt:
Please squash.

ratmice:
Done, and fwiw thanks for the time spent reviewing this, hopefully my next patch will go smoother :crossed_fingers:

ltratt:
Thanks!

bors r+

PR #291, opened 2022-05-15
Spans for Symbols

ratmice:
Marking as a draft, I would like to 
1. test this more, so I would like to just work of this branch for a bit in my project, exercising it.
2. think about the second patch a bit more.

Regarding the second patch -- what I was after was spans for `productions`, adding them directly to productions seemed fraught (can't seem remember why though, I will have to check). adding them to symbols *seemed* to be the way, then get the Symbol for the production and you have it's span.

But it is kind of roundabout, and also caused the custom `hash` and `eq` impls.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -131,13 +131,14 @@ impl GrammarAST {
         }
     }
 
-    pub fn add_rule(&mut self, name: String, actiont: Option<String>) {
+    pub fn add_rule(&mut self, name: String, actiont: Option<String>, span: Option<Span>) {
         self.rules.insert(
             name.clone(),
             Rule {
                 name,
                 pidxs: Vec::new(),
                 actiont,
+                span,
             },
         );
     }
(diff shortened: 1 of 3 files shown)


ltratt:
https://github.com/softdevteam/grmtools/pull/291/commits/5e6b1a10952d95309d2ba879d65b847d01887222 makes sense to me and we can probably get it in straight away.

I'm less sure about https://github.com/softdevteam/grmtools/pull/291/commits/b0fa770f16ac9399c4f4c9994203c6ee1e8ec64a in part because of the `PartialEq` hack: that feels like it could cause some weird problems in the future. I think that's because `Symbol` is the "abstract" notion of a name without being tied to a particular input. That means, perhaps, that if someone wants to carry around `Span` information they should do `(Span, Symbol)`?

ratmice:
I'll have a go at `(Span, Symbol)`, and either close this or rebase on top of the new PR once that is ready.

ratmice:
I just close this for now, easist to just start fresh, but I'll leave it around for a while in case we decide to reverse course.

PR #292, opened 2022-05-15
span for rules

ratmice:
This pulls 5e6b1a10952d95309d2ba879d65b847d01887222 out of #291 since it is more ready than the subsequent patch in that branch yet free standing.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -33,7 +33,7 @@ pub struct GrammarAST {
 
 #[derive(Debug)]
 pub struct Rule {
-    pub name: String,
+    pub name: (String, Span),
     pub pidxs: Vec<usize>, // index into GrammarAST.prod
     pub actiont: Option<String>,
 }
@@ -131,11 +131,11 @@ impl GrammarAST {
         }
     }
 
-    pub fn add_rule(&mut self, name: String, actiont: Option<String>) {
+    pub fn add_rule(&mut self, (name, name_span): (String, Span), actiont: Option<String>) {
         self.rules.insert(
             name.clone(),
             Rule {
-                name,
+                name: (name, name_span),
                 pidxs: Vec::new(),
                 actiont,
             },
@@ -260,7 +260,7 @@ impl GrammarAST {
 mod test {
     use super::{
         super::{AssocKind, Precedence},
-        GrammarAST, GrammarValidationError, GrammarValidationErrorKind, Symbol,
+        GrammarAST, GrammarValidationError, GrammarValidationErrorKind, Span, Symbol,
     };
 
     fn rule(n: &str) -> Symbol {
@@ -286,8 +286,9 @@ mod test {
     #[test]
     fn test_invalid_start_rule() {
         let mut grm = GrammarAST::new();
+        let empty_span = Span::new(0, 0);
         grm.start = Some("A".to_string());
-        grm.add_rule("B".to_string(), None);
+        grm.add_rule(("B".to_string(), empty_span), None);
         grm.add_prod("B".to_string(), vec![], None, None);
         match grm.complete_and_validate() {
             Err(GrammarValidationError {
@@ -301,8 +302,9 @@ mod test {
     #[test]
     fn test_valid_start_rule() {
         let mut grm = GrammarAST::new();
+        let empty_span = Span::new(0, 0);
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string(), None);
+        grm.add_rule(("A".to_string(), empty_span), None);
         grm.add_prod("A".to_string(), vec![], None, None);
         assert!(grm.complete_and_validate().is_ok());
     }
@@ -310,9 +312,10 @@ mod test {
     #[test]
     fn test_valid_rule_ref() {
         let mut grm = GrammarAST::new();
+        let empty_span = Span::new(0, 0);
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string(), None);
-        grm.add_rule("B".to_string(), None);
+        grm.add_rule(("A".to_string(), empty_span), None);
+        grm.add_rule(("B".to_string(), empty_span), None);
         grm.add_prod("A".to_string(), vec![rule("B")], None, None);
         grm.add_prod("B".to_string(), vec![], None, None);
         assert!(grm.complete_and_validate().is_ok());
@@ -321,8 +324,9 @@ mod test {
     #[test]
     fn test_invalid_rule_ref() {
         let mut grm = GrammarAST::new();
+        let empty_span = Span::new(0, 0);
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string(), None);
+        grm.add_rule(("A".to_string(), empty_span), None);
         grm.add_prod("A".to_string(), vec![rule("B")], None, None);
         match grm.complete_and_validate() {
             Err(GrammarValidationError {
@@ -336,9 +340,10 @@ mod test {
     #[test]
     fn test_valid_token_ref() {
         let mut grm = GrammarAST::new();
+        let empty_span = Span::new(0, 0);
         grm.tokens.insert("b".to_string());
         grm.start = Some("A".to_string());
-        grm.add_rule("A".to_string(), None);
+        grm.add_rule(("A".to_string(), empty_span), None);
         grm.add_prod("A".to_string(), vec![token("b")], None, None);
         assert!(grm.complete_and_validate().is_ok());
     }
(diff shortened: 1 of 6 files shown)


cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -131,13 +132,14 @@ impl GrammarAST {
         }
     }
 
-    pub fn add_rule(&mut self, name: String, actiont: Option<String>) {
+    pub fn add_rule(&mut self, name: String, actiont: Option<String>, span: Option<Span>) {
```
ltratt, line 135: Something that just occurred to me: what is the span of? Is it the rule's name or the whole rule (including productions etc)?

ratmice: Ahh, this could be commented better, it is the LHS of the rule so it's name, the easiest place to see it is (I now see got cut off outside the diff context)

https://github.com/softdevteam/grmtools/pull/292/files/3f9278a8c9a06cacb99014bc7cfd3c0d8174314c#diff-f19cd37f82592a87f9e6c986053e2b5e5194c270f9ba8c6185ea5633125e0b74R1478-R1491

ltratt: OK, so if it's the name (which makes sense) then the signature should, I think, be:

```rust
 pub fn add_rule(&mut self, name: String, name_span: Option<Span>, actiont: Option<String>)
```

ratmice: Well, my link didn't seem to work, but it's line 1478-1491 of the function `test_token_spans` in grammar.rs

ltratt:
Please squash.

ltratt:
I've spent a bit of time going through the codebase to remind myself exactly what things like `Symbol` mean and do (it's been a while...). I have the following thoughts:

* I think you were right that adding `Span` *inside* `Symbol` is the right thing to do -- my suggestion of moving it outside is less good (sorry!).
* I'm not sure that `Option<Span>` is ideal (in part because of space worries: it will add a whole extra machine word to each `Symbol`: would it be better to make use of an empty `Span` to signify the same thing?

ratmice:
> I've spent a bit of time going through the codebase to remind myself exactly what things like `Symbol` mean and do (it's been a while...). I have the following thoughts:
> 
>     * I think you were right that adding `Span` _inside_ `Symbol` is the right thing to do -- my suggestion of moving it outside is less good (sorry!).

No worries, I was pretty uncertain of that myself, so it helps to actually see the alternative.
 
>     * I'm not sure that `Option<Span>` is ideal (in part because of space worries: it will add a whole extra machine word to each `Symbol`: would it be better to make use of an empty `Span` to signify the same thing?
 
So, for `Option<Span>`, we have a couple of different cases (In particular for `Rules`), we could easily drop the option,
i.e. we only ever use the `None` value in the testsuite when generating rules for test cases, So for `Rules` this is only because there doesn't seem to be a requirement that rules are actually backed by any source file. 

For `Rule`, my initial attempt left out the `Option`, and I made up spans for the test cases.  It works totally works, but is a bit inconvenient and odd.

For `Symbol` this is somewhat different, because we add `implicit` symbols.  So even if we have a backing source file, there is not necessarily a valid `Span` for a given symbol, *I* think using a `Span::new(0, 0)` or some other empty span seems fine.
It's generally the case that we only see these when they produce erroneous results, and we shouldn't be encountering spans for `implicit` in those.

>     * I think it might make sense to bundle `(String, Span)` as a tuple rather than two named arguments.
> 
> Any thoughts?

Seems fine

ltratt:
It sounds like we're more-or-less on the same page here. How about we recast this PR (and the others) so that:

* `Span` in inside `Symbol`.
* `Span::new(0, 0)` conventionally means "this symbol didn't come from a user's source file".
* Rather than having two fields/parameters we always have a tuple `(String, Span)`.

Sound like a plan? Sorry for creating more work!

ratmice:
Sorry, Now that I think about it I don't think the `Borrow<Name>` idea would end up how we want since it'd allow you to look up e.g. `Symbols` given a `Rule` by name, so we'd end up with `Borrow<RuleName>`, and `Borrow<SymbolName>`, giving me pause on top of it being a relatively little used/known rust `HashMap` feature.

ltratt:
> But I do want to double check if I could get an ACK on the Eq/Hash hack.

I think we need to find a better way of doing this: it's just too dangerous IMHO. For example we could have, say, a `eq_str_name` method that only does equality on the string.

ratmice:
Glad I asked, I'm not positive that actually putting it in `Symbol` will work without that, I have to check the collections they get stored in.

ratmice:
FWIW, Here is the only explanation I know of for the `Hash` trick I was referring to, in case it is unfamiliar (should have probably dug this up when I mentioned it).  http://idubrov.name/rust/2018/06/01/tricking-the-hashmap.html

ltratt:
That trick is used extensively in CPCT+ so I know it well :)

ratmice:
Yeah it definitely seems like we can't just add it to symbol without some form of hack or trick.  Because at least the stategraph is going to behave differently once we add `Span`'s to `Symbol`, because of `edges: Vec<HashMap<Symbol<StorageT>, StIdx>>,`.

I think it is plausible though that there is a way in which the `Borrow` trick might be tolerable (or more tolerable than I had originally thought), so I think i will at least give it a try.

ltratt:
> Yeah it definitely seems like we can't just add it to symbol without some form of hack or trick. Because at least the stategraph is going to behave differently once we add Span's to Symbol, because of edges: Vec<HashMap<Symbol<StorageT>, StIdx>>,.

Hmm, I hadn't thought of that.

ratmice:
Just noticed something which at least explains some confusion on my part :thinking: "I thought Symbol had a name String",
because there are both [`cfgrammar::Symbol`](https://docs.rs/cfgrammar/latest/cfgrammar/yacc/ast/enum.Symbol.html) and [`cfgrammar::yacc::ast::Symbol`](https://docs.rs/cfgrammar/latest/cfgrammar/enum.Symbol.html).

Which may/may not change things if they have to be in correspondence.

ltratt:
Aha! (You can tell it's a long time since I've looked at this part of the code!) They don't need to be in correspondence, I think (though we should check). cfgrammar::yacc::ast::Symbol](https://docs.rs/cfgrammar/latest/cfgrammar/enum.Symbol.html) should definitely have a `Span` associated with it, so we could perhaps do that as a first step then think about `cfgrammar::Symbol` afterwards.

[And perhaps in another PR we can think of a way to better differentiate these two things!]

ratmice:
So, I updated both this and the `Symbol` branch, overall things went okay even with no `Hash`/`Eq` hack.  I had to add a `Hash` impl to `Span` though merely so the `derive` works.  But both are the standard, I think this should at least be close.

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -414,7 +414,12 @@ where
 
     /// Return the name of rule `ridx`. Panics if `ridx` doesn't exist.
     pub fn rule_name(&self, ridx: RIdx<StorageT>) -> &str {
```
ltratt, line 416: For consistency (and clarity) I reckon this should be either:

```rust
pub fn rule_name(&self, ridx: RIdx<StorageT>) -> (&str, Span)
```

or the two functions are `rule_name_str` and `rule_name_span`.

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -426,7 +431,7 @@ where
     pub fn rule_idx(&self, n: &str) -> Option<RIdx<StorageT>> {
         self.rule_names
             .iter()
-            .position(|x| x == n)
+            .position(|(x, _span)| x == n)
```
ltratt, line 434: For consistency with the rest of the code, let's use `_` rather than `_span`.

ltratt:
>  I had to add a Hash impl to Span

Would I be an idiot if I said I couldn't see this?

ratmice:
> > I had to add a Hash impl to Span
> 
> Would I be an idiot if I said I couldn't see this?

Sorry, it's in the other branch, I'm getting a little branch fatigued remembering what is where.

ltratt:
So that change isn't needed in this branch at all? Let's quickly check :)

bors try

cfgrammar/src/lib/yacc/grammar.rs
ratmice: Sounds okay to me, do we just want that one `rule_name (..) -> (&str, Span)` or all three?
I had mostly left it how it was before strictly for backwards compatibility, but I don't think this ever gets called except by things like nimbleparse.

ratmice: But yeah thinking about it, since we changed the signature of `add_rule` (and probably other things) incompatibly, it's probably worth changing this now.  To conform to a similar signature

ltratt: I think we either want one combined method or two separate ones (but not three). I'm neutral between one method or two.

ltratt:
Please squash.

ltratt:
Are you done? [If you don't ping me on PRs I don't know to look :)]

ratmice:
Sorry, yes I believe I am done.

ltratt:
Could we tweak the commit title to say something like "Add span to Yacc rule names" and maybe quickly justify why this is useful? Please force push an update.

ratmice:
Done

PR #293, opened 2022-05-15
Associate spans with production symbols

ratmice:
Here is the attempt at using `(Span, Symbol)` instead of adding a `span` field to symbol.
The main reason I was trying to add it to field, is it seems to require us to change the type of the [`prod`](https://docs.rs/cfgrammar/latest/cfgrammar/yacc/grammar/struct.YaccGrammar.html#method.prod) function.

As such there is quite a bit more fallout in callers, and i'm generally not terribly fond of all these calls like `grm.prod(pindex)[i].1`
But for now I just went with it, though perhaps a `struct SpannedSymbol` with `symbol` and `span` fields would be preferred with all this indexing into slices.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -379,7 +379,7 @@ impl YaccParser {
                 if self.ast.tokens.insert(sym.clone()) {
                     self.ast.spans.push(span);
                 }
-                syms.push(Symbol::Token(sym));
+                syms.push(Symbol::Token(sym, span));
             } else if let Some(j) = self.lookahead_is("%prec", i) {
                 i = self.parse_ws(j, true)?;
                 let (k, sym, _) = self.parse_token(i)?;
@@ -394,11 +394,11 @@ impl YaccParser {
                 i = j;
                 action = Some(a);
             } else {
-                let (j, sym, _) = self.parse_token(i)?;
+                let (j, sym, span) = self.parse_token(i)?;
                 if self.ast.tokens.contains(&sym) {
-                    syms.push(Symbol::Token(sym));
+                    syms.push(Symbol::Token(sym, span));
                 } else {
-                    syms.push(Symbol::Rule(sym));
+                    syms.push(Symbol::Rule(sym, span));
                 }
                 i = j;
             }
@@ -698,7 +698,7 @@ mod test {
             ast::{GrammarAST, Production, Symbol},
             AssocKind, Precedence, YaccKind, YaccOriginalActionKind,
         },
-        YaccParser, YaccParserError, YaccParserErrorKind,
+        Span, YaccParser, YaccParserError, YaccParserErrorKind,
     };
 
     fn parse(yacc_kind: YaccKind, s: &str) -> Result<GrammarAST, YaccParserError> {
@@ -708,23 +708,30 @@ mod test {
     }
 
     fn rule(n: &str) -> Symbol {
-        Symbol::Rule(n.to_string())
+        Symbol::Rule(n.to_string(), Span::new(0, 0))
+    }
+
+    fn rule_span(n: &str, span: Span) -> Symbol {
+        Symbol::Rule(n.to_string(), span)
     }
 
     fn token(n: &str) -> Symbol {
-        Symbol::Token(n.to_string())
+        Symbol::Token(n.to_string(), Span::new(0, 0))
+    }
+    fn token_span(n: &str, span: Span) -> Symbol {
+        Symbol::Token(n.to_string(), span)
     }
 
     #[test]
-    fn test_macro() {
-        assert_eq!(Symbol::Token("A".to_string()), token("A"));
+    fn test_helper_fn() {
+        assert_eq!(Symbol::Token("A".to_string(), Span::new(0, 0)), token("A"));
     }
 
     #[test]
     fn test_symbol_eq() {
         assert_eq!(rule("A"), rule("A"));
-        assert!(rule("A") != rule("B"));
-        assert!(rule("A") != token("A"));
+        assert_ne!(rule("A"), rule("B"));
+        assert_ne!(rule("A"), token("A"));
     }
 
     #[test]
@@ -740,14 +747,16 @@ mod test {
         )
         .unwrap();
         assert_eq!(grm.get_rule("A").unwrap().pidxs, vec![0]);
+        let a_span = Span::new(33, 34);
         assert_eq!(
             grm.prods[grm.get_rule("A").unwrap().pidxs[0]],
             Production {
-                symbols: vec![token("a")],
+                symbols: vec![token_span("a", a_span)],
                 precedence: None,
                 action: None
             }
         );
+        assert_eq!(&src[a_span.start()..a_span.end()], "a");
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


ratmice:
So I think this one is probably ready for review once you're done with the other PR, and have the time.

I'm just going to preemptively add some notes:
* I added `Hash` to `Span` because `Symbol` derives it.
   If you remove the derive from `Symbol` though cargo build still works works so nothing inside grmtools appears to be depending upon the `Hash` impl producing the same hash value for two different productions.
* The `Eq` impl is primarily used in tests where we are either producing a `Span` directly, or can derive it from a source file.
* I can't think of anything else span-related which *I* am going to be needing, we could do actions I don't really need them,
  but if you think we should add them I could do so.

ltratt:
Somehow I missed this -- sorry!

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -707,24 +707,29 @@ mod test {
         Ok(yp.ast())
     }
 
-    fn rule(n: &str) -> Symbol {
-        Symbol::Rule(n.to_string())
+    fn rule(n: &str, span: Span) -> Symbol {
```
ltratt, line 710: AFAICS we always call this with an empty span, so maybe remove the `span: Span` argument and document that the rule it creates has an empty span? It'll save space below!

ltratt:
> I added Hash to Span because Symbol derives it. If you remove the derive from Symbol though cargo build still works works so nothing inside grmtools appears to be depending upon the Hash impl producing the same hash value for two different productions.

Good spot! I'd be inclined to remove `Hash` from both, simply because I worry that people migrating older code might assume that two `Symbol`s from different parts of the input should compare/hash equal.

> The Eq impl is primarily used in tests where we are either producing a Span directly, or can derive it from a source file.

Is it only used in tests? Again, I have a gut feeling that it might be better to remove `Eq`/`PartialEq` for the same reasons as above.

Otherwise, this looks reasonable to me!

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -858,14 +873,18 @@ mod test {
             &src,
         )
         .unwrap();
+        let a_span = Span::new(8, 9);
+        let b_span = Span::new(11, 12);
         assert_eq!(
             grm.prods[grm.get_rule("A").unwrap().pidxs[0]],
             Production {
-                symbols: vec![token("a"), rule("B")],
+                symbols: vec![token("a", Span::new(8, 9)), rule("B", Span::new(11, 12))],
```
ratmice, line 881: The span argument to rule is I believe used here, where we are comparing equality with span's generated by the parser.

incidentally this should be `a_span`  and `b_span` instead of `Span::new(...)`

ltratt: If we're only using this in tests, we can make our own helper function `token_eq` which compares spans.

*Alternatively* (and I haven't thought this through!) we could define `Symbol` to be "an instance of a symbol at a given point in a file" at which point `Eq` & `Hash` do make sense.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Since it is used in it seems at least one place how about i add `fn rule_span(n, span)` and leave `rule` with an empty span?
These are just private functions in the test module IIRC

ltratt: Good idea!

ratmice:
> > The Eq impl is primarily used in tests where we are either producing a Span directly, or can derive it from a source file.
> 
> Is it only used in tests? Again, I have a gut feeling that it might be better to remove `Eq`/`PartialEq` for the same reasons as above.

It appears to be just used in tests afaict.

The only way we can remove `Eq` is to remove it from all of `Span`, `Symbol`, and `Production`, AFAICT it stops there.
So we'd have to provide equality stubs for all of those for testing, i believe.

One way we could maybe do this is using `#[cfg_attr(test, derive(Eq))]`?

ratmice:
Well, `#[cfg_attr(test, derive(Eq, PartialEq))]` almost works, I had to add a feature though for lrlex.
It only required it for `Span`, but it seemed prudent to me to add it for `Production` and `Rule` as well

ratmice:
So, having thought it through a bit I think my preference is definitely no on the feature, merely because it makes it harder to change the `Eq` behavior if we do decide to implement it one way or another.

I think `Eq` , `PartialEq` for `Span` is much less controversial than `Eq` for `Symbol`, or `Production` now that it contains `Span`s.
I had tried alternately implementing `span_eq`, in `lrlex` but then came across `lrpar/cttests` where it does a bunch of `Eq` over `Vec<Span>`.

I went ahead and pushed that over enabling it via feature, because it seems like the right thing to me.

ratmice:
So, I think this is ready for another round of review when you have the time, as nothing relevant has some to mind since my last push.  (And I believe the situation regarding `Eq` is okay in regards my previous span patches, since plain rules tokens are just bare Strings + Spans).

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -707,29 +707,31 @@ mod test {
...
-    fn token(n: &str, span: Span) -> Symbol {
+    fn token(n: &str) -> Symbol {
+        Symbol::Token(n.to_string(), Span::new(0, 0))
+    }
+    fn token_span(n: &str, span: Span) -> Symbol {
         Symbol::Token(n.to_string(), span)
     }
 
     #[test]
     fn test_macro() {
```
ltratt, line 726: This isn't really testing a macro, so maybe `test_helper_fn` or similar?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -707,29 +707,31 @@ mod test {
...
     }
 
     #[test]
     fn test_symbol_eq() {
-        let empty_span = Span::new(0, 0);
-        assert_eq!(rule("A", empty_span), rule("A", empty_span));
-        assert!(rule("A", empty_span) != rule("B", empty_span));
-        assert!(rule("A", empty_span) != token("A", empty_span));
+        assert_eq!(rule("A"), rule("A"));
+        assert!(rule("A") != rule("B"));
```
ltratt, line 733: `assert_ne`

ltratt:
https://github.com/softdevteam/grmtools/pull/293/commits/a1c7039d8a3828df5c35b51d3c5d8a29644fce64 is amusingly evil and a much smaller hack than the one I had suggested!

ratmice:
Good catch, I fixed those 2 things.

ltratt:
Please squash.

ratmice:
Done.

ratmice:
Thank you for the review, I do believe that this is probably the last major change this series on spans,

There may be some places where we are returning error `Results`, where it would be nice to also include spans.
But I need to go through and evaluate those, since they may have just picked up `Span` from `Symbol` now.

But that really shouldn't affect any major types used throughout the crate.
Let me know if you prefer this to be done e.g. before any form of release, 
or alternately if it is okay to do this on an ad-hoc basis as I run across errors that could be improved.

ratmice:
Okay, I will do a pass over that then.

PR #294, opened 2022-05-15
Turn StateTableError to an enum, handle self-loop for #290

### LINKED ISSUE #290: Apparently infinite recursive rule
One of the "fun" things about my project is running the parser on strange, half edited/incomplete changes.
Here is one such case I encountered that way, and have minimized.

given the input character `a`, this will cause an infinite loop pushing to `pstack` between 
https://github.com/softdevteam/grmtools/blob/master/lrpar/src/lib/parser.rs#L297
https://github.com/softdevteam/grmtools/blob/master/lrtable/src/lib/statetable.rs#L461-L466
adding a case like: `Some(i) if i == usize::from(stidx) + 1 => None,` to `goto` fixes it, (i.e. the return value of goto == `prior`).

Filing this as a bug report rather than sending a PR though, because I haven't yet tested it against valid parsers, or
as of yet tried to surmise if this case can _only_ and  _always_ lead to infinite recursion or if it ever actually comes up in a valid way.

```
%%
a "a"
[\t\n ] ;
```

```
%%
Start: Bar;
Foo: "a" | ;
Bar: Foo | Foo Bar;
```

ratmice:
Here is at least an attempt to fix #290,
It appeared to me that bison ignores this case entirely (no additional errors or self recursion, besides the conflicts, and if you run the parser it won't loop forever), and I believe I see in the bison code where it ignores it.

It's worth pointing out the Accept/Reduce errors here with it's "Infinitely recursive rule let through",
I haven't yet gone and tried to fully understand what makes this case different from an Accept/Reduce.

I had to switch the error around here, because unless I am mistaken, at that point we don't appear to have any way to get a `PIdx`.

### DIFF
--- a/lrtable/src/lib/mod.rs
+++ b/lrtable/src/lib/mod.rs
@@ -16,7 +16,7 @@ pub mod statetable;
 
 pub use crate::{
     stategraph::StateGraph,
-    statetable::{Action, StateTable, StateTableError, StateTableErrorKind},
+    statetable::{Action, StateTable, StateTableError},
 };
 use cfgrammar::yacc::YaccGrammar;
 
--- a/lrtable/src/lib/statetable.rs
+++ b/lrtable/src/lib/statetable.rs
@@ -94,25 +94,20 @@ where
     }
 }
 
-/// The various different possible Yacc parser errors.
-#[derive(Debug)]
-pub enum StateTableErrorKind {
-    AcceptReduceConflict,
-}
-
 /// Any error from the Yacc parser returns an instance of this struct.
 #[derive(Debug)]
-pub struct StateTableError<StorageT> {
-    pub kind: StateTableErrorKind,
-    pub pidx: PIdx<StorageT>,
+pub enum StateTableError<StorageT> {
+    AcceptReduceConflict(PIdx<StorageT>),
+    StateGraphSelfLoop(RIdx<StorageT>),
 }
 
 impl<StorageT: Debug> Error for StateTableError<StorageT> {}
 
 impl<StorageT> fmt::Display for StateTableError<StorageT> {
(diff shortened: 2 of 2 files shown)


ltratt:
Interestingly if I take your example:

```
%%
Start: Bar;
Foo: "a" | ;
Bar: Foo | Foo Bar;
```

and run it on (traditional-ish) yacc I get:

```
yacc: 1 rule never reduced
tmp.y: yacc finds 2 shift/reduce conflicts
tmp.y: yacc finds 1 reduce/reduce conflict
```

but on Bison:

```
tmp.y: warning: 2 shift/reduce conflicts [-Wconflicts-sr]
tmp.y: warning: 1 reduce/reduce conflict [-Wconflicts-rr]
tmp.y: note: rerun with option '-Wcounterexamples' to generate conflict counterexamples
tmp.y:4.6-8: warning: rule useless in parser due to conflicts [-Wother]
    4 | Bar: Foo | Foo Bar;
      |    
```

In both cases we have a warning, not an error, but the yacc warning is IMHO clearer.

So we probably have two questions to think about:

1. Should this be an error or a warning?
2. What should we call it / pretty-print to the user etc?

ratmice:
I think if we want to turn it into a warning we need to do something about the infinitely growing action vector when it hits that state.

(Although I guess my program should probably refuse to try and parse files, unless the number of conflicts to ignore is set appropriately).

I'm not exactly sure what the right way to go about I would assume replacing it with State `0`?

ltratt:
We should probably do whatever yacc/bison does, at least if we're in `YaccKind::Original` mode. In `YaccKind::Grmtools` I think I'd be fine with throwing an error.

ltratt:
Does that mean we should close this PR or ... ?

ltratt:
OK, agreed, we might as well close this -- it'll still be here.

PR #295, opened 2022-05-16
Move all references to lrpar::Span to cfgrammar::Span.

ltratt:
This also allows us to mark lrpar::Span as deprecated, so people get a (hopefully) gentle upgrade path.

@ratmice Could you review this please? You'll have rights with bors to merge it if/when you think it's ready.

bors delegate=ratmice


ratmice:
seems straight-forward and lgtm.

bors r+

Issue #296, opened 2022-05-23
token_span sometimes returns the wrong span


ratmice:
I haven't gotten to the bottom of this yet, but `token_span` added in my first series of patches,
appears as though it can sometimes return the wrong span.

I also think that perhaps with #293 token_span may actually be superfluous,
given that it should return the first production rule where it occurs, between rule_span's and symbol spans.
I think you could probably replace it with a running `find` over symbol spans.

I will get to the bottom of what is wrong with it regardless of whether we think it should be kept.
Apologies, I kind of thought approaching this in stages would help but it really hasn't seemed to.

ltratt:
Bugs happen -- thanks for being honest enough to point this out and try finding a fix for it!

ratmice:
FWIW, I believe I see now why symbol spans may not be an adequate replacement for token/token name, 
In the absence of a `%token X` directive a `token_span` will correspond to one from a symbol.

But when there are `%token` directives it should match one of those, so perhaps it isn't so superfluous.

PR #297, opened 2022-05-23
fix for issue #296

ratmice:
This adds a minimized testcase for issue #296, but I still need to look into the cause.


ltratt:
I think you're saying this should bork as-is? No need to verify as bors will tell me :)

bors try

ratmice:
Working on a fix, it happens when we insert a duplicate into the indexset, but then add the span to a vec regardless.
Then the indexes into the vec are off by one for each dupe.

ratmice:
FYI, this one should be ready, just force pushing to keep in sync with master.

ltratt:
Please squash (it would be nice to have the test and fix in one commit IMHO).

PR #299, opened 2022-06-02
Add a span field to LexBuildError

ratmice:
This currently uses the offset given to build an empty span at `(off, off)`.
I've commented in the tests, some spans that I *believe* might be the right thing, but without really
digging into the parser code for each case and checking out the results, I'm not certain of the accuracy of these comments.

I mainly wanted to push this before starting any work on that, because that will probably take a bit of work,
increase the size of the patch, and that work should be isolated private API.

My feeling is the next step would be to remove the `line`/`col` fields from the structure and the error text, moving that to callers using the span, in this patch we still need those, because by the time we format the message we no longer have the text to count newlines.  Does that sound reasonable?


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1332,36 +1285,37 @@ x"
...
             Ok(_) => panic!("Incorrect %prec parsed"),
             Err(YaccParserError {
                 kind: YaccParserErrorKind::IllegalString,
-                line: 3,
+                span,
                 ..
-            }) => (),
+            }) => assert_eq!(line_of_offset(src, span.start()), 3),
             Err(e) => panic!("Incorrect error returned {}", e),
         }
```
ratmice, line 1303: It might be a good idea to macroize these blocks so we can do the `let src` inside the macro,
since there are a few blocks with multiple `let src` shadowing.  Once we bind it like this there is potential to lookup newlines in the wrong `src` variable.

ratmice:
One thing we should consider since it impacts public types if if we want to add a second span to `cfgrammar::yacc::parser::YaccParserErrorKind::Duplicate*`  and `lrlex::LexErrorKind::Duplicate*`.

So we could point not just to the duplicate occurrence but the original as well, (perhaps in a follow up patch?)

lrpar/src/lib/ctbuilder.rs
```diff
@@ -357,7 +357,40 @@ where
...
+            (line_m1 + 1, c_off + 1)
+        }
+
+        let grm = YaccGrammar::<StorageT>::new_with_storaget(yk, &inc).map_err(|e| match e {
+            YaccGrammarError::YaccParserError(e) => {
+                let (column, line) = off_to_line_col(&inc, e.span.start());
+                format!("{} at column {column} line {line}", e)
+            }
+            e => e.to_string(),
+        })?;
```
ratmice, line 393: This is a bit hacky going from `YaccGrammarError -> String -> Box<dyn Error>`, but I'm not so sure what we could actually do otherwise

ratmice:
@ltratt So, I think this at least covers the user-facing components (Though i'll keep poking around to make sure I haven't missed anything). This will have inferior test suite output at the many `YaccGrammar::new(...).unwrap()` though (at least *this* patch doesn't try to add any niceties there).

So it's probably a good point where we can decide if this seems like a reasonable approach, as I think it is relatively complete/hits  the points of interest.

lrlex/src/lib/mod.rs
```diff
@@ -29,12 +29,15 @@ pub use crate::{
...
 
+use cfgrammar::Span;
+
 pub type LexBuildResult<T> = Result<T, LexBuildError>;
 
 /// Any error from the Lex parser returns an instance of this struct.
 #[derive(Debug)]
 pub struct LexBuildError {
     pub kind: LexErrorKind,
+    pub span: Span,
```
ltratt, line 40: This is interesting because we now have two ways of locating the error: `span` and `line`/`col`. The challenge that brings is that in order to make sense of `span` you need to have the input available, which perhaps suggests that `LexBuildError` should have a reference to the source input. Ironically that's quite easy because at the moment you need to move ownership of the source to the parser:

```rust
impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
    pub(super) fn new(src: String) -> LexBuildResult<LexParser<StorageT>> {
        ...
    }
}
```

In retrospect that should probably have been `src: &str`, but that's not directly important here.

So one possibility is that `LexBuildError` should be:

```rust
pub struct LexBuildError {
    pub kind: ...,
    span: Span,
    src: String
}
```

since it can then rederive `line`/`col` automatically. *But* that might not make sense in the context of LPS sorts of things... Any thoughts?

lrpar/src/lib/parser.rs
```diff
@@ -839,7 +839,7 @@ pub enum ParseRepair<LexemeT: Lexeme<StorageT>, StorageT: Hash> {
 #[derive(Clone, Debug, PartialEq)]
 pub struct ParseError<LexemeT: Lexeme<StorageT>, StorageT: Hash> {
     stidx: StIdx,
-    lexeme: LexemeT,
+    lexeme: LexemeT, // Can get the span through lexeme.
```
ltratt, line 842: I agree with the comment, but we shouldn't leave it in. [Same with the same comment on `statetable.rs`.]

lrlex/src/lib/parser.rs
```diff
@@ -297,7 +320,13 @@ mod test {
                 kind: LexErrorKind::DuplicateName,
                 line: 3,
                 col: 7,
-            }) => (),
+                span,
+            }) => {
+                // I would suggest that this should actually span around int
```
ltratt, line 287: Is this comment saying "the error message isn't great"? If so, we should file this as an issue and fix it later rather than leave a comment.

lrlex/src/lib/parser.rs
```diff
@@ -244,9 +225,11 @@ mod test {
             Ok(_) => panic!("Broken rule parsed"),
             Err(LexBuildError {
                 kind: LexErrorKind::MissingSpace,
-                line: 2,
-                col: 1,
-            }) => (),
+                span,
```
ltratt, line 228: Might this be nicer as `span if span == Span::new(3, 3)`? Maybe not!

lrlex/src/main.rs
```diff
@@ -36,6 +36,32 @@ fn read_file(path: &str) -> String {
     s
 }
 
+fn off_to_line_col(src: &str, off: usize) -> (usize, usize) {
```
ltratt, line 39: The repetition of this function throughout the codebase isn't great. We should work out a way of having one copy only, if at all possible.

One possibility might be to have (probably) three methods on `Span` something like:

```
impl  Span {
  fn start_to_line_col(src: &str) -> (usize, usize) { ... }
  fn end_to_line_col(src: &str) -> (usize, usize) { ... }
  fn start_end_to_line_col(src: &str) -> ((usize, usize), (usize, usize)) { ... }
}
```

The third of those isn't exactly necessary, but would be an optimisation if you need to know the start and end of a `Span`.

lrlex/src/lib/mod.rs
ltratt: Clearly not only did you have thoughts, but your later commits address them -- mea culpa! Still, it might be worth at least briefly thinking about whether we should carry around `src` or not. [I'm inclined to think "don't carry it around" but I haven't thought things through fully.]

lrlex/src/lib/parser.rs
ratmice: Yeah, more-or-less, I'd go with "this error message could be improved", but I will pull these comments out, and file an issue.

lrlex/src/main.rs
ratmice: Indeed, I guess what bugged me about making a public API for this, is that if you start dealing with multiple spans, this API will have to rebuild the newlines offset cache vector for each span,

I'd thought about wrapping the `src` argument in some type `Source`, and putting the newline cache, and these functions there, but it *does* add quite a bit of API surface to do that and further there are quite a few crates out there like `codespan`, and `miette` for producing really nice diagnostics, I'm guessing these have their own newline cache built in.

So, I was kind of just trying to not regress the output, but I wholeheartedly agree that this repetition is terrible.

ratmice: Would you be okay with some kind of documentation notice that the above functions are `helper functions but inefficient for repeated calls`, with pointers to these other crates as alternatives?  I think that would alleviate my concerns and strike a balance of convenience, without having to introduce an elaborate API surface.

ltratt: > Indeed, I guess what bugged me about making a public API for this, is that if you start dealing with multiple spans, this API will have to rebuild the newlines offset cache vector for each span,

That's a very good point! That rules out this idea.

Maybe what we need is some sort of `NewlineToLineColCache` thing that can encapsulate newlines for a given source input and translate spans into (line, col)s?

ratmice: I'll give it a go, this was exactly what I was thinking with the mention of a `Source` type above, it is at least worth trying!

ltratt: Cool! I guess its most natural home is `cfgrammar`.

lrlex/src/lib/mod.rs
ratmice: Yeah, I hadn't occurred to me to try and make it owned, so I haven't considered it to this point enough to have really formed an opinion.  :thinking:

lrlex/src/main.rs
ratmice: I did this, and noticed a couple issues while doing so (flipped column, line -> line/column ), and a mixing of `lex_src` where it should be `yacc_src`.

So I think this is done.

cfgrammar/src/lib/span.rs
```diff
@@ -43,3 +43,56 @@ impl Span {
         self.len() == 0
     }
 }
+
+pub struct NewlineToLineColCache<'a> {
```
ratmice, line 47: Actually, did you want this pub or not?

lrlex/src/lib/parser.rs
ratmice: Hmm, I could go either way on this, I had done pattern guards a few places, but I ran into issues where the test was failing, and the `assert_eq!` produced output about the failure, where a pattern guard just wouldn't match, so I'd moved them all to assertions.

So i'm kind of hesitant, let me know if you still want me to try it out that way though.

lrlex/src/lib/mod.rs
ratmice: So thought about this, and my general thinking is that as you said it really doesn't make much sense in LSP sorts of contexts. There you'd have to keep a clone around for your own use, and then rely on the fact that spans between the two clones still correlate because they don't mutate the owned value.

One thing i'm wondering about `&str` wise is if it would be good to use `AsRef<str>`, (not sure it belongs in this, perhaps another time/patch?)

ltratt: I agree: let's keep `kind` and `span` and not add `src`.

lrlex/src/lib/parser.rs
ratmice: So, I have come around to your suggestion with the addition that we format the error like it used to be,
and working on an implementation of that..

ratmice: Alright, pushed that and feel that it is an improvement, but feel free to unresolve or change your mind if you feel otherwise

cfgrammar/src/lib/yacc/parser.rs
ratmice: I think there is probably too much magic to stuff in a macro, it may shorten the amount of repetition it but it's probably clearer without, so going to just resolve this.

ratmice:
So, I believe that i've resolved all the things from your previous review now.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1008,8 +1019,8 @@ mod test {
             Err(YaccParserError {
                 kind: YaccParserErrorKind::IllegalString,
                 span,
-            }) => assert_eq!(span, Span::new(28, 28)),
-            Err(e) => panic!("Incorrect error returned {}", e),
+            }) if span == Span::new(28, 28) => (),
```
ratmice, line 1022: Now that we have all the line/column helpers, instead of hard coding these spans we should probably check against those?

cfgrammar/src/lib/span.rs
ltratt, line 47: Hmm, here's something interesting. When I first (naively) thought about this, I assumed that the cache wouldn't need a copy of the source. That's wrong, of course, because `offset_to_line_col` needs that to calculate column numbers.

But now I'm wondering if/how this could support incrementally built-up newline caches (e.g. for a streaming parser). Any thoughts?

ratmice: I had considered it but (not in the streaming context though, i'm just not sure what type the streaming context would/should give for src exactly),

The way I as considering was moving the building of `newlines` into `offset_to_line_col` and checking the `off` parameter against `newlines.last()` in an a branch (after the `self.src.len()` check but before the line/col calculation. then using `extend` on `newlines`, using `self.src[last..off].char_indices.map(|cidx| cidx.o + last)`.

or something to that effect, the danger here is that off needs to be on an `str::is_char_boundary` for that slice to work I believe.  I don't see anywhere that the current mechanism relies on that, but it seems like it might be a reasonable requirement.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Finally had some time to figure out the bug that i ran into implementing this...

cfgrammar/src/lib/span.rs
```diff
@@ -51,13 +51,15 @@ pub struct NewlineToLineColCache<'a> {
...
 impl<'a> NewlineToLineColCache<'a> {
     pub fn new(src: &'a str) -> Self {
-        let newlines: Vec<usize> = src
-            .char_indices()
-            .filter_map(|c| match c {
-                (offset, '\n') | (offset, '\r') => Some(offset),
-                _ => None,
-            })
-            .collect::<Vec<_>>();
+        let mut newlines: Vec<usize> = vec![0];
```
ratmice, line 54: Not really a good place to add this note, but it occurred to me I need to check the old `newlines` vec it seems likely we can now remove them if the only uses are `push`

ratmice: One was unused, and the other could be replaced by a count of newlines used in some whitespace sensitive parsing.

cfgrammar/src/lib/span.rs
ratmice: Well, i've been thinking this morning about this and noticed that there is one issue with the sort of lazy building of the newlines vector, specifically I think in cases where a file ends without a newline, it'll try and lazily find more newlines for any offset on the last line.

This would seem like it could be an issue for code like minified javascript and json in such cases where you have excessively long lines.

In the non-streaming case we could just add an index into newlines at EOF (which besides variable name should be harmless)... That kind of seems like a poor approximation for just keeping a specific `usize` around for the furthest byte we've looked for a newline at.

ltratt:
Sorry for the delay! I started chewing on this and then hit a mental dead end, but now I think I've got a handle on things.

Summary: https://github.com/softdevteam/grmtools/pull/299/commits/e7da5fdb56fe79807d18d0eea93573774365765e is quite cool, but I thought we wanted incremental building of `src` (incremental building of `newlines` is a pleasant optimisation, but not fundamental). But that isn't going to work with lifetimes in a sensible way and we don't want to cache some sort of `Vec<String>` as that will probably duplicate what the user already has.

What I've now realised is that we need to get rid of `src` in the newlines cache entirely. I think the minimal API should be, roughly, something like:

```rust
struct NewLinesCache {
    newlines: Vec<usize>
}

impl NewLinesCache {
    /// Feed more input into the cache, calculating newlines data from it.
    fn feed(&mut self, s: &str) { ... }
    /// Convert a byte offset in the input to a logical line number. Returns None if
    /// the byte offset exceeds the known input length.
    fn byte_to_line_num(&self, byte: usize) -> Option<usize> { ... }
}
```

Importantly we don't store `src`: the user just keeps pushing strings into the cache, we keep adding newline data, but don't store the user's string.

The obvious downside is that because we don't have `src` we can no longer calculate column offsets. We can offer a function that makes it easier for the user to do so:

```rust
impl NewLinesCache {
    ... // As above

    /// Convert a byte offset in the input to the byte offset of the beginning of its line.
    /// Returns None if the byte offset exceeds the known input length.
    fn byte_to_line_byte(&self, byte: usize) -> Option<usize> { ... }
}
```

Assuming the user has their own cache of the input, they can then use the byte of the beginning of the line to easily calculate the column offset.

Of course we could add a convenience function along the lines of to do this on their behalf:

```rust
impl NewLinesCache {
    ... // As above

    /// Convert a byte offset in the input to logical line and column numbers.
    /// Returns None if the byte offset exceeds the known input length.
    fn byte_to_line_and_col(&self, s: &str, byte: usize) -> Option<(usize, usize)> { ... }
}
```

I am now very aware that we are dragging this PR ever further away from its original intent. Do you think it's perhaps now sensible to strip this PR back so it just does the original "add a span field" and then do a second PR with a newline cache along the lines of the above?

ratmice:
Don't worry about the delay, sometimes things just need to simmer for a bit.
One thing I definitely notice here is the similarity to the API for [`ropey`](https://docs.rs/ropey) which I really like. 

In ropey `byte_to_line_byte` is achieved by the composition of 2 separate functions `line_to_byte` and `byte_to_line`, I bring it up because `byte_to_line_num` is modulo panics vs option similar to `byte_to_line`, and we could also easily implement `line_num_to_byte` which is a bit more primitive than the functions proposed above.

I just mention it because I've found the `line_to_byte` naming to be quite self evident, where `byte_to_line_byte` requires a bit more thought.  Perhaps it would be more efficient than `byte_to_line_num` and `byte_to_line_byte` since with `line_to_byte` we can just do random access into the newlines `Vec`, where `byte_to_line_byte` would require a second search of the `newlines` vector?

My thinking is with these changes we are still within the realm of reasonable, while 'add a span field' isn't a regression, it really feels incomplete. 
And I think that these changes we're discussing are still pretty localized, and won't impact anything above beyond what this series is already changing.  So I kind of lean towards reaching parity with the existing code and once squashed it seems like it should still be on the cusp of being a reasonable patch...

ratmice:
So I did a first pass at this, perhaps I misunderstood something, I had to make 2 minor changes (both are probably untested and need some test cases).

* I needed a way to keep track of the number of bytes trailing a fed string after the last newline, this is `trailing_bytes`.
* I wasn't certain how to deal with the `src` parameter to `bytes_to_line_and_col`, whether this `src` parameter is to be the cumulative string of all values passed via `feed` concatenated together
  or instead `src` being one of the parameters which was passed to feed in which case, I needed a byte position reflecting the start of `src`.

I had gone with the latter meaning of src, did you mean the first?
I haven't yet tested this outside of the single string usage where `src_pos` is 0 though,

perhaps some of this additional `Option` handling in callers could be turned into unwraping particularly in the tests.

cfgrammar/src/lib/span.rs
```diff
@@ -44,72 +44,98 @@ impl Span {
...
+impl NewlineToLineColCache {
+    /// Feed more input into the cache, calculating newlines data from it.
+    /// The `src` string given is treated as though it is contiguous with previous a call to `feed`.
+    /// subsequent logical lines, and logical bytes reflecting the sum of total of offsets which
+    /// have been fed.
+    pub fn feed(&mut self, src: &str) {
+        let start_pos = self.newlines[self.newlines.len() - 1] + self.trailing_bytes;
+        self.newlines
+            .extend(src.char_indices().filter_map(|c| match c {
+                (offset, '\n') | (offset, '\r') => {
```
ltratt, line 70: Will this count a `\n\r` newline (the or `\r\n` or whatever the old DOS standard is) as two newlines?

cfgrammar/src/lib/span.rs
```diff
@@ -44,72 +44,98 @@ impl Span {
     }
 }
 
-pub struct NewlineToLineColCache<'a> {
-    src: &'a str,
-    high_water: usize,
+pub struct NewlineToLineColCache {
     newlines: Vec<usize>,
+    trailing_bytes: usize,
```
ltratt, line 49: That's very clever. I hadn't even thought of this!

cfgrammar/src/lib/span.rs
```diff
@@ -44,72 +44,98 @@ impl Span {
...
-
-            self.high_water = off;
-        }
+impl NewlineToLineColCache {
+    /// Feed more input into the cache, calculating newlines data from it.
+    /// The `src` string given is treated as though it is contiguous with previous a call to `feed`.
+    /// subsequent logical lines, and logical bytes reflecting the sum of total of offsets which
+    /// have been fed.
+    pub fn feed(&mut self, src: &str) {
+        let start_pos = self.newlines[self.newlines.len() - 1] + self.trailing_bytes;
```
ltratt, line 67: Later you use `self.newlines.last().unwrap()` which is probably clearer.

cfgrammar/src/lib/span.rs
```diff
@@ -44,72 +44,98 @@ impl Span {
...
-        let c_off = self.src[line_off..]
-            .char_indices()
-            .position(|(c_off, _)| c_off == off - line_off)
-            .unwrap();
-        (line_m1 + 1, c_off + 1)
     }
 
-    pub fn start_to_line_col(&mut self, span: Span) -> (usize, usize) {
-        self.offset_to_line_col(span.start())
+    /// Convert a logical line number into a logical byte offset. Returns None
```
ltratt, line 100: I'm not sure what a "logical byte" is.

cfgrammar/src/lib/span.rs
ratmice: It'll count it as two newlines, I actually wondered about that, and had started working on code to `peek`, but then when i looked it seemed to me that the existing code was doing this treating them separately, thanks for bringing it up I kind of forgot to.

cfgrammar/src/lib/span.rs
ratmice: Yeah, i wasn't certain if I should introduce that or not, but I meant it as indicating that it was a byte relative to the concatenation of all strings fed.  I could try and improve that or remove it wtdyt?

ltratt: I think "logical" is misleading here, because it's just the "byte offset since the start of the input". The reason we have "logical lines" is because they don't match up 1:1 to bytes.

cfgrammar/src/lib/span.rs
ratmice: Sounds good, I thought I had encountered something, but I don't see it anymore/in the tests.

cfgrammar/src/lib/span.rs
```diff
@@ -103,7 +103,7 @@ impl NewlineToLineColCache {
         if line_num > self.newlines.len() {
             None
         } else {
-            Some(self.newlines[line_num - 1])
+            Some(self.newlines[line_num.saturating_sub(1)])
```
ratmice, line 106: Not sure if we need this `saturating_sub`, since this fn isn't `pub`, so we don't need to worry about being called with `0`.  I suppose this is a good reason to keep it private.

cfgrammar/src/lib/span.rs
```diff
@@ -126,12 +126,21 @@ impl NewlineToLineColCache {
     ) -> Option<(usize, usize)> {
         let line_num = self.byte_to_line_num(byte);
         line_num.map(|line_num| {
-            if src_pos + byte == src.len() {
-                let line_off = self.newlines.last().unwrap();
-                return (self.newlines.len(), src[*line_off..].chars().count() + 1);
+            let line_byte = self.line_num_to_byte(line_num);
```
ratmice, line 129: It strikes me that this function isn't taking `trailing_bytes` into account, and I feel like probably needs to, i'll have to think it over

cfgrammar/src/lib/span.rs
ratmice: It seemed okay to conservatively special case this to return None, but still zero trailing bytes, even though it'll probably be zero'd by a subsequent `\n` on the next character.

ratmice: Going to resolve with that, but feel free to unresolve if you feel otherwise of course!

cfgrammar/src/lib/span.rs
ratmice: Hopefully not too clever, there may be some more places we need to use it beyond where we currently do, I'll try and think about this tonight, and come up with some tests which deal with multiple strings, including needing the trailing_bytes.

cfgrammar/src/lib/span.rs
```diff
@@ -60,25 +60,29 @@ impl Default for NewlineToLineColCache {
...
-        let start_pos = self.newlines[self.newlines.len() - 1] + self.trailing_bytes;
+        let start_pos = self.newlines.last().unwrap() + self.trailing_bytes;
         self.newlines
             .extend(src.char_indices().filter_map(|c| match c {
-                (offset, '\n') | (offset, '\r') => {
+                (offset, '\n') => {
                     self.trailing_bytes = 0;
                     Some(start_pos + offset + 1)
                 }
+                (_, '\r') => {
```
ltratt, line 74: Without multiple tests to tell what this actually does, I'm not super happy with this: it looks like it's going to have some weird consequences. But I might be wrong! So let's add tests please :)

cfgrammar/src/lib/span.rs
ltratt: I think being driven by tests would be a good thing here!

cfgrammar/src/lib/span.rs
ratmice: Sorry about that thanks for bringing it to my attention, I'll just say when something 'I believe this should be resolved by commit id', in the future and let you resolve them if you agree.

cfgrammar/src/lib/span.rs
ratmice: Yeah, you are correct here this is going to mess up byte offsets, and if we aren't dealing with `\r` by itself the corner case I was imagining won't actually be a case were going to handle.

The thing I was worried about is '\r' contributing to the column calculation, in particular if we have input like:
`cache.feed("aaaa\nbbbb"); cache.feed("mnop"); cache.byte_to_line_and_column(.., "mnop", ...);`
we need to (but aren't currently) dealing with the need for the characters "bbbb" to contribute to the column count.

to deal with this case we need *both* a `trailing_bytes` and `trailing_chars`, and `\r` *should* contribute to `trailing_bytes` so as to avoid messing up byte offsets.  But it seems like it should *not* contribute to `trailing_chars`. If we were dealing with `\r` by itself and `\r\n`, this would seem to be much more problematic.  As then we get cases like `cache.feed("\r"); cache.feed("\n");`.

At least I feel like I understand the case that lead me to make this change now, so i'll get to work on codifying them as tests...

ratmice: So it's actually worse than needing `trailing_chars` I believe the tests I've added need a `Vec<usize>` i.e. a vector of trailing chars one for each call to `feed`, or perhaps a `HashMap<usize, usize>` to map `src_pos` to `trailing_chars`.

This is seeming to get a bit hairy so i am going to wait to proceed in that until you have a chance to weigh in.

ratmice: Well I just went ahead and implemented the above in 0f8f14f so we could at least look at the damage before deciding...

cfgrammar/src/lib/span.rs
```diff
@@ -46,14 +48,20 @@ impl Span {
...
+    trailing: HashMap<usize, (usize, usize)>,
 }
 
 impl Default for NewlineToLineColCache {
     fn default() -> NewlineToLineColCache {
         NewlineToLineColCache {
             newlines: vec![0],
             trailing_bytes: 0,
+            trailing_chars: 0,
+            trailing: HashMap::new(),
```
ltratt, line 63: What are `trailing_chars` and (in particular) `trailing`?

ratmice: `trailing_chars` and `trailing_bytes` are accumulators, where `trailing` is a "snapshot" of `(trailing_bytes, trailing_chars)` at a point where feed was called, the key of trailing is the `src_pos` of the *next* call to feed beyond the current one.

`trailing_chars`, is just the number of characters trailing the previous newline, we can't calculate it from bytes because of unicode characters, the copy in trailing is used to calculate the column in cases where the newline is in a previous string than the one `byte_to_line_and_col` is passed as `src`. The copy of `trailing_bytes` in `trailing` is used similarly to slice into the `src` string passed into `bytes_to_line_and_col`.  We need a snapshot of each `trailing_bytes` and `trailing_chars` for each potential call to `bytes_to_line_and_col`

for example:
`cache.feed("\nabc"); cache.feed("mnop"); cache.feed("xyz");`

`trailing` should (I believe be) `{ 4 = (3,3), 8 = (7,7)}`, We could add the first `0 = (0, 0)`, but it is unnecessary the previous has never has trailing bytes or chars.
With 4 being the `src_pos` for `"mnop"` and 8 for `"xyz"`, we can then calculate that `'n'` of `"mnop"` is at the 5th column, even though `"abc"` was never passed to `bytes_to_line_and_col`.

In these ASCII cases they'll be equal, but should differ for unicode (another test).

ltratt: Is the idea here that we want to cope with strings that end in non-valid-unicode characters?

ratmice: No, just valid multibyte where each one which only add `1` to the column offset, `trailing_bytes` part of this is an artifact of the way we know we are slicing at a unicode `char_boundary` by doing so after the newline we know it is a valid `char_boundary`.  

Your questions trigger me to think that we can perhaps get rid of this usage of `trailing_bytes` though.
`if line_off < src_pos` my feeling is I'm probably doing a bunch of extra work to calculate `src[0..]`, aka `src` given a `line_off` by subtracting out stuff...

cfgrammar/src/lib/span.rs
```diff
@@ -152,3 +152,37 @@ impl NewlineToLineColCache {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::NewlineToLineColCache;
+    use std::ops::Range;
+    #[test]
+    fn newline_cache_splits() {
+        let same_line_col_1 = [
```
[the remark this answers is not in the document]
ratmice: I'll have to digest this in the morning, sorry for adding more copy/pastes of the above test you quote, but I wanted to fix/check that case I just added so I could sleep :rofl:

cfgrammar/src/lib/span.rs
ratmice: Indeed, now that I have that one major edge case under wraps, I will add unicode tests, test the `None` cases, and some more general testing, and stuff at `EOF`.  But I don't know of any more major edge cases lurking, but no proof they don't exist...

cfgrammar/src/lib/span.rs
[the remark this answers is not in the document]
ratmice: I've added `newline_test_helper`, I went with the `&[Vec<(usize, usize)>]`, but I didn't have a strong preference.
It shows the difference between these `["a", "b"]` cases, but maybe flattening these is preferred, we wouldn't have to repeat so much, but it wouldn't show the difference anymore :shrug:.

I also called `feed` with all the arguments first, because I think it will catch it would catch cases where we could `trailing_bytes` when it is out of date because another string has been fed, rather than doing it in the single loop.

Anyhow just added a couple of cases, until we've decided

cfgrammar/src/lib/span.rs
```diff
@@ -237,35 +234,61 @@ mod tests {
...
+        newline_test_helper(&["ab", "c"], abc);
+
+        let ab = &[(1, 1), (1, 2), (2, 1)];
+        newline_test_helper(&["a\n", "b"], ab);
+        newline_test_helper(&["a", "\nb"], ab);
+
+        let cr = &[(1, 1), (1, 2), (1, 3), (2, 1)];
+        newline_test_helper(&["a\r\n", "b"], cr);
+        newline_test_helper(&["a", "\r\nb"], cr);
+        newline_test_helper(&["a\r", "\nb"], cr);
```
ratmice, line 269: So, the lack of incrementing trailing chars removed in this patch is because,
I kind of feel like this should test should be for `&[(1, 1), (1, 2), (1, 2), (2, 1)]` rather than `... (1, 2), (1, 3), ..`

But doing it uniformly across split strings is a bit of work, but we should decide what we want the output to be here.
With special casing `\r` in feed, it was giving `&[.., (1, 2), (1, 2), ...]` for the 3rd `cr` case, but `(1, 2), (1, 3)` for the first 2.

Any thoughts?

ltratt: Personally I'm fine with just ignoring `\r` when calculating newlines and treating it as a zero width character for column positions. It's simple and has the same effect (AFAICT) as "complicated" solutions.

ltratt:
I'm still really unsure what the `trailing` hashmap is doing or why it's needed. That might be me being stupid, but it doesn't *feel* like it should be necessary.

cfgrammar/src/lib/span.rs
ratmice: Alright, that is going to take some work though.

ratmice:
I believe it is necessary, I believe that It should probably be renamed  `preceding`, or `precedeing_bytes_and_characters`, when 2 strings are fed across a single line, when we go to calculate columns, we only have the offset of the line we're on.
When that line starts on a previous string, there is a bunch of bytes and characters potentially after the newline.

At the point of column calculation we don't have that string containing '\n', so for column calculation we need to know how many bytes the string of characters after the newline there are, and how many characters are in it.

I agree it doesn't feel necessary, it took me a while to come up with something that works though, and I certainly don't think it is obvious.  But I currently don't see any other way around it.

cfgrammar/src/lib/span.rs
ratmice: Fixed by [a243de6](https://github.com/softdevteam/grmtools/pull/299/commits/a243de6db79874edba92b68d444e715cf2ffdafe)

ltratt:
I understand that we might need to keep some of the previous count (though do we need to keep *character* counts? I think we just need byte counts) but why a `HashMap`?

cfgrammar/src/lib/span.rs
```diff
@@ -232,4 +236,44 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn all_one_line() {
+        newline_test_helper(
+            &["a", "b", "c"],
+            &[vec![(1, 1)], vec![(1, 2)], vec![(1, 3)]],
+        );
+        newline_test_helper(&["abc"], &[vec![(1, 1), (1, 2), (1, 3)]]);
```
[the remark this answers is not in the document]
ratmice: I did the first part in [f995d7f](https://github.com/softdevteam/grmtools/pull/299/commits/f995d7f87b9f6d106c858dd211f168e870200347) still thinking through the latter "negative" tests part.

ratmice:
I think we need both, for the unicode tests, because a multibyte character only contributes a single character to the column count,
and we can't recover the number of characters from bytes nor the other way around.

`HashMap` is definitely debatable, and picked by default, I was thinking `BTreeMap` would be better.
we Could perhaps use a `Vec<(usize, usize)>`, we'd need to return some kind of `SrcIdx` from `feed`, to be passed back in to `bytes_to_line_and_col`. I believe that would work.

ratmice:
I wonder if where we are differing in our thinking due to sequential vs non-sequential,
I'd agree that if we have

```
cache.feed(x);
cache.line_num(.., x, ...);
```

We wouldn't need the `trailing`, if it was only called sequentially like this, but we usually don't get spans until multiple strings will have been fed.  So we need to the state at a previous call to feed.

once we have
```
cache.feed(x);
cache.feed(y);
cache.feed(z);
cache.line_num(..., y, ...); 
```
We need to look back at the preceeding bytes/characters when *x* was fed.
anyhow it would help if I could pinpoint we seem to be out of sync.

ltratt:
Nope, I still don't get it! I think you should be able to call any interleaving you want of `feed` and `line_num` and I don't really see why keeping previous state is necessary. Can you maybe spell out (bearing in mind that I'm an idiot!) what happens with/without the hashmap?

ratmice:
Sure, 
lets say
```
let x = "a";
let y = "b";
let z = "c";
// after this call cache.trailing_bytes == 1
cache.feed(x);
// after this call cache.trailing_bytes == 2
cache.feed(y);
// after this call cache.trailing_bytes == 3
cache.feed(z);

let (line, col) = cache.byte_to_line_and_col(1, y, 1); 
```
when you look up a byte in `y`, you want a window into the past where `trailing_bytes == 1`, but by this time it has moved beyond that due to the call to `feed(z)`

ratmice:
I should probably continue:

So now looking into `byte_to_line_and_col`, we have as parameters:
`src_pos = 1`, `y = "b"`, `byte = 1`.

The output we want looking up `b`, from `cache.newlines` we know the `line_byte = 0`,
given our `src_pos = 1` we know that the current line started in a preceding string.
The column we expect is `2`, so we want to add the trailing character 'a' to the column count.

When it comes to looking up `'c'` in `z`, though we want the trailing characters to reflect both the trailing characters in
`x`, and `y`.  The previous newline is 2 strings ago, and to calculate the column 3 for `'c'` with 2 of those characters preceding.

ratmice:
I tried to rename the variables in a way which more reflects their usage, and added a large comment explaining the purpose and layout of it in this case, I think/hope I do better in that format than in comment type thread even if we decide to move that comment out elsewhere, I hope that either the above or the last commit may help in conveying the need for it.

ratmice:
So, I did manage in 11bc6c9 to remove the `trailing_bytes` from what was trailing.  So your intuition that something was unnecessary there was right on.  Sorry I didn't see it earlier, too much thinking relative to line offsets.

Do you mind if I reorder and squash some commits out, to get rid of a bunch of interim changes?

ltratt:
> Do you mind if I reorder and squash some commits out, to get rid of a bunch of interim changes?

Please go ahead.

ratmice:
Getting started on that,

I'm not going to change anything during the rebase, but here are some thoughts/ideas on
getting rid of the `HashMap`, I see two options to do that:

1. return `trailing_characters` from `feed`, and pass it back in to `bytes_to_line_and_col`.
    i.e. the first call to feed returns `0`, the second returns the `trailing_chars` accumulated by the first call.
3. variation of 1, where we change feed to something like
   `feed<'a>(&mut self, &'a str) -> LineColCacheEntry<'a>`, and pass it back to `bytes_to_line_and_col`

```
struct NewlineCacheEntry<'a> {
  src_pos: usize,
  src: &'a str,
  preceeding_cols: usize,
}
```

Would something to that effect still incur the wrath of the lifetimes, which you were trying to avoid?
It would at reduce the number of arguments to `bytes_to_line_and_col(&self, entry: &LineColCacheEntry<'a>, byte: usize)`.

*note*: given the `\r` contributing `0` to the column count, a better name for `trailing_chars` is going to be `trailing_cols`,
So I should probably rename that.

ltratt:
I wonder if the difference in our approaches is in part of a comment:

> We need the preceding characters to calculate column numbers.

My original suggestion in https://github.com/softdevteam/grmtools/pull/299#issuecomment-1153105846 was that we don't try and do anything about column numbers because the only way to do that in UTF-8 is to store the entire string (or something that's close to it). We don't want to do that because we'll roughly double the amount of memory a given input takes. I suggested a convenience function `byte_to_line_and_col` where the user can pass a `&str` in to get column numbers; but if they don't do that, all they get by default is line numbers.

ratmice:
Part of the issue is that the `&str` alone isn't enough to calculate valid column numbers, in cases where lines span multiple strings.
given a `src: &str`, columns and character positions are relative to that `&str`, so we can calculate that "a" is at `1`, but if that string is preceeded by `"foo"`, the actual column for `a` is `4`.

So without _something_ there is no way to know whether a given column number is accurate, So the problem is that `byte_to_line_and_col` needs a bit of information about the very end of the previous string which was fed before it, to actually produce a valid column number.

ltratt:
The idea of the `byte_to_line_and_col` function is that you have to pass the *whole* string in (i.e. if you did `feed("a"); feed("b")` you need to do `byte_to_line_and_col("ab")`). I can't see any way around this that doesn't involve caching the entire string.

ratmice:
Aha, now I see what you mean, and that would work.

Yeah, the only way I can think of which doesn't involve caching the entire string is the approach i've been currently undertaking or some variation on it... Now I also finally understand why you didn't need the `src_pos` argument denoting its offset either.

cfgrammar/src/lib/span.rs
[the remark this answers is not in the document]
ratmice: I believe that I covered all the cases we should return `None` in `newline_cache_negative` of 
[fb14a8d](https://github.com/softdevteam/grmtools/pull/299/commits/fb14a8d8681c3939f7b052f8c58930948a31b2fb)

I'm not sure though I understand this case about `(1, 4)`, I.e. the above test should fail an assert_eq ` left: [..., (1, 3)]  right: [..., (1,4)]`, it's possible I am missing something.

cfgrammar/src/lib/span.rs
ratmice: Well, I'd tried ignoring `\r`, but ended up ignoring '\n' if '\n' is preceded by '\r', primarily because,
if '\r' appears at the beginning of a line, we could end up with a `column: 0`, which didn't seem right.

ratmice: I believe this is fixed in [21d2f1f](https://github.com/softdevteam/grmtools/pull/299/commits/21d2f1ffac040002ba90eecd49b09b2140145f5a).

ratmice:
FWIW, Once we're done here i'll probably finish that other implementation in it's own freestanding crate for fun...

I just wanted to mention that I don't think the space overhead is that bad, if we compare it to the `newlines` vector, except maybe with very short strings.  This is primarily because the overhead needed for `newlines` is one usize per newline, while the overhead for `preceding_columns` is less than one per-string.  The map only needs to contain 1 usize per string where the column position at the end of the string is non-zero otherwise you may assume that if it is not in the map, that it is zero.

So I think in the scheme of things, it is really tiny compared to newlines. I am not sure if you were comparing to the size of the reference, or the size of the string contents if comparing to the string contents it seems negligible.

That said the additional complexity, and additional arguments are non-negligible and it seems perfectly acceptable for it to be in it's own freestanding crate.

ltratt:
Please squash.

[FWIW, I think having an external version of this with more features, but a slight performance penalty, is a good idea. Choice for users is always a good thing!]

ratmice:
Squashed.

ltratt:
Can we expand the commit message a bit because "A struct NewlineToLineColCache is added to convert the byte offsets of spans into line and column values." is sorta true, but also sorta misleading :) I think it would be nice to get across that, as it stands, it only helps with line numbers, but we provide a convenience method that also helps with column numbers for cases where the user has access to the complete input. Please force push an update.

ratmice:
That is a good point, updated.

PR #302, opened 2022-06-16
Tweak the newlinecache

ltratt:
This PR tweaks the newline cache stuff a bit: hopefully each commit explains sensibly what it's doing. I do wonder if we should replace the `byte_to` stuff with `span_to`? But that would be a future PR...

@ratmice can you review this please? If/when you're happy with it, you can bors it.

bors delegate=ratmice


cfgrammar/src/lib/newlinecache.rs
```diff
@@ -3,16 +3,14 @@ pub struct NewlineCache {
     trailing_bytes: usize,
 }
 
-impl Default for NewlineCache {
-    fn default() -> NewlineCache {
-        NewlineCache {
+impl NewlineCache {
+    pub fn new() -> Self {
```
ratmice, line 10: :+1:  :100: I only did this so clippy wouldn't trigger on it,
I only now realize that `#![allow(clippy::new_without_default)]` is enabled.

cfgrammar/src/lib/newlinecache.rs
```diff
@@ -11,6 +11,14 @@ impl NewlineCache {
         }
     }
 
+    /// Construct a `NewlineCache` directly from a `&str`. This is equivalent to creating a blank
+    /// `NewlineCache` and [`feed()`]ing the string directly in.
+    pub fn from_str(s: &str) -> Self {
```
ratmice, line 19: Clippy is going to want this to implement,
https://doc.rust-lang.org/stable/std/str/trait.FromStr.html

But that returns a `Result`, we could use `type Error = std::convert::Infallible;`
I'm not sure how that works/whether it e.g. allows use to use `?` in functions that don't return results though.

Another option would be `#[allow(clippy::should_implement_trait)]` on this function, or a name like `cache_str`?
I'm not sure about allowing that lint globally though, if there may be cases we want it to catch.

ltratt: That's unfortunate. My gut instinct is to just ignore Clippy for this specific function but let me chew on it.

ratmice: Another option is that we could implement this using the `From` trait.

cfgrammar/src/lib/newlinecache.rs
```diff
@@ -84,9 +89,12 @@ impl NewlineCache {
...
     }
 
-    /// Given a `src` string, which should have strings previously passed to `feed` concatenated together.
-    /// Convert the given `byte` offset in the input to logical line and column numbers.
-    /// Returns None if the byte offset exceeds the known input length, or the src and input length differ.
+    /// A convenience method to return the logical line and logical column number of a byte. This
+    /// requires passing a `&str` which *must* be equivalent to the string(s) passed to `feed`.
+    //
+    /// # Panics
+    ///
```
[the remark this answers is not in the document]
ltratt: We can't *guarantee* to panic if they pass in an incorrect string but we *might* panic (e.g. because we try to index in the middle of a unicode char).

Sort-of clarified in c9731d1.

cfgrammar/src/lib/newlinecache.rs
ltratt: I think `FromStr` is the least worst option. Fixed in 73f40a1.

ratmice:
Sorry, should have asked if you wanted to squash anything first.

ltratt:
Hopefully https://github.com/softdevteam/grmtools/pull/302/commits/7ab42aa9305d11e73df17034842bd86212c0730e fixes bors. If you're OK with that, I will need to do a bit of squashing.

bors try

ratmice:
Absolutely OK with that.

ltratt:
Squashed.

PR #303, opened 2022-06-16
Implement FromIterator<&str> for NewlineCache.

ratmice:
Adding a `FromStr` impl made me wonder if we couldn't clean up `newline_test_helper` a little with an `FromIterator` implementation.  It ends up adding more code than it removes, but does reduce code at the call site...
I think it makes the call site clearer.

I haven't really looked at the code generation and whether `feed` + `push_str` in a single loop yields any better code though.
But I guess the presence of `FromIterator` stops anyone from writing loops manually.

Curious if you think it is worthwhile?

### DIFF
--- a/cfgrammar/src/lib/newlinecache.rs
+++ b/cfgrammar/src/lib/newlinecache.rs
@@ -135,17 +135,26 @@ impl FromStr for NewlineCache {
     }
 }
 
+impl<'a> FromIterator<&'a str> for NewlineCache {
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: IntoIterator<Item = &'a str>,
+    {
+        let mut nlcache = NewlineCache::new();
+        for s in iter.into_iter() {
+            nlcache.feed(s)
+        }
+        nlcache
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::NewlineCache;
 
     fn newline_test_helper(feed: &[&str], tests: &[(usize, usize)]) {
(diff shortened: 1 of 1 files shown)


cfgrammar/src/lib/newlinecache.rs
```diff
@@ -135,17 +135,27 @@ impl FromStr for NewlineCache {
...
+    use std::ops::Deref;
 
     fn newline_test_helper(feed: &[&str], tests: &[(usize, usize)]) {
-        let mut cache = NewlineCache::new();
-        let mut full_string = String::new();
-        for f in feed {
-            cache.feed(f);
-            full_string.push_str(f);
-        }
+        let cache: NewlineCache = feed.iter().map(|s| s.deref()).collect();
```
ratmice, line 157: I guess `feed.iter().copied().collect()` might be preferred?

ltratt: Yes, `copied` is Clippy-friendly, so let's do that.

ratmice: Alright, I pushed that, and squashed.

ltratt: OK. Though please don't squash without the reviewer asking next time, as it makes reviewing in general very difficult. This one time, fortunately, even my pea brain could cope with things :)

ratmice: Ok, I thought the intermediate commit would help, but will avoid that in the future.

Issue #300, opened 2022-06-05
Error span improvements


ratmice:
In pr #299 which adds spans to various Error types, the Spans returned are based off of the existing
offset data from which we can derive a line & column.  As it is we currently always return a span where start == end, since it is just getting us to the desired semver ABI.

1. SemVer compatible changes (after we add Spans to Errors): 
    After that PR we could include in the error more information from the `parse` functions into `YaccParserError` and `LexBuildError`.
   This may require some reorganization of the various private `parse` functions.
3. Potential SemVer incompatible changes (after we add Spans to Errors):
   `YaccErrorKind`, and `LexErrorKind` could sometimes have useful additional spans, for instance `LexErrorKind::DuplicateName`
  Could have a span pointing to the first occurrence of the duplicate entry.

- [ ] SemVer compatible improvements
- [x] SemVer incompatible improvements

ratmice:
Closing as I believe that we got all this finished.

PR #304, opened 2022-06-17
YaccParserErrorKind::DuplicateRule add an original_span field

ratmice:
While YaccParserError has the span of the duplicate, original_span contains a span for the rule being duplicated.
Issue #300 task #2 (SemVer incompatible) extra spans, has quite a few variants which could potentially use extra spans.

This patch is kind of an exploratory commit which implements it just for `YaccParserErrorKind::DuplicateRule`, 
so that we could review it and future patches for other `Duplicate*` error kinds, could perhaps follow a similar structure.

There is one place in the doc comment specific to the variant it refers to a *rule*, and one it is specific to the error which contains the kind. I couldn't think of any good variant neutral word, and figured that clarity was more important than convenience here.

*note*: I am uncertain if all `Duplicate*` and similar variants have easily accessible spans, for instance `MissingBrace` may not be tracking the original brace.

I need to update my project to master, so I can see how these end up looking in context...

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -22,12 +22,7 @@ pub enum YaccParserErrorKind {
     IllegalName,
     IllegalString,
     IncompleteRule,
-
-    DuplicateRule {
-        /// While [`YaccParserError`] has the span of the duplicate.<br/>
-        /// `original_span` contains a span for the rule being duplicated.
-        original_span: Span,
-    },
+    DuplicateRule,
     IncompleteComment,
     IncompleteAction,
     MissingColon,
@@ -49,18 +44,38 @@ pub enum YaccParserErrorKind {
     InvalidString,
 }
 
+#[derive(Debug)]
+pub enum YaccParserAdditionalErrorKind {
+    /// This kind correlates to an existing YaccParserErrorKind.
+    Error(YaccParserErrorKind),
+    /// A more specialized error kind which may not be a valid top-level error kind.
+    Specialized(YaccParserSpecializedErrorKind),
+}
+
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum YaccParserSpecializedErrorKind {
+    OriginalDeclaration,
+}
+
 /// Any error from the Yacc parser returns an instance of this struct.
 #[derive(Debug)]
 pub struct YaccParserError {
     pub kind: YaccParserErrorKind,
     pub span: Span,
+    pub related_errors: Vec<(YaccParserAdditionalErrorKind, Span)>,
 }
 
 impl Error for YaccParserError {}
 
 impl fmt::Display for YaccParserError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let s = match self.kind {
+        write!(f, "{}", self.kind)
+    }
+}
+impl fmt::Display for YaccParserErrorKind {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let s = match self {
             YaccParserErrorKind::IllegalInteger => "Illegal integer",
             YaccParserErrorKind::IllegalName => "Illegal name",
             YaccParserErrorKind::IllegalString => "Illegal string",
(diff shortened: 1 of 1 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -22,7 +22,12 @@ pub enum YaccParserErrorKind {
     IllegalName,
     IllegalString,
     IncompleteRule,
-    DuplicateRule,
+
+    DuplicateRule {
+        /// While [`YaccParserError`] has the span of the duplicate.<br/>
+        /// `original_span` contains a span for the rule being duplicated.
+        original_span: Span,
```
ratmice, line 29: I guess now, I wonder if perhaps this would be better as something like `original: DuplicatedItem{DuplicateRule, span}`.
which could allow us to implement an `Error` like `format`, and message?

ltratt: I'm not going to pretend I've followed that!

I'm broadly supportive of the general idea behind this PR *but* if we're going to break the API, let's make it more future proof. It should cope with multiple copies, so something like `DuplicateRules(Vec<Span>)` would be safer in the long run.

ratmice: Sorry about that, was attempting to sleep and write stuff down at the same time.

The general gist of it is, I was wondering if we want to provide a string message along with the span,
that mirrors the original "Duplicate rule" string but provides some context for the span.

like "The first instance of the duplicated rule", "additional duplicate rule",
we'll need some enum to match on, for implementing format strings, and it might make sense to just use the `newtype` pattern and use `YaccParserErrorKind` as exists in master for that.

So we could have something like `Vec<Newtype(YaccParserErrorKind, Span)>` in error where `newtype` provides:

```
// in the newtype
match self.kind {
   YaccParserErrorKind::DuplicateRule = f("the original rule duplicated"),
   _ => { "Does not have additional data" },
}
```

The `Other` variant with `YaccParserErrorRelatedDataKind`, was for fields which don't exactly correspond to one of the top-level error kinds, like "additional duplicate rule".

So basically I was thinking that perhaps we shouldn't add the span to kind, but put it along side a kind in the `YaccParserError`.

Part of the thinking is that this would be more future proof for errors we can't easily get spans for now, but possibly could
in the future, we can just leave an empty `Vec<(newtype, span)>` when we can't.

ltratt: Nope, I still don't understand :)

ratmice:
Going to go mark this as draft, as I would like to play with some designs locally which provide a string context for the additional span.

cfgrammar/src/lib/yacc/parser.rs
ratmice: No worries, trying to implementing it, will perhaps be better than trying to explain it, and it may end up a terrible idea.

ratmice:
So, i went ahead and pushed a commit for which lays out the general idea.

I kind of want to lay out the thought process behind it:

1. It is much easier to be semver compatible and incremental work towards adding spans, if we don't have to add the span to each `YaccParserErrorKind` variants.  So instead this adds a Vec<(kind, Span)> as a field to the error.
2. It makes it non-exhaustive so we can add additional Specialized errors in the future, in theory you shouldn't need to match on it    Just turn them into strings?

This also changes the `Display` impl of `YaccParserError` around, moving it mostly into `YaccParserErrorKind`,
So that we can reuse it.

It is certainly much less polished than my original commit, but I think it may be a more flexible mechanism in the future?

Anyhow, I'm curious your thoughts on the matter.

ltratt:
I must admit that I think this is too complex, and we're also trying to do two things in one PR, when it would be better to split them in two. I would prefer this PR *only* tries to deal with "here are the >1 place you tried to define rule R". Then a second PR could try to deal with "how do we reflect multiple errors in the grammar?" [An obvious thing to try for the second PR would be to return `Result<..., Vec<ParserError>>` or similar rather than a substruct.]

ratmice:
Sounds like a plan, I'm going to go ahead and close this, since that seems like it won't have much in common with either of the previous two commits.

I agree that it is probably too complex too, i.e. that we only actually use the `Kind` now in the `Display` impl of `ParserError` does seem like it could allow for some simplifications, and that some of the complexity here is probably incidental to the previous design.

So this seems like a good call.

PR #305, opened 2022-06-17
Use NewlineCache inside lrlex.

ltratt:
This does require adding the `span_line_bytes` method, but is still a simplification overall.

@ratmice can you review this please? If/when you're happy with it, you can bors it.

bors delegate=ratmice

### DIFF
--- a/cfgrammar/src/lib/newlinecache.rs
+++ b/cfgrammar/src/lib/newlinecache.rs
@@ -1,3 +1,4 @@
+use crate::Span;
 use std::str::FromStr;
 
 /// Cache newlines from an input. These can be used to turn UTF-8 byte offsets into user-friendly
@@ -121,6 +122,26 @@ impl NewlineCache {
             (line_num, column)
         })
     }
+
+    /// Return the (start byte, end byte) of the lines containing `span`. This will always cover
+    /// at least 1 logical line.
+    pub fn span_line_bytes(&self, span: Span) -> (usize, usize) {
+        let (st, st_line) = match self.newlines.binary_search(&span.start()) {
+            Ok(j) => (self.newlines[j], j + 1),
+            Err(j) => (self.newlines[j - 1], j),
+        };
+        let en = match self.newlines[st_line..].binary_search(&span.end()) {
+            Ok(j) if st_line + j == self.newlines.len() - st_line => {
+                self.newlines.last().unwrap() + self.trailing_bytes
+            }
+            Ok(j) => self.newlines[st_line + j + 1] - 1,
+            Err(j) if st_line + j == self.newlines.len() => {
+                self.newlines.last().unwrap() + self.trailing_bytes
+            }
+            Err(j) => self.newlines[st_line + j] - 1,
+        };
+        (st, en)
+    }
 }
 
 impl FromStr for NewlineCache {
@@ -151,6 +172,8 @@ impl<'a> FromIterator<&'a str> for NewlineCache {
 #[cfg(test)]
 mod tests {
     use super::NewlineCache;
+    use crate::Span;
+    use std::str::FromStr;
 
     fn newline_test_helper(feed: &[&str], tests: &[(usize, usize)]) {
         let cache: NewlineCache = feed.iter().copied().collect();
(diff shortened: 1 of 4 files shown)


cfgrammar/src/lib/newlinecache.rs
```diff
@@ -121,6 +122,24 @@ impl NewlineCache {
...
+            Ok(j) => (self.newlines[j], j + 1),
+            Err(0) => (0, 0),
+            Err(j) => (self.newlines[j - 1], j),
+        };
+        let en = match self.newlines[st_line..].binary_search(&span.end()) {
+            Ok(j) => self.newlines[st_line + j + 1] - 1,
+            Err(j) if st_line + j == self.newlines.len() => {
+                self.newlines.last().unwrap() + self.trailing_bytes
+            }
+            Err(j) => self.newlines[st_line + j] - 1,
```
ratmice, line 139: I've had a little trouble convincing myself the `Ok(j)` and `Err(j)` cases where the slicing or subtraction could cause a panic are unreachable, and covered by the pattern guard.

But, given that this is movement from below, and the conditions of the `newlines` should be the same, 
It now also handles `trailing_bytes`, So I wouldn't consider any issues here to be a regression.

So, I think this patch as it is an improvement, but it might be nice to have some clarification around these aspects, but given that it is moving code perhaps that could be a future improvement

cfgrammar/src/lib/newlinecache.rs
```diff
@@ -121,6 +122,24 @@ impl NewlineCache {
             (line_num, column)
         })
     }
+
+    /// Return the (start byte, end byte) of the lines containing `span`. This will always cover
+    /// at least 1 logical line.
+    pub fn span_line_bytes(&self, span: Span) -> (usize, usize) {
+        let (st, st_line) = match self.newlines.binary_search(&span.start()) {
+            Ok(j) => (self.newlines[j], j + 1),
+            Err(0) => (0, 0),
```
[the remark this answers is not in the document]
ratmice: Ahh, I see now this was reachable old code, but I think it no longer is.

cfgrammar/src/lib/newlinecache.rs
ratmice: It seems like `span_line_eq("aa\n", 2, 3, "aa\n");` hits one of the panics I was worried about in the `Ok(j)` branch.

cfgrammar/src/lib/newlinecache.rs
```diff
@@ -251,4 +272,37 @@ mod tests {
...
+        span_line_eq(" a\nb\n  c d", 3, 4, "b");
+        span_line_eq(" a\nb\n  c d", 7, 8, "  c d");
+        span_line_eq(" a\nb\n  c d", 9, 10, "  c d");
+
+        let s = "\na\na a\na a a\na a a a";
+        span_line_eq(&s, 0, 0, "");
+        span_line_eq(&s, 0, 2, "\na");
+        span_line_eq(&s, 0, 4, "\na\na a");
+        span_line_eq(&s, 0, 7, "\na\na a\na a a");
+        span_line_eq(&s, 4, 7, "a a\na a a");
```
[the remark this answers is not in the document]
ltratt: Fixed in 75f8374.

cfgrammar/src/lib/newlinecache.rs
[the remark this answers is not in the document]
ltratt: Agreed. I *think* this code originally assumed a possibly-empty `Vec`.

Fixed in 2222013.

cfgrammar/src/lib/newlinecache.rs
ltratt: Very good spot. Fixed (with a mild difference, but the same in spirit) in 01c24eb.

ratmice:
Please squash.

cfgrammar/src/lib/newlinecache.rs
```diff
@@ -131,6 +131,9 @@ impl NewlineCache {
             Err(j) => (self.newlines[j - 1], j),
         };
         let en = match self.newlines[st_line..].binary_search(&span.end()) {
+            Ok(j) if st_line + j == self.newlines.len() - st_line => {
+                self.newlines.last().unwrap() + self.trailing_bytes
```
ratmice, line 135: Only one minor thing, now that we have 2 of these cases, should we consider making
`input_length` public, and perhaps renaming it `input_len` (been kicking myself over that naming choice).

But i'll leave it up to you if we want to do this.

ratmice: Actually, cancel that these are in the module so, we could just use `self.input_length()`.

ltratt: I plan on doing a pass over the method names soon-ish now that the API is starting to look stable-ish. We can leave this for now though.

ltratt:
Squashed.

PR #307, opened 2022-06-19
Collect spans for multiple duplicate rules during yacc parsing

ratmice:
This series adds display in support of some tests,
then adds tests, then modifies `DuplicateRule` to collect spans into a vector.

It doesn't modify the return value, but does change behavior when you have a syntax error after a duplicate rule.
Now only triggers `DuplicateRule` error *after* successfully parsing all of the rules.

Not sure if we really want the macro, the reason it is a macro is because of the `$error_kind:path` which allows the match to work.
To turn it into a function I think we need to add `Eq`, and use the `if kind == error_kind` pattern guards.

The tests don't check the spans in anticipation of them changing.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -1,10 +1,14 @@
 // Note: this is the parser for both YaccKind::Original(YaccOriginalActionKind::GenericParseTree) and YaccKind::Eco yacc kinds.
 
-use std::{collections::HashSet, error::Error, fmt, str::FromStr};
-
 use lazy_static::lazy_static;
 use num_traits::PrimInt;
 use regex::Regex;
+use std::{
+    collections::{HashMap, HashSet},
+    error::Error,
+    fmt,
+    str::FromStr,
+};
 
 type YaccResult<T> = Result<T, YaccParserError>;
 
@@ -16,13 +20,14 @@ use super::{
 };
 
 /// The various different possible Yacc parser errors.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum YaccParserErrorKind {
     IllegalInteger,
     IllegalName,
     IllegalString,
     IncompleteRule,
-    DuplicateRule,
+    /// Contains the spans of all duplicate rules.
+    DuplicateRule(Vec<Span>),
     IncompleteComment,
     IncompleteAction,
     MissingColon,
@@ -45,7 +50,7 @@ pub enum YaccParserErrorKind {
 }
 
 /// Any error from the Yacc parser returns an instance of this struct.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct YaccParserError {
     pub kind: YaccParserErrorKind,
     pub span: Span,
@@ -55,12 +60,18 @@ impl Error for YaccParserError {}
 
 impl fmt::Display for YaccParserError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let s = match self.kind {
+        write!(f, "{}", self.kind)
+    }
+}
+
+impl fmt::Display for YaccParserErrorKind {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let s = match self {
             YaccParserErrorKind::IllegalInteger => "Illegal integer",
             YaccParserErrorKind::IllegalName => "Illegal name",
             YaccParserErrorKind::IllegalString => "Illegal string",
             YaccParserErrorKind::IncompleteRule => "Incomplete rule",
-            YaccParserErrorKind::DuplicateRule => "Duplicate rule",
+            YaccParserErrorKind::DuplicateRule(_) => "Duplicated rule",
             YaccParserErrorKind::IncompleteComment => "Incomplete comment",
             YaccParserErrorKind::IncompleteAction => "Incomplete action",
             YaccParserErrorKind::MissingColon => "Missing ':'",
@@ -99,6 +110,9 @@ pub(crate) struct YaccParser {
     num_newlines: usize,
     ast: GrammarAST,
     global_actiontype: Option<String>,
+    /// The key is the span of the [Rule](crate::yacc::ast::Rule) being duplicated.
+    /// The value contains one span for every duplicate of the key.
+    duplicate_rule_spans: HashMap<Span, Vec<Span>>,
 }
 
 lazy_static! {
@@ -116,6 +130,7 @@ impl YaccParser {
             num_newlines: 0,
             ast: GrammarAST::new(),
             global_actiontype: None,
+            duplicate_rule_spans: HashMap::new(),
         }
     }
 
@@ -315,6 +330,14 @@ impl YaccParser {
             i = self.parse_rule(i)?;
             i = self.parse_ws(i, true)?;
         }
+
+        if let Some((orig_span, spans)) = self.duplicate_rule_spans.iter().next() {
+            return Err(YaccParserError {
+                kind: YaccParserErrorKind::DuplicateRule(spans.clone()),
+                span: *orig_span,
+            });
+        }
+
         Ok(i)
     }
 
@@ -333,9 +356,6 @@ impl YaccParser {
                 i = j;
             }
             YaccKind::Grmtools => {
-                if self.ast.get_rule(&rn).is_some() {
-                    return Err(self.mk_error(YaccParserErrorKind::DuplicateRule, i));
-                }
                 i = self.parse_ws(j, true)?;
                 if let Some(j) = self.lookahead_is("->", i) {
                     i = j;
@@ -344,7 +364,14 @@ impl YaccParser {
                 }
                 i = self.parse_ws(i, true)?;
                 let (j, actiont) = self.parse_to_single_colon(i)?;
-                self.ast.add_rule((rn.clone(), span), Some(actiont));
+                match self.ast.get_rule(&rn) {
+                    None => self.ast.add_rule((rn.clone(), span), Some(actiont)),
+                    Some(orig) => self
+                        .duplicate_rule_spans
+                        .entry(orig.name.1)
+                        .or_insert_with(Vec::new)
+                        .push(span),
+                }
                 i = j;
             }
         }
@@ -1792,6 +1819,32 @@ x"
         assert_eq!(
             grm.parse_param,
             Some(("a::b".to_owned(), "(u64, u64)".to_owned()))
+        );
+    }
+
+    #[test]
+    fn test_multi_span_error_kinds() {
+        let e = parse(
+            YaccKind::Grmtools,
+            "%start A
+             %%
+             A -> (): 'a1';
+             B -> (): 'bb';
+             A -> (): 'a2';
+             C -> (): 'cc';
+             A -> (): 'a3';
+            ",
+        )
+        .expect_err("parse successful while expecting error");
+        assert_eq!(
+            e,
+            YaccParserError {
+                span: Span::new(38, 39),
+                kind: YaccParserErrorKind::DuplicateRule(vec![
+                    Span::new(94, 95),
+                    Span::new(150, 151)
+                ])
+            }
         );
     }
 }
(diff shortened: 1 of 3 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -310,20 +310,29 @@ impl YaccParser {
...
+                kind: YaccParserErrorKind::DuplicateRule,
+                span: *span,
+            });
+        }
+
         Ok(i)
     }
 
-    fn parse_rule(&mut self, mut i: usize) -> YaccResult<usize> {
+    fn parse_rule(&mut self, duplicate_spans: &mut Vec<Span>, mut i: usize) -> YaccResult<usize> {
```
ratmice, line 335: This seems like it may want to be a `HashMap<Name, Vec<Span>>`, if we want to collate these by name.
rather than the order they appear in?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -55,7 +55,12 @@ impl Error for YaccParserError {}
 
 impl fmt::Display for YaccParserError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let s = match self.kind {
+        write!(f, "{}", self.kind)
+    }
+}
```
ltratt, line 60: Missing blank line.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -305,20 +310,29 @@ impl YaccParser {
     }
 
     fn parse_rules(&mut self, mut i: usize) -> YaccResult<usize> {
+        let mut duplicate_spans = Vec::new();
```
ltratt, line 313: It would be easier to work with IMHO if this was an attribute on `YaccParser` (we'll have more of these in the future, so having them all as parameters will get messy).

cfgrammar/src/lib/yacc/parser.rs
ratmice: Should be fixed in https://github.com/softdevteam/grmtools/pull/307/commits/e3770abb08029f1f5e54609c4283fce28cb6d4b5

cfgrammar/src/lib/yacc/parser.rs
ratmice: Changed the name to `duplicate_rule_spans` while doing that, and should be fixed in https://github.com/softdevteam/grmtools/pull/307/commits/d5ffadc4e2c426092604332bbcbb81a7a8e91adc

ratmice:
There is perhaps one thing worth asking about, I *think* it is intended for yacc compatibility,
But if we look at https://github.com/softdevteam/grmtools/blob/88d85957e324b2e7ca758051323d7c6f4d85361a/cfgrammar/src/lib /yacc/parser.rs#L329 we see error only happens with `YaccKind::Grmtools` and not `YaccKind::Original`.

I didn't change it assuming it was intended.

cfgrammar/src/lib/yacc/parser.rs
ltratt: Good point -- I agree.

ltratt:
> There is perhaps one thing worth asking about, I think it is intended for yacc compatibility,
But if we look at parser.rs:329 we see error only happens with YaccKind::Grmtools and not YaccKind::Original.

I no longer remember! Maybe see what BSD yacc and GNU Bison do? Then we can make a decision about whether this is allowed or not.

ratmice:
I haven't yet checked if ignoring them is the right thing to do, but they are definitely both silent.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -354,11 +354,13 @@ impl YaccParser {
                 i = j;
             }
             YaccKind::Grmtools => {
-                if self.ast.get_rule(&rn).is_some() {
+                let mut is_dupe = false;
+                if let Some(original_rule) = self.ast.get_rule(&rn) {
```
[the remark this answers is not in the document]
ratmice: much nicer, fixed in https://github.com/softdevteam/grmtools/pull/307/commits/468426670d67ae711c94d47fa7df5da025838456

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -99,6 +109,7 @@ pub(crate) struct YaccParser {
     num_newlines: usize,
     ast: GrammarAST,
     global_actiontype: Option<String>,
+    duplicate_rule_spans: HashMap<Span, Vec<Span>>,
```
ltratt, line 112: Let's add a comment/docstring that explains what the two `Span`s mean (otherwise we'll forget!).

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1794,4 +1819,30 @@ x"
...
+
+        expect_yacc_error_kind!(
+            YaccKind::Grmtools,
+            "%start A
+             %%
+             A -> (): 'a1';
+             A -> (): 'a2';
+             A -> (): 'a3';
+            ",
+            YaccParserErrorKind::DuplicateRule
```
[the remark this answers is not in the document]
ratmice: Yeah, it can't yet because we only ever return the one of these errors from `parse_rules`.
I've a local `assert_eq!(errors.len(), 2);` check, but that check really doesn't belong there.

We figure out the error type to return multiple things, before we can do that.

cfgrammar/src/lib/yacc/parser.rs
ratmice: fixed in https://github.com/softdevteam/grmtools/pull/307/commits/a39ed7e4baf5a69a04964daaed69c0667d0ee0ae

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -315,6 +327,15 @@ impl YaccParser {
...
             i = self.parse_ws(i, true)?;
         }
+
+        if let Some((_, spans)) = self.duplicate_rule_spans.iter().next() {
+            let span = spans.first().unwrap();
+            return Err(YaccParserError {
+                kind: YaccParserErrorKind::DuplicateRule,
+                span: *span,
+            });
+        }
```
ratmice, line 339: This checks whether there exists an error, and returns the first duplicate span,
as `YaccParserError` still just contains a single `kind`/`span`, as there isn't a way to return multiple errors yet.

ratmice: Yeah, the big question here is if we should do it on a per-`YaccParserErrorKind` basis, or on a `YaccParserError` basis.

Doing it in `YaccParserError` means, *any* error can produce multiple errors.
for instance, we can still perhaps produce syntax errors (e.g. `MissingRightArrow` after a `DuplicateRule` is encountered, and it may be nice to produce a  `vec![(DuplicateRule, span), (MissingRightArrow, span]`.

For this i'm kind of partial to changing `YaccParserError` to have something like a `Vec<YaccParserErrorKind, Span>`.
Making it part of `YaccParserError` also means, we don't have to change the semver compatibility for other `Duplicate*` errors like `DuplicatePrecedence`.

I think that both of these aspects (semver, and encountering multiple errors) make me lean towards keeping `DuplicateRule` as-is, and adding it to the `YaccParserError`.

ltratt: There are two different things here:

1. Grammars can have multiple errors.
2. A given rule can be duplicated more than once.

In this case, we're not yet handling #1 but we are handling #2. So `YaccParserErrorKind::DuplicateRule(Vec<Span>)` makes sense here IMHO.

ratmice: Right, the only qualm I have with this is that it isn't reusable for other `Duplicate*` types, of which there are quite a few.
So adding the `Vec<Span>` has to be done for each one of these individually as a semver breaking change.

Perhaps we could consider adding an `additional_spans: Vec<Span>` in `YaccParserError` to sit along side the `kind`?
The major difference being with this solution _every_ error, even ones which will never produce multiple spans, have an empty vector of additional spans.

ltratt: > So adding the Vec<Span> has to be done for each one of these individually as a semver breaking change.

Yes, that's true. Of course there's no reason why we can't add a `Vec` to them now, even if we don't really populate it. [Not in this PR! But in the next PR or two.]

> Perhaps we could consider adding an additional_spans: Vec<Span> in YaccParserError to sit along side the kind?

I'm not keen on this approach, because it only applies to a subset of errors.

ltratt: Yep!

ratmice: Fixed in https://github.com/softdevteam/grmtools/pull/307/commits/8cb0b6746e9b46b2b7ce860c63fa0337b24bbea2

I managed to salvage the macro, but it isn't exactly pretty.
The `spans.clone()` is kind of regrettable here, I think the only way to avoid that clone is to move it back out of `self`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1844,7 +1852,8 @@ x"
              A -> (): 'a2';
              A -> (): 'a3';
             ",
-            YaccParserErrorKind::DuplicateRule
+            YaccParserErrorKind::DuplicateRule,
+            2
```
ltratt, line 1856: We really need to test a) that the duplicated rule is `A` b) the locations of the duplicates. [We should also add perhaps rules `B` and `C` between each `A`.] Without that more specific test, we could quite easily break the code and return duplicates that are actually of the original rule.

ratmice: I tried to do this in a way that doesn't require me to go hard code spans individually, and still hopefully be reusable
for other duplicate error kinds in the future.  But I believe this is fixed in https://github.com/softdevteam/grmtools/pull/307/commits/add3e128ad789316d955fe79556ba67b3a2e438e

ltratt: I'm afraid we really do have to check the `Span`s (lots of other grmtools tests do this, for good reason).

ltratt: Yep, it's tedious, but we have to do it, because without that we can't a) be sure there aren't bugs now b) we don't introduce bugs later. That's why there are *lots* of hard-coded `Span` tests already.

ratmice: fixed in https://github.com/softdevteam/grmtools/pull/307/commits/fe8974a76d5e28c4465daae75726630f3aaeb4d6

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1862,14 +1860,15 @@ x"
...
              A -> (): 'a2';
-             C -> (): 'c';
+             C -> (): 'cc';
              A -> (): 'a3';
             ",
             YaccParserErrorKind::DuplicateRule,
             "A",
-            2
+            Span::new(38, 39),
+            vec![Span::new(94, 95), Span::new(150, 151)]
```
ratmice, line 1871: IMO this macro has gotten a bit ugly now, it may just be better to add `Eq` to `YaccParserErrorKind`,
and then pass `YaccParserErrorKind::DuplicateRule([...])`, It means dropping the

"expected DuplicateRule(vec![])", on test failure, i'm not sure it is worth the complexity it adds by splitting these apart...

ratmice: I had to add a trait, to get the spans out of a kind, for testing against the expected string `"A"`,
I just added it private to the testsuite module, but was able to keep everything.

ltratt: I wonder if we should simplify this even further? Ultimately wouldn't something super simple like this work?

```rust
let r = yacc_kind, "%start A ...");
assert_eq!(r,
  YaccParserErrorKind::DuplicateRule { span: Span::new(...), kind: DuplicateRule(vec![Span::new(...), Span::new(...)]});
```

ltratt:
Now for the magic words: please squash :)

ratmice:
Squashed

ltratt:
Good stuff :)

bors r+

PR #308, opened 2022-06-21
Add a Vec<Span> to YaccErrorKind::Duplicate* variants

ratmice:
It seems like it is best to add other YaccParserErrorKind variants for a nice clean history/merge.
(At least as long the variant isn't turning into a any kind of grand adventure).

* [X] `DuplicateRules` https://github.com/softdevteam/grmtools/commit/36534ca45807579adbc2907bf0aee6705dd3e1dc (In a previous PR #307),
* [X] `DuplicateAvoidInsert`
* [X] `DuplicatePrecedence`,
* [X] `DuplicateImplicitTokensDeclaration`,
* [X] `DuplicateExpectDeclaration`,
* [X] `DuplicateExpectRRDeclaration`,
* [X] `DuplicateStartDeclaration`,
* [X] `DuplicateActiontypeDeclaration`,
* [X] `DuplicateEPP`,

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -51,7 +51,8 @@ pub enum YaccParserErrorKind {
     DuplicateStartDeclaration(Vec<Span>),
     /// Contains the spans of all duplicate action type declarations.
     DuplicateActiontypeDeclaration(Vec<Span>),
-    DuplicateEPP,
+    /// Contains the spans of all duplicate epp declaration.
+    DuplicateEPP(Vec<Span>),
     ReachedEOL,
     InvalidString,
 }
@@ -105,7 +106,7 @@ impl fmt::Display for YaccParserErrorKind {
             YaccParserErrorKind::DuplicateActiontypeDeclaration(_) => {
                 "Duplicate %actiontype declaration"
             }
-            YaccParserErrorKind::DuplicateEPP => "Duplicate %epp declaration for this token",
+            YaccParserErrorKind::DuplicateEPP(_) => "Duplicate %epp declaration for this token",
             YaccParserErrorKind::ReachedEOL => {
                 "Reached end of line without finding expected content"
             }
@@ -131,6 +132,7 @@ pub(crate) struct YaccParser {
     duplicate_expectrr_declarations: Option<(Span, Vec<Span>)>,
     duplicate_start_declarations: Option<(Span, Vec<Span>)>,
     duplicate_actiontype_declarations: Option<Vec<Span>>,
+    duplicate_epp_declarations: HashMap<Span, Vec<Span>>,
 }
 
 lazy_static! {
@@ -156,6 +158,7 @@ impl YaccParser {
             duplicate_expectrr_declarations: None,
             duplicate_start_declarations: None,
             duplicate_actiontype_declarations: None,
+            duplicate_epp_declarations: HashMap::new(),
         }
     }
 
@@ -197,6 +200,12 @@ impl YaccParser {
         if let Some((orig_span, spans)) = &self.duplicate_start_declarations {
             return Err(YaccParserError {
                 kind: YaccParserErrorKind::DuplicateStartDeclaration(spans.clone()),
+                span: *orig_span,
+            });
+        }
+        if let Some((orig_span, spans)) = self.duplicate_epp_declarations.iter().next() {
+            return Err(YaccParserError {
+                kind: YaccParserErrorKind::DuplicateEPP(spans.clone()),
                 span: *orig_span,
             });
         }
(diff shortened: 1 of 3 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -228,19 +236,24 @@ impl YaccParser {
...
                 while j < self.src.len() && self.num_newlines == num_newlines {
                     let (j, n, span) = self.parse_token(i)?;
                     if self.ast.tokens.insert(n.clone()) {
                         self.ast.spans.push(span);
                     }
-                    if self.ast.avoid_insert.as_ref().unwrap().contains(&n) {
-                        return Err(
-                            self.mk_error(YaccParserErrorKind::DuplicateAvoidInsertDeclaration, i)
-                        );
+                    match self.ast.avoid_insert.as_ref().unwrap().get(&n) {
```
ltratt, line 246: May be better written as `self.ast.avoid_insert.entry(...)` (or one of the other combinators?) so you don't have to do a second `self.ast.avoid_insert.as_mut().unwrap().insert(n, span)`.

ratmice:
Let me know if you want me to keep adding variants to this series for less merge commits, or if we want to just merge now.
Otherwise it is squashed.

ltratt:
Ah there are other places where we need to do the same thing? Then let's add extra commits here -- it makes sense to do it in one PR.

ratmice:
Will do, I'll just add a task list to the initial pull-request and perhaps rename the PR.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -316,15 +325,15 @@ impl YaccParser {
...
+                    let (j, n, span) = self.parse_token(i)?;
+                    if let Some((_, orig_span)) = self.ast.precs.get(&n) {
+                        self.duplicate_precedence_spans.entry(*orig_span).or_insert_with(Vec::new).push(span);
                     }
                     let prec = Precedence {
                         level: prec_level,
                         kind,
                     };
-                    self.ast.precs.insert(n, prec);
+                    self.ast.precs.insert(n, (prec, span));
```
ratmice, line 337: Now that we don't return early, this needs to be and else/match and probably use entry.

ratmice:
About half-way done with the variants, but I think we are good with these.
 I wouldn't mind reordering/squashing these commits, then updating the top comment.

ltratt:
Let's keep going with the commits as-is, then we can do squashing at the end.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -43,7 +43,8 @@ pub enum YaccParserErrorKind {
     DuplicateAvoidInsertDeclaration(Vec<Span>),
     /// Contains the spans of all duplicated implicit token declarations.
     DuplicateImplicitTokensDeclaration(Vec<Span>),
-    DuplicateExpectDeclaration,
+    /// Contains the spans of all duplicated expect declarations.
```
ratmice, line 46: `duplicated` -> `duplicate`.  been using the wrong tense here in a few places.

ltratt: Good point.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -244,12 +253,17 @@ impl YaccParser {
                 continue;
             }
             if let Some(j) = self.lookahead_is("%expect", i) {
-                if self.ast.expect.is_some() {
-                    return Err(self.mk_error(YaccParserErrorKind::DuplicateExpectDeclaration, i));
-                }
                 i = self.parse_ws(j, false)?;
                 let (j, n) = self.parse_int(i)?;
-                self.ast.expect = Some(n);
+                let span = Span::new(i, j);
```
ratmice, line 258: So this is referring to the span surrounding `"1"` in `"%expect 1"`.
Up until this point this has been the right thing, but because it is the expect declaration which is unique rather than the value.
This should span "%expect 1" or "%expect".

ratmice: Cool, then I won't change anything.

ltratt:
Works for me! Any duplicate-possibilities left?

ratmice:
Just 3 more, `DuplicateStart`, `DuplicateActionTypeDeclaration`, and `DuplicateEPP`.
After that I think it's just `lrlex::LexErrorKind::DuplicateName`, not sure if we want to handle that here, or in another PR.

Start is taking a bit because of its pervasiveness across the library/testsuite.

ratmice:
Alright, well that is it I believe. At least I am not aware that the others can return multiple locations, but I haven't done a thorough examination there.

The last one, I ended up adding a second span, just because I thought it would be really confusing if I were to put a tuple of String and Span together if the Span was not actually for the String, but the key.

ltratt:
Please squash.

ratmice:
Squashed, but do let me know if you prefer it squashed further into one commit.

ltratt:
It looks like they're correct now!

bors r+

ratmice:
My fault pushed the wrong branch accidentally, fixed.

PR #309, opened 2022-06-22
Add a Vec<Span> to LexErrorKind::DuplicateName

ratmice:
Previously this error returned a span which was at the initial quotation
mark.  Now it points to the string inside the quotation mark.

In this case it would be a bit of work to keep this behavior since wouldn't directly correlate to any `name_span`.
I figured it would be easier to do it later, than do it and undo it later because we don't want that complexity.

let me know if you would like to keep that behavior the same.

### DIFF
--- a/lrlex/src/lib/mod.rs
+++ b/lrlex/src/lib/mod.rs
@@ -50,7 +50,7 @@ pub enum LexErrorKind {
     UnknownDeclaration,
     MissingSpace,
     InvalidName,
-    DuplicateName,
+    DuplicateName(Vec<Span>),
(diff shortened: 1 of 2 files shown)


ltratt:
This looks fine to me! OK to merge?

>So, now that my mind wanders towards returning multiple errors, I notice that all these duplicate_*: HashMap<..> should likely be IndexMap's to preserve insertion order.

I tend to agree that this isn't *too* important, at least at the moment.

ratmice:
I squashed out the second commit, but I think it is OK with me.

PR #310, opened 2022-06-23
Prototype multiple errors for YaccParserError

ratmice:
I may be absent for a few days, Here is at least an experiment towards returning multiple errors.
It also tries to do some methods for integrating the newline cache/custom display methods,
I ended up with an unsafe line, and don't have time to see if there is a nice way to remove it.

It basically is just a newtype around Vec (with a pub vec even), that implements `Display + Debug`.
I'd added a trait for primary & duplicate spans...

Perhaps I got carried away, and it could be split up, below is as far as I got on the output.

```
foo.y: 3 Errors were encountered
  Duplicated %avoid_insert declaration at line 4 column 16
    1 duplicate at:
      line 5 column 16
  Duplicated %start declaration at line 1 column 8
    2 duplicates at:
      line 2 column 8
      line 3 column 8
```

### DIFF
--- a/cfgrammar/src/lib/mod.rs
+++ b/cfgrammar/src/lib/mod.rs
@@ -52,11 +52,13 @@
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+pub mod errors;
 mod idxnewtype;
 pub mod newlinecache;
 pub mod span;
 pub mod yacc;
 
+pub use errors::Errors;
 pub use newlinecache::NewlineCache;
 pub use span::Span;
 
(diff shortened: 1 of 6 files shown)


ltratt:
I think we might be trying to do a bit too much in one PR here. Returning `Vec<Error>` seems like a sensible first step. Then we can work out if/how we might want to format it and so on.

ratmice:
I went ahead and pushed that #311 
I'm not certain how I feel about it (e.g. the situation with YaccGrammarError and display) where we can't actually implement Display for `Vec`.  So e.g. `parse` previously implemented something which implemented display, now it doesn't.

Anyhow closing this in favor of that.

PR #311, opened 2022-06-24
Return a Vec for multiple YaccParserErrors errors.

ratmice:


cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -1025,14 +1025,14 @@ where
 
 #[derive(Debug)]
 pub enum YaccGrammarError {
```
ltratt, line 1027: This might be a dumb question (I don't remember this bit of the code well enough): could/should we fold `YaccParserError` and `GrammarValidationError` into one? Or, perhaps, should we return `Vec<YaccGrammarError>`?

ratmice: I'm not terribly familiar with `GrammarValidationError` either, 

It seems plausible to me, the line between `GrammarValidationErrors` and `YaccParserErrors` is blurry.
For example, `Duplicate`* could be seen as validation errors, since they do not cause us to fail to parse later stages.

But `Duplicate`* can be checked as we parse, and doing so is beneficial because it means we don't add state which is known to be invalid, e.g. to the `rules`. field.

So, `GrammarValidationError` is kind of a niche thing, for errors which cannot be checked at parse time.
I.e. `NoStartRule` can't be determined until *after* we've parsed all the rules.  

If we take the position (hypothetically) that `Duplicate`* should really be validation errors.
We find that we are potentially returning errors, when we *know* about other errors.

I believe that the point/benefit of `GrammarValidationError` might just be that they *don't* get added checked,
Unless we successfully parsed.  I.e. If we hit a syntax error parsing a rule, we don't want to *also* emit a `NoStartRule` error which we could do if we moved the `GrammarValidationError` checks to `parse`.

This doesn't necessarily mean that they shouldn't or couldn't be in the same `enum`, but being separate we can't make this mistake of trying to perform grammar validation in parse, and emitting cascading errors that only fail because of previous errors.

I believe that is how I read the situation, anyways.

ratmice: Sorry for perhaps excessive detail, there is a counter-argument that I failed to notice above:

We currently don't emit `NoStartRule` when we could, for instance given:
```
%avoid_insert a
%avoid_insert a
%%
```
only when we fix the error we then get 'No Start Rule'.

```
%%
Start-Rule:;
```
The rule will fail to parse because of the `-` in the rule, but it looks for
bar.y: Missing ':' at line 2 column 6

And currently *won't* emit `NoStartRule`, and I would say arguably it shouldn't.
Currently the types prevent us from emitting `NoStartRule` in the first case.
I'm not exactly sure the heuristics involved with trying to emit it in the first while avoiding it in the second.

This seems like a decent argument for combining them though, and perhaps doing the checking in the
`Ok(i) => if errors.len() { ... } else { check start rule(...); Ok(...) } branch`

ltratt: Would it make sense to try combining them "before" this PR (if that makes sense)?

ratmice: From the perspective of doing both changes I don't think it makes much difference the order we do the changes in, the overlap is minimal, we're going to have to make 2 sets of changes either order we go.  

Unless what we're worried about is the affect on the enum size. like if for a currently unforeseen reason the second one we encounter an issue that causes us not to want to combine these.  And we don't want to end up with this enum layout it may not make sense to introduce it in the first place, and if that fails, we can think about what other alternatives we might have.

ltratt: > Unless what we're worried about is the affect on the enum size. like if for a currently unforeseen reason the second one we encounter an issue that causes us not to want to combine these.

Yes, that's what I'm (mildly) concerned about.

ratmice: Alright, I'll go take a stab at that first. I'll probably leave this open for now in hopes that the conflicts will be minimal enough that it's still salvageable :pray:

ltratt: Works for me!

ratmice:
I went ahead and rebased this on top of master.

ltratt:
I *think* there's something left unstated in the description which is that this PR keeps trying to process errors even after it's hit the first error?

ratmice:
Yeah, that is a good point, it is fairly implicit in "multiple".

Perhaps something like "When parsing yacc grammars, after encountering a non-fatal error, continue parsing and return all the errors encountered."

ltratt:
Right, so since we do now keep going, we're going to need a lot more tests to test as many of the "keep going" combinations as is practical. This will probably highlight some amusing one-correct-but-now-not assumptions!

ratmice:
> Right, so since we do now keep going, we're going to need a lot more tests to test as many of the "keep going" combinations as is practical. This will probably highlight some amusing one-correct-but-now-not assumptions!

I did 2 of these for `%avoid_insert`, so we could assess the damage and refine the tests before doing them all.
I *think* the `HashMap` -> `IndexMap` thing for storing duplicates could probably clean these up a little, and avoid having to create these temporary `HashSets`?

ltratt:
I'm starting to wonder if we're trying to do two (or more...) PRs worth of work in one PR? It might make more sense to structure this as:

1. A PR (or commit) which does the bare minimum to return `Vec<YaccGrammarError>` but never actually returns more than one `YaccGrammarError`.
2. One more PRs (or commits) which starts to change `return vec![<error>]` to `errors.push(error)`, each time adding a test that something sensible is happening.

What do you think?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1698,6 +1699,76 @@ x"
...
+                ]);
+                assert_eq!(
+                    spans_set,
+                    HashSet::from_iter([
+                        (3, [Span::new(77, 78)].as_slice()),
+                        (2, &[Span::new(103, 104)])
+                    ])
+                )
+            }
+            Err(e) => incorrect_errs!(src, e),
```
ratmice, line 1768: Can actually refine that to something like below, but I feat it is a lot packed into a match case.
```
 if HashSet::<(usize, &[Span])>::from_iter([
                (line_of_offset(src, span.start()), spans.as_slice()),
                (line_of_offset(src, span2.start()), spans2.as_slice()),
            ]) == HashSet::from_iter([
                (2, [Span::new(77, 78)].as_slice()),
                (3, &[Span::new(103, 104)]),
            ]) => {}
```

ratmice:
Sounds like a plan, I was trying to think if it makes sense to do one PR with 2 commits that switch both `YaccGrammarError`, and `LexBuildError`, to `Vec`, to get the ABI stuff out of the way since they are going to be similar changes, but probably best left to separate PRs.

PR #313, opened 2022-06-25
merge YaccParserError and GrammarValidationError

ratmice:
Here is kind of an exploratory patch, which we discussed in #311 
I say exploratory because it hits some rough edges, I had failed to notice there the `sym` field in `GrammarValidationError`,
this causes us 3 issues:

1. `Symbol` isn't `Eq` while `YaccParserError` is.  Used a manually derived `Eq`.
2. `Symbol` has a span, so the errors coming from `GrammarValidationError`, end up with both a span in the symbol,
    and a span in the error.
3. `YaccParserErrorKind` is `Display` and `GrammarValidationErrorKind` is not, I guess we could do better than I did,
   by printing the kind without the associated `Symbol` though.

For my own sanity I went ahead and prefixed all the `GrammarValidationErrorKind`s with `Validation` so, `NoStartRule` becomes `ValidationNoStartRule`, at least as a temporary thing.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -55,20 +55,76 @@ pub enum YaccParserErrorKind {
     DuplicateEPP(Vec<Span>),
     ReachedEOL,
     InvalidString,
+    ValidationNoStartRule,
+    ValidationInvalidStartRule,
+    ValidationUnknownRuleRef,
+    ValidationUnknownToken,
+    ValidationNoPrecForToken,
+    ValidationUnknownEPP,
 }
 
 /// Any error from the Yacc parser returns an instance of this struct.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub struct YaccParserError {
     pub kind: YaccParserErrorKind,
     pub span: Span,
+    pub sym: Option<Symbol>,
 }
 
+impl PartialEq for YaccParserError {
+    fn eq(&self, other: &Self) -> bool {
+        if !(self.kind == other.kind) {
+            return false;
+        }
+        if self.span != other.span {
+            return false;
+        }
+        match (&self.sym, &other.sym) {
+            (Some(sym1), Some(sym2)) => match (sym1, sym2) {
+                (Symbol::Token(t1, span1), Symbol::Token(t2, span2)) => t1 == t2 && span1 == span2,
+                (Symbol::Rule(r1, span1), Symbol::Rule(r2, span2)) => r1 == r2 && span1 == span2,
+                _ => false,
+            },
+            (None, None) => true,
+            _ => false,
+        }
+    }
+}
+
+impl Eq for YaccParserError {}
 impl Error for YaccParserError {}
 
 impl fmt::Display for YaccParserError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.kind)
+        match self.kind {
+            YaccParserErrorKind::ValidationNoStartRule => write!(f, "No start rule specified"),
+            YaccParserErrorKind::ValidationInvalidStartRule => write!(
+                f,
+                "Start rule '{}' does not appear in grammar",
+                self.sym.as_ref().unwrap()
+            ),
+            YaccParserErrorKind::ValidationUnknownRuleRef => write!(
+                f,
+                "Unknown reference to rule '{}'",
+                self.sym.as_ref().unwrap()
+            ),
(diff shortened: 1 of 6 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -55,20 +55,76 @@ pub enum YaccParserErrorKind {
...
+    ValidationUnknownEPP,
 }
 
 /// Any error from the Yacc parser returns an instance of this struct.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub struct YaccParserError {
     pub kind: YaccParserErrorKind,
     pub span: Span,
+    pub sym: Option<Symbol>,
```
ratmice, line 71: So, my thinking is the right thing to do here, might be to just move this `sym` into the `Kind`, so these become e.g. `InvalidStartRule(Symbol)`, all these except `NoStartRule`, (which ignores sym) panic if `sym` is `None`,
so we could drop the `Option` then.

ratmice: So, that would basically fix *1.* and *3.*, one option for fixing *2.* is also getting rid putting the span field on `YaccParserError` moving it too into the variants, basically we'd have 3 cases:

```
IllegalInteger(Span)`
// Lots more Foo(Span)
DuplicateRule(Span, Vec<Span>)`
// Few more with (Span, Vec<Span>)
ValidationInvalidStartRule(Symbol),
// few more with `Symbol`
```

Which leaves `YaccParserError { kind: YaccParserErrorKind }`, at which point we can remove it `YaccParserError`,
and likely rename `YaccParserErrorKind` to an `YaccParserError`.

There is then some trouble because we no longer have a way to get at the span for an arbitrary error, but we wouldn't be duplicating it any longer, which the testsuite relies on.

I assume we'd just add an `fn span() -> Span` either as a trait but probably just as an `impl YaccParserError`.
I'd like to play around with this path, and see what the memory layout looks like.

Probably push something to look at in another branch starting from here, unless it sounds revolting.

ratmice: FWIW, here are a couple of commits implementing the above 2 comments.

https://github.com/ratmice/grmtools/commit/9ce3a49ea66a02d9ef6789c0ec4fc64514b867dd
https://github.com/ratmice/grmtools/commit/4c02011acb12a5fa1e12ad681472c866eddd8503

ratmice: It doesn't really seem terribly great memory layout, at `56`, if we limit it to `(Span, String)` the enum reduces to 48, the size of `Symbol`, It seems plausible that we can reduce it by making `ValidationInvalidStartRule(Rule)`, instead of using `Symbol`.

it may be that e.g. we return `Symbol` because some errors return `Rule` and some `Token`, while we could then use the specific type for each error, removing the need for the tag bits.

ratmice: Does the `(Span, String)` thing via a new `TokenSymbol` and `RuleSymbol`.
Anyhow I think that is as compact as it gets, and at `48` compared to `64` with `YaccGrammarError` + `GrammarValidationError` 

https://github.com/ratmice/grmtools/commit/9b6e932a0a9c8d240775a6fcd48f85676cb8ecfe

ltratt:
This is interesting stuff. Of course, now we're shaving the yak, and finding it has many coats ;) It seems to me that the key problem you're facing is that currently we have:

```
pub struct GrammarValidationError {
    pub kind: GrammarValidationErrorKind,
    pub sym: Option<Symbol>,
}
```

which is horrible because `Symbol` really should be `Span`. So, probably, the first thing we should do is change `pub sym: Option<Symbol>` to `pub sym: Span`. AFAICS the only reason `Option` is there is because there's no obvious place to attach `InvalidStartRule` to. I'd be fine saying the relevant `Span` is (0, 0) for that.

If we do think this is the right thing to do then (and I'm hating myself for typing this...), we should probably open a new PR to first change `Option<Symbol>` to `Span` (assuming that's easy to do! if it's not... we should think again), then come back to this PR.

Any thoughts?

ratmice:
My main thought/note that I would make is that the `String` of `sym`, could go in the `Kind` variant, without any impact on the size of the merged enum, then we can still refer to the name in the output even without the source available.

So we should discuss whether we want to keep the string, and just move it into the variant before moving directly to span?

ratmice:
Well, I just went ahead and made a separate PR, since it seemed like that approach should be pretty smooth, and it seemed to go fairly smoothly, so just going to close this.

PR #314, opened 2022-06-26
merge YaccParserError and GrammarValidationError take 2

ratmice:
Here is basically take 2, I think this is the simplest way to merge these 2 structures, and remove the enum combining them.
I did it over 2 commits.  If we really don't want the `String` maybe we could remove that in a follow up, but it shouldn't affect the size of `YaccParserError` in the 2nd commit.

I'm thinking perhaps this combined structure should be named `YaccGrammarError` rather than `YaccParserError`,
It would be easy to do that in a additional commit with search/replace.

I didn't want to rename it here, because it would add a lot of noise due to the prevalence of `YaccParserError`.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -1,8 +1,8 @@
-use std::{collections::HashMap, error::Error, fmt};
+use std::{collections::HashMap, fmt};
 
 use indexmap::{IndexMap, IndexSet};
 
-use super::Precedence;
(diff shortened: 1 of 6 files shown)


ltratt:
>I'm thinking perhaps this combined structure should be named YaccGrammarError rather than YaccParserError,
It would be easy to do that in a additional commit with search/replace.

I tend to agree this would be a sensible next PR.

This PR seems ready for merging?

ratmice:
I'm happy with it, so if you are too, it does seem so.

PR #315, opened 2022-06-26
Rename yaccparsererror

ratmice:
The first commit is just a search/replace of `YaccParserError` -> `YaccGrammarError`,
since `YaccParserError` now contains the sum of what was the `YaccGrammarError` enum variants.

I wasn't sure If I should move it, and if so, where...
i'd probably think given that it is shared across modules, `errors`, or `yacc/mod.rs`, before `grammar`?

It is easy enough to add another commit to this series if we want to move it, so I just left it for now.


ltratt:
>I wasn't sure If I should move it, and if so, where...

We *could*, but I'm not too worried about it at the moment.

bors r+

PR #316, opened 2022-06-28
minimal changes to return Vec<YaccGrammarError> from parse.

ratmice:
This changes the return value of `YaccParser::parse` from
`YaccGrammarError` to `Vec<YaccGrammarError>, currently
this vector will contain a single `YaccGrammarError`.

Existing tests are changed to expect one error,
callers are changed to now loop over the vector.

And the creation of `DuplicateRule` errors was moved from
`parse_rules` into `parse`.

supercedes #311

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -188,65 +188,71 @@ impl YaccParser {
         }
     }
 
-    pub(crate) fn parse(&mut self) -> YaccResult<usize> {
+    pub(crate) fn parse(&mut self) -> Result<usize, Vec<YaccGrammarError>> {
         // We pass around an index into the *bytes* of self.src. We guarantee that at all times
         // this points to the beginning of a UTF-8 character (since multibyte characters exist, not
         // every byte within the string is also a valid character).
-        let mut i = self.parse_declarations(0)?;
+        let mut i = self.parse_declarations(0).map_err(|e| vec![e]);
         if let Some((orig_span, spans)) = self.duplicate_avoid_insert_spans.iter().next() {
-            return Err(YaccGrammarError {
+            return Err(vec![YaccGrammarError {
                 kind: YaccGrammarErrorKind::DuplicateAvoidInsertDeclaration(spans.clone()),
                 span: *orig_span,
-            });
+            }]);
         }
         if let Some((orig_span, spans)) = self.duplicate_precedence_spans.iter().next() {
-            return Err(YaccGrammarError {
+            return Err(vec![YaccGrammarError {
                 kind: YaccGrammarErrorKind::DuplicatePrecedence(spans.clone()),
                 span: *orig_span,
-            });
+            }]);
         }
         if let Some((orig_span, spans)) = self.duplicate_implicit_token_spans.iter().next() {
-            return Err(YaccGrammarError {
+            return Err(vec![YaccGrammarError {
                 kind: YaccGrammarErrorKind::DuplicateImplicitTokensDeclaration(spans.clone()),
                 span: *orig_span,
-            });
+            }]);
         }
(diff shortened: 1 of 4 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -188,65 +188,71 @@ impl YaccParser {
...
                     .global_actiontype
                     .as_ref()
                     .map(|(_, span)| span)
                     .unwrap(),
-            });
+            }]);
         }
-        i = self.parse_rules(i)?;
-        self.parse_programs(i)
+        i = self.parse_rules(i?).map_err(|e| vec![e]);
```
ltratt, line 248: I think this chunk has been moved to the wrong place? It looks like it should be at line 483?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -360,15 +360,19 @@ where
...
-            let mut line_cache = NewlineCache::new();
-            line_cache.feed(&inc);
-            if let Some((line, column)) =
-                line_cache.byte_to_line_num_and_col_num(&inc, e.span.start())
-            {
-                format!("{} at line {line} column {column}", e)
-            } else {
-                format!("{}", e)
+            let mut s = String::new();
+            for e in e {
```
ltratt, line 364: I'm not sure I like `for e in e` ;)

A possibly bigger problem is that I *think* it leaves one too many `\n` at the end of the string?

Could we simplify to something like:

```rust
errs.map(|e| if let Some(...) = ... { format!(...) } else { format!(...) }).join("\n")
```

[The type checker might need a `collect::<Vec<_>>()` (or similar in) the middle.]

nimbleparse/src/main.rs
```diff
@@ -130,18 +130,20 @@ fn main() {
...
-                writeln!(
-                    stderr(),
-                    "{}: {} at line {line} column {column}",
-                    &yacc_y_path,
-                    &s
-                )
-                .ok();
-            } else {
-                writeln!(stderr(), "{}: {}", &yacc_y_path, &s).ok();
+            for s in s {
```
ltratt, line 133: Again, `s` probably isn't the right variable name for `Err(s)` any more. Presumably `Err(errs)` or something like that might make more sense?

cfgrammar/src/lib/yacc/parser.rs
ratmice: this is "the creation of DuplicateRule errors was moved from parse_rules into parse." part of the commit message,

This movement should be entirely equivalent to what was there, since we move it from right before returning `Ok`,
to right before we check the result returned by the function it is moved from.

There are 2 reasons for it:
1. It allows us to keep all of these checks in the same place.
2. Allows us to avoid switching the return type of `parse_rules`, and allocating a temporary vector for it in the next patch.

nimbleparse/src/main.rs
ratmice: Argh, I meant to go and fix these after getting it all working, but forgot :(

I figured since we're touching every line here, it's also good to rename `s` to `e`,
it seems like maybe it used to be a string at some point.

fixed in edc3d0fbeb9f3711ca3380cef0ea24f4bd4ba660

lrpar/src/lib/ctbuilder.rs
ratmice: fixed in edc3d0fbeb9f3711ca3380cef0ea24f4bd4ba660

cfgrammar/src/lib/yacc/parser.rs
ltratt: Ah, I get it.

ltratt:
Please squash.

ratmice:
Squashed.

PR #317, opened 2022-06-29
Minimal changes to return Vec<LexBuildError>.

ratmice:
While we now return a vector, it currently never contains more than
a single error within that vector, in preparation to eventually return
multiple errors at a time.

In support of this:
* The `LexerDef` trait's `from_str` method, has been changed to
  return a `Result<.., Vec<LexBuildError>` instead of a `LexBuildResult`.
* The deprecated function `build_lex` has been removed.
* Existing tests expecting the current behavior `LexBuildError`
  now expect a vector with a single error.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -306,17 +306,24 @@ where
         let lex_src = read_to_string(&lexerp)?;
         let mut lexerdef: Box<dyn LexerDef<StorageT>> = match self.lexerkind {
             LexerKind::LRNonStreamingLexer => Box::new(
-                LRNonStreamingLexerDef::<LexemeT, StorageT>::from_str(&lex_src).map_err(|e| {
-                    let mut line_cache = NewlineCache::new();
-                    line_cache.feed(&lex_src);
-                    if let Some((line, column)) =
-                        line_cache.byte_to_line_num_and_col_num(&lex_src, e.span.start())
-                    {
-                        format!("{} at line {line} column {column}", e)
-                    } else {
-                        format!("{}", e)
-                    }
-                })?,
+                LRNonStreamingLexerDef::<LexemeT, StorageT>::from_str(&lex_src).map_err(
+                    |errs| {
+                        errs.iter()
+                            .map(|e| {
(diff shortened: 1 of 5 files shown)


ratmice:
I figured it might be good to get SemVer changes over with, before working on individual error sites, and returning multiple errors.
So here is a similar change, as my previous patch except for `LexBuildError`.  I'm not sure i'm happy with how `parse` turned out, but it was the best I could come up with.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -306,17 +306,24 @@ where
...
-                        format!("{} at line {line} column {column}", e)
-                    } else {
-                        format!("{}", e)
-                    }
-                })?,
+                LRNonStreamingLexerDef::<LexemeT, StorageT>::from_str(&lex_src).map_err(
+                    |errs| {
+                        errs.iter()
+                            .map(|e| {
+                                let mut line_cache = NewlineCache::new();
```
ltratt, line 313: It would be good to a) convert this to `NewlineCache::from` (rather than `new` and `feed`) b) hoist this out of the `map` so that we only pay the cost once. But I'd like to do both in a second commit rather than overloading the first commit.

ltratt:
Should we be changing the definition of `LexBuildResult` in this PR? It *might* make things a bit tidier (or not...).

lrlex/src/lib/ctbuilder.rs
ratmice: Should be fixed in f594e65

ratmice:
Squashed/Reordered those 3 commits, which gets rid of any interim changes across commits.

ltratt:
OK to merge?

ratmice:
I think so

ratmice:
Thank you for the review, just wanted to do a quick update, in that this is the last semver affecting change I had wanted to get in, and so everything beyond this should be semver compatible changes to utilize the new return types, which should be easy to split into small chunks and fairly localized changes.

I went ahead and marked off the semver incompatible changes checklist from #300

ltratt:
Thanks for the pointer!

PR #318, opened 2022-07-01
Add start states (in progress)

SMartinScottLogic:


lrlex/src/lib/lexer.rs
```diff
@@ -651,4 +687,115 @@ b 'B'
             .collect::<Vec<_>>();
         assert_eq!(anonymous_rules[0].name_span, Span::new(21, 21));
     }
+
+    #[test]
+    fn test_token_start_states() {
+        let src = "%x EXCLUSIVE_START
```
ltratt, line 693: What does `%exclusive_start` mean vs. `%inclusive_start`?

ratmice: This is perhaps the feature I miss the most, so thank you for working on it Simon!
The flex documentation is here:
https://westes.github.io/flex/manual/Start-Conditions.html#Start-Conditions

The short answer is that the inclusive state `<Inc>` state matches *both* `b` and  `c`, while the exclusive only matches `a`
```
%x Ex
%s Inc
<Ex>a "A"
<Inc>b "B"
c "C"
```

It is really useful for lexing things like SML style nested comments `(* (* commented out comment *) outer comment *)`
and probably rust-like raw string literals `r#"""#`, `r##""""##`.

I haven't used it with a lexer that doesn't have actions, e.g. in flex you can see that switching between states is generally done in action code.  I haven't thought too much about it, but given the actionless style of lrlex, in both the cases I mentioned it seems there is a counter number of `#` signs and the number of open comments.

E.g. we would re-enter the initial state after seeing the appropriate number of '#'s preceeded by a quote, or a balanced number of `(*` and `*)`.

I guess I struggle somewhat to see transformations for doing this in an actionless way.  Towards the end of that flex documentation it mentions `yy_push_state`/`yy_pop_state`.

In the calc example in your commit, there is these snippets below switching states,
```
/\*            <COMMENT>;
<COMMENT>\*/   <INITIAL>;
```

I would be kind of curious about a push/pop like version of these, with `+` for push and `-` for pop.
```
/\*            +<COMMENT>;
<COMMENT>\*/   -<COMMENT>;
```

And whether something like that would allow for a translation of these generally counter dependent state exits.

ltratt: Ah! I have never used these. AIUI these are part of "normal" lex and aren't a flex extension? I'm certainly happy to improve compatibility with lex!

ratmice: I'm uncertain about that, some reading (o'reilly lex & yacc) seems to indicate that at least `exclusive` states originate with *flex*, and insinuates that some have `inclusive` without `exclusive`.  The *POSIX* lex specification doesn't seem to describe either `inclusive` or `exclusive` as far as I can tell.

ltratt: Right. I've long thought (and there was a partial PR from someone to do this) that lex should, like lrpar, support different flavours of lexer. This might be a good candidate for that. There again, lexing is simpler than parsing, and given we don't formally support lex's regex language, pretending that we have greater compatibility than we actually have might be a fool's mission!

SMartinScottLogic: Still some code clean-up work I want to do on this work (plus rebasing onto most recent commits in trunk).

Reading https://pubs.opengroup.org/onlinepubs/007904875/utilities/lex.html, it mentions both inclusive and exclusive start states (or start conditions, they appear to be used interchangeably).

Would it be worth me adding some documentation into this Pull Request - how inclusive and exclusive start states work, with an example?

ltratt: Pointing to the definitions elsewhere of inclusive/exclusive is fine with me! We will, of course, need tests that cover all the possibilities so that any code proposed for merging does what we think it should :)

[I really need to duplicate this for grmtools, but while I'm here: the process we use for PRs is documented at https://ykjit.github.io/ykdocs/contributing/prs.html. Please let me know if things aren't clear!]

SMartinScottLogic: I believe this is ready for a review - I've added the tests I consider necessary, and updated the documentation.
I've been consulting the flex source (especially https://github.com/westes/flex/blob/master/src/scan.l), and it doesn't explicitly include **%start** for declaring start conditions (in the definitions section) - should I adjust to include the possibilities as in http://dinosaur.compilertools.net/lex/index.html (Section 10)?

ltratt:
It would be great to get comments from @ratmice on this too (as he clearly understands this feature better than I do!).

lrlex/src/lib/parser.rs
```diff
@@ -93,7 +146,26 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+            match line[rspace + 1..].find('>') {
+                Some(l) => {
+                    let state = self.get_start_state_by_name(
+                        i + rspace + 1,
+                        &line[rspace + 2..rspace + 1 + l],
+                    )?;
+                    target_state = Some(state.id);
+                    &line[rspace + 1 + l + 1..]
+                }
+                None => {
```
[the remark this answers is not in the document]
ltratt: Definitely more in keeping to catch the error :)

ltratt:
Hello @SMartinScottLogic, I guess this is ready for rereview? I've unresolved conversations as it's the reviewer's job to say whether things are resolved or not (I now realise that https://ykjit.github.io/ykdocs/contributing/prs.html doesn't make that clear -- mea culpa!). For each comment could you point us at a part(s) of the commit(s) that address the comment? That way it's easiest for us reviewers to understand what's going on. Thanks!

SMartinScottLogic:
Sure - sorry for the error - on my other projects, we use 'Tasks' in BitBucket, and check them off as we've 'actioned'

lrlex/src/lib/mod.rs
```diff
@@ -62,6 +65,9 @@ impl fmt::Display for LexBuildError {
             LexErrorKind::MissingSpace => "Rule is missing a space",
             LexErrorKind::InvalidName => "Invalid rule name",
             LexErrorKind::UnknownStartState => "Start state not known",
+            LexErrorKind::DuplicateStartState => "Start state already exists",
+            LexErrorKind::InvalidStartState => "Invalid start state",
+            LexErrorKind::InvalidStartStateName => "Invalid start state",
```
ltratt, line 70: Should this be `"Invalid start state name"`?

lrlex/src/lib/parser.rs
```diff
@@ -1,14 +1,31 @@
...
 use try_from::TryFrom;
 
 use crate::{lexer::Rule, LexBuildError, LexBuildResult, LexErrorKind};
 
 type LexInternalBuildResult<T> = Result<T, LexBuildError>;
+
+lazy_static! {
+    static ref RE_START_STATE_NAME: Regex = Regex::new(r"^[a-zA-Z][a-zA-Z0-9_.]*$").unwrap();
+}
+const INITIAL_START_STATE_NAME: &str = "INITIAL";
```
ltratt, line 14: Is it part of the `lex` specification that the initial state is called `INITIAL` or is it normally anonymous or ... ?

lrlex/src/lib/parser.rs
```diff
@@ -96,21 +113,27 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+                for name in start_states {
                     let id = self.start_states.len();
                     let start_state = StartState::new(id, name, true);
                     self.start_states.push(start_state);
                 }
                 Ok(i + line_len)
             }
-            "%s" => {
-                for name in self.src[i + declaration_len..i + line_len]
+            _ if (declaration.to_lowercase()=="%s" || declaration.to_lowercase()=="%start") => {
```
ltratt, line 129: I *think* the other declarations are case sensitive (or we treat them as such). Does lex treat them case sensitive/insensitive?

lrlex/src/lib/parser.rs
```diff
@@ -121,6 +144,26 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+        self.validate_start_state_name(off, name)?;
+        if self.start_states.iter().any(|state| state.name == name) {
+            Err(self.mk_error(LexErrorKind::DuplicateStartState, off))
+        } else {
+            Ok(name)
+        }
+    }
+
+    fn validate_start_state_name(&self, off: usize, name: &str) -> LexInternalBuildResult<()> {
+        match RE_START_STATE_NAME.is_match(name) {
```
ltratt, line 161: I think this would be clearer as just `if RE_START_STATE_NAME.is_match(name)` (rather than a `match`).

ltratt:
@ratmice have you had a look at https://github.com/softdevteam/grmtools/pull/318/commits/6131713664496f6ffaa0ff25c439d3657322388d? Any comments?

lrlex/src/lib/parser.rs
ratmice: Yeah, it is typically called initial, the specification at https://pubs.opengroup.org/onlinepubs/007904875/utilities/lex.html says

> The initial state is indicated by the digit '0' or the token INITIAL.

ratmice:
Yeah, I've had a look, but haven't noticed anything beyond what has been mentioned

lrlex/src/lib/parser.rs
ratmice: So IIUC the above correctly I believe this should allow digits in the first character?

ltratt: Good question, it does look like it should. The easiest thing is probably to try this out on "real" C-lex.

lrlex/src/lib/parser.rs
SMartinScottLogic: According to https://pubs.opengroup.org/onlinepubs/007904875/utilities/lex.html#tag_04_73_13_01

> ...any line beginning with a '%' (percent sign) character and followed by an alphanumeric word beginning with either 's' or 'S' shall define a set of start conditions. Any line beginning with a '%' followed by a word beginning with either 'x' or 'X' shall define a set of exclusive start conditions. When the generated scanner is in a %s state, patterns with no state specified shall be also active; in a %x state, such patterns shall not be active. The rest of the line, after the first word, shall be considered to be one or more <blank>-separated names of start conditions. 

I hadn't seen this before - if you agree, I'll adjust the logic to match?

lrlex/src/lib/mod.rs
SMartinScottLogic: Yes, I'll fix

lrlex/src/lib/parser.rs
SMartinScottLogic: My understanding (may have issues) is that a start state is either:
1. A name (so must be a valid c identifier) - I deliberately omitted non-alpha start characters to allow 'special' names at a later date
2. Purely numeric

Thus, the regex would be ^([0-9]+|[a-zA-Z][a-zA-Z0-9_.]*)$
But in the numeric case, I'd prefer to have special handling (so the id is aligned with the parsed result of the input state name)
I feel this could be done as a later enhancement, especially as I've never used numbered start states - they feel much harder to work with (as with any 'magic' numbers)

lrlex/src/lib/ctbuilder.rs
```diff
@@ -419,26 +430,32 @@ pub fn lexerdef() -> {lexerdef_type} {{
...
+            };
             let n_span = format!(
                 "::cfgrammar::Span::new({}, {})",
                 r.name_span.start(),
                 r.name_span.end()
             );
             outs.push_str(&format!(
                 "
-        Rule::new({}, {}, {}, \"{}\".to_string()).unwrap(),",
+        Rule::new({}, {}, {}, \"{}\".to_string(), {:?}.to_vec(), {}).unwrap(),",
```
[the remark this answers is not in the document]
SMartinScottLogic: I agree it's a little strange, but I couldn't come up with a clean way to do equivalent, and as you say, it's consistent with the existing code.

lrlex/src/lib/lexer.rs
```diff
@@ -29,6 +30,8 @@ pub struct Rule<StorageT> {
     pub name_span: Span,
     pub(super) re_str: String,
     re: Regex,
+    pub start_states: Vec<usize>,
```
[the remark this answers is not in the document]
SMartinScottLogic: Added docstrings in https://github.com/softdevteam/grmtools/pull/318/commits/6131713664496f6ffaa0ff25c439d3657322388d

lrlex/src/lib/lexer.rs
```diff
@@ -29,6 +30,8 @@ pub struct Rule<StorageT> {
...
+    pub target_state: Option<usize>,
```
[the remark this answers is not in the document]
SMartinScottLogic: From docs in https://github.com/softdevteam/grmtools/pull/318/commits/6131713664496f6ffaa0ff25c439d3657322388d:

> If Some(state), successful matching of this rule will cause the current start condition of the lexer to be changed to the enclosed value.
> If none, successful matching causes no change to the current start condition.

lrlex/src/lib/lexer.rs
```diff
@@ -217,11 +237,22 @@ impl<
     ) -> LRNonStreamingLexer<'lexer, 'input, LexemeT, StorageT> {
         let mut lexemes = vec![];
         let mut i = 0;
+        let state = self.get_start_state_by_name("INITIAL");
```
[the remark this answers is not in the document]
SMartinScottLogic: Pulled the constant out in https://github.com/softdevteam/grmtools/pull/318/commits/6131713664496f6ffaa0ff25c439d3657322388d#diff-44796c04a2099ba8168addfa723b939dfc1f097419459ba8dfccc4b83da68eb6

From https://pubs.opengroup.org/onlinepubs/007904875/utilities/lex.html:
> The initial state is indicated by the digit '0' or the token INITIAL.

lrlex/src/lib/parser.rs
SMartinScottLogic: I agree - will modify to an if + else

lrlex/src/lib/parser.rs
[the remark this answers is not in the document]
SMartinScottLogic: Updated to error, under https://github.com/softdevteam/grmtools/pull/318/commits/6131713664496f6ffaa0ff25c439d3657322388d#diff-44796c04a2099ba8168addfa723b939dfc1f097419459ba8dfccc4b83da68eb6R203

lrlex/src/lib/parser.rs
```diff
@@ -125,18 +197,59 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+    fn parse_start_states<'a>(
+        &self,
+        off: usize,
+        re_str: &'a str,
+    ) -> LexInternalBuildResult<(Vec<usize>, &'a str)> {
+        if !re_str.starts_with('<') {
+            Ok((vec![], re_str))
+        } else {
+            match re_str.find('>') {
+                None => Ok((vec![], re_str)),
```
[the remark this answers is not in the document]
SMartinScottLogic: Updated to error, c.f. https://github.com/softdevteam/grmtools/pull/318/commits/6131713664496f6ffaa0ff25c439d3657322388d#diff-44796c04a2099ba8168addfa723b939dfc1f097419459ba8dfccc4b83da68eb6R268

lrlex/src/lib/lexer.rs
ltratt: > should I adjust to include the possibilities as in http://dinosaur.compilertools.net/lex/index.html (Section 10)?

Oops, sorry, I missed this question! I don't think a PR needs to do everything in one go: making things better is enough :) I would, though, like to make sure that we don't do anything that's in conflict with the "standard" -- but I'm happy for you to say whether you think the PR is consistent with the spec or not.

lrlex/src/lib/ctbuilder.rs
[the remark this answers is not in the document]
ltratt: Agreed. I've raised an issue for this https://github.com/softdevteam/grmtools/issues/323, but let's not worry about it in this PR.

lrlex/src/lib/lexer.rs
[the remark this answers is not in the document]
ltratt: Got it!

lrlex/src/lib/parser.rs
ltratt: I'm fine with not handling numeric states in this PR (but I'd ask that we raise an issue to remind us that this is a limitation).

lrlex/src/lib/parser.rs
ltratt: Works for me!

ltratt:
Added a few comments -- this is progressing nicely!

lrlex/src/lib/parser.rs
SMartinScottLogic: Updated in https://github.com/softdevteam/grmtools/pull/318/commits/6ace9f373b657ebe18660cf0c5d0bd2010cc7ba9

lrlex/src/lib/mod.rs
SMartinScottLogic: Fixed under https://github.com/softdevteam/grmtools/pull/318/commits/9bf38da33e6078106a1963382ecd6f09eb4cdd7a

lrlex/src/lib/parser.rs
SMartinScottLogic: Updated in https://github.com/softdevteam/grmtools/pull/318/commits/c91dab0b6433d3650ab08dae6172849d7d569444

lrlex/src/lib/parser.rs
SMartinScottLogic: Raised an issue to track this enhancement: #325

SMartinScottLogic:
I think I've now addressed all concerns raised.

lrlex/src/lib/parser.rs
ltratt: Looks good, but could we add a negative test or two for names that shouldn't be accepted? That'll stop us accidentally mucking this up in the future!

lrlex/src/lib/parser.rs
ltratt: Thanks!

ltratt:
One minor comment about a few negative tests, but we're close!

lrlex/src/lib/parser.rs
SMartinScottLogic: See tests in c2348a3a6118a0c797d7202a7b9de9b578b8229c

ltratt:
bors try

ltratt:
We're almost there. You'll need to push a new `cargo fmt` commit first so that I can see if bors will accept the PR.

SMartinScottLogic:
Code reformatted, under https://github.com/softdevteam/grmtools/pull/318/commits/ab71b6d07fdf49486678941afee6c8f82be15f6f

ltratt:
bors try

ltratt:
Please squash (this doesn't mean you necessarily have to squash down to 1 commit, but simply that the commits need to be tidied up into a coherent subset of commits).

SMartinScottLogic:
Squashed down to 1 commit in https://github.com/softdevteam/grmtools/commit/4b680254775bbede702cca485385abf2299e7bf0.

Cleanest as the code changes weren't tightly coupled to enhancements (especially in the beginning)

ltratt:
Thanks!

bors r+

PR #319, opened 2022-07-06
add a YaccResult

ratmice:
It occurred to me we might want to the same internal/public type alias for errors as we do with `LexBuildResult`, for *yacc*.

I'm not sure if this is absolutely necessary, e.g. `YaccResult` is currently private unlike how `LexBuildResult` was already public.
But perhaps it is worthwhile having the crates to conform to similar patterns.

I don't believe that this should conflict with anything in the branch of @SMartinScottLogic , which seems fairly isolated to *lrlex*.
But if in doubt deference to that work, since this is mostly search/replace, I just wanted to bring it up while I was thinking about it, so it didn't slip through the cracks.

### DIFF
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -6,7 +6,11 @@ use serde::{Deserialize, Serialize};
 use vob::Vob;
 
 use super::{
-    ast, firsts::YaccFirsts, follows::YaccFollows, parser::YaccParser, YaccGrammarError, YaccKind,
+    ast,
+    firsts::YaccFirsts,
+    follows::YaccFollows,
+    parser::{YaccGrammarResult, YaccParser},
+    YaccKind,
 };
 use crate::{PIdx, RIdx, SIdx, Span, Symbol, TIdx};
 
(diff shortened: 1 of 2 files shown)


ratmice:
I could rename it if you like,  unsure about the internal one `YaccInternalGrammarResult` is only saving like 2 characters over `Result<_, YaccGrammarError>`, which might be the purpose, and why the existing name is so short.

ltratt:
On reflection, why don't we delete the current `YaccResult` first and expand it to `Result<T, YaccGrammarError>`? The short form doesn't really help much.

I think I like `YaccGrammarResult` as a name (and as a `type`) because it slightly abstracts away from the `Vec` aspect, so if we ever find that we need to tweak that, it will cause slightly less breakage in user code.

ratmice:
Sounds good to me, so I did that. The commit messages are a bit mislabeled until I squash the others out.

ltratt:
Please squash.

ratmice:
Squashed.

ltratt:
Silly question possibly: is https://github.com/softdevteam/grmtools/pull/319/commits/69fa93fedeca2e613d819e36c75eab2f1a7cc630 semver compatible? I *think* it is but I might be wrong! Also, could you force push updates to the commit messages to add a bit of motivation about why we want to do these things? I always find that helpful!

ratmice:
> Silly question possibly: is [69fa93f](https://github.com/softdevteam/grmtools/commit/69fa93fedeca2e613d819e36c75eab2f1a7cc630) semver compatible? I _think_ it is but I might be wrong!

I *think* so too, but there are cases where adding a type alias is *not* semver compatible (beyond naming clashes, e.g. replacing an enum with an alias to an enum, perhaps something involving where types since they get ignored for aliases), so it is definitely not *generally* true.

I had actually wondered this too earlier, and looked but couldn't find anything to confirm or deny.

ratmice:
Forgot to mention, Force pushed again with commentary in the commit messages.

ltratt:
Agreed (and thanks for the new commit messages!).

bors r+

PR #320, opened 2022-07-07
Rename cfgrammar::yacc::ast::Symbol to ASTSymbol.

ratmice:
Previously we had both `cfgrammar::Symbol` and
`cfgrammar::yacc::ast::Symbol`. The rename is intended to reduce
confusion this could cause.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -40,23 +40,23 @@ pub struct Rule {
 #[derive(Debug)]
 #[cfg_attr(test, derive(Eq, PartialEq))]
 pub struct Production {
-    pub symbols: Vec<Symbol>,
+    pub symbols: Vec<ASTSymbol>,
     pub precedence: Option<String>,
     pub action: Option<String>,
 }
 
 #[derive(Clone, Debug)]
 #[cfg_attr(test, derive(Eq, PartialEq))]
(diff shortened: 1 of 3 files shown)


ratmice:
I remembered that one of the big confusions that caused some havoc in my patch series were these different `Symbol` types, and that we were deriving and using `Hash`/`Eq` on one, leading me to think we needed to keep these trait impls intact on the other.

Since this is a semver incompatible change, it seemed like a good time to make some kind of decision about what to do here.
Perhaps a different approach would be to just use `ast::Symbol` everywhere?

ltratt:
I'm not convinced by this one, to be honest. Conceptually there could be several different kinds of grammars in grmtools, and they'll all have some some sort of "Symbol" concept. If the struct name was to be unique it would have to be YaccASTSymbol, which is a mouthful. I agree with you that the clash of `Symbol` isn't my favourite, but I think that Rust's module system gives us a nice way of disambiguating them.

ratmice:
Understandable, I just wanted to throw the idea out there while we were already breaking semver, it seemed a good time to consider it.

I'll go ahead and just close this since we can switch to `ast::Symbol` at any time without breaking semver,
If that is what you meant by using Rust's module system to distinguish them?

ltratt:
>  If that is what you meant by using Rust's module system to distinguish them?

Exactly!

PR #321, opened 2022-07-10
Return multiple YaccGrammarErrors in YaccGrammarResult.

ratmice:
So these are an initial pair of commits for returning multiple `YaccGrammarError` in the vector of `YaccGrammarResult`,
It goes through the last 2 errors (from bottom up) and adds test cases for these.

These two patches cover all the changes to flow control, what is left is just calls to `errs.extend(...)` and test additions for the other `YaccGrammarErrorKind`s. And should be similar to the test additions and calls to `es.extend(...)`  for `DuplicateRule` and `DuplicateActiontypeDeclaration` herein.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -193,7 +193,9 @@ impl YaccParser {
         // We pass around an index into the *bytes* of self.src. We guarantee that at all times
         // this points to the beginning of a UTF-8 character (since multibyte characters exist, not
         // every byte within the string is also a valid character).
-        let i = self.parse_declarations(0).map_err(|e| vec![e]);
+        let mut r = self.parse_declarations(0);
+        // As a side-effect the above call to `parse_declarations` may have added items to `self.duplicate_*`.
+        // Convert into errors, while continuing to process the file as much as possible.
         if let Some((orig_span, spans)) = self.duplicate_avoid_insert_spans.iter().next() {
             return Err(vec![YaccGrammarError {
                 kind: YaccGrammarErrorKind::DuplicateAvoidInsertDeclaration(spans.clone()),
@@ -236,17 +238,23 @@ impl YaccParser {
                 span: *orig_span,
             }]);
         }
-        if let Some(spans) = &self.duplicate_actiontype_declarations {
-            return Err(vec![YaccGrammarError {
+        errs.extend(self.duplicate_actiontype_declarations.iter().map(|spans| {
+            YaccGrammarError {
                 kind: YaccGrammarErrorKind::DuplicateActiontypeDeclaration(spans.clone()),
                 span: *self
                     .global_actiontype
                     .as_ref()
                     .map(|(_, span)| span)
                     .unwrap(),
-            }]);
-        }
-        let r = self.parse_rules(i?);
+            }
+        }));
+        r = self.parse_rules(match r {
+            Ok(i) => i,
+            Err(e) => {
+                errs.push(e);
+                return Err(errs);
+            }
+        });
         // As a side-effect the call to `parse_rules` may add items to self.duplicate_rule_spans.
         // Convert into errors, while continuing to process the file as much as possible.
         errs.extend(
@@ -2200,6 +2208,41 @@ x"
             ) if line_of_offset(src, span.start()) == 2 => {
                 assert_eq!(spans, &[Span::new(46, 48), Span::new(70, 72)])
             }
+            Err(e) => incorrect_errs!(src, e),
+        }
+    }
+
+    #[test]
+    fn test_duplicate_avoid_insert_and_premature_end() {
+        let src = "
+         %actiontype T1
+         %actiontype T2
+         %actiontype T3";
+        match parse(
+            YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
+            src,
+        )
+        .as_ref()
+        .map_err(Vec::as_slice)
+        {
+            Ok(_) => panic!(),
+            Err(e)
+                if check_errors(
+                    e,
+                    &[
+                        YaccGrammarError {
+                            kind: YaccGrammarErrorKind::DuplicateActiontypeDeclaration(vec![
+                                Span::new(46, 48),
+                                Span::new(70, 72),
+                            ]),
+                            span: Span::new(22, 24),
+                        },
+                        YaccGrammarError {
+                            kind: YaccGrammarErrorKind::PrematureEnd,
+                            span: Span::new(71, 71),
+                        },
+                    ],
+                ) => {}
             Err(e) => incorrect_errs!(src, e),
         }
     }
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -913,6 +913,12 @@ mod test {
         s[..off].lines().count()
     }
 
+    /// Helper function which tests for equivalence between slices containing YaccGrammarErrors.
+    /// Does not assume that the slices are in the same order.
+    /// `errs` is the `&[YaccGrammarError]` obtained from a call to parse
+    /// `errs_expect` is the expected result of that call
+    /// Returns `false` if `errs` contains duplicates or the slices contain different
+    /// elements using an unordered comparison.
     fn check_errors(errs: &[YaccGrammarError], errs_expect: &[YaccGrammarError]) -> bool {
         let es = errs.iter().collect::<HashSet<&YaccGrammarError>>();
         let es_expect = errs_expect.iter().collect::<HashSet<&YaccGrammarError>>();
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -20,7 +20,7 @@ use super::{
 };
 
 /// The various different possible Yacc parser errors.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Hash)]
 pub enum YaccGrammarErrorKind {
     IllegalInteger,
     IllegalName,
@@ -64,7 +64,7 @@ pub enum YaccGrammarErrorKind {
 }
 
 /// Any error from the Yacc parser returns an instance of this struct.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Hash)]
 pub struct YaccGrammarError {
     pub kind: YaccGrammarErrorKind,
     pub span: Span,
@@ -189,10 +189,11 @@ impl YaccParser {
     }
 
     pub(crate) fn parse(&mut self) -> YaccGrammarResult<usize> {
+        let mut errs = Vec::new();
         // We pass around an index into the *bytes* of self.src. We guarantee that at all times
         // this points to the beginning of a UTF-8 character (since multibyte characters exist, not
         // every byte within the string is also a valid character).
-        let mut i = self.parse_declarations(0).map_err(|e| vec![e]);
+        let i = self.parse_declarations(0).map_err(|e| vec![e]);
         if let Some((orig_span, spans)) = self.duplicate_avoid_insert_spans.iter().next() {
             return Err(vec![YaccGrammarError {
                 kind: YaccGrammarErrorKind::DuplicateAvoidInsertDeclaration(spans.clone()),
@@ -245,14 +246,32 @@ impl YaccParser {
                     .unwrap(),
             }]);
         }
-        i = self.parse_rules(i?).map_err(|e| vec![e]);
-        if let Some((orig_span, spans)) = self.duplicate_rule_spans.iter().next() {
-            return Err(vec![YaccGrammarError {
-                kind: YaccGrammarErrorKind::DuplicateRule(spans.clone()),
-                span: *orig_span,
-            }]);
+        let r = self.parse_rules(i?);
+        // As a side-effect the call to `parse_rules` may add items to self.duplicate_rule_spans.
+        // Convert into errors, while continuing to process the file as much as possible.
+        errs.extend(
+            self.duplicate_rule_spans
+                .iter()
+                .map(|(orig_span, spans)| YaccGrammarError {
+                    kind: YaccGrammarErrorKind::DuplicateRule(spans.clone()),
+                    span: *orig_span,
+                }),
+        );
+        let i = match r {
+            Err(e) => {
+                errs.push(e);
+                return Err(errs);
+            }
+            Ok(i) => i,
+        };
+        match self.parse_programs(i) {
+            Ok(i) if errs.is_empty() => Ok(i),
+            Err(e) => {
+                errs.push(e);
+                Err(errs)
+            }
+            _ => Err(errs),
         }
-        self.parse_programs(i?).map_err(|e| vec![e])
     }
 
     pub(crate) fn ast(self) -> GrammarAST {
@@ -859,6 +878,7 @@ mod test {
         },
         Span, YaccGrammarError, YaccGrammarErrorKind, YaccParser,
     };
+    use std::collections::HashSet;
 
     fn parse(yacc_kind: YaccKind, s: &str) -> Result<GrammarAST, Vec<YaccGrammarError>> {
         let mut yp = YaccParser::new(yacc_kind, s.to_string());
@@ -883,6 +903,12 @@ mod test {
 
     fn line_of_offset(s: &str, off: usize) -> usize {
         s[..off].lines().count()
+    }
+
+    fn check_errors(errs: &[YaccGrammarError], errs_expect: &[YaccGrammarError]) -> bool {
+        let es = errs.iter().collect::<HashSet<&YaccGrammarError>>();
+        let es_expect = errs_expect.iter().collect::<HashSet<&YaccGrammarError>>();
+        errs.len() == es.len() && es == es_expect
     }
 
     macro_rules! incorrect_errs {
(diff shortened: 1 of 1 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -245,14 +246,32 @@ impl YaccParser {
...
-        i = self.parse_rules(i?).map_err(|e| vec![e]);
-        if let Some((orig_span, spans)) = self.duplicate_rule_spans.iter().next() {
-            return Err(vec![YaccGrammarError {
-                kind: YaccGrammarErrorKind::DuplicateRule(spans.clone()),
-                span: *orig_span,
-            }]);
+        let r = self.parse_rules(i?);
+        // As a side-effect the call to `parse_rules` may add items to self.duplicate_rule_spans.
+        // Convert into errors, while continuing to process the file as much as possible.
+        errs.extend(
```
ltratt, line 260: Is it just me or is `errs` always empty at line 250?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -885,6 +905,12 @@ mod test {
         s[..off].lines().count()
     }
 
+    fn check_errors(errs: &[YaccGrammarError], errs_expect: &[YaccGrammarError]) -> bool {
```
ltratt, line 916: Function needs a doc string.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yes, but in subsequent patches this will change e.g. patch 2 moves errs being empty up to line 241, by the final patch it will be populated at line 197.

ltratt: You'll have to forgive me, as I'm not very good at imaging commits that don't exist yet :)

This mix of local variables and stored-in-attribute errors feels like an accident waiting to happen. One possibility (off the top of my head!) is that we get rid of `duplicate_rule_spans` and friends as explicit attributes and just have `errors`. When you detect a duplicate rule span, you first of all search `errors` for an existing duplicate rule span error for that name; if none exists, you append to the end. I *think* that works, and would be simpler -- any thoughts?

ratmice: Yeah, i've thought of it too, the 2 main difficulties the array we want to append to is within the enum structure variaint.
So we need to reach into the specific variant to pull that out to append to it, it makes it a bit difficult to generalize.

Additionally, we either need another kind-like-type to do the lookup, like `YaccGrammarErrorSort::DuplicateRule` with *only* unitary variants, or otherwise construct a `YaccGrammarError` whose kind has an empty or one element `Vec<Span>` and do a partial comparison using `std::mem::discriminant` to ignore any spans therein, but either way we have to allocate a vector just to append to the existing one.  

We can avoid that allocation by using the `YaccGrammarErrorSort` unitary variant to do lookup, alternately if we just use the unitary variants entirely and move the `spans` field to `Option<Vec<Span>>` inside the `YaccGrammarError`, we avoid having two types one type `Sort` for lookup and another `Kind` for holding `spans`.

These are basically the aspects which make it difficult to generalize so we don't have to do the entire lookup + insert or append hard coded at every error site with match statements and all.

cfgrammar/src/lib/yacc/parser.rs
ratmice: I went ahead and pushed a fix for this, but it will likely be impacted by the field removal of `duplicate_*` discussion,
since this is just avoiding ordering instability.  If we switch to an `errors` field, that will likely go away.

cfgrammar/src/lib/yacc/parser.rs
ltratt: I'm not totally certain I've understood. AFAICS all we need is something like:

```rust
enum GrammarError {
  DuplicateSpan{name: String, spans: Vec<Span>}
  ...
}

struct S {
  errors: Vec<GrammarError>
}

impl S {
  fn parse_blah(&mut self, ...) {
    // found dup named n
    for e in self.errors.iter_mut() {
      match e {
        DuplicateSpan{name, ref mut spans} if name = n => spans.push(...)
      }
    }
  }
}
```

but perhaps I've misunderstood something!

ratmice: I don't think you've misunderstood anything, I think it is probably just personal preference, where I would heavily lean towards something like
```
// found a dup 
self.add_duplicate_occurrence(DuplicateSpan, orig_span, dupe_span);
```
I'm somewhat predisposed to trying avoid adding these for loops everywhere.
I guess what I was trying to say, is the current situation makes the creation of a function like `add_duplicate_occurrence`,
awkward.

```
add_duplicate_occurrence(DuplicateSpan(vec![dup_span]), orig_span);
// This isn't particularly elegant but would avoid for loops everywhere
fn add_duplicate_occurrence(ekind, span) {
  for e in self.errors.iter_mut() {
     if std::mem::discriminant(e.kind) == std::mem::discriminant(ekind) && orig_span == e.span {
        match (ekind, e.kind) { 
          (Duplicate_spans(new_spans),  DuplicateSpans(spans)) => spans.extend(new_spans),
          ...
        }
     }
  }
}
```
If you're fine with adding for loops everywhere I don't think there are any problems wrong with your approach.

In the spirit of further brainstorming, one other thing we still along the same lines as our current code,
but perhaps less error prone than what we have, is just moving these variables into their own structures, one per error producer.

```
struct YaccGrammarDeclarationsErrorContext {
    duplicate_avoid_insert_spans: HashMap<Span, Vec<Span>>,
    duplicate_precedence_spans: HashMap<Span, Vec<Span>>,
    duplicate_implicit_token_spans: HashMap<Span, Vec<Span>>,
    duplicate_expect_declarations: Option<(Span, Vec<Span>)>,
    duplicate_expectrr_declarations: Option<(Span, Vec<Span>)>,
    duplicate_start_declarations: Option<(Span, Vec<Span>)>,
    duplicate_actiontype_declarations: Option<Vec<Span>>,
    duplicate_epp_declarations: HashMap<Span, Vec<Span>>,
}
fn parse_declarations(&mut self, mut i: usize, ctxt: &mut YaccGrammarDeclarationsErrorContext) -> Result<usize, YaccGrammarError> {
   ...
}
#[derive(Default)]
struct YaccGrammarRulesErrorContext {
    duplicate_rule_spans: HashMap<Span, Vec<Span>>,
}

fn parse_rules(&mut self, mut i: usize, ctxt: &mut YaccGrammarRulesErrorContext) -> Result<usize, YaccGrammarError> {
  ...
}
```

These types would then have code for producing something like `fn into_yacc_grammar_error(Self) -> Iterator<Item=YaccGrammarError>` which we chain together.  Anyhow just figured it was worth exploring ideas that are more refinements on what we have.

ratmice: I went ahead and did it for one (DuplicateActionTypeDeclaration) to just get an idea of generally how much code per error occurrence adding the for loops manually at each site causes, because we still need to insert it if it isn't found...
Seems like it is roughly 20 loc per error occurrence site.

```
                        let mut added = false;
                        for e in errs.iter_mut() {
                            match &mut e.kind {
                                YaccGrammarErrorKind::DuplicateActiontypeDeclaration(spans) => {
                                    spans.push(span);
                                    added = true;
                                    break;
                                }
                                _ => continue,
                            }
                        }
                        if !added {
                            errs.push(YaccGrammarError {
                                kind: YaccGrammarErrorKind::DuplicateActiontypeDeclaration(vec![
                                    span,
                                ]),
                                span: *self
                                    .global_actiontype
                                    .as_ref()
                                    .map(|(_, span)| span)
                                    .unwrap(),
                            })
                        }
```

ltratt: >  would heavily lean towards something like
>
>// found a dup 
>self.add_duplicate_occurrence(DuplicateSpan, orig_span, dupe_span);

Agreed.

Most errors would, of course, just be a simple `push`: it's only those (like duplicate names) where we'd want to do some searching.

>but perhaps less error prone than what we have, is just moving these variables into their own structures, one per error producer.

Interesting idea! I dunno, though, the difficulty with this is that later you want to merge them into some sort of sorted list by location, and that looks like a copy and paste nightmare! But perhaps a `From` conversion would make it OK? In an ideal world we'd try both and see which sucks less, but that's a lot of work!

ratmice: That is a good point, that suggestion indeed does nothing to alleviate the ordering issues.

ratmice: Here is a different `YaccGrammarErrorKind` layout which manages to keep it the same size,
But I think should make it easier to write a nicer API wrapping the lookup/insertion code.
Curious if you think that might be a viable solution, but sadly means revisiting the error kind :(

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e677c97a5873e81306906936c755b6a6

ltratt: I'm not sure I understand the need for such a change in the `enum`(s), but that might be me being dense! Care to motivate?

ltratt: Ah, you're trying to make a generic `add_duplicate` function I *think*?

Another way might be to have something like (very very pseudo-codey!):

```rust
fn add_dup<F: FnOnce>(&mut self, f: F) {
   for e in errs.iter_mut() {
     if f(e) {
        span.push(...);
     }
   }
}
```

where you'd pass a closure like `|e| e.kind = =... && e.span == ...` or similar.

ratmice: Yeah exactly, trying to make it more generic, so we don't have to copy/paste that loop all over.
I'm not sure if that closure would work, I think you need at least 2 closures, one to perform the check and return an existing spans vector, and one to add a new one if it isn't present.

ratmice: FWIW i've also been from the start on the same "merge duplicate errors" page, since it is something that LSP errors directly support, these heirarchical errors. In the `related_information` field of https://docs.rs/lsp-types/latest/lsp_types/struct.Diagnostic.html

So there *is* some editor benefit from it, here is an image (It only shows one error with multiple entries in the related_information field). https://user-images.githubusercontent.com/17038621/49050943-f90a5500-f1a2-11e8-8192-ad10dc04070e.png

So i'm definitely for keeping it, if we can find a nice way.

FWIW, I redid on top of the first two commits, with the errors inlined removing the duplicate_* fields.
Just to see what the total damage is, if you'd like to have a look.

https://github.com/ratmice/grmtools/tree/return_multiple_errors_inlined

I could also do an exploratory patch with that modified `YaccGrammarErrorKind` if you'd like.

ratmice: the last commit is going to be representative of conversion of the removal of the all rest of the `duplicate*` fields I haven't gotten to,
The 2 previous to that are have some setup to get up to that point.

It added arguments instead of `self.errors` so it could move instead of `self.errs.clone()` avoid,
So definitely the last commit but the previous two perhaps.

ratmice: FWIW I did another exploratory branch with the modified `YaccGrammarErrorKind`, 

All the `Kind` modification churn is in here:
https://github.com/softdevteam/grmtools/commit/803bccc38366c89acf8617b489d1b20a8e0949c3

And conversion of `actiontype` to use a new `add_duplicate_occurrence`...
https://github.com/softdevteam/grmtools/commit/3292fff4726a590e33e497be2b983526433ec17e

ratmice: Sorry, should have just linked them directly:

last commit (reflective of all the work being done):
https://github.com/softdevteam/grmtools/commit/d688e6c40c7dc63870ffae2fd6a26b408e9e940a

`parse_declarations` setup and `DuplicateActiontypeDeclaration`:
https://github.com/softdevteam/grmtools/commit/e81d73260ffa0e12c95b61c8352728cf07562ae1
`parse_rules setup`, and DuplicateRules:
https://github.com/softdevteam/grmtools/commit/1a683d5b879426dbe2992584a00ae43e70baef5b

ltratt: I wonder if we're being too clever for our own good and what we really need to do is make `YaccGrammarError` an `enum` along the lines of:

```rust
enum YaccGrammarError {
  DuplicateEPP(Vec<Span>),
  IncompleteComment(Span)
  ...
}
```

or something like that?

ratmice: I think if we want to do that we need to provide some kind of a trait or impl for `YaccGrammarError`,
otherwise code like https://github.com/softdevteam/grmtools/blob/master/lrpar/src/lib/ctbuilder.rs#L368
is going to have to match on error to obtain spans.

This actually made me think of one more thing we should also take into consideration, rust enum variants for `pub` enums, are implicitly `pub`.  So we will *always* have to bump semver to add any error variant.

So it might be worthwhile to also consider a *newtype* around the enum, and just providing a trait or some other API to access the duplicated spans, and the primary span.

So I'm not opposed to it if we want to go that route, but I think it pushes things in a direction where we would have to consider the consumer interface, because as-is that would push us to a position where consumers would have to `match` on every variant.

ratmice: FWIW, I went ahead and did an experiment with the combination of your idea (moving span into the enum),
with my thoughts above (keep YaccGrammarError as a newtype with a private kind) with a proposed `impl YaccGrammarError` adding the requisite methods.

I based that one off of master, though it has some hacks for all the `Vec<Span>` ones, which would need following up on.

https://github.com/softdevteam/grmtools/commit/45c54cc52493ccc3c9f351e321722136c55b50d1

ltratt: I must admit that I didn't at first understand your semver point about the `enum` but https://github.com/softdevteam/grmtools/commit/45c54cc52493ccc3c9f351e321722136c55b50d1 makes it really clear. Hmm, that's a really good point!

OK, so you've convinced me that we should use a `struct`. Now I wonder if we should do:

```rust
struct YaccGrammarError {
    span: Span,
    kind: YaccGrammarErrorKind, // might contain multiple `Span`s
```

or:

```rust
struct YaccGrammarError {
    spans: Vec<Span>, // contains at least one `Span`
    kind: YaccGrammarErrorKind,
```

I *think* that would make it easier to operate on things like "multiple duplicate names"?

ratmice: yeah, I think it would make things easier too, 

however i'm *also* very susceptible to the argument that errors such as `InvalidXYZ` would then need to allocate a vector when it wouldn't actually use it, 

but I'm also susceptible to the argument that it is the a failure path, and so performance is less of a concern.

There is a small amount of `Span` is nicer to deal with than `Vec<Span>` in the testsuite, because it doesn't have to deal with e.g. `[Span 2]` vs `&[Span]`, but that pain is perhaps quite overshadowed by all the construction in the actual code?

This really seems like much more of a tradeoff, but if both are private I guess we can change our minds if we end up not liking one of these choices?  
I don't really seem to have a firm position on either choice I suppose the first one kind of forces us to provide an method for `spans`, while the second one we could consider making spans `pub` or not.

ltratt: > but I'm also susceptible to the argument that it is the a failure path, and so performance is less of a concern.

I tend to agree. If it ever proved to be a bottleneck we could use `SmallVec` or something similar that doesn't allocate until you have more than *n* elements. But I doubt we'd ever get to that point.

>There is a small amount of Span is nicer to deal with than Vec<Span> in the testsuite, because it doesn't have to deal with e.g. [Span 2] vs &[Span], but that pain is perhaps quite overshadowed by all the construction in the actual code?

That's my working guess too. Maybe it's worth trying to convert a tiny subset of the code and see if this hypothesis proves true? If so, maybe we should push forward in this direction?

ratmice: I'll have a look into trying that option at it a bit later.

ratmice: I went ahead and did this in https://github.com/ratmice/grmtools/commit/f929552392c73173c60cd9fd3c8b52e547a6c0bb
`&[Span]` vs `&[Span; 2]` types of errors weren't a problem because we don't have `Option<&[Span]>`.
I thought it turned out well.

That patch is based off of master, and the stuff in *this* PR isn't reflected there yet.

There are also some low-hanging fruit cleanups that could be done to the testsuite,
like we could perhaps now add helper functions for `expect_error_at_line(kind: YaccGrammarErrorKind)` etc, for all these tests which hard code a pattern match.

I didn't want to do all the possible cleanups in one patch though, so it seemed like a good place to stop.

ltratt: I have some comments on https://github.com/ratmice/grmtools/commit/f929552392c73173c60cd9fd3c8b52e547a6c0bb but they're fairly minor. I tend to agree that it looks like the right way to go. What's next?

ratmice: Well, I went ahead and implemented the testsuite cleanups I had considered,

If you could have a look at https://github.com/softdevteam/grmtools/commit/303f30acf35cf06758a620c122062e4d8e9fab0b before I close this in favor of the other branch.

I guess it has the unfortunate side effect of making test failure line numbers point into the helper function rather than the test.  So if that is intolerable, I want to remove that patch before I make the PR.

After that is basically replacing the `YaccParser{ duplicate_* }` fields with a `Vec<YaccGrammarError>`, and producing multiple errors.

ltratt: > I guess it has the unfortunate side effect of making test failure line numbers point into the helper function rather than the test.

This is never perfect, but `RUST_BACKTRACE=1` solves the problem adequately enough IMHO.

Given that the commit has an add:delete ratio of 1:2, I am very happy with it! Please go ahead!

PR #322, opened 2022-07-17
move span fields to YaccGrammarError from kind, make fields private

ratmice:
This reorganizes things moving the duplicate spans fields from kind into YaccGrammarError,
and adds an impl to replace direct access to spans. Also adds some subsequent cleanups which this allows.

This is intended to provide a better basis #321 although it doesn't implement that PR yet.
That could be either done here or in a subsequent PR.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -985,6 +985,71 @@ mod test {
         }};
     }
 
+    trait ErrorsHelper {
+        fn expect_error_at_line(self, src: &str, kind: YaccGrammarErrorKind, line: usize);
+        fn expect_error_at_line_col(
+            self,
+            src: &str,
+            kind: YaccGrammarErrorKind,
+            line: usize,
+            col: usize,
+        );
+        fn expect_error_at_lines_cols(
+            self,
+            src: &str,
+            kind: YaccGrammarErrorKind,
+            lines_cols: &mut dyn Iterator<Item = (usize, usize)>,
+        );
+    }
+
+    impl ErrorsHelper for Result<GrammarAST, Vec<YaccGrammarError>> {
+        fn expect_error_at_line(self, src: &str, kind: YaccGrammarErrorKind, line: usize) {
+            match self.as_ref().map_err(Vec::as_slice) {
+                Ok(_) => panic!("Parsed ok while expecting error"),
+                Err([e])
+                    if e.kind == kind
+                        && line_of_offset(src, e.spans().next().unwrap().start()) == line
+                        && e.spans.len() == 1 => {}
+                Err(e) => incorrect_errs!(src, e),
+            }
+        }
+
+        fn expect_error_at_line_col(
+            self,
+            src: &str,
+            kind: YaccGrammarErrorKind,
+            line: usize,
+            col: usize,
+        ) {
+            self.expect_error_at_lines_cols(src, kind, &mut std::iter::once((line, col)))
+        }
+
+        fn expect_error_at_lines_cols(
+            self,
+            src: &str,
+            kind: YaccGrammarErrorKind,
+            lines_cols: &mut dyn Iterator<Item = (usize, usize)>,
+        ) {
+            match self.as_ref().map_err(Vec::as_slice) {
+                Ok(_) => panic!("Parsed ok while expecting error"),
+                Err([e])
+                    if e.kind == kind
+                        && line_col!(src, e.spans().next().unwrap())
+                            == lines_cols.next().unwrap() =>
+                {
+                    assert_eq!(
+                        e.spans()
+                            .skip(1)
+                            .map(|span| line_col!(src, span))
+                            .collect::<Vec<(usize, usize)>>(),
+                        lines_cols.collect::<Vec<(usize, usize)>>()
+                    )
+                }
+                Err(e) => incorrect_errs!(src, e),
+            }
+        }
+    }
+
     #[test]
     fn test_helper_fn() {
         assert_eq!(Symbol::Token("A".to_string(), Span::new(0, 0)), token("A"));
@@ -1266,43 +1331,21 @@ mod test {
     #[test]
     fn test_unicode_err1() {
         let src = "%token '❤' ❤;".to_string();
-        match parse(
+        parse(
             YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
             &src,
         )
-        .as_ref()
-        .map_err(Vec::as_slice)
-        {
-            Ok(_) => panic!("Incorrect token parsed"),
-            Err(
-                [e @ YaccGrammarError {
-                    kind: YaccGrammarErrorKind::IllegalString,
-                    ..
-                }],
-            ) if line_col!(src, e.spans().next().unwrap()) == (1, 12) => (),
-            Err(e) => incorrect_errs!(src, e),
-        }
+        .expect_error_at_line_col(&src, YaccGrammarErrorKind::IllegalString, 1, 12);
     }
 
     #[test]
     fn test_unicode_err2() {
         let src = "%token '❤'\n%%\nA : '❤' | ❤;".to_string();
-        match parse(
+        parse(
             YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
             &src,
         )
-        .as_ref()
-        .map_err(Vec::as_slice)
-        {
-            Ok(_) => panic!("Incorrect token parsed"),
-            Err(
-                [e @ YaccGrammarError {
-                    kind: YaccGrammarErrorKind::IllegalString,
-                    ..
-                }],
-            ) if line_col!(src, e.spans().next().unwrap()) == (3, 11) => (),
-            Err(e) => incorrect_errs!(src, e),
-        }
+        .expect_error_at_line_col(&src, YaccGrammarErrorKind::IllegalString, 3, 11);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -66,8 +57,10 @@ pub enum YaccGrammarErrorKind {
 /// Any error from the Yacc parser returns an instance of this struct.
 #[derive(Debug, PartialEq, Eq)]
 pub struct YaccGrammarError {
-    pub kind: YaccGrammarErrorKind,
-    pub span: Span,
+    pub(crate) kind: YaccGrammarErrorKind,
+    // contains at least 1 span.
```
ltratt, line 61: These need to be (nicely formatted :)) `///` doc strings.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -142,6 +133,16 @@ impl fmt::Display for YaccGrammarErrorKind {
     }
 }
 
+impl YaccGrammarError {
+    pub fn span(&self) -> Span {
```
ltratt, line 137: I'm not sure about the existence of this method to be honest. At least it would have to be called `first_span` or `base_span` or similar? I think I would prefer not to privilege any of the spans unless we have a good use case for doing so beyond the test suite.

ltratt:
The message for https://github.com/softdevteam/grmtools/pull/322/commits/089316434ccd1203cee38f0141a4a4ce13a9d971 doesn't seem to match the (1 line?) diff in it?

ratmice:
Those two will need to be squashed into their previous commit I was just somewhat hesitant to squash it since you'd looked at the first two.  I don't understand the commit message not matching though?

cfgrammar/src/lib/yacc/parser.rs
ratmice: I was thinking `primary_span`, if that is ok?

So, the reason for separating these is I wanted to send a patch to format strings for extra spans,
Currently I hard code a string such as "Additionally declared here", to `extra_spans`.

Further we don't see *all* the spans as being an error, for instance if there are 2 duplicates, only one of those
needs to be removed, so it seemed to me there is a semantic difference between the spans.  That is generally why I had kept them separate, so that in a future patch I could work towards improving a better reflection of this semantic difference.

ltratt:
>Those two will need to be squashed into their previous commit I was just somewhat hesitant to squash it since you'd looked at the first two. I don't understand the commit message not matching though?

Oh, I see what you mean! Ignore my comment then!

cfgrammar/src/lib/yacc/parser.rs
ltratt: I'm thinking of some compilers where the "main" error is at line 10 but it will tell you "lines 7 and 14 are also related". I suppose we *could* differentiate "primary" spans (notice the plural!) from "secondary" spans but I'm not sure it's worth the complication: I think I'm fine having an ordered sequence of spans. It also means that if someone knows something special about one of the kinds, they might be able to interpret the sequence of spans in a clever way (whereas having methods to access them does implicitly place emphasis on the first entry).

ratmice: Ok, one thing to consider is that with `kind  now private, We don't have any mechanism to differentiate span order relative to kind, except by querying the `errs.len()`.
We could consider a hint like
```
/// Gives an indication of the error type 
enum YaccGrammarErrorType {
  DuplicationError,
  Error,
}

fn error_type(&self: YaccGrammarError) -> YaccGrammarErrorType { ... }
```

I'm a little uncomfortable interpreting the sequence of spans, without a mechanism in place to differentiate them.

ltratt: I like the documentation idea -- that gives us lots of flexibility but isn't too onerous!

ratmice: Should be fixed in https://github.com/softdevteam/grmtools/pull/322/commits/65d12bfc088d9a8d69cf7824da79daa22424666b

After this patch it might take me a bit to figure out if there is a nice way to avoid squashing this all into a single commit,
But i'm not sure.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Should be fixed in https://github.com/softdevteam/grmtools/pull/322/commits/65d12bfc088d9a8d69cf7824da79daa22424666b

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -938,6 +938,91 @@ mod test {
...
+            }
+        }
+
+        fn expect_error_at_line_col_extra(
+            self,
+            src: &str,
+            kind: YaccGrammarErrorKind,
+            line: usize,
+            col: usize,
+            extra_spans: &[Span],
```
ratmice, line 1038: Now that we removed the `extra_spans` function in favor of a single span, this function needs to be rethought.

I think `expect_error_kind_with_lines_cols(&self, &str, kind, &[(usize, usize)])`
and `expect_multiple_errors(&self, &str, &[(kind, &[(usize, usize)]])` or something to that effect

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -133,13 +136,57 @@ impl fmt::Display for YaccGrammarErrorKind {
...
+    /// Returns the spans associated with the error, always containing at least 1 span.
+    ///
+    /// Refer to [YaccGrammarErrorType] via [error_type](Self::error_type)
+    /// for the meaning and interpretation of spans and their ordering.
+    pub fn spans(&self) -> &[Span] {
+        self.spans.as_slice()
+    }
+
+    /// Returns the error type associated with this error.
+    pub fn error_type(&self) -> YaccGrammarErrorType {
```
ltratt, line 157: I'm not sure I see the need for this function outside of tests?

cfgrammar/src/lib/yacc/parser.rs
ltratt: I now realise I had completely misunderstood your message -- mea culpa! What I *thought* you meant was that we would do:

```rust
enum YaccGrammarErrorKind {
   /// Only has 1 span
   IllegalInteger,
   /// First Span is the original, subsequent spans are duplicates.
   DuplicateRule,
   ...
}
```

and so on (hence my "documentation idea" comment).

I think your approach in https://github.com/softdevteam/grmtools/commit/65d12bfc088d9a8d69cf7824da79daa22424666b is a better approach. I suggest we push it a bit further by:

* `error_type` should be `spans_kind` returning a `SpansKind` which would be along the lines of your existing `enum`:
```rust
/// Gives an indication of the error type and interpretation of it's data.
pub enum SpansKind {
    /// The first span is the first occurrence, and a span for each subsequent occurrence.
    DuplicationError,
    /// Contains a single span at the site of the error.
    Error,
}
```
* *Perhaps* (this might not be a good idea!) we turn the `spans` method into `iter_spans` so that we don't have to expose the underlying `Vec` quite so obviously? But that might a bit annoying to use...

cfgrammar/src/lib/yacc/parser.rs
ltratt: Ignore this comment.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yeah `SpansKind` is a better name,

Just for clarification the reason against documenting via the `YaccGrammarErrorKind` variants is because they are now private.  So in practice if it is documented there it is hard leverage that documentation, or even make it visible.

These two things should be fixed in https://github.com/softdevteam/grmtools/pull/322/commits/d70d0357dbca29ba90223a52d6183b418dfb56d0 (rename), and https://github.com/softdevteam/grmtools/pull/322/commits/6ff70b3dc8488bcc2f4c8ad47198594824bea205 (use iter)

cfgrammar/src/lib/yacc/parser.rs
ratmice: I tried to make the first more consistent with the new names etc,
https://github.com/softdevteam/grmtools/pull/322/commits/743ca1c49e05f811aaeec5d61f0ee9b6aaeecc93

I removed the second fn until we have a use for it, expect to readd it in a subsequent patch when we return multiple errors.
https://github.com/softdevteam/grmtools/pull/322/commits/cc445e189b3d9568a85712272138b1bfbd0a3223

cfgrammar/src/lib/yacc/parser.rs
ratmice: I guess I should summarize that kind of long winded text (sorry)...

1. I kind of don't see the need outside of the testsuite to access kind, they generally just get used via conversion to `String`
2. There is some benefit to making them private in that we can add/remove without bumping semver.

I'm not really opposed to making them `pub` again though, it isn't important to me that these be private, just seemed like it was worth considering for the semver flexibility.

ltratt: Duh, I forgot the semver thing. Agreed, let's keep them private at least for the time being. I *think* one day we'll want to open them up but, hopefully, when someone actually needs that, the set of errors will have stabilised.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -142,6 +136,60 @@ impl fmt::Display for YaccGrammarErrorKind {
...
+    /// Returns the spans associated with the error, always containing at least 1 span.
+    ///
+    /// Refer to [SpansKind] via [spans_kind](Self::spans_kind)
+    /// for the meaning and interpretation of spans and their ordering.
+    pub fn spans(&self) -> impl Iterator<Item = Span> + '_ {
+        self.spans.iter().copied()
+    }
+
+    /// Returns the kind of spans associated with this error.
+    pub fn spans_kind(&self) -> SpansKind {
```
ltratt, line 157: Hmm, should this be `spans_kind` or `spanskind`? I'm not sure!

ltratt:
One minor comment and then I think we're pretty much done?

cfgrammar/src/lib/yacc/parser.rs
ratmice: Good catch, normally I would go with `spans_kind`, I could give some other projects that use `foo_kind` in similar situations,
But to be self-consistent there are already other methods, which would imply it should be `spanskind`.

https://docs.rs/lrpar/latest/lrpar/struct.CTParserBuilder.html#method.yacckind
https://docs.rs/lrlex/latest/lrlex/struct.CTLexerBuilder.html#method.lexerkind

ratmice: Fixed in https://github.com/softdevteam/grmtools/pull/322/commits/ed97c1b38b3b2dfd03adae370faf4b2b2f685cff

ratmice:
Squashed, I had to do some surgery on the first commit to try and keep the testsuite cleanup separate.

PR #324, opened 2022-07-21
Return multiple errors take 2

ratmice:
Following the YaccGrammarError cleanups, here is an attempt to remove the fields from `YaccParser`, and return multiple errors from `parse`.

When looking at the changes to `parse` it may be easier to look at the combined cumulative diff of the entire series.
But it seemed worthwhile to keep the for each `Kind` to it's own patch.

Rather than previous PRs, where I did the first 2 patches in this series, I just went ahead and  did all duplicate error kinds.
So there shouldn't be any additional patches for `YaccGrammarError` expected, besides the normal post-review ones.


ratmice:
Rebased this onto master, but no other changes.

ltratt:
Erp, I missed this! Best to "assign" me (then it shows up on one of my daily checks).

Reviewing now.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -224,109 +234,55 @@ impl YaccParser {
...
-        i = self.parse_rules(i?).map_err(|e| vec![e]);
-        if let Some((orig_span, spans)) = self.duplicate_rule_spans.iter().next() {
-            let mut tmp = vec![*orig_span];
-            tmp.extend(spans);
-            return Err(vec![YaccGrammarError {
-                kind: YaccGrammarErrorKind::DuplicateRule,
-                spans: tmp,
-            }]);
+        let mut result = self.parse_declarations(0, &mut errs);
+        result = self.parse_rules(
```
ltratt, line 246: I wonder if we can squeeze this down to something like `self.parse_rules(result.map_err(|e| errs.push(e); Err(errs))?, &mut errs)`? The borrow checker might not like it though.

ltratt:
One minor comment but otherwise looks good!

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yeah, that won't work exactly because of the borrow checker, and most of the other combinators won't allow a `return` from within, so I've been perplexed as to whether there is a nicer way.

ltratt: Let's stick with it as-is. We can fix it in a later PR if we think of something clever.

PR #326, opened 2022-07-24
Remove StIdxStorageT and replace it with StorageT

ltratt:
The main bulk of this PR is in https://github.com/softdevteam/grmtools/commit/af8c85ede6bfafe4b6828063a3672519e1833b4c and partially ameliorates a long-time grmtools wart, which is that the maximum number of parsing states was hardcoded to a `u16`. This commit moves that to `StoargeT` maximum states. In one sense that overloads `StorageT` but at least the user now has *some* control over how many states grmtools can deal with.

Thoughts welcome: this is a part of the code I haven't looked at in some time!

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -147,14 +147,16 @@ where
 {
     /// Create a new `CTParserBuilder`.
     ///
-    /// `StorageT` must be an unsigned integer type (e.g. `u8`, `u16`) which is big enough to index
-    /// (separately) all the tokens, rules, and productions in the grammar and less than or
-    /// equal in size to `usize` (e.g. on a 64-bit machine `u128` would be too big). In other
-    /// words, if you have a grammar with 256 tokens, 256 rules, and 256 productions, you
-    /// can safely specify `u8` here; but if any of those counts becomes 256 you will need to
-    /// specify `u16`. If you are parsing large files, the additional storage requirements of
-    /// larger integer types can be noticeable, and in such cases it can be worth specifying a
-    /// smaller type. `StorageT` defaults to `u32` if unspecified.
+    /// `StorageT` must be an unsigned integer type (e.g. `u8`, `u16`) which is:
+    ///   * big enough to index (separately) all the tokens, rules, productions in the grammar,
+    ///   * big enough to index the state table created from the grammar,
+    ///   * less than or equal in size to `u32`.
+    /// In other words, if you have a grammar with 256 tokens, 256 rules, and 256 productions,
+    /// which creates a state table of 256 states you can safely specify `u8` here; but if any of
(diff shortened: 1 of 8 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -147,14 +147,16 @@ where
...
-    /// specify `u16`. If you are parsing large files, the additional storage requirements of
-    /// larger integer types can be noticeable, and in such cases it can be worth specifying a
-    /// smaller type. `StorageT` defaults to `u32` if unspecified.
+    /// `StorageT` must be an unsigned integer type (e.g. `u8`, `u16`) which is:
+    ///   * big enough to index (separately) all the tokens, rules, productions in the grammar,
+    ///   * big enough to index the state table created from the grammar,
+    ///   * less than or equal in size to `u32`.
+    /// In other words, if you have a grammar with 256 tokens, 256 rules, and 256 productions,
+    /// which creates a state table of 256 states you can safely specify `u8` here; but if any of
+    /// those counts becomes 256 you will need to specify `u16`. If you are parsing large files,
```
ratmice, line 156: I think this intends to say "but if any of those counts becomes 257 you will need".

lrtable/src/lib/pager.rs
```diff
@@ -257,7 +260,7 @@ where
                         panic!("StorageT is not big enough to store this stategraph.");
                     }
                     // The assert above guarantees that the cast below is safe.
-                    let stidx = StIdx(core_states.len().as_());
+                    let stidx = StIdx(core_states.len());
```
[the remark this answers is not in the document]
ltratt: Fixed in 970f5b5ed36234b95204a5482b548ec6dca68d74.

lrpar/src/lib/ctbuilder.rs
ltratt: Fixed in 9502f8d.

ratmice:
Well, this isn't really an area I'm very familiar with, and I've never shied away from adding generic type parameters (for better or worse).  But it doesn't seem like adding another type parameter in this case would actually buy you much, partially because the trait bounds don't actually enforce these sizing invariants, so adding another would seem to come with _additional_ dynamic checks.

One thing I notice that we could use to reduce the `StorageT` bounds, is something like:
https://www.worthe-it.co.za/blog/2017-01-15-aliasing-traits-in-rust.html

But I kind of don't think that adding these `aliasing traits`, would actually tip this towards adding generic type parameters.
Perhaps we could consider that in a separate patch though?

ltratt:
I did not know about the trait alias trick! I agree that it's probably something to consider in a separate tidy-up PR.

ratmice:
Cool, well please squash, if you feel it is ready.

ltratt:
I think this is an improvement (albeit there's more to be done!). Squashed.

PR #327, opened 2022-07-25
Keep updating CHANGES.md.

ltratt:
@ratmice do you think there will be further semver-breaking changes? If so, I'm happy to keep waiting to make a release (but I'll keep updating CHANGES.md intermittently while I vaguely remember what's happened.)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,33 @@
 # grmtools 0.13.0 (XXXX-XX-XX)
 
+This release contains a number of new features and breaking changes -- most of
+the latter are in advanced/niche parts of the API which don't affect most users
+(with the exception of the `StorageT` change, the first noted in "breaking
+changes" below, which may affect a minority of users).
+
+
+## New features
+
+* Improved error messages, with various parts of grammar and lexer files now
+  carrying around `Span` information to help pinpoint errors.
+
+* A single error can be related to multiple `Span`s. For example, if you
+  duplicate a rule name in a grammar, all duplicates are reported in a single
+  error.
+
+* The new `cfgrammar::NewlineCache` struct makes it easier to store the
+  minimal information needed to convert byte offsets in an input into logical
+  line numbers.
+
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -19,6 +46,14 @@
     pub fn add_rule(&mut self, (name, name_span): (String, Span), actiont: Option<String>) {
   ```
 
+* `YaccParserError` has been renamed to `YaccGrammarError` and changed from an
+  `enum` (which frequently changed, breaking semver) to a `struct`. The `kind`
+  field is now internal (so won't break semver as often).
```
ratmice, line 51: There is also some removals associated with this which should probably be noted: 
Perhaps something to the effect of:

 `GrammarValidationError` and `YaccParserError` have been combined as `YaccGrammarError`, replacing the enum `YaccGrammarError` which had a variant for each of those.

CHANGES.md
```diff
@@ -19,6 +46,14 @@
...
+
+* `LexBuildresult` returns on failure `Err(Vec<LexBuildError>)` rather than
+  `Err(LexBuildError)`.
+
```
ratmice, line 55: This one I still need to look at and see if we need to give `LexBuildResult` the same treatment regarding `spans` field and `kind` as we did to `YaccGrammarError`, so there may be one more semver change here perhaps.

ltratt: Ah, OK. Should I delete this line or ... ?

ratmice: No I think what is written shouldn't be changing, 
Just wanted to point out one last semver change might be needed.

ltratt: Got it!

CHANGES.md
ltratt: Fixed in 0b1d465.

ratmice:
Please squash

ltratt:
Squashed.

PR #329, opened 2022-07-25
Move from usize to u64 backing storage for Vobs.

ltratt:
By using variable sized storage, we couldn't serialise a `Vob` on a 64-bit machine and deserialise on a 32-bit machine (i.e. we couldn't cross-compile grmtools to WASM). This relatively simple change makes that possible (since the backing is now always `u64` no matter the machine). [See https://github.com/softdevteam/sparsevec/pull/21 which made this PR possible.]

This may stop people asking me when grmtools will support wasm (it always "has" in the sense that if you compiled it on a 32-bit machine it worked. But, in fairness, I don't have access to a 32-bit development environment any more!

### DIFF
--- a/lrtable/Cargo.toml
+++ b/lrtable/Cargo.toml
@@ -18,5 +18,5 @@ fnv = "1.0"
 num-traits = "0.2"
 cfgrammar = { path="../cfgrammar", version = "0.12", features=["serde"] }
 serde = { version="1.0", features=["derive"], optional=true }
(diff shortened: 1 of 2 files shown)


ratmice:
looks pretty straight forward to me, do we need to rebase on master, or does bors manage to take care of that?

ltratt:
It'll create a merge commit (so if it can't merge it safely, it'll bork in one way another).

PR #330, opened 2022-07-27
Return multiple lexbuildresult

ratmice:
This implements the returning of multiple errors for `LexBuildResult`,
The collection of multiple spans for `DuplicateStartState`, and testsuite cleanups.

There is quite a bit of duplicate copy/paste between cfgrammar/lrlex for the testsuite helpers, but I didn't want to undertake something like adding an entire helper crate for tests.

All the TODO/FIXME added get removed by subsequent patches.

@SMartinScottLogic  This modifies your start state code which was recently added, to collect `Span`s, mostly in the last patch, but some parameters are added in the first, if you don't mind having a look.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -313,7 +313,10 @@ where
                         errs.iter()
                             .map(|e| {
                                 if let Some((line, column)) = line_cache
-                                    .byte_to_line_num_and_col_num(&lex_src, e.span.start())
+                                    .byte_to_line_num_and_col_num(
+                                        &lex_src,
+                                        e.spans().next().unwrap().start(),
+                                    )
                                 {
                                     format!("{} at line {line} column {column}", e)
                                 } else {
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -419,8 +419,12 @@ pub fn lexerdef() -> {lexerdef_type} {{
             write!(
                 outs,
                 "
-        StartState::new({}, {:?}, {}),",
-                ss.id, ss.name, ss.exclusive
+        StartState::new({}, {:?}, {}, ::cfgrammar::Span::new({}, {})),",
+                ss.id,
+                ss.name,
+                ss.exclusive,
+                ss.name_span.start(),
+                ss.name_span.end()
             )
             .ok();
         }
--- a/lrlex/src/lib/mod.rs
+++ b/lrlex/src/lib/mod.rs
@@ -27,6 +27,7 @@ pub use crate::{
     parser::StartState,
 };
 
+use cfgrammar::yacc::parser::SpansKind;
 use cfgrammar::Span;
 
 pub type LexBuildResult<T> = Result<T, Vec<LexBuildError>>;
@@ -34,8 +35,8 @@ pub type LexBuildResult<T> = Result<T, Vec<LexBuildError>>;
 /// Any error from the Lex parser returns an instance of this struct.
 #[derive(Debug)]
 pub struct LexBuildError {
-    pub kind: LexErrorKind,
-    pub span: Span,
+    pub(crate) kind: LexErrorKind,
+    pub(crate) spans: Vec<Span>,
 }
 
 impl Error for LexBuildError {}
@@ -52,8 +53,34 @@ pub enum LexErrorKind {
     DuplicateStartState,
     InvalidStartState,
     InvalidStartStateName,
-    DuplicateName(Vec<Span>),
+    DuplicateName,
     RegexError,
+}
+
+impl LexBuildError {
+    pub fn spans(&self) -> impl Iterator<Item = Span> + '_ {
+        self.spans.iter().copied()
+    }
+
+    pub fn spanskind(&self) -> SpansKind {
+        match self.kind {
+            LexErrorKind::PrematureEnd
+            | LexErrorKind::RoutinesNotSupported
+            | LexErrorKind::UnknownDeclaration
+            | LexErrorKind::MissingSpace
+            | LexErrorKind::InvalidName
+            | LexErrorKind::UnknownStartState
+            | LexErrorKind::InvalidStartState
+            | LexErrorKind::InvalidStartStateName
+            | LexErrorKind::RegexError
+            // TODO DuplicateStartState should be a DuplicationError,
+            // but currently behaves as a SpansKind::Error.
+            | LexErrorKind::DuplicateStartState => SpansKind::Error,
+            LexErrorKind::DuplicateName => {
+                SpansKind::DuplicationError
+            }
+        }
+    }
 }
 
 impl fmt::Display for LexBuildError {
@@ -68,7 +95,7 @@ impl fmt::Display for LexBuildError {
             LexErrorKind::DuplicateStartState => "Start state already exists",
             LexErrorKind::InvalidStartState => "Invalid start state",
             LexErrorKind::InvalidStartStateName => "Invalid start state name",
-            LexErrorKind::DuplicateName(_) => "Rule name already exists",
+            LexErrorKind::DuplicateName => "Rule name already exists",
             LexErrorKind::RegexError => "Invalid regular expression",
         };
         write!(f, "{s}")
--- a/lrlex/src/lib/mod.rs
+++ b/lrlex/src/lib/mod.rs
@@ -72,11 +72,8 @@ impl LexBuildError {
             | LexErrorKind::UnknownStartState
             | LexErrorKind::InvalidStartState
             | LexErrorKind::InvalidStartStateName
-            | LexErrorKind::RegexError
-            // TODO DuplicateStartState should be a DuplicationError,
-            // but currently behaves as a SpansKind::Error.
-            | LexErrorKind::DuplicateStartState => SpansKind::Error,
-            LexErrorKind::DuplicateName => {
+            | LexErrorKind::RegexError => SpansKind::Error,
+            LexErrorKind::DuplicateName | LexErrorKind::DuplicateStartState => {
                 SpansKind::DuplicationError
             }
         }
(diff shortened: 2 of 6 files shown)


lrlex/src/lib/ctbuilder.rs
```diff
@@ -399,6 +402,7 @@ where
 
         outs.push_str(&format!(
             "{mod_vis} mod {mod_name} {{
+use cfgrammar::Span;
```
ltratt, line 405: I've just realised: we shouldn't *really* use global imports here. [Ideally we'd banish the other ones too, but not in this PR!] It's better if we always use `::cfgrammar::Span` in the generated code as then we can't have name clashes (i.e. the `::` syntax, though laborious, allow us to largely maintain hygiene when we copy the user's code in).

lrlex/src/lib/mod.rs
```diff
@@ -27,15 +27,16 @@ pub use crate::{
     parser::StartState,
 };
 
+use cfgrammar::yacc::parser::SpansKind;
```
ratmice, line 30: Should we export this from top-level, so we don't have to reach into parser from within lrlex?

lrlex/src/lib/mod.rs
```diff
@@ -52,10 +53,32 @@ pub enum LexErrorKind {
...
     InvalidStartStateName,
-    DuplicateName(Vec<Span>),
+    DuplicateName,
     RegexError,
 }
 
+impl LexBuildError {
+    pub fn spans(&self) -> impl Iterator<Item = Span> + '_ {
+        self.spans.iter().copied()
+    }
```
ltratt, line 63: Missing newline separator.

lrlex/src/lib/parser.rs
```diff
@@ -359,6 +436,90 @@ mod test {
         }};
     }
 
+    fn line_of_offset(s: &str, off: usize) -> usize {
+        s[..off].lines().count()
+    }
```
ltratt, line 441: Missing newline separator.

lrlex/src/lib/mod.rs
ltratt: Not quite sure I understand?

ltratt:
This looks mostly good to me, modulo some minor nitpicks (though I agree that it would be good to hear from @SMartinScottLogic).

lrlex/src/lib/mod.rs
ratmice: Just that it is reusing this type which is declared within `cfgrammar::yacc`, from within the `lrlex` crate,
rather than a neutral place like `cfgrammar::span::SpanKind` where it can be used by *both* `cfgrammar::yacc` and `lrlex`, but keeping it where it is (close to the errors from which it derives its meaning might be preferred organizationally).

I think it is harmless, but dependency-wise I wasn't certain it made sense to actually depend upon the `yacc` module here.

ltratt: Oh, I'm with you! So make this a `pub use` in `cfgrammar/mod.rs`?

ratmice: Yeah, that is what I was thinking!

ltratt: Having double-checked this, I think the top-level of cfgrammar should be somewhat independent of yacc, and `SpansKind` is (at least at the moment) seemingly yacc specific, so I think `use cfgrammar::yacc::parser::SpansKind;`, while long, is the right way to do things.

lrlex/src/lib/ctbuilder.rs
ratmice: This and the whitespace changes fixed in https://github.com/softdevteam/grmtools/pull/330/commits/d050bbfd082436697ce9852c71c3547fd71c50ef

lrlex/src/lib/mod.rs
ratmice: Alright, I'll leave this as-is then.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -402,7 +402,7 @@ where
 
         outs.push_str(&format!(
             "{mod_vis} mod {mod_name} {{
-use cfgrammar::Span;
+use ::cfgrammar::Span;
```
ltratt, line 405: Ah, no what I mean is: we don't want to introduce `Span` into the namespace at all. In other words, we should kill this `use`, and wherever we reference `Span` change it to `::cfgrammar::Span` (there are some instances of this already IIRC).

ratmice: Aha, now I see should be fixed in [860575e](https://github.com/softdevteam/grmtools/pull/330/commits/860575eaf7b85a2ebf99762d66ffa9e4917597c9)

lrlex/src/lib/parser.rs
```diff
@@ -46,56 +48,104 @@ impl StartState {
...
-            Some(j) => {
-                let k = self.parse_ws(j).map_err(|e| vec![e])?;
-                if k == self.src.len() {
-                    Ok(i)
-                } else {
-                    Err(vec![self.mk_error(LexErrorKind::RoutinesNotSupported, i)])
+        let mut errs = Vec::new();
+        let mut result = self.parse_declarations(0, &mut errs);
+        result = self.parse_rules(
+            match result {
```
SMartinScottLogic, line 103: As we're NOT parsing rules when the `parse_declarations` returns an error, would it not be more legible to pull out this match - possibly into something like
```
if let Err(e) = result {
   errs.push(e);
   return Err(errs);
} else {
  result = self.parse_rules(result.unwrap(), // unwrap safe due to the above
               &mut errs,
  );
}

File: lrlex/src/lib/parser.rs [hunk 7]
```diff
@@ -104,11 +154,15 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
             if let Some(j) = self.lookahead_is("%%", i) {
                 break Ok(j);
             }
-            i = self.parse_declaration(i)?;
+            i = self.parse_declaration(i, errs)?;
```
SMartinScottLogic, line 157: Do we not want to gather all declaration errors into `errs`, rather than aborting on the first failure?
Or is the intent to get the initial code refactoring into place, then gradually widen the permitted collection?

ratmice: It really depends upon the error, there is primarily two factors
1. whether we can continue parsing after the error occurs at all.
2. if the continuing to parse will cause a cascade of subsequent errors if we ignore it

All of the "Duplicated" errors, tend to more obviously fall in the category of being able to continue parsing without causing a cascade of errors, since we can just not add the duplicate state.

Other errors are likely to require some experimentation to see how they end up reacting.
`UnknownDeclaration` seems like it could probably be ignored, I'm less certain how `InvalidStartStateName`, since it is likely to occur because of a syntax error during e.g. an incomplete edit.

So yeah, the intent is to just get the initial obvious cases working, enough so that it is tested, and probably fixing more cases as we uncover them.

lrlex/src/lib/parser.rs
```diff
@@ -46,56 +48,104 @@ impl StartState {
...
-            None => {
-                assert_eq!(i, self.src.len());
-                Ok(i)
-            }
+            Ok(i) => match self.lookahead_is("%%", i) {
+                Some(j) => {
+                    let k = self.parse_ws(j).map_err(|e| vec![e])?;
+                    if k == self.src.len() {
+                        if errs.is_empty() {
+                            Ok(result.unwrap())
```
ratmice, line 124: `result.unwrap()` can be `i`

lrlex/src/lib/parser.rs
ratmice: That is an option, I mostly went with the match to avoid `unwrap()`, but another option instead of embedding the match in `parse_rules` we could

```rust
let mut i = match self.parse_declarations(0) {
        Ok(i) => i,
        Err(e) => {
           errs.push(e); 
           return Err(errs);
        }
};
```

It ended up this primarily as a consequence of the way `YaccGrammarError` was at the start of a series,
with things happening between `result = parse_foo(..);` and the subsequent `match`, but that has been refactored out of `YaccGrammarError`, and isn't the case here at all.  I just hadn't reconsidered it since the start of that other series.

lrlex/src/lib/parser.rs
```diff
@@ -71,37 +90,55 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+                errs.push(e);
+                Err(errs)
             }
-            None => {
-                assert_eq!(i, self.src.len());
-                Ok(i)
-            }
+            Ok(i) => match self.lookahead_is("%%", i) {
+                Some(j) => {
+                    let k = self.parse_ws(j).map_err(|e| vec![e])?;
```
ratmice, line 121: this `map_err` should `errs.push(..); ...`, will try and fix these issues tomorrow

lrlex/src/lib/parser.rs
SMartinScottLogic: Definitely better than the embedded match

lrlex/src/lib/parser.rs
SMartinScottLogic: Sounds good. I'm happy with that as a plan for this first refactor. Once in, much easier for multiple people to pick up individual pieces separately and work through the failure modes, and consequences of continuing after each possible error.

lrlex/src/lib/parser.rs
ratmice: I implemented that in https://github.com/softdevteam/grmtools/pull/330/commits/5f6a164f261ef08fe42b74afdb6ded8ab6f9986c along with the other issues I noticed in the `parse` function.

ltratt:
Works for me. @SMartinScottLogic ?

SMartinScottLogic:
Looks good to me too

ltratt:
Please squash (looks like a merge conflict too: sorry!).

ratmice:
Squashed

ltratt:
I have the nittiest of nitpicks: the commit message for https://github.com/softdevteam/grmtools/pull/330/commits/eb2fae5d9e86eacdec3b3bb944cb42f24f72a1be has a leading space on the line beginning "multiple spans" I think? If that is the case, please force push an update.

ratmice:
Fixed that and pushed.

PR #331, opened 2022-07-28
Clippy fixes

ltratt:
https://github.com/softdevteam/grmtools/commit/f4c2fbf6ad943e27f44c0908eb9e075e26769649 is more extensive than usual, but I generally prefer to keep Clippy warnings to zero, because it does highlight some useful things, and it's worth not losing them amongst the noise.

### DIFF
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, collections::HashMap};
+use std::{cell::RefCell, collections::HashMap, fmt::Write};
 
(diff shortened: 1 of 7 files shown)


ratmice:
The only thing I would note is if you run with `cargo clippy --enable-targets=all` 
one more `s.push_str(format!(...))` appears in lrlex/src/lib/parser.rs `exceed_tok_id_capacity()`.

Other than that, looks fine

ltratt:
Weirdly, that invocation of clippy doesn't work for me, but I see the `push_str` anyway: fixed in f358cec.

ratmice:
Sorry, that should have been *--all-targets*.

Please squash

ltratt:
Squashed.

PR #332, opened 2022-07-29
One-line dup start states

ratmice:
I hadn't thought about this before, but noticed one case where we weren't catching duplicate start states, because they would be deduplicated by `HashSet` before being added to `start_states`, So if all the duplicates occur on one line, they wouldn't raise an error.

This adds a test for it, and a proposed fix.

Would @SMartinScottLogic mind having a look?

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -1,7 +1,6 @@
 use cfgrammar::Span;
 use lazy_static::lazy_static;
 use regex::Regex;
-use std::collections::HashSet;
 use try_from::TryFrom;
 
 use crate::{lexer::Rule, LexBuildError, LexBuildResult, LexErrorKind};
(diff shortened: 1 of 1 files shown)


lrlex/src/lib/parser.rs
```diff
@@ -201,20 +200,22 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
             .split_whitespace()
             .map(|name| {
                 let off = name.as_ptr() as usize - self.src.as_ptr() as usize;
                 let span = Span::new(off, off + name.len());
-                self.validate_start_state(span, name, errs)
+                (name, span)
             })
-            .filter_map(Result::transpose)
-            .collect::<LexInternalBuildResult<HashSet<(&str, Span)>>>()?;
+            .collect::<Vec<(&str, Span)>>();
```
SMartinScottLogic, line 210: Do you really want to *both* explicitly type in the variable `start_states`' declaration, and turbo-fish in the `collect`?

lrlex/src/lib/parser.rs
```diff
@@ -201,20 +200,22 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
 
         for (name, name_span) in start_states {
             let id = self.start_states.len();
-            let start_state = StartState::new(id, name, exclusive, name_span);
-            self.start_states.push(start_state);
+            if let Some(result) = self.validate_start_state(name_span, name, errs).transpose() {
+                let (name, name_span) = result?;
+                let start_state = StartState::new(id, name, exclusive, name_span);
+                self.start_states.push(start_state);
+            }
```
SMartinScottLogic, line 218: Is it possible for there to be a `None` case here? I note it's silently ignored if so

ratmice: Yeah, that is the case where we encountered a duplicate, and the error was added to `errs` which is eventually caught in `parse`, so ignoring `None` is the right thing.

lrlex/src/lib/parser.rs
ratmice: I'd failed to notice because of inline hints, we could even just simplify it to just `Vec<_>` in collect.

ratmice: Should be fixed in [658db78](https://github.com/softdevteam/grmtools/pull/332/commits/658db78e67fd8c0ebb1df5a5dd2c27535bd182ef)

lrlex/src/lib/parser.rs
ratmice: One simplification here, could perhaps do is just return `Result<bool>` from `validate_start_state` since it is just returning its arguments?

ratmice: I went ahead and added a comment to this effect, and tried to simplify this branch in https://github.com/softdevteam/grmtools/pull/332/commits/661fc9340e80186728b974d90c35c18200493ee7.

lrlex/src/lib/parser.rs
```diff
@@ -201,20 +200,22 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
             })
-            .filter_map(Result::transpose)
-            .collect::<LexInternalBuildResult<HashSet<(&str, Span)>>>()?;
+            .collect::<Vec<_>>();
 
         for (name, name_span) in start_states {
             let id = self.start_states.len();
-            let start_state = StartState::new(id, name, exclusive, name_span);
-            self.start_states.push(start_state);
+            // If false, assume an err was added to errs that will be handled later.
```
ltratt, line 214: I think the right place for this comment is a doc string on `validate_start_date` so that everyone knows what behaviour to expect from a return of `false`.

ratmice: Fixed in [58a53e7](https://github.com/softdevteam/grmtools/pull/332/commits/58a53e75f14401753958da26411b8e0fdefc233d)

ltratt:
Please squash.

ratmice:
Squashed.

PR #333, opened 2022-07-30
Test for start state operators at start of regex

ratmice:
Updating my various projects to run on master, in one of my projects,
I noticed that the addition of start states has caused a change in parsing lex files I hadn't anticipated.

https://github.com/ratmice/crimson/blob/main/src/crimson.l#L12

There is a comment in posix lex specification about this, which I've quoted in the test.

One workaround which should work with both old and new versions of lrlex is using `[<]` instead of `<`.
which would be one way of avoiding special casing the escaping of '\<'.

It also seems from the posix spec they mention '\>' I guess that is just for parity with '\<' I don't see why it is actually required, or why `>` wouldn't be able to start a regex.

I haven't looked yet at where might be the right place to add this special cased escaping though.

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -1,6 +1,8 @@
 use cfgrammar::Span;
 use lazy_static::lazy_static;
 use regex::Regex;
+use std::borrow::{Borrow as _, Cow};
+use std::ops::Not;
 use try_from::TryFrom;
 
 use crate::{lexer::Rule, LexBuildError, LexBuildResult, LexErrorKind};
@@ -13,6 +15,9 @@ lazy_static! {
         Regex::new(r"^%[sS][a-zA-Z0-9]*$").unwrap();
     static ref RE_EXCLUSIVE_START_STATE_DECLARATION: Regex =
         Regex::new(r"^%[xX][a-zA-Z0-9]*$").unwrap();
+    // Documented in `Escape sequences` in lexcompatibility.m
+    static ref RE_LEX_ESC_LITERAL: Regex =
+        Regex::new(r"^(([xuU][[:xdigit:]])|[[:digit:]]|[afnrtv\\]|[pP]|[dDsSwW]|[AbBz])").unwrap();
 }
 const INITIAL_START_STATE_NAME: &str = "INITIAL";
 
@@ -361,9 +366,100 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
         &self,
         off: usize,
         re_str: &'a str,
-    ) -> LexInternalBuildResult<(Vec<usize>, &'a str)> {
+    ) -> LexInternalBuildResult<(Vec<usize>, std::borrow::Cow<'a, str>)> {
         if !re_str.starts_with('<') {
-            Ok((vec![], re_str))
+            /// This implements the 'Table: Escape Sequences in lex' from POSIX lex specification
+            ///
+            /// Most of the escape handling is left to regex, except this part:
+            ///
+            /// Escape: \c
+            ///
+            /// Description:
+            /// A <backslash> character followed by any character not described in this table or in the table in
+            /// XBD File Format Notation ( '\\', '\a', '\b', '\f' , '\n', '\r', '\t', '\v' ).
+            ///
+            /// Meaning: The character 'c', unchanged.
+            fn unescape(re: Cow<str>) -> Cow<str> {
+                // POSIX lex has two layers of escaping, there are escapes for the regular
+                // expressions themselves and the escapes which get handled by lex directly.
+                // We can find what the `regex` crate needs to be escaped with `is_meta_character`.
+                //
+                // We need to avoid sending Regex an escaped character which it does not consider
+                // a meta_character. As that would fail to compile. While ensuring we retain
+                // the escape sequences for the meta characters that it does require.
+                //
+                // '<' is interesting because when used at the beginning of a regex, intended to
+                // be part of the regex, it needs to be escaped, otherwise lex will interpret it
+                // as the beginning of a [`StartState`](StartState)
+                //
+                // If regex_syntax changes behavior here, it is highly likely that we should just
+                // remove this assertion, and send regex `\<`. Instead of unescaping `\<` into `<`.
+                // still it may be worthwhile to ensure that we notice such a change.
+                debug_assert!(regex_syntax::is_meta_character('<').not());
+                let re_str: &str = re.borrow();
+                let mut re_chars = re_str.char_indices();
+
+                // Look for an escape sequence which needs unescaping
+                let mut cursor = loop {
+                    if let Some((i, c)) = re_chars.next() {
+                        if c == '\\' {
+                            // Look at the next character and whether it is something we need to unescape
+                            if let Some((j, c2)) = re_chars.next() {
+                                let s = &re_str[j..];
+                                if !(regex_syntax::is_meta_character(c2)
+                                    || RE_LEX_ESC_LITERAL.is_match(s))
+                                {
+                                    break (Some((i, s, j, c2)));
+                                }
+                            }
+                        }
+                    } else {
+                        break None;
+                    }
+                };
+
+                if cursor.is_none() {
+                    // There is nothing to unescape, return the original parameter
+                    return re;
+                }
+
+                // At this point we have found something to unescape
+                let mut unescaped = String::new();
+                let mut last_pos = 0;
+
+                'outer: while let Some((i, s, j, c)) = cursor {
+                    if regex_syntax::is_meta_character(c) || RE_LEX_ESC_LITERAL.is_match(s) {
+                        // For both meta characters and literals we want to push the entire substring
+                        // up to and including the c match back into the string still escaped.
+                        unescaped.push_str(&re_str[last_pos..j + c.len_utf8()]);
+                        last_pos = j + c.len_utf8();
+                    } else {
+                        // Given '\c' in the original string, push 'c' to the new string.
+                        unescaped.push_str(&re_str[last_pos..i]);
+                        last_pos = j + c.len_utf8();
+                        unescaped.push_str(&re_str[j..last_pos]);
+                    }
+
+                    // Continue looking for the next escape sequence in the original unmodified string.
+                    loop {
+                        if let Some((step1_pos, step1)) = re_chars.next() {
+                            if step1 == '\\' {
+                                cursor = re_chars.next().map(|(step2_pos, step2)| {
+                                    (step1_pos, &re_str[step2_pos..], step2_pos, step2)
+                                });
+                                continue 'outer;
+                            }
+                        } else {
+                            // Hit the end without finding another escape sequence. Copy over the trailing string.
+                            unescaped.push_str(&re_str[last_pos..]);
+                            break 'outer;
+                        }
+                    }
+                }
+                Cow::from(unescaped)
+            }
+
+            Ok((vec![], unescape(Cow::from(re_str))))
         } else {
             match re_str.find('>') {
                 None => Err(self.mk_error(LexErrorKind::InvalidStartState, off)),
@@ -374,7 +470,7 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
                         .map(|s| self.get_start_state_by_name(off, s))
                         .map(|s| s.map(|ss| ss.id))
                         .collect::<LexInternalBuildResult<Vec<usize>>>()?;
-                    Ok((start_states, &re_str[j + 1..]))
+                    Ok((start_states, Cow::from(&re_str[j + 1..])))
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


ltratt:
I think we should follow the POSiX spec unless there's a compelling reason not to. If that does break some old lrlex lexers, we can document it carefully, and maybe even do an intermediate release which warns about the problem.

ratmice:
Cool, I believe that these tests I added should check that we conform to POSIX spec.
But I should double check with a conforming lexer, and figure out where we need to add this escaping.

ratmice:
Well, I think I was entirely misunderstanding what that portion of the lex spec was saying,
It turns out it was entirely handled by this bit of the 'Escape sequences in lex" table.

>The character 'c', unchanged.

That allows you to use `\<` anywhere, but it is only *required* at the beginning of a regex.
This doesn't implement the entire `Escape sequences in lex` table, I'm uncertain about the Regex <--> Lex divide for the rest.
But this does seem to work, handles this aspect of the table, and appears to match *Flex*'s behavior.

This code isn't real pretty, but it does avoid allocating a string if there is no escaping to do,
It could perhaps be simplified if we didn't bother with that.

Any chance you could have a look @SMartinScottLogic?

lrlex/src/lib/parser.rs
```diff
@@ -13,6 +14,9 @@ lazy_static! {
         Regex::new(r"^%[sS][a-zA-Z0-9]*$").unwrap();
     static ref RE_EXCLUSIVE_START_STATE_DECLARATION: Regex =
         Regex::new(r"^%[xX][a-zA-Z0-9]*$").unwrap();
+    static ref RE_ERE_QUOTED_CHARS: Regex = Regex::new(r#"([\^|\.|\[\$|\(|\)|\*|\+|\?])"#).unwrap();
```
ratmice, line 17: Hmm, well this currently isn't quite enough to get crimson working after escaping `<`,
because it has `\{` and '\}', which aren't referenced anywhere in POSIX lex or POSIX ERE `QUOTED_CHARS`.

Simple enough to add, but I suppose I should dig through the regex crate, to figure out if there are more differences lurking about.

ratmice: I just pushed a commit to use `regex_syntax::is_meta_character`, which has some risks e.g. if `regex_syntax` decides that '<' should be a `meta_character`.  But it should be the most compatible with existing lrlex files, if we want to keep using the rust regex format over POSIX ERE.

I figured it is easy to roll back if we want to go another route, but it seems like the most maintainable/compatible option to me

lrlex/src/lib/parser.rs
```diff
@@ -361,9 +365,68 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+                };
+                if cursor.is_none() {
+                    return re;
+                }
+                let mut unescaped = re_str.to_owned();
+                let mut diff = 0usize;
+                'outer: loop {
+                    if let Some((i, s, j)) = cursor {
+                        if !(RE_ERE_QUOTED_CHARS.is_match(s) || RE_LEX_ESC.is_match(s)) {
+                            unescaped.replace_range(i - diff..*j - diff, "");
```
ratmice, line 411: I guess it'd be more efficient to use `with_capacity` to allocate `unescaped` above, copying up to `cursor`, and then instead of `replace_range`, call `push_str` here.

ratmice:
Sorry, this heat has my brain a bit more scattered than usual, I've converted to a draft for the time being.
This will probably want a good rebase before review, because the history is a bit jumbled.

lrlex/src/lib/parser.rs
```diff
@@ -14,7 +14,6 @@ lazy_static! {
         Regex::new(r"^%[sS][a-zA-Z0-9]*$").unwrap();
     static ref RE_EXCLUSIVE_START_STATE_DECLARATION: Regex =
         Regex::new(r"^%[xX][a-zA-Z0-9]*$").unwrap();
-    static ref RE_ERE_QUOTED_CHARS: Regex = Regex::new(r#"([\^|\.|\[\$|\(|\)|\*|\+|\?])"#).unwrap();
     static ref RE_LEX_ESC: Regex =
         Regex::new(r#"([:xdigit:]|[digit]|[\\|a|b|f|n|r|t|v])"#).unwrap();
```
ratmice: So I think that this needs some work, In particular the `xdigit`and `digit` should if I'm understanding the lex spec should
be matching words rather than characters, so I don't think this is right yet.

Or at least I still need to look into this better.

ltratt:
It sounds like @SMartinScottLogic is best placed to comment on this one (but I will do a quick review myself too).

lrlex/src/lib/parser.rs
```diff
@@ -361,9 +365,73 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+                            || RE_LEX_ESC.is_match(s))
+                        {
+                            unescaped.replace_range(i - diff..*j - diff, "");
+                            diff += 1;
+                        }
+                    } else {
+                        break;
+                    }
+
+                    loop {
```
ltratt, line 418: Can this be `while let Some(i, c) = re_chars.next()` ?

lrlex/src/lib/parser.rs
```diff
@@ -361,9 +365,73 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+            /// Description:
+            /// A <backslash> character followed by any character not described in this table or in the table in
+            /// XBD File Format Notation ( '\\', '\a', '\b', '\f' , '\n', '\r', '\t', '\v' ).
+            ///
+            /// Meaning: The character 'c', unchanged.
+            fn unescape(re: Cow<str>) -> Cow<str> {
+                let re_str: &str = re.borrow();
+                let mut re_chars = re_str.char_indices().peekable();
+                debug_assert!(regex_syntax::is_meta_character('<').not());
+                let mut cursor = loop {
```
ltratt, line 385: I wouldn't mind a comment or two in here to explain to an idiot like me roughly what's going on :)

lrlex/src/lib/parser.rs
ratmice: I'm actually surprised, but it appears that it can be!

lrlex/src/lib/parser.rs
ratmice: Yeah, I was actually just in the process of writing one.

lrlex/src/lib/parser.rs
ratmice: Should be fixed in [cf5b0b1](https://github.com/softdevteam/grmtools/pull/333/commits/cf5b0b177f76defff3a0afd47295226af0fc444e)

lrlex/src/lib/parser.rs
ratmice: Should be fixed in [cf5b0b1](https://github.com/softdevteam/grmtools/pull/333/commits/cf5b0b177f76defff3a0afd47295226af0fc444e),

lrlex/src/lib/parser.rs
ratmice: So, perhaps I went overboard with the comments in [cf5b0b1](https://github.com/softdevteam/grmtools/pull/333/commits/cf5b0b177f76defff3a0afd47295226af0fc444e), but it is kind of a funny loop.

lrlex/src/lib/parser.rs
```diff
@@ -379,51 +379,85 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
             ///
             /// Meaning: The character 'c', unchanged.
             fn unescape(re: Cow<str>) -> Cow<str> {
+                // POSIX lex has two layers of escaping, there are escapes for the regular expressions themselves
+                // And the escapes which get handled by lex.  We can find what the `regex` crate considers
+                // an escaped regular expression meta_character which requires escaping. through `is_meta_character`.
```
ltratt, line 384: Extraneous `.` before "through".

lrlex/src/lib/parser.rs
```diff
@@ -379,51 +379,85 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
+                //
+                // We generally want to avoid sending Regex an escaped character which it does not consider a
```
ltratt, line 386: Why "generally"?

lrlex/src/lib/parser.rs
```diff
@@ -379,51 +379,85 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
...
-                        {
-                            unescaped.replace_range(i - diff..*j - diff, "");
-                            diff += 1;
-                        }
-                    } else {
-                        break;
+
+                // At this point we have found something to unescape, allocate a string for it.
+                // We know it'll be shrinking by at least '1' backslash.
+                let mut unescaped = String::with_capacity(re_str.len() - 1);
```
ltratt, line 431: I think `String::new()` is *probably* better here since we don't know by how much less than `re_str.len()-1` we actually need. I can be convinced otherwise though.

lrlex/src/lib/parser.rs
```diff
@@ -379,51 +379,85 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
             ///
             /// Meaning: The character 'c', unchanged.
             fn unescape(re: Cow<str>) -> Cow<str> {
+                // POSIX lex has two layers of escaping, there are escapes for the regular expressions themselves
```
ltratt, line 382: I'm a nitpicker, but can we format the comment to a consistent line length?

lrlex/src/lib/parser.rs
ratmice: Easy enough to change back if we can find some evidence that avoiding resizing is, beneficial.

Should fixed in https://github.com/softdevteam/grmtools/pull/333/commits/819f4e9b626ae3e7f5d631757611c349a0013953

lrlex/src/lib/parser.rs
ratmice: Removed the word, and expanded on the alternate situation in the comment in https://github.com/softdevteam/grmtools/pull/333/commits/819f4e9b626ae3e7f5d631757611c349a0013953.

lrlex/src/lib/parser.rs
ratmice: Should be fixed in https://github.com/softdevteam/grmtools/pull/333/commits/819f4e9b626ae3e7f5d631757611c349a0013953

ltratt:
I'm happy. Over to @SMartinScottLogic !

lrlex/src/lib/parser.rs
ratmice: > \xdigits A <backslash> character followed by the longest sequence of hexadecimal-digit characters
> (01234567abcdefABCDEF). If all of the digits are 0 (that is, representation of the NUL character), the behavior is undefined.

So I guess part of my confusion with this regex is my reading of the above says this should be:
`\a' should be interpreted as a ascii bell and as a `\a` xdigit.

I think it is just not mentioning `\xa` is hex and `\a` is bell.  I.e. I think it should be "a <backslash> followed by an 'x' followed by ...', alas I haven't really used hex escaping in lex and haven't found any examples of it that might clarify things.

ratmice: So, the only other thing i've found so far is:
> \xDigits The character whose encoding is represented by the sequence of hexadecimal characters specified by the Digits string. 

from https://www.ibm.com/docs/en/aix/7.1?topic=information-extended-regular-expressions-in-lex-command

So I am thinking that this is just the *normal* `\xf00`, and not some ancient lex mishap implying that that regex should be `x[:xdigits:]+`

I have changed this, in https://github.com/softdevteam/grmtools/pull/333/commits/6020226bd8fa32a57554e6e4a5666f20f667659e In doing so I'm convinced that we can remove some TODO I had added in previous commits.

ratmice:
FWIW, I believe I have gotten through the remaining issues knock on wood.

lrlex/src/lib/parser.rs
```diff
@@ -15,8 +15,9 @@ lazy_static! {
         Regex::new(r"^%[sS][a-zA-Z0-9]*$").unwrap();
     static ref RE_EXCLUSIVE_START_STATE_DECLARATION: Regex =
         Regex::new(r"^%[xX][a-zA-Z0-9]*$").unwrap();
-    static ref RE_LEX_ESC: Regex =
-        Regex::new(r#"^([:xdigit:]|[digit]|[\\|a|b|f|n|r|t|v])"#).unwrap();
+    // In addition to traditional lex \123 \x1234, accept the \u1234 and \U12345678
+    static ref RE_LEX_ESC_LITERAL: Regex =
+        Regex::new(r"^(([xuU][[:xdigit:]])|[[:digit:]]|[abfnrtv\\])").unwrap();
```
ratmice, line 20: So I guess that there are 2 things I should mention here,

1. the `uU` in the `xdigit` part is I believe currently just passed through to regex (e.g. in the previous release), I believe i've used it in a project but have to look/update it.  As an extension beyond lex.
2. there is also `\p`, `\P`, `\d`, `\D` https://docs.rs/regex/1.6.0/regex/#syntax the new code will convert to `p`, `P`, `d`, `D`.

We probably should make a decision on whether these `regex` crate additional escape sequences that currently get passed through, but will no longer will without explicit code to allow them.

Unfortunately I don't believe that there is anything in the `regex-syntax` crate which can test these in the way that `is_meta_character` can test for regex escaping.

ratmice: I added another patch, which just allows all the escape sequences which it would previously implicitly supported through regex and documents the differences.  If regex changes in the future to add escape sequences though we would have to add them manually before they can be used.

https://github.com/softdevteam/grmtools/pull/333/commits/52d78b9da56cd531ee69a9ee99588cf0ddfe83b7

SMartinScottLogic:
Looks good to me. Sorry for the delay - I've been swamped this week.

ltratt:
Please squash.

ratmice:
Squashed.

Issue #325, opened 2022-07-21
Permit stack operations on start conditions


SMartinScottLogic:
In #318, start state logic was added for start states defined by name.
In the POSIX lex standard, start states can be used by numeric id

Q: Should this include support for expanding the target start state logic to support increment and decrementing the current start state, as well as setting to an explicit target?

ratmice:
Regarding your Q, I'd perhaps just go with the former for now, I'm not quite sure if just incrementing/decrementing a counter for the start state will suffice, I think that a stack (possibly with a counter to reduce depth?) might be necessary if there is a need to enter the same state multiple times, and only exiting the state after the same number of state exits.

Once you do that over a pair of states, a simple counter wouldn't suffice but a stack of them would.
Having said that, all the cases in which I've wanted this manipulation of start state logic via operators, I've since realized I would also need custom lex action code above and beyond operators for target state manipulation.

I'm curious though, if you have any cases where operators for manipulating target states are sufficient?

SMartinScottLogic:
The classic one is nested comments

ratmice:
Ahh yes, I had forgot about that I had been thinking about rusts raw strings like `r##"r#""""#"##` where the number of `#` characters on the outside affects the inner match, it was these that i've had trouble doing without actions as well as some form of start state manipulation.

SMartinScottLogic:
Misunderstanding of the spec - can't see anything in the spec to permit numeric start conditions. 

Remaining: stack of start conditions (push + pop operations)

SMartinScottLogic:
I'm considering a +/- as prefixes before the start condition name in the rule action, e.g.
`%x bracket brace
%%
<*>{       <+brace>'OPEN BRACE'
<brace>}   <-brace>'CLOSE BRACE'
<*>[       <+bracket>'OPEN BRACKET'
<bracket>] <-bracket>'CLOSE BRACKET'
`

ratmice:
Sounds good, and it seems like the obvious choice i would go with as well.

The only thing I wonder about is how to handle states that apply to multiple conditions, if there is a need for a variable indicating the current state.  Something to the effect of:

`<StateA, StateB>] <-$CurrentState>'End State A or B'`

SMartinScottLogic:
Technically, the pop will always pop the top of the stack - so could just be `<->` - thought it made sense to be consistent with push, but possibly consider everything until the close `>` as unprocessed text (effectively a comment)

ratmice:
[...]

PR #334, opened 2022-08-07
Allow start state specification by id.

SMartinScottLogic:
Start of implementation against #325


SMartinScottLogic:
Cancelled - numeric start conditions don't appear in spec that I can see after all. Instead, I'll look into push/pop logic for stacking current start condition.

ratmice:
FWIW, I'm fine with not having them, I've never actually seen it used and they seem less clear than a text start condition.
below is the text in the spec where I had seen them mentioned

> The initial state is indicated by the digit '0' or the token INITIAL.

PR #335, opened 2022-08-07
Fix panics caused by lex rule names with spaces.

ratmice:
I noticed a panic could be caused by the first test this patch adds.  While we have made changes to this area recently,
this appears to have been possible to trigger before these changes. 

The panic in question is because `line.rfind(' ')`, causes a rule like `a " "`, to get an `orig_name` of `"` finding the space between the quotes, instead of the one after `a`.  Subsequently the line: `Some(orig_name[1..orig_name.len() - 1].to_string());` ends up with `orig_name[1..0]` produces a slice error.

The match statements I added could probably be improved, not sure with what yet though.
This now will produce an invalid name error, rather than accept it, but I think that is preferred to a panic.

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -311,15 +311,15 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
             &line[rspace + 1..]
         };
         let name_span;
-        let dupe = if orig_name == ";" {
+        let dupe = if orig_name == ";" || orig_name == r#""""# || orig_name == "''" {
             name = None;
             let pos = i + rspace + 1;
             name_span = Span::new(pos, pos);
             false
         } else {
-            debug_assert!(!orig_name.is_empty());
-            if !((orig_name.starts_with('\'') && orig_name.ends_with('\''))
-                || (orig_name.starts_with('\"') && orig_name.ends_with('"')))
+            if orig_name.len() <= 2
+                || !((orig_name.starts_with('\'') && orig_name.ends_with('\''))
+                    || (orig_name.starts_with('\"') && orig_name.ends_with('"')))
             {
                 return Err(self.mk_error(LexErrorKind::InvalidName, i + rspace + 1));
             }
@@ -733,6 +733,55 @@ mod test {
             2,
             7,
         )
+    }
+
+    #[test]
+    fn test_broken_rule_names_with_spaces() {
+        let srcs = [r#"a ""#, "a '", r#"a '""#, r#"a "'"#];
+        for line_two in srcs {
+            let src = format!("%%\n{line_two}");
+            LRNonStreamingLexerDef::<DefaultLexeme<u8>, u8>::from_str(&src)
+                .expect_error_at_line_col(&src, LexErrorKind::InvalidName, 2, 3);
+        }
+
+        let srcs = [
+            r#"a " ""#,
+            "a ' '",
+            r#"a ' "'"#,
+            r#"a " Name"#,
+            r#"a " Name""#,
+            "a ' Name",
+            "a ' Name'",
+            r#"a " '""#,
+        ];
+
+        for line_two in srcs {
+            let src = format!("%%\n{line_two}");
+            LRNonStreamingLexerDef::<DefaultLexeme<u8>, u8>::from_str(&src)
+                .expect_error_at_line_col(&src, LexErrorKind::InvalidName, 2, 5);
+        }
+    }
+
+    #[test]
+    fn test_unusual_ok_rules() {
+        // Unusual cases of valid regexes and names, that make splitting them tricky.
+        let srcs = [
+            r#"[ "] 'a'"#,
+            r#"[ "] "a""#,
+            r#"[ "] ;"#,
+            r#"[ "] """#,
+            r#"[ "] ''"#,
+            r#"[ '] "a""#,
+            r#"[ '] ;"#,
+            r#"[ '] """#,
+            "[ '] ''",
+            "[ '] 'a'",
+            "[ '] ;",
+        ];
+        for line_two in srcs {
+            let src = format!("%%\n{line_two}");
+            assert!(LRNonStreamingLexerDef::<DefaultLexeme<u8>, u8>::from_str(&src).is_ok());
+        }
     }
 
     #[test]


lrlex/src/lib/parser.rs
```diff
@@ -287,9 +287,21 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
             .find(|c| c == '\n')
             .unwrap_or(self.src.len() - i);
         let line = self.src[i..i + line_len].trim_end();
-        let rspace = match line.rfind(' ') {
-            Some(j) => j,
-            None => return Err(self.mk_error(LexErrorKind::MissingSpace, i)),
+        let rspace = match line.rfind(|ch| ch == '\'' || ch == '"') {
+            Some(x) => match line[0..x].rfind(|ch| ch == '\'' || ch == '"') {
```
SMartinScottLogic, line 291: will this not match non-matching quoting?
e.g. it reads like
.          "a'test 'b"
will produce a _name_ of **b**

SMartinScottLogic: I am starting to get the feeling that this is complex enough that it may be cleaner to switch to regex for rule parsing

ratmice: These mismatched quotes get caught in a later pass when the error `InvalidName` is produced, so while this will make it through this early pass looking for space, it'll fail later.

That is why i hadn't bothered to check that they actually matched, but I think you're probably right e.g. given that "'" appears to currently be a valid rule we should probably check for matching quotes here too.

SMartinScottLogic: Other option: could we use the string parsing from the rust compiler, then we'd support **all** permissible rust strings, which may align with users' expectations.

ltratt: > could we use the string parsing from the rust compiler

Is that something that rustc exposes? [I'm not sure if it's a good idea or not, but my working guess is that either way, we can't actually get at it.]

ratmice: I'll have to think about it a bit, the difficulty in using both something like regex, or a string parsing crate is that this currently needs to be matched in reverse, because the regex itself can contain an unbounded number of ` ` characters.
such as in `([a-c] |[w-z] ) "foo"`, 

I don't see any regex API for matching in reverse, and it seems unlikely that any string parsing crate would have one.
Further is that regex also can contain quotes, so I'm a bit hesitant to do something to the effect of `^.*[ ]("[^"]*"|'[^']*')$`.
because it seems like it could match in the middle of a regex perhaps like `[ "] "space_or_quote"`.

So I think that this, being easiest while matching in reverse is probably the biggest hurdle to doing something more principled here.

ltratt: > So I think that this, being easiest while matching in reverse is probably the biggest hurdle to doing something more principled here.

That sounds plausible.

> I think you're probably right e.g. given that "'" appears to currently be a valid rule we should probably check for matching quotes here too.

Good idea.

ratmice: So, I believe this is resolved in https://github.com/softdevteam/grmtools/pull/335/commits/6606b79e505e9cc1f1d19eabbf44d54ada9b9156 there is some info in the commit message.

But essentially this attempt to match the space before the quotes turned out to just not be possible.
At the same time the additional error checking needed to cover `a "` can also cover `a " "`.

So I reverted back to just finding the right-most space, and added tests for some cases that the first patch in this series had broken by skipping over the rightmost space and finding one in a regex.

ratmice: I suppose one other thing which we could consider is requiring escaping for `"`, `'` and `' '` in a regex.
So these would become '\"`, etc. with that we could search forwards to find the first non-escaped space.

It may produce better errors for these invalid name situations, but less readable regular expressions and compatibility.
So there is definitely a trade off in doing so.

ltratt: This looks like a meaningful improvement to me -- @SMartinScottLogic ?

SMartinScottLogic:
Looks good to me - small but significant improvement to erroneous edge cases

lrlex/src/lib/parser.rs
```diff
@@ -750,34 +737,54 @@ mod test {
...
-        LRNonStreamingLexerDef::<DefaultLexeme<u8>, u8>::from_str(src).expect_error_at_line_col(
-            src,
-            LexErrorKind::InvalidName,
-            2,
-            3,
-        );
+        let srcs = [r#"a ""#, "a '", r#"a '""#, r#"a "'"#];
+        for line_two in srcs {
+            let src = format!("%%\n{line_two}");
+            eprintln!("lexing: {src} ");
```
ltratt, line 743: I think we need to remove this?

ratmice: I don't mind removing them if you'd like I left them for 2 reasons

1. Cargo by default will capture `stderr` not displaying it unless you run `cargo test -- --nocapture` (IIRC)
    or if the test fails in which it displays a portion of the scrollback, so you can see the src in question.
2. Since we currently aren't actually printing the source line where failure occurrs.
    and this runs through multiple source files in a loop when it fails it is currently it can be difficult to see which
    item in the array actually caused the failure.

So, I'd left it in because I thought it was harmless, and helpful.  But I could remove it and file an issue for printing the source lines on error instead, which would probably be a better fix.

ltratt: I quite often use `--nocapture` when I'm debugging, so I've developed an allergy to unnecessary chattiness :) If/when a test fails, I think it's reasonable to expect a developer to have to put various `print`s in to work out what's going on, but we can leave it to that future person :)

ratmice: Aha yes, I do see how that could be annoying then, will remove.

ltratt:
Please squash.

ratmice:
Squashed.

ratmice:
Build failure looks pretty unrelated to the patch, but I haven't had any luck in reproducing it locally, 
by running `RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps` anywhere. Either because of a 
rustdoc version difference locally, or because I haven't figured out the right directory to run it in.

ltratt:
With luck once https://github.com/softdevteam/grmtools/pull/339 this problem should go away (I think it's because we've just picked up rust 1.63.0 in CI).

PR #336, opened 2022-08-10
Fix panic on empty lex file

ratmice:
I hadn't run across any panics while editing after my other PR.  Figured I might run https://rust-fuzz.github.io/book/ with lrlex,
which found one more panic involving empty files.

Fixing that, both AFL and the random input fuzzers have been running lrlex while without incident.
with AFL given a corpus of various input files collected from github, and erroneous inputs from the testsuite.

With both fuzzers catching this and the panic from my previous PR fairly quickly. I haven't really looked at doing so for yacc yet though.

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -157,7 +157,7 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
         loop {
             i = self.parse_ws(i)?;
             if i == self.src.len() {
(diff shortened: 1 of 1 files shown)


ltratt:
I have a suspicion this might not pass `cargo fmt` but let's see!

bors r+

ratmice:
That should be fixed in https://github.com/softdevteam/grmtools/pull/336/commits/30116601bf32c218d96153bf129cb2496ac4dcca

ltratt:
Please squash.

ratmice:
Squashed

PR #338, opened 2022-08-12
Fix doc links to macros

ratmice:
This should fix the build failure encountered in https://github.com/softdevteam/grmtools/pull/335
I was able to reproduce it after updating my toolchain..


ratmice:
Closing in favor of #339

PR #340, opened 2022-08-12
Use the offset at EOF for LexErrorKind::PrematureEnd.

ratmice:
Previously this used the character before EOF. While there
is currently no way to produce this error such that it ends
in a multibyte character. This makes the behavior match yacc.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -1037,7 +1037,11 @@ mod test {
                             .map(|span| line_col!(src, span))
                             .collect::<Vec<(usize, usize)>>(),
                         lines_cols.collect::<Vec<(usize, usize)>>()
-                    )
+                    );
+                    // Check that it is valid to slice.
+                    for span in e.spans() {
+                        let _ = &src[span.start()..span.end()];
+                    }
                 }
                 Err(e) => incorrect_errs!(src, e),
             }
@@ -1054,6 +1058,10 @@ mod test {
                     if errs
                         .iter()
                         .map(|e| {
+                            // Check that it is valid to slice the source with the spans.
+                            for span in e.spans() {
+                                let _ = &src[span.start()..span.end()];
+                            }
                             (
                                 e.kind.clone(),
                                 e.spans()
@@ -1477,6 +1485,12 @@ A:
             &src,
         )
         .expect_error_at_line_col(&src, YaccGrammarErrorKind::PrematureEnd, 1, 17);
+        let src = "// 🦀".to_string();
+        parse(
+            YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
+            &src,
+        )
+        .expect_error_at_line_col(&src, YaccGrammarErrorKind::PrematureEnd, 1, 5);
     }
 
     #[test]
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -157,7 +157,7 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
         loop {
             i = self.parse_ws(i)?;
             if i == self.src.len() {
-                break Err(self.mk_error(LexErrorKind::PrematureEnd, i.saturating_sub(1)));
+                break Err(self.mk_error(LexErrorKind::PrematureEnd, i));
             }
             if let Some(j) = self.lookahead_is("%%", i) {
                 break Ok(j);
@@ -601,7 +601,11 @@ mod test {
                             .map(|span| line_col!(src, span))
                             .collect::<Vec<(usize, usize)>>(),
                         lines_cols.collect::<Vec<(usize, usize)>>()
-                    )
+                    );
+                    // Check that it is valid to slice the source with the spans.
+                    for span in e.spans() {
+                        let _ = &src[span.start()..span.end()];
+                    }
                 }
                 Err(e) => incorrect_errs!(src, e),
             }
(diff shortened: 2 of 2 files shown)


ratmice:
> While there is currently no way to produce this error such that it ends in a multibyte character.

It looks like this might be a misstatement, just beginning to analyze the results of a fuzzing run where this appears to have failed
and I believe it might be possible to end in a multibyte whitespace character presumably because of `is_whitespace()`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1485,6 +1485,12 @@ A:
             &src,
         )
         .expect_error_at_line_col(&src, YaccGrammarErrorKind::PrematureEnd, 1, 17);
+        let src = "// 🦀".to_string();
```
ratmice, line 1488: While this is a Yacc test rather than a Lex one, It appears to be a nice way to produce this error that doesn't involve invalid code gen.

I added it here, because i'm not certain that accepting multibyte whitespace is actually intended, or something we just accidentally accept.
If that is the case, (I don't believe lex accepts comments), but if we added them we could reproduce PrematureEnd with a comment like this.

ratmice:
@ltratt mind chiming in on whether we should be accepting multibyte whitespace characters here or not?

ltratt:
I didn't know there was such a thing as multibyte whitespace! Since we do aim to support unicode properly, I think the right thing to do is to deal with multibyte whitespace properly, but I can be persuaded otherwise.

ratmice:
Okay, I believe that the extent of that might be just changing the `parse_ws` function to use

```
if c.is_whitespace() {
 break 
}
```

But I'll have to see what the callers think of that change.

ratmice:
Contrary to the above, I believe it would be worthwhile to just follow https://www.unicode.org/reports/tr31/ which is intended for parsing/lexing of unicode
and the Pattern_Whitespace property defined in, https://www.unicode.org/Public/13.0.0/ucd/PropList.txt

In addition to e.g. 'is_whitespace' we would need to identify the SpaceSeparator `Zs` class subset of Whitespace or Pattern_Whitespace.

For Pattern_Whitespace this SpaceSeparator looks to be just the typical ASCII space character. Pattern_Whitespace *does* contain some multibyte characters still, e.g. LineSeparator and ParagraphSeparator so we'd still encounter this issue with those.

There is an issue in unicode-xid, pointing to some rustc code which might be helpful,
(Or I could try and solve it there, if you feel like having it as a dependency would be ok, it looks like we actually already depend on it for the serde feature).
https://github.com/unicode-rs/unicode-xid/issues/1

Anyhow it seems to me, this would be the relevant unicode guidelines

ltratt:
Unicode makes my head hurt, so I'm happy to go with what you think best (provided it doesn't break anything obvious, and we have some new tests to make sure we don't introduce new oddities)!

ratmice:
So, I think one of the reasons it makes sense not to use `is_whitespace()` is things like Ogham Space Mark,
which is to be read right to left (or bottom to top I would guess?), so it is kind of strange to allow it in something that isn't really right-to-left etc.

Anyhow, editor support for all these non-ASCII characters is pretty terrible, i'm really not sure how I would actually feel about people using them. But even if we decide to not use the `\p{Pattern_white_Space}` things, it at least *should* be fairly simple now to change them to whatever we decide on and handle it uniformly across the code base.

The regex's could be a bit nicer, but I don't think there is a `\p{gc=Space_Separator}` etc in rusts regex, I at least couldn't find anything... :shrug:

lrlex/src/lib/parser.rs
```diff
@@ -1311,4 +1320,16 @@ a\[\]a 'aboxa'
         }
         LRNonStreamingLexerDef::<DefaultLexeme<u8>, u8>::from_str(&src).ok();
     }
+
+    /// Test with various /// [Pattern_White_Space](https://unicode.org/reports/tr31/) separators.
```
ltratt: I'm not sure about the inner `///`?

ratmice: Fixed in https://github.com/softdevteam/grmtools/pull/340/commits/7635044013a546b6c90e145c7cbcc650d3de337a

ltratt:
I had to look up Ogham -- I had no idea it existed!

It feels to me like this PR probably does enough as-is to make the situation meaningfully better. If anyone does try to write lex/yacc files in the same manner as 4th-6th century Irish, we can revisit this. [Via the wonders of wikipedia I now know that I can find such writing in Cornwall...]

ltratt:
I think this is ready for squashing? If so, please go ahead.

lrlex/src/lib/parser.rs
```diff
@@ -221,7 +232,7 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
                 self.start_states.push(start_state);
             }
         }
-        Ok(off + line_len)
+        self.parse_ws(i)
```
ratmice, line 235: The purpose of this change is the `let src = "%S X ";` test in, which I had forgotten to re-add.
Otherwise this accepts `X` as a state name and skips right over the `En Space` at the end of the line without producing an error, there might be a better error we could do here, if test something like`line_end == self.parse_ws(i)`.

But I need a better understanding of what has gone awry in that test.

https://github.com/softdevteam/grmtools/pull/340/commits/7635044013a546b6c90e145c7cbcc650d3de337a

ratmice:
Before I squash, 

I would like to understand why in that last testcase I added, the first one gets `InvalidStartStateName`, while the second one gets `UnknownDeclaration`, I think it would be ideal if they both got `InvalidStartStateName`. but i'll have to look at it tomorrow.

lrlex/src/lib/parser.rs
ratmice: Should be fixed in https://github.com/softdevteam/grmtools/pull/340/commits/54825ad126c18ee75def82d06d38986e79d6f8d6 This is an issue which was much more prevalent in the code than I had anticipated, essentially we need to use `trim_matches` instead of `trim` and friends, which work on the unicode whitespace.

ltratt:
Please squash.

ratmice:
Squashed.

PR #341, opened 2022-08-14
Add operations to start condition processing.

SMartinScottLogic:


SMartinScottLogic:
Pending: tests in `lrlex/src/lib/lexer.rs`

Question: If the input to `fn lexer(...)` leaves the lexer with a non-empty stack, should it be retained for (optional) resumption with a following call to the lexer?

ratmice:
Interesting question, I know that if it can be retained that behavior being optional seems necessary.
my nimbleparse_lsp project generally should assume files start in the `Initial` state, and calls `lexer` multiple times.

But for streaming lexers (which I don't believe exist), some retention between `&str`'s would seem necessary but that would likely *not* be the `lexer()`  function, but a function on `lexer()`'s return value, akin to the `NewlineCache::feed()` function.

One place I could imagine using this is in my crimson project, each file currently starts with an optional header/schema where type declarations go, which specifies the type of the top-level value.  Something to retain the stack would allow me to specify a header/schema at the top of the first *file* then disallow it for subsequent files.  To ensure things aren't sending redundant data or changing the schema between files.

So I guess my personally I think that if it isn't troublesome to implement, it seems like it could be an interesting feature.

ltratt:
> Question: If the input to fn lexer(...) leaves the lexer with a non-empty stack, should it be retained for (optional) resumption with a following call to the lexer?

The intention (though the API is definitely not complete for it!) would be that this is a "streaming lexer" (vs. the current `NonStreamingLexer`s).

SMartinScottLogic:
My understanding from the above, is that it would be worth extending the return value from `lexer()` to include details of the final state of the start condition stack. I like it, it will help with tests as well!

ratmice:
So, I think through it maybe that `lexer()` might not be the place where this is needed, as that takes the *.l file* input, rather than
The input which is going to be performing start condition operations, perhaps I am misunderstanding something about your question though, but I don't see how the input to `lexer()`.

Once I look at the `LRNonStreamingLexer` though and it's combination of `new()` and `iter()`, I don't see a good way to expose it without, turning that into an something like `Iterator<Item=Result<Lexeme ...>, StartState)>` , exposing the target state at every iteration.

Perhaps you have something in mind, but now that I think i'm looking in the right place I don't quite see any convenient way to just get just the final state, without doing something that I imagine might add overhead to each iteration.  So, I'm no longer sure if this is something we can do right now, or just need to keep in mind when designing the "streaming lexer", like Laurie mentioned.

ltratt:
I'm willing to be proven wrong on this, but I designed the `NonStreaming` interface to be as simple as possible, knowing that it would almost certainly be too simple for streaming. I don't know what the streaming interface should look like: it came under "think about it one day in the far future" :)

For example, if the current lexer can return "half way through", I think we would force the user to add an extra check "the lexer stack is empty" as a proxy for "was my file completely lexed"? That seems like something most people (I assume `NonStreaming` is the more common case!) would not expect to have to deal with themselves.

lrlex/src/lib/lexer.rs
```diff
@@ -281,13 +284,37 @@ impl<
                         }
                     }
                 }
-                if let Some(target_state_id) = r.target_state {
-                    state = match self.get_start_state_by_id(target_state_id) {
+                if let Some((target_state_id, op)) = &r.target_state {
+                    // TODO Handle state stack operations
```
ratmice, line 288: Still a TODO, seems like it is?

lrlex/src/lib/ctbuilder.rs
```diff
@@ -403,7 +403,7 @@ where
         write!(
             outs,
             "{mod_vis} mod {mod_name} {{
-use lrlex::{{LexerDef, LRNonStreamingLexerDef, Rule, StartState}};
+use lrlex::{{LexerDef, LRNonStreamingLexerDef, Rule, StartState, StartStateOperation}};
```
ratmice, line 406: I think we were moving to fully qualified paths here in generate code, so i believe this should be `::lrlex::StartStateOperation` below, besides these minor things I think this is looking good!

ratmice (changes requested):

lrlex/src/lib/parser.rs
```diff
@@ -362,6 +372,15 @@ impl<StorageT: TryFrom<usize>> LexParser<StorageT> {
         Ok(i + line_len)
     }
 
+    fn parse_start_state_ops<'a>(&self, start_state_str: &'a str) -> (&'a str, StartStateOperation) {
+        let (left_delta, operation) = match start_state_str.chars().nth(0).unwrap_or_default() {
```
ratmice, line 376: Seems, I lost one comment in my review, 

`cargo clippy --all-targets` is going to want `nth(0)` -> `next()`, here and in the tests

lrlex/src/lib/lexer.rs
```diff
@@ -242,14 +243,16 @@ impl<
     ) -> LRNonStreamingLexer<'lexer, 'input, LexemeT, StorageT> {
         let mut lexemes = vec![];
         let mut i = 0;
-        let state = self.get_start_state_by_id(0);
-        let mut state = match state {
+        let mut state_stack: Vec<(usize, &StartState)> = Vec::new();
+        let state = match self.get_start_state_by_id(0) {
```
ratmice, line 247: I notice there is one more use of this, `state` variable, which should be using the top of the stack.

Perhaps this `state` variable could be in it's own `{ let state = ...; ... state_stack.push((1, push)) }` block so `state` goes out of scope? I don't think it needs to be in scope after initialization.

An alternative might be putting the initial state into a closure and calling `vec![initial_state()];` or something.

lrlex/src/lib/lexer.rs
```diff
@@ -281,13 +284,37 @@ impl<
...
+                        StartStateOperation::Push => {
+                            match head {
+                                Some((count, s)) if s.id == state.id => *count += 1,
+                                _ => state_stack.push((1, state)),
+                            }
+                        },
+                        StartStateOperation::Pop => {
+                            match head {
+                                Some((count, s)) if *count > 1 => {*count -= 1;},
+                                Some(_) => {state_stack.pop();},
```
ratmice, line 311: Should this be something to the effect of:

```
state_stack.pop();
if state_stack.is_empty() { state_stack.push((1, initial_state)) }
```

So there is always an entry in the state_stack?

ratmice:
I pushed a commit here, based on trying to get this working in comments example if it helps, but most of it is mentioned in my review.
https://github.com/ratmice/grmtools/commit/67a57c6137a857ca13a3164590ad65d9c1c8df9d

lrlex/src/lib/lexer.rs
ratmice: So, I guess that is a real question of whether we should fall back to the initial state on an empty stack, or throw an error when there is an empty stack.  I feel, like with the default behavior being replace, it is going to be very easy to `ReplaceStack` and then `Pop`,

Here https://github.com/ratmice/grmtools/commit/0bc7c1c682e26315af40a34bfb976be2d38a7b72 was a second attempt to simplify things around this, by using `.unwrap_or(initial_state)`: 

The nice thing about it I guess is if you aren't using start states, you won't need to allocate the stack,
I'm not sure if this leads to any win in regards to not having to go to the heap if start_stack is empty.

Anyhow I think you should take this suggestion as something that we may need to decide between producing an error on an empty stack is the right behavior or not.

ltratt:
Hello @SMartinScottLogic, I wondered how this PR is getting on?

SMartinScottLogic:
Sorry, I've been on family vacation - I'll aim to get the next commits in in the next couple of days

ltratt:
Hello @SMartinScottLogic, are you still chugging away on this one?

ltratt:
@ratmice has very patiently and kindly explained what's going in this PR to the point that I've realised my comment in https://github.com/softdevteam/grmtools/pull/341#issuecomment-1214969503 was entirely bone-headed! This PR has nothing to do with streaming / non-streaming and I'm an idiot!

One thing that Matt and I have realised -- although we haven't thrashed out all the details -- relates to the comment in https://github.com/softdevteam/grmtools/pull/341#issuecomment-1214716803. The return type needs to have something like an `Option<LexingState>` attached to it. My working guess is that we might just be able to do something as simple as extending `LexError`:

```rust
pub struct LexError {
  span: Span,
  lexing_state: Option<LexingState>
}
```

Basically I think I'd like to get the functionality of this PR into grmtools before we make the next release if possible. I think there are people who will find this useful, and if (as is possible) it requires an API tweak or two, we can bite that bullet alongside the other API changes we have on the master branch.

lrlex/src/lib/lexer.rs
ratmice: I think as per https://github.com/softdevteam/grmtools/pull/341#issuecomment-1277830747
we can disregard my suggestions here, and this can just produce a `LexError { lexing_state: None, ...}`

ltratt:
@SMartinScottLogic I'd love to see this PR get into grmtools before we do the next release (hopefully soon). Is that plausible, do you think?

SMartinScottLogic:
Sorry for the delay on this one - I've had a number of distractions come up in my personal and professional life in the last few months. I'll hopefully be able to address the concerns on this PR in the next week or so.

ltratt:
Excellent :)

lrlex/src/lib/lexer.rs
SMartinScottLogic: Resolved in f3794c8343d0c1ee1980237527b5fd8218290ad1

lrlex/src/lib/ctbuilder.rs
SMartinScottLogic: Fixed under ddaae44153ae8441522a4b20b8a7593f22a9eb6c

lrlex/src/lib/parser.rs
SMartinScottLogic: Done, in 6da39479cf110d8bc9ffd9dea3a2bcf0badc8f8d

lrlex/src/lib/lexer.rs
SMartinScottLogic: Adjusted under 8a321950eab128b3f694d58f3aff61532bf86ef4

SMartinScottLogic:
> I guess, I have one minor question that isn't quite about this patch the example, _lrpar/examples/start_states/comment.y_ Wondering if it is intended, or if it would be better to include epsilon in the rules, so that comments alone/lines with no text parse successfully, something like:
> 
> ```
> %start Expr
> %%
> Expr: Expr Text | ;
> Text: 'TEXT';
> ```

Changed under 145f8d68eaf5aa0de3338ea56e692211f889f8cc

SMartinScottLogic:
Could you clarify exactly what you'd expect to be in `LexingState` under https://github.com/softdevteam/grmtools/pull/341#issuecomment-1277830747 ?

ratmice:
@SMartinScottLogic  I believe that adding the `Option<LexingState>`, is for the following cases, where it can be none

```
 // On receiving 'a' clear the implicit INITIAL state from the stack then push the B state, stack looks like [B].
a <B>'a' 
 // on receiving 'b' in the B state, Pop the B state, leaving an empty stack due to push/pop imbalance.
<B>b <-B>'b' 
```

```
  // on receiving 'a' when in the INITIAL state Pop the INITIAL state off stack leaving an empty stack.
<INITIAL>a <-INITIAL>'A' 
```

Initially these caused a panic,  I had suggested that these could fall back to the initial state via `unwrap_or(initial_state)`,
But I believe the final suggestion was to make these return a `LexError`, with a `lexing_state: None`

ltratt:
I think @ratmice has summed up my thinking too. Apart from that I don't have any major comments.

SMartinScottLogic:
I understand the failure cases in https://github.com/softdevteam/grmtools/pull/341#issuecomment-1304813657, what I'm unsure of is - in other cases where we produce a `LexError` (i.e. where `lexing_state` should be present) what type is within the `Option`? `LexingState` is **not** an existing type - shall I create a new type, and put current start state details in for the current PR?

ltratt:
Good question, and I don't immediately have a good suggestion. @ratmice might?

ratmice:
Aha, I was thinking just the `id:usize`, with the contents being `state_stack.last()` and the user would then have to call `get_start_state_by_id`, but now I see that it isn't `pub` (I'm running a bit behind on a holiday so I don't have a chance to see if there are any difficulties with making it `pub` though).

There are also some questions if we are making the `id` public if it should also have an `idxnewtype` associated with it I suppose?
This at least seemed better to me than making the `StartState` clone, and putting that directly in the LexError?

ltratt:
> There are also some questions if we are making the id public if it should also have an idxnewtype associated with it I suppose?

What do we expect people to do with that state? Is a fully opaque state OK or do they need to peek at the contents to make use of it?

ratmice:
> > There are also some questions if we are making the id public if it should also have an idxnewtype associated with it I suppose?
> 
> What do we expect people to do with that state? Is a fully opaque state OK or do they need to peek at the contents to make use of it?

I think it is just needed to get a `struct StartState` via `get_start_state_by_id`, and so fully opaque ID would seem fine, once they get the `struct StartState`, primarily the printing of the `name`, and `name_span` fields I believe just so you can discern which state the lexer was in when the error ocurred.

ltratt:
Ah, I see, so I *think* we're saying:

1. Ideally we want to return `Option<StartState>`.
2. But that would require `clone`ing a `StartState` which commits us to an ugly API.

I think I would thus suggest:

* Adding a `StartStateId` opaque type, that's currently a wrapper around `usize`.
* Converting `fn get_start_state_by_id(&self, id: usize) -> Option<&StartState>` to `pub fn get_start_state_by_id(&self, id: StartStateId) -> Option<&StartState>`.
* Adding a `name(&self) -> &str` (or similar) and `name_span(&self) -> Span` methods to `StartState` so users can work out what start state caused the problem.

Does that sound sensible?

ratmice:
> Does that sound sensible?

Yeah, sounds alright to me.

SMartinScottLogic:
Sorry, I'm not understanding the 'opaque type' bit above - I haven't encountered this in rust - so I'm not certain of the desired code.
I understand and agree with the remainder, and I've largely completed (using usize for StartStateId at the moment) - c.f. 224744d96caae7d33322b25ddd73666d196110ea

lrlex/src/lib/lexer.rs
```diff
@@ -293,15 +293,16 @@ impl<
...
                             break;
                         }
                     }
                 }
                 if let Some((target_state_id, op)) = &r.target_state {
                     let state = match self.get_start_state_by_id(*target_state_id) {
                         None => {
-                            lexemes.push(Err(LexError::new(Span::new(old_i, old_i))));
+                            // TODO: I can see an argument for lexing state to be either `None` or `Some(target_state_id)` here
+                            lexemes.push(Err(LexError::new(Span::new(old_i, old_i), None)));
```
ratmice, line 305: Hmm, is this case with the TODO normally reachable when using lrlex, e.g. for `state` to be None at this point wouldn't one of `LexErrorKind::UnknownStartState`, or `InvalidStartState` would have previously triggered, implying some internal inconsistency around target_state_id?

Quite possible i'm misreading this, but if so maybe some thing like  `assert_eq!(None, Some(target_state_id))`, because I'm not sure what the user could do with a `StartStateId` if it isn't possible to get a `StartState` from it here, which looks to be the case.

ratmice: This might be another way of asking if we should just `self.get_start_state_by_id(*target_state_id).unwrap()`?

lrpar/src/lib/lex_api.rs
```diff
@@ -5,15 +5,28 @@ use std::{cmp, error::Error, fmt, hash::Hash, marker};
 use cfgrammar::Span;
 use num_traits::{PrimInt, Unsigned};
 
+#[derive(Copy, Clone, Debug)]
+pub struct StartStateId {
+    id: usize,
+}
```
ratmice, line 11: So I'm now seeing there is a bit of difficulty because we're cross-crate issues, in getting a `StartState` here, from `StartStateId`. Compounded somewhat by *lrpar* not depending on *lrlex*.  I also noticed that `get_start_state_by_id` is in a trait `LexerDef` so seems to be implicitly `pub` already.

Just not sure how to deal with this situation where `LexError`/`StartStateId` are in *lrpar*, but `StartState` is in *lrlex*.
I'm going to have to think on it.

SMartinScottLogic: Okay, I'll hold off on further code-churn while you consider!

ratmice: Well, I pushed a first attempt here which just makes a bunch of accessor functions for `LexError`/`StartStateId` https://github.com/ratmice/grmtools/tree/start_state_ids
Which seems to work i'm not sure it is super elegant, I hadn't really anticipated that we can't just use `StartStateId.id` directly inside lrlex.

I guess though, where I'm having trouble with this, is after the addition of these accessors + `LexError::new` + `StartState::try_from` with infallible errors, we don't actually gain much in the way invariants ensuring that `StartStateId`s are valid.

However I guess they still ensure that it is difficult to mix them up with arbitrary usizes?

ltratt: > Just not sure how to deal with this situation where LexError/StartStateId are in lrpar, but StartState is in lrlex.

Hmm, yeah, good point -- I hadn't thought about that.

What operations does a user need to do on a `StartState`? Is it just "compare two state states for equality and print it out for debugging purposes"? If so, we could e.g. make `LexError` something like:

```rust
pub struct LexError<StartStateId> {
    span: Span,
    lexing_state: Option<StartStateId>,
}
```

and then lrlex can return `LexError<StartStateId>` where it defines `struct StartStateId(usize)` or similar?

[Warning: this might be a terrible idea! It's off the top of my head.]

ratmice: It does look like adding that as a generic is not going to be a walk in the park, as it entails threading the generic throughout everything starting with the LexError within https://softdevteam.github.io/grmtools/master/api/lrpar/trait.Lexer.html

From there it spreads pretty heavily throughout all the other generics, in ways which I haven't succeeded in adding it to yet.

ltratt: Another thing we could do if we wanted is have a `trait StartStateId` and `lexing_state: Option<Box<dyn StartStateId>>` but we'd to know what functionality a `StartStateId` should have. If it's just `Eq + Debug + Display`, that should be easy?

ratmice: Well, part of the difficulty here is, you need a reference to the concrete `LRNonStreamingLexerDef`, to actually `Debug`/`Display` anything meaningful.  If we revisited putting `StartState` directly in `LexError`, we would have something meaningful to print for `Debug`/`Display` though, it might also want to be `Eq + Debug + Display + Spanned`.

E.g. the `Debug` impl we're after first requires a call to something like `fn (lex: lrlex::LrNonStreamingLexerDef, state: lrpar::StartStateId) -> lrlex::StartState`.  Because `StartStateId` is just kind of a internal implementation detail which only has meaning to the lexer itself, and not so much to the user.

ltratt: If it's purely internal, is there even any point exposing this to the user?

[In retrospect, I suspect the `Copy` on `LexError` was a mistake. I don't know how difficult removing it now might be though...]

ratmice: Yeah currently there is a point, in that we can get something relevant to the lex file from it.
But if we can just embed `StartState` directly, there is no point in exposing it.

When developing lex files with start states here, I've found it can be fairly confusing when you get a `LexError`, and a regex doesn't match the regex you expect, but not because the regex is wrong but because the lexer is in a different state than you expect.

So I think it is important to give a hint here regarding which state the lexer was in, because the typical reaction is that there is something wrong about the regex.

ltratt: OK, so then I guess the question is: should `StartStartId` just be a "token" (no pun intended) that you can hand back to the lexer to say "give me the name of the state you're in"? Because if that's all we have to do with it, we can probably hack it with a trait that's an `Any` (it's not pretty... but it would work). Either way, I think I'm inclined to get this PR in if we're happy with it, and then fix the `LexError` issue later. Any thoughts?

ratmice: That is kind of how https://github.com/ratmice/grmtools/commit/ca3444534b84a8dca736e9275b7bfe376bf9ce4c worked except `StartStateId` is concrete rather than `Any`, from there the name can be gotten by `lexer.error_lexing_state(err).map(|state| state.name())` or something to that effect.

So yes something along those lines of it being just a token seems like it should work fine.
I'd also be fine if we want to fix the `LexError` issue in another patch, and am happy with the state of things in this one.

ltratt:
@SMartinScottLogic Please squash (you'll need to rebase against master and resolve whatever the conflict(s) is/are).

SMartinScottLogic:
Squashed, as requested

ltratt:
I think you'll need to add a new commit for `cargo fmt`.

ratmice:
I thought I was assigned, but perhaps I'm misremembering. Anyway I don't think bors will listen to me because of that.

PR #344, opened 2022-08-20
Fix paste issue with iter_stidx() documentation

ratmice:


lrtable/src/lib/stategraph.rs
```diff
@@ -36,8 +36,8 @@ where
         self.start_state
     }
 
-    /// Return an iterator which produces (in order from `0..self.rules_len()`) all this
-    /// grammar's valid `RIdx`s.
+    /// Return an iterator which produces (in order from `StorageT::zero()..self.all_states_len()`)
+    /// all this grammar's valid `StIdx`s.
```
ltratt, line 40: The code below says `self.states.len()`?

ratmice: Sorry, I hadn't noticed the other uses of `self.states.len()`, my thinking was that the `states` field is private, while `all_states_len()` is public, however it brings with it the `zero()` weirdness.

Could change it to match the others, or vice versa. Whichever you prefer

ltratt: Good spot -- we should only document using the external API!

PR #345, opened 2022-08-23
Unused symbol

ratmice:
I wasn't having much luck with *Unused due to conflicts*, which I believe i'm starting to understand why bison's production of that warning is limited (to Shift/Reduce conflicts). So here is a prototype for warnings, which adds an `unused symbol` warning for yacc which are *not* due to conflicts.  

It tries to include a mechanism through which we can also produce warnings as errors, and largely follows from the all the YaccGrammarError work that has been done minus the `Error` trait.

A few things are noteworthy `spanskind` wanted to be on `YaccGrammarWarningKind` as well as `YaccGrammarWarning`,
so that `YaccGrammarError` could leverage it when treating warnings as errors, `SpansKind::Error` is also an awkward choice of naming here.

There is some subsequent patches left to do:
1. Integrating this into various tools
2. ~~I noticed that `complete_and_validate` is still returning a `YaccGrammarError`, and that we should probably make it that `Vec<YaccGrammarError>`.~~ Realized this is just `pub(crate)` not `pub`, so can do this whenever.
3. How or if we should pass a `warnings_as_errors` flag to `new_with_storaget`, and whether the place I implemented `warnings()` on `YaccGrammarError` is actually the right place, or if this is something that *callers* should do using the `From` impl given.
4. When we `complete_and_validate()` and there *are* errors, this code is only callable too late to actually have both warnings and errors, but implementing it to be callable earlier seemed tougher since it relies on PIdx's a bit.

As such I'm going to mark this as draft pre-emptively, since it seems like there is some thought that needs to go into
what the best thing to do here is, and whether this is even worth the additional types/impls?

### DIFF
--- a/cfgrammar/Cargo.toml
+++ b/cfgrammar/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/lib/mod.rs"
 
 [dependencies]
 lazy_static = "1.0"
-indexmap = "1"
+indexmap = "1.9"
 num-traits = "0.2"
 regex = "1"
 serde = { version="1.0", features=["derive"], optional=true }
--- a/cfgrammar/Cargo.toml
+++ b/cfgrammar/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/lib/mod.rs"
 
 [dependencies]
 lazy_static = "1.0"
-indexmap = "1.9"
+indexmap = "1"
 num-traits = "0.2"
 regex = "1"
 serde = { version="1.0", features=["derive"], optional=true }
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -234,109 +234,83 @@ impl GrammarAST {
     }
 
     pub(crate) fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
-        #[derive(Hash, PartialEq, Eq, Debug, Copy, Clone)]
-        enum SymbolKind {
-            Rule,
-            Token,
-        }
         let start_rule_name = self.start.as_ref().map(|(name, _)| name.clone());
         let start_rule = self
             .rules
             .iter()
             .find(|(rule_name, _)| start_rule_name.as_ref() == Some(rule_name));
-        let mut used_symbols = HashSet::new();
-        let mut expect_unused = self
-            .expect_unused
-            .iter()
-            .filter_map(|sym| match &sym {
+        let mut seen_rules = HashSet::new();
+        let mut seen_tokens = HashSet::new();
+        let mut expected_unused_tokens = HashSet::new();
+        let mut expected_unused_rules = HashSet::new();
+        for sym in &self.expect_unused {
+            match sym {
                 Symbol::Rule(sym_name, _) => {
-                    if self.rules.contains_key(sym_name) {
-                        Some((sym_name, SymbolKind::Rule))
-                    } else {
-                        None
-                    }
+                    expected_unused_rules.insert(sym_name);
                 }
                 Symbol::Token(sym_name, _) => {
-                    if self.tokens.contains(sym_name) {
-                        Some((sym_name, SymbolKind::Token))
-                    } else {
-                        None
-                    }
+                    expected_unused_tokens.insert(sym_name);
                 }
-            })
-            .collect::<Vec<(&String, SymbolKind)>>();
+            }
+        }
         // If a rule is specified in `%expect-unused`, also add the tokens associated with it.
         // This perhaps should add rules as well, but i'm not quite certain.
-        let mut extra_unused = Vec::new();
-        for (unused_sym, _) in expect_unused.iter() {
-            if let Some(rule) = self.rules.get(*unused_sym) {
+        for unused_rule in expected_unused_rules.iter() {
+            if let Some(rule) = self.rules.get(*unused_rule) {
                 for pidx in &rule.pidxs {
                     for symbol in &self.prods[*pidx].symbols {
                         if let Symbol::Token(tok, _) = symbol {
-                            extra_unused.push((tok, SymbolKind::Token));
+                            expected_unused_tokens.insert(tok);
                         }
                     }
                 }
             }
         }
-        expect_unused.extend(extra_unused);
-
         if let Some(implicit_tokens) = self.implicit_tokens.as_ref() {
-            expect_unused.extend(
-                implicit_tokens
-                    .iter()
-                    .map(|(key, _)| (key, SymbolKind::Token)),
-            )
+            expected_unused_tokens.extend(implicit_tokens.keys())
         }
-
         if let Some((start_name, start_rule)) = start_rule {
-            let mut queue = Vec::new();
-            let start_sym = (start_name, SymbolKind::Rule);
-            queue.extend(start_rule.pidxs.iter().copied());
-            used_symbols.insert(start_sym);
+            let mut todo = Vec::new();
+            todo.extend(start_rule.pidxs.iter().copied());
+            seen_rules.insert(start_name);
 
-            while let Some(pidx) = queue.pop() {
+            while let Some(pidx) = todo.pop() {
                 let prod = &self.prods[pidx];
                 for sym in &prod.symbols {
-                    let (sym_name, sym_kind) = match sym {
-                        Symbol::Rule(name, _) => (name, SymbolKind::Rule),
-                        Symbol::Token(name, _) => (name, SymbolKind::Token),
-                    };
-
-                    if used_symbols.insert((sym_name, sym_kind)) && sym_kind == SymbolKind::Rule {
-                        if let Some(rule) = self.rules.get(sym_name) {
-                            queue.extend(&rule.pidxs);
+                    match sym {
+                        Symbol::Rule(name, _) => {
+                            if seen_rules.insert(name) {
+                                if let Some(rule) = self.rules.get(name) {
+                                    todo.extend(&rule.pidxs);
+                                }
+                            }
                         }
-                    }
+                        Symbol::Token(name, _) => {
+                            seen_tokens.insert(name);
+                        }
+                    };
                 }
             }
         }
         self.rules
             .iter()
             .filter_map(|(rule_name, rule)| {
-                if expect_unused.contains(&(rule_name, SymbolKind::Rule)) {
+                if expected_unused_rules.contains(rule_name) || seen_rules.contains(rule_name) {
                     None
                 } else {
-                    Some((rule_name, SymbolKind::Rule, rule.name.1))
+                    Some(rule.name.1)
                 }
             })
             .chain(self.tokens.iter().enumerate().filter_map(|(tok_idx, tok)| {
-                if expect_unused.contains(&(tok, SymbolKind::Token)) {
+                if expected_unused_tokens.contains(tok) || seen_tokens.contains(tok) {
                     None
                 } else {
-                    Some((tok, SymbolKind::Token, self.spans[tok_idx]))
+                    Some(self.spans[tok_idx])
                 }
             }))
-            .filter_map(|(sym_name, kind, span)| {
-                let used_sym = used_symbols.get(&(sym_name, kind));
-                if used_sym.is_none() {
-                    Some(YaccGrammarWarning {
-                        kind: YaccGrammarWarningKind::UnusedSymbol,
-                        spans: vec![span],
-                    })
-                } else {
-                    None
-                }
+            .map(|span| YaccGrammarWarning {
+                kind: YaccGrammarWarningKind::UnusedSymbol,
+                spans: vec![span],
             })
             .collect::<Vec<_>>()
             .into_iter()
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -29,8 +29,8 @@ pub struct GrammarAST {
     pub expectrr: Option<(usize, Span)>,
     pub parse_param: Option<(String, String)>,
     pub programs: Option<String>,
-    // Unchecked `Symbol` names, with spans pointing into the `%allow-unused` declaration.
-    pub allow_unused: Vec<Symbol>,
+    // Unchecked `Symbol` names, with spans pointing into the `%expect-unused` declaration.
+    pub expect_unused: Vec<Symbol>,
 }
 
 #[derive(Debug)]
@@ -81,7 +81,7 @@ impl GrammarAST {
             expectrr: None,
             parse_param: None,
             programs: None,
-            allow_unused: Vec::new(),
+            expect_unused: Vec::new(),
         }
     }
 
@@ -207,7 +207,7 @@ impl GrammarAST {
             });
         }
 
-        for sym in &self.allow_unused {
+        for sym in &self.expect_unused {
             match sym {
                 crate::yacc::ast::Symbol::Rule(sym_name, sym_span) => {
                     if self.get_rule(sym_name).is_none() {
@@ -241,8 +241,8 @@ impl GrammarAST {
             Rule,
             Token,
         }
-        let allow_unused = self
-            .allow_unused
+        let expect_unused = self
+            .expect_unused
             .iter()
             .filter_map(|sym| match &sym {
                 Symbol::Rule(sym_name, _) => {
@@ -274,7 +274,7 @@ impl GrammarAST {
             .iter()
             .filter_map(|(rule_name, rule)| {
                 if start_rule_name.as_ref() == Some(rule_name)
-                    || allow_unused.contains(&(rule_name, SymbolKind::Rule))
+                    || expect_unused.contains(&(rule_name, SymbolKind::Rule))
                 {
                     None
                 } else {
@@ -282,7 +282,7 @@ impl GrammarAST {
                 }
             })
             .chain(self.tokens.iter().enumerate().filter_map(|(tok_idx, tok)| {
-                if allow_unused.contains(&(tok, SymbolKind::Token)) {
+                if expect_unused.contains(&(tok, SymbolKind::Token)) {
                     None
                 } else {
                     Some((tok, SymbolKind::Token, self.spans[tok_idx]))
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -253,19 +253,6 @@ impl GrammarAST {
                 }
             }
         }
-        // If a rule is specified in `%expect-unused`, also add the tokens associated with it.
-        // This perhaps should add rules as well, but i'm not quite certain.
-        for unused_rule in expected_unused_rules.iter() {
-            if let Some(rule) = self.rules.get(*unused_rule) {
-                for pidx in &rule.pidxs {
-                    for symbol in &self.prods[*pidx].symbols {
-                        if let Symbol::Token(tok, _) = symbol {
-                            expected_unused_tokens.insert(tok);
-                        }
-                    }
-                }
-            }
-        }
         if let Some(implicit_tokens) = self.implicit_tokens.as_ref() {
             expected_unused_tokens.extend(implicit_tokens.keys())
         }
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -28,6 +28,8 @@ pub struct GrammarAST {
     pub expectrr: Option<(usize, Span)>,
     pub parse_param: Option<(String, String)>,
     pub programs: Option<String>,
+    // Unchecked `Symbol` names, with spans pointing into the `%allow-unused` declaration.
+    pub allow_unused: Vec<Symbol>,
 }
 
 #[derive(Debug)]
@@ -78,6 +80,7 @@ impl GrammarAST {
             expectrr: None,
             parse_param: None,
             programs: None,
+            allow_unused: Vec::new(),
         }
     }
 
@@ -201,6 +204,27 @@ impl GrammarAST {
                 kind: YaccGrammarErrorKind::UnknownEPP(k.clone()),
                 spans: vec![Span::new(0, 0)],
             });
+        }
+
+        for sym in &self.allow_unused {
+            match sym {
+                crate::yacc::ast::Symbol::Rule(sym_name, sym_span) => {
+                    if self.get_rule(sym_name).is_none() {
+                        return Err(YaccGrammarError {
+                            kind: YaccGrammarErrorKind::UnknownRuleRef(sym_name.clone()),
+                            spans: vec![*sym_span],
+                        });
+                    }
+                }
+                crate::yacc::ast::Symbol::Token(sym_name, sym_span) => {
+                    if !self.has_token(sym_name) {
+                        return Err(YaccGrammarError {
+                            kind: YaccGrammarErrorKind::UnknownToken(sym_name.clone()),
+                            spans: vec![*sym_span],
+                        });
+                    }
+                }
+            }
         }
         Ok(())
     }
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, fmt};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt,
+};
 
 use indexmap::{IndexMap, IndexSet};
 
@@ -236,7 +239,7 @@ impl GrammarAST {
     }
 
     fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
-        #[derive(Ord, PartialOrd, PartialEq, Eq, Debug)]
+        #[derive(Hash, PartialEq, Eq, Debug)]
         enum SymbolKind {
             Rule,
             Token,
@@ -262,7 +265,7 @@ impl GrammarAST {
                 }
             })
             .collect::<Vec<(&String, SymbolKind)>>();
-        let mut used_symbols = std::collections::BTreeSet::new();
+        let mut used_symbols = HashSet::new();
         self.prods.iter().for_each(|prod: &Production| {
             used_symbols.extend(prod.symbols.iter().map(|sym| match sym {
                 Symbol::Rule(sym_name, _) => (sym_name, SymbolKind::Rule),
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -285,20 +285,22 @@ impl GrammarAST {
                 if expected_unused_rules.contains(rule_name) || seen_rules.contains(rule_name) {
                     None
                 } else {
-                    Some(rule.name.1)
+                    Some(YaccGrammarWarning {
+                        kind: YaccGrammarWarningKind::UnusedRule,
+                        spans: vec![rule.name.1],
+                    })
                 }
             })
             .chain(self.tokens.iter().enumerate().filter_map(|(tok_idx, tok)| {
                 if expected_unused_tokens.contains(tok) || seen_tokens.contains(tok) {
                     None
                 } else {
-                    Some(self.spans[tok_idx])
+                    Some(YaccGrammarWarning {
+                        kind: YaccGrammarWarningKind::UnusedToken,
+                        spans: vec![self.spans[tok_idx]],
+                    })
                 }
             }))
-            .map(|span| YaccGrammarWarning {
-                kind: YaccGrammarWarningKind::UnusedSymbol,
-                spans: vec![span],
-            })
             .collect::<Vec<_>>()
             .into_iter()
     }
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -212,7 +212,7 @@ impl GrammarAST {
 
         for sym in &self.expect_unused {
             match sym {
-                crate::yacc::ast::Symbol::Rule(sym_name, sym_span) => {
+                Symbol::Rule(sym_name, sym_span) => {
                     if self.get_rule(sym_name).is_none() {
                         return Err(YaccGrammarError {
                             kind: YaccGrammarErrorKind::UnknownRuleRef(sym_name.clone()),
@@ -220,7 +220,7 @@ impl GrammarAST {
                         });
                     }
                 }
-                crate::yacc::ast::Symbol::Token(sym_name, sym_span) => {
+                Symbol::Token(sym_name, sym_span) => {
                     if !self.has_token(sym_name) {
                         return Err(YaccGrammarError {
                             kind: YaccGrammarErrorKind::UnknownToken(sym_name.clone()),
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -233,18 +233,13 @@ impl GrammarAST {
         Ok(())
     }
 
-    /// Return an iterator over all the `YaccGrammarWarning` produced by this grammar.
-    pub fn warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
-        self.unused_symbol_warnings()
-    }
-
-    fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
+    pub(crate) fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
         #[derive(Hash, PartialEq, Eq, Debug)]
         enum SymbolKind {
             Rule,
             Token,
         }
-        let expect_unused = self
+        let mut expect_unused = self
             .expect_unused
             .iter()
             .filter_map(|sym| match &sym {
@@ -265,6 +260,11 @@ impl GrammarAST {
                 }
             })
             .collect::<Vec<(&String, SymbolKind)>>();
+
+        if let Some(implicit_tokens) = self.implicit_tokens.as_ref() {
+            expect_unused.extend(implicit_tokens.keys().map(|key| (key, SymbolKind::Token)))
+        }
+
         let mut used_symbols = HashSet::new();
         self.prods.iter().for_each(|prod: &Production| {
             used_symbols.extend(prod.symbols.iter().map(|sym| match sym {
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -231,11 +231,11 @@ impl GrammarAST {
     }
 
     /// Return an iterator over all the `YaccGrammarWarning` produced by this grammar.
-    pub fn warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> + '_ {
+    pub fn warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
         self.unused_symbol_warnings()
     }
 
-    fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> + '_ {
+    fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
         #[derive(Ord, PartialOrd, PartialEq, Eq, Debug)]
         enum SymbolKind {
             Rule,
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -2,6 +2,7 @@ use std::{collections::HashMap, fmt};
 
 use indexmap::{IndexMap, IndexSet};
 
+use super::parser::{YaccGrammarWarning, YaccGrammarWarningKind};
 use super::{Precedence, YaccGrammarError, YaccGrammarErrorKind};
 
 use crate::Span;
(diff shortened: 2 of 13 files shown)


ratmice:
So, one thought I have after kind of thinking this through, is something that sits between `GrammarAST` and `YaccGrammar`,
which has fields like `prods: Vec<Vec<Option<Symbol<StorageT>>>>`, where `YaccGrammar`'s existing `prods: Vec<Vec<Symbol<StorageT>>>`, just checks that everything in the `ProtoGrammar` is `Some`, and `unwraps`.

Seems like some work, but it would be one way of being able to run this earlier while minimizing the duplicated logic,
something to think about I guess.

ltratt:
Before I try and make my brain hurt by thinking deeply about this I think:

1. The idea here is to report things defined in the lexer but not used in the grammar?
2. In keeping with yacc would `%expect-unused` be more in keeping perhaps?

ratmice:
Sorry, I didn't read that carefully enough, with regards to 1.

> 1. The idea here is to report things defined in the lexer but not used in the grammar?

Actually we already have that `missing_from_lexer`, `missing_from_parser`,
What this does is `Rule`s defined in the parser, not used as productions in the parser.

ltratt:
> What this does is Rules and Tokens defined in the parser, not used in any Productions of the parser.

Aha, got it! How does that interact with the [[Turning lexing errors into parsing errors](https://softdevteam.github.io/grmtools/master/book/errorrecovery.html#turning-lexing-errors-into-parsing-errors)](https://softdevteam.github.io/grmtools/master/book/errorrecovery.html#turning-lexing-errors-into-parsing-errors) trick?

ratmice:
For that we'd have to add `%expect-unused Unmatched` née allow on the `Unmatched` rule to keep it from emitting a warning.

One question which occurred to me is whether `%expect-unused` alone with no parameters should disable the warning altogether?  We could do that which might be useful while developing grammars.

ltratt:
> For that we'd have to add %expect-unused Unmatched née allow on the Unmatched rule to keep it from emitting a warning.

Right, gotcha.

> One question which occurred to me is whether %expect-unused alone with no parameters should disable the warning altogether?

I'm not super keen on that: it sounds like something people would leave in accidentally!

ltratt:
Obvious question: what's the best next step?

ratmice:
Well, either move forward and integrate into the tools, or take a step back and try and instead of doing the checking over `Pidx`'s and `cfgrammar::Symbol`, just try and work it out from the `String` comparisons and `ast::Symbol`.

ltratt:
>  Well, either move forward and integrate into the tools, or take a step back and try and instead of doing the checking over Pidx's and cfgrammar::Symbol, just try and work it out from the String comparisons and ast::Symbol on GrammarAST.

I must admit that I don't have a good intuition as to which route will produce the tidiest outcome, so I don't know what to suggest!

ratmice:
I *just* messed around/pushed an implementation of that in [86271ce](https://github.com/softdevteam/grmtools/pull/345/commits/86271cef08d4a5781b1fb8b48dd5e8a2c452ef27), So we can at least look at it and decide, It didn't turn out *that* involved.

One of the few things I could think of that could go wrong is that `start: Option` could be `None` and so this would complain about start rules, but that doesn't actually appear to be the case, `complete_and_validate` also checks that it is not `None`.  So it doesn't actually end up redoing much of any logic.

There are a couple of less elegant aspects regarding the need to introduce a local type for `Ord`, where `Symbol` isn't `Ord`.

ltratt:
Why do we need `Ord` on `SymbolKind`?

ratmice:
Because it goes into the `used_symbols` `BTreeSet`, (and anticipating why we need the `BTreeSet` question) we need that one for `Extend`.

ltratt:
You are getting used to my silly questions :) It's OK, I have an infinite supply!

Why are we using a `BTreeSet`? [I'm not even sure I know what that is!]

ratmice:
It's basically like a `HashSet` instead of using `Hash` it uses `Ord`, but where `HashSet` doesn't implement `extend`, `BtreeSet` does :)

ltratt:
Dumb question (I told you have lots :)): why don't we just use `HashSet` and not have the extra complication?

ratmice:
We could that too, it just shortens
```
for symbol in prod.symbols.iter().map {
   used_symbols.insert(...) 
}
```

To the `used_symbols.extend(prod.symbols.iter().map(...)` call, so it is only a tiny little win in code length really.

But we'd need to add `Hash` to `SymboKind` instead I believe, I don't think `ast::Symbol` derives `Hash` either.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -228,6 +229,79 @@ impl GrammarAST {
         }
         Ok(())
     }
+
+    /// Return an iterator over all the `YaccGrammarWarning` produced by this grammar.
+    pub fn warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> + '_ {
```
ratmice, line 234: I think we can get rid of this `'_` lifetime

cfgrammar/Cargo.toml
```diff
@@ -16,7 +16,7 @@ path = "src/lib/mod.rs"
 
 [dependencies]
 lazy_static = "1.0"
-indexmap = "1"
+indexmap = "1.9"
```
ratmice, line 19: This version bump was for the newer `first()` function, since I realized `self.start` should be valid I no longer call it,
So we can revert back to `indexmap = "1"` here I think

cfgrammar/src/lib/yacc/ast.rs
ratmice: Should be fixed in [1de1fc0](https://github.com/softdevteam/grmtools/pull/345/commits/1de1fc06e43373e771c3cde9f8a246a25f3dfe11)

cfgrammar/Cargo.toml
ratmice: Should be fixed in [6652f53](https://github.com/softdevteam/grmtools/pull/345/commits/6652f530ce579b706c4c6f8a590fd04f64902a35)

ratmice:
FWIW, double checking the docs I was wrong about `HashSet` not implementing `extend` I must have gotten that impression from some type error, apologies.

 should be fixed in [b2c8186](https://github.com/softdevteam/grmtools/pull/345/commits/b2c81863ca9c6aa6d658fa0c8bdfcd9265cc2835).

ratmice:
So, the `pub`/`pub (crate)` visibility of this has bitten me again, for some reason I was assuming that `GrammarAST` being public, we could get hold of one but `YaccParser`, and it's `ast()` function being both private/`pub (crate)`, puts a damper on how to best expose this along side errors.

one potential solution which isn't just *make stuff public*, could be having `YaccGrammar` accept a parameter `YaccGrammarParsingContext`, That would give us a caller owned structure to put these warnings in, making them accessible whether there are errors, or otherwise?

```
fn new(yacc_kind: YaccKind, src: &str, ctxt: &mut YaccGrammarWarningAccumulator)
```

ltratt:
I had to remind myself of this, but the conflicts API currently says:

> The conflicts feature is currently unstable and may change in the future.

https://softdevteam.github.io/grmtools/master/api/lrpar/struct.CTParser.html#method.conflicts

On that basis, I'm fine with rethinking how we do things like report conflicts and unused symbols to the user. That might then tell us the best way to integrate this into grmtools internals?

ratmice:
Sounds good, I'll go ahead and give what I was thinking a try so we can see if that seems like a worthwhile approach, I'll make an effort to try and think of some better names.

cfgrammar/src/lib/yacc/firsts.rs
```diff
@@ -310,6 +315,7 @@ mod test {
         let grm = YaccGrammar::new(
             YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
             "
+          %expect-unused G
```
ratmice, line 318: This one shows some of the limitations of the current mechanism for finding unused rules.

It only triggers on `G`, while `C`, `D`, and `F`,  are only used by `G`.

ratmice:
In [c27e1fc](https://github.com/softdevteam/grmtools/pull/345/commits/c27e1fc31720017f37795ef63623ce68b721881b) I didn't try and integrate `Conflicts` into this first attempt, primarily because of the dependency situation, where
`Conflicts` is declared in `lrtable`, which isn't currently depended upon by `cfgrammar`.

I'm a bit worried once we try and depend upon it we'll get in a situation where crates have trouble resolving the same types through different import resolution rules.  So I didn't want to try and jump into that rabbit hole quite yet.

ratmice:
Ran cargo test in the wrong directory, so there looks to be a bunch more work to be done.

That explains the surprisingly limited number of test failures encountered.

cfgrammar/src/lib/yacc/firsts.rs
ratmice: So I updated the mechanism to a recursive one, which catches a lot more things.

I went and ignored any *token* also specified in an `expect-unused`, which seemed like an improvement.

So we don't have to `%expect-unused Unmatched "UNMATCHED"` just the rule.
It was less clear to me whether `%expect-unused` should also expect that any subsequent rules can also be unused, so I left a comment there.

So I think this should be fixed in [903f5a4](https://github.com/softdevteam/grmtools/pull/345/commits/903f5a4d1c0bf4e1c4e0211c8fdf9dbf47ce3428)

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -377,6 +446,22 @@ impl YaccParser {
                 i = self.parse_ws(j, true)?;
                 continue;
             }
+            if let Some(j) = self.lookahead_is("%expect-unused", i) {
+                i = self.parse_ws(j, false)?;
+                while i < self.src.len() {
+                    if self.lookahead_is("%", i).is_some() {
+                        break;
+                    }
+                    let (j, n, span) = self.parse_token(i)?;
```
ltratt, line 455: There is `parse_rule` so I think what we really want is something like:

```rust
match self.parse_token(i) {
  Ok(...) => ...,
  Err(_) => {
    match self.parse_rule(i) {
      Ok(...) => ...,
      Err(_) => return Err("Expected token or rule at {i}")
  }
};
```

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -202,8 +209,138 @@ impl GrammarAST {
...
+                            spans: vec![*sym_span],
+                        });
+                    }
+                }
+            }
+        }
         Ok(())
     }
+
+    pub(crate) fn unused_symbol_warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> {
```
ltratt, line 236: I'm a bit slow, so I didn't realise at first quite what this is doing. Basically it's a reachability algorithm. I *think* we can simplify it a bit. In pseudocode something like this might work?

```rust
let expected_unused_rules = HashSet()...;
let expected_unused_tokens = HashSet()...;

let mut todo = vec![start_rule];
let mut seen_rules = HashSet::new();
let mut seen_tokens = HashSet::new();
while !todo.is_empty() {
    let rule = todo.pop();
    for sym in rule {
    	match sym {
	    Rule(x) => {
	       if !seen_rules.insert(x) {
	           todo.push(x);
	       }
	    }
	    Token(x) => seen_tokens.insert(x)
	}
    }
}

let unexpected_unused_rules = all_rules.difference(seen_rules).difference(expected_unused_rules);
let unexpected_unused_tokens = all_tokens.difference(seen_tokens).difference(expected_unused_tokens);
```

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -202,8 +209,138 @@ impl GrammarAST {
                 spans: vec![Span::new(0, 0)],
             });
         }
+
+        for sym in &self.expect_unused {
+            match sym {
+                crate::yacc::ast::Symbol::Rule(sym_name, sym_span) => {
```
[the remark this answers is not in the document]
ratmice: This can be shortened to `Symbol::Rule`, or maybe `self::Symbol::` but not `ast::Symbol` because it *is* the `ast` module, and unfortunately something like `use self as ast;` isn't allowed.

I expect it is probably fine to just shorten it to `Symbol::Rule`, but during the first patches I was dealing with both `Symbol` types a bit.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -202,8 +209,138 @@ impl GrammarAST {
...
+                Symbol::Token(sym_name, _) => {
+                    if self.tokens.contains(sym_name) {
+                        Some((sym_name, SymbolKind::Token))
+                    } else {
+                        None
+                    }
+                }
+            })
+            .collect::<Vec<(&String, SymbolKind)>>();
+        // If a rule is specified in `%expect-unused`, also add the tokens associated with it.
```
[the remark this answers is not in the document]
ratmice: Hmm, I think there are 2 kinds of changes that can cause unused token errors to appear here 

1. When you remove a rule `%expect-unused` while it is still unreachable, you now would get an errors for the rule and the tokens.
2. When you remove a token from a `%expect-unused` unreachable rule, while it is still used by another unreachable rule not present in `%expect-unused`.  You'd get 1 new error.

I'm pretty certain at least that actually fixing an error here can only cause (zero or more) errors to be fixed, but never introduced, since making the rule reachable makes the tokens in it reachable.

Are these cases similar enough to the situations you described?

cfgrammar/src/lib/yacc/ast.rs
ratmice: I'll give it a try, but I think part of what is making my impl a bit ugly, is the need to have `Span` in the return value,
these can't go in the `HashSet`'s .

So my `all_rules` and `all_tokens` equivalents have a different type than `seen_rules` and `expect_unused_rules` equivalents.  So for this to work I think i'd have add an extra pass over `unexpected_unused_*` to get spans.

ratmice: So I did a start at using separate rules/tokens sets in [49645ce](https://github.com/softdevteam/grmtools/pull/345/commits/49645ce2e0a0a61233f625a38ee0bff7b05503d5). 

I'll have to look into whether it will be possible to use something like `difference` later though.

cfgrammar/src/lib/yacc/ast.rs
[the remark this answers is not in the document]
ratmice: I went ahead and removed this in [edf85b9](https://github.com/softdevteam/grmtools/pull/345/commits/edf85b973db8fac2b7d23fc50da5a7ea59a8bceb), I'm definitely in the for doing something to reduce the number of warnings here, as it seems like perhaps token rules caused by unused rules *seem* superfluous and noisy in many cases.

But either way, it is better to start off simple, and we can always reduce the noise later after we've gained some experience with using it.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Modulo `s/parse_rule/parse_name/` and inverting the order so rule names are parsed first, I believe this works, pushed in [bef29c0](https://github.com/softdevteam/grmtools/pull/345/commits/bef29c05d0f6628ea314131357342fbe58362fe7)

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -2,6 +2,7 @@ use std::{collections::HashMap, fmt};
 
 use indexmap::{IndexMap, IndexSet};
 
+use super::parser::{YaccGrammarWarning, YaccGrammarWarningKind};
```
ratmice, line 8: Should add these to `mod.rs`, like the `Error`s?

cfgrammar/src/lib/yacc/info.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Controls the collection of and collects information about warnings and other information
+/// that result during the construction of a `YaccGrammar`.  If `warnings_as_errors` is `true`
+/// does not collect any warnings.
+pub struct YaccGrammarInfo {
+    pub(crate) warnings_as_errors: bool,
+    warnings: Vec<YaccGrammarWarning>,
+}
+
+impl YaccGrammarInfo {
+    pub fn new(warnings_as_errors: bool) -> Self {
```
ratmice, line 12: So, I think it'd probably be better to go for a full builder pattern here, with a `warnings_as_errors().build()` function here.
I was hesitant to just add `warnings_as_errors()` on `YaccGrammarInfo` itself it seems like it would only be confusing if it was callable after construction, since it may not have the desired effect if called too late.

But using the full builder pattern here would solve that, and be easier to add more flags in the future.

ratmice:
FWIW, I figured now that I have something in roughly a working state, it might be a good time to look into `Conflicts`, and the cross crate issues, I'm not sure I have any good ideas here.

I see a couple of different problems, each with a few potential solutions
1. Trying to turn `Conflicts` into a `Warning`, and turn `Warning` into a trait, so lrtable can declare it's own instance of the trait.
2. Alternately, move the `Conflicts` and `StIdx` types into cfgrammar, and add that directly to `Info`..

Issues impacting leveraging the Info structure:
3. Generic over the warning trait with a different instance of `Info` for each crate, and that crates warning/error type
4. Alternately, have a shared instance of `Info` across crates, when then would take `dyn Warning` trait objects
5. Or, duplicate the the entire Info structure for each crate (lrtable)

all (5) issues above except for 3 run into an instances of the same problem in that they require making `add_warning` or `add_conflicts` or Conflict construction public.  One way around this is to provide a `freeze()` function, after which subsequent calls would cause a panic. Ensuring that it is in fact empty upon entry makes them not publicly callable in any way.
Or just let them be public, or public but also have an opaque type indicating the producer of the warning . 

There are some weirder options involving the typetag, inventory, or linkme crates which could eliminate some need make things public potentially, but their target specific code doesn't support important targets like web-assembly.

In general I get the impression that every option has some drawbacks. My thinking was to try 1 + 3, along with `freeze()`.
Curious if i'm missing anything obvious, or need to clarify any of the above.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -377,6 +453,33 @@ impl YaccParser {
                 i = self.parse_ws(j, true)?;
                 continue;
             }
+            if let Some(j) = self.lookahead_is("%expect-unused", i) {
+                i = self.parse_ws(j, false)?;
+                while i < self.src.len() {
+                    if self.lookahead_is("%", i).is_some() {
```
ltratt, line 459: I *think* this will allow `%expect-unused %` to be parsed successfully, which doesn't feel quite right.

ltratt:
> Trying to turn Conflicts into a Warning, and turn Warning into a trait, so lrtable can declare it's own instance of the trait.

I think my feeling now is that Conflicts and Warnings are just specific instances of "stuff we've learnt about the grammar after analysing it". That suggests that what we really want to do is return some sort of "errors and warnings" struct that the user can then query to determine if things went wrong or not (and, if it did, how it went wrong). Most users will probably want to bork on any error; some might want to bork on warnings; but that's up to them.

In other words, perhaps what we should do is rip out the (unstable) conflicts API and find a way of returning "warnings and errors" in such a way that "errors cause failure" is still the default, but more advanced users have access to deeper information if they want it.

cfgrammar/src/lib/yacc/info.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// towards caution. For example, the treatment of warnings as errors would be enabled.
+    ///
+    /// The exact set of flags enabled by this function are not specified and subject to change.
+    pub fn cautious() -> Self {
+        Self::builder().warnings_as_errors().build()
+    }
+
+    /// If `warnings_as_errors` is `true` converts the warning to an error and returns `Some`.
+    /// Otherwise adds the warning for later usage and returns `None`.
+    pub(crate) fn add_warning(&mut self, warning: YaccGrammarWarning) -> Option<YaccGrammarError> {
```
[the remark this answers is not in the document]
ratmice: We can do that, but it does mean we'll no longer produce an `Err(...)` branch from the callee.  It is probably fine though if the callee knows nothing about this and produces `Ok` though.

cfgrammar/src/lib/yacc/info.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+impl YaccGrammarInfo {
+    /// Returns a `YaccGrammarInfoBuilder` for configuring the behavior of a `YaccGrammarInfo`
+    pub fn builder() -> YaccGrammarInfoBuilder {
+        Default::default()
+    }
+
+    /// Returns a `YaccGrammarInfo` settings that are enabled are selected with a disposition
```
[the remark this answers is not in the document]
ratmice: Rather than clarify it, since we're moving this condition out to the caller, it is probably better to just remove the function.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -131,11 +134,82 @@ impl fmt::Display for YaccGrammarErrorKind {
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum YaccGrammarWarningKind {
+    UnusedRule,
+    UnusedToken,
+}
+
+/// Any Warning from the Yacc parser returns an instance of this struct.
+#[derive(Debug, PartialEq, Eq)]
+pub struct YaccGrammarWarning {
+    /// Uniquely identifies each Warning.
```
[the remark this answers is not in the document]
ratmice: We'll want to change this for `YaccGrammarError` too.

ratmice:
I think that once we remove the warnings/errors conversion, we should still be able to use this basic skeleton implementing that directly in the caller.  Since we merely borrow `Info`. Once that borrow goes out of scope the caller still owns it, and so they should be able to enumerate "stuff we've learnt", and if they want errors they'll have that `From` impl.

I'd like to see whether that seems sufficient though, before we try to `return` one.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Indeed.

ratmice:
Sorry missed a `--amend` before I pushed :(

cfgrammar/src/lib/yacc/parser.rs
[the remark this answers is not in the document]
ratmice: These should be fixed in [5324d5a](https://github.com/softdevteam/grmtools/pull/345/commits/5324d5ace5b3afb740ede7e353e81a14e65cbeb8)

cfgrammar/src/lib/yacc/info.rs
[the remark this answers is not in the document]
ratmice: Removed in [c47e5c4](https://github.com/softdevteam/grmtools/pull/345/commits/c47e5c4135be21a4db2d57f1af6488b4488d55b0)

cfgrammar/src/lib/yacc/info.rs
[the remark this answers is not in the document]
ratmice: I removed all this in [c47e5c4](https://github.com/softdevteam/grmtools/pull/345/commits/c47e5c4135be21a4db2d57f1af6488b4488d55b0), some error conversion had to be added to `new_with_storaget`.

At some point I was worried that we get into a situation where we report an error, and then it *also* exists in warnings.
But that actually can't happen because `YaccGrammarInfo` can't escape from `new_with_storaget`.

But `new_with_storaget` is kind of the underlying need for conversion internally, I had tried to keep compatibility with the existing API here.

cfgrammar/src/lib/yacc/parser.rs
ratmice: I added a test for this, the way it works is that since `%` is just lookahead, it tries to parse it as a declaration on the next iteration and then throws an `UnknownDeclaration`.

ratmice:
Well, this definitely needs a rebase/squashing now, but the `test_info_collection`, does show we can get warnings out to the caller where they can decide how to handle them, without explicitly having to `return` them.

So I don't think we *have* to return Warnings, but we could at the expense of some moving them, and changing some signatures. Let me know if you still think returning warnings is preferable.

I think my preference towards passing it by mutable borrow is just that we want the caller to eventually own the warnings, so
to me it is simpler if they just own it from the beginning.  Rather than creating it and moving it to the caller. But that is perhaps just preference and may not be a good justification.

ltratt:
I think we can simplify both the internals and the external(ish) facing API if we make the following choices:

1. By default, warnings should be errors: users should have to say "I don't want warnings to be errors" (in the spirit of the existing https://softdevteam.github.io/grmtools/master/api/lrpar/struct.CTParserBuilder.html#method.error_on_conflicts).
2. We shouldn't turn warnings *into* errors as we go: we should differentiate them and allow users to tease them apart later.

That, to me, suggests a relatively simple API. On CTParserBuilder we add a function `warnings_are_errors(bool)` which defaults to `true`. In `new_with_storaget` we then do something like:

```
struct YaccGrammarAnalysis {

**Comment by ratmice** [comment]
I at least get the idea in spirit, and it'll give me something to chew on.

I believe we actually want to return `YaccGrammarAnalysis` from `build()` rather than `new_with_storaget`. As that is where it constructs a `YaccGrammar`, and currently returns a `Result<CTParser, Box<dyn Error>>` with errors from disparate sources.
This `build()` is also convenient because it won't have to figure out how to return a `Self` and warnings or errors and warnings.

IIUC, the block here: https://softdevteam.github.io/grmtools/master/api/src/lrpar/ctbuilder.rs.html#365-379 It seems would have to move outside to the caller, or perhaps moved to some new function like `pp_analysis(analysis: &YaccGrammarAnalysis)` or the like, for people to use in build.rs which we could think about later.

My main thought at the moment is that `SpannedError` and `SpannedWarning` traits, so we could use `spanned_errors: Vec<Box<SpannedError>>` in `YaccGrammarAnalysis`, and `errors: Vec<Box<dyn Error>>` for the other errors without spans.

Those are at least the hurdles that initially stand out to me.

**Comment by ltratt** [comment]
> I believe we actually want to return YaccGrammarAnalysis from build() rather than new_with_storaget

Possibly, yes. Those two functions' return types are fairly interrelated, and perhaps should be made even more so!

> My main thought at the moment is that SpannedError and SpannedWarning traits, so we could use spanned_errors: Vec<Box<SpannedError>> in YaccGrammarAnalysis, and errors: Vec<Box<dyn Error>> for the other errors without spans.

Not quite sure I understood this bit?

**Comment by ratmice** [comment]
> > My main thought at the moment is that SpannedError and SpannedWarning traits, so we could use spanned_errors: Vec<Box> in YaccGrammarAnalysis, and errors: Vec<Box> for the other errors without spans.
> 
> Not quite sure I understood this bit?

`build()` has an existing return type of `Box<Error>`, which coalesces both errors containing spans, and some errors without spans[^1] into a `Box<Error>`, If we intermix these it hides any ability to call `span()`, and friends, which is currently all done *inside* `build()` before being boxed into the error.

So, the current thing I am thinking of trying something to the effect of:
```
trait Spanned {
   fn spans(&self) -> &[Span]; // I changed the return type here from an iterator since we can't return impl from trait (seems nicer than Box<dyn Iterator=Item=Span>.
   fn spanskind(&self) -> SpansKind;
}
```

```
trait YaccGrammarAnalysis: Into<Result<Self, Self>> {
  fn add_warning(&mut self, Box<dyn Spanned>);
}
```

Then what is *currently* `CTParser` can `impl YaccGrammarAnalysis`[^2]. That allows what is `CTParser` to add it's own `Box<Error>` like values along side spanned warnings and errors.

I suppose that the purpose of using `dyn Spanned` and `SpannedError`/`SpannedWarning` was so we can add things like `Conflicts` to warnings, where those are emitted by `lrtable` rather than `cfgrammar`, sorry if I mixed up these two trains of thought.

[^1]: These seem to just be `self.set_output_file(...)?` and `"Generating two parsers to the same path..."`
[^2]: This trait would also be useful for my project, with a different `impl` than `CTParser`, which rather than having vectors of warnings and errors would have a channel which errors get sent through.

**Comment by ltratt** [comment]
I think I'm missing out on something, as I don't get why `Spanned` is needed. I don't *think* the fact that some errors have `Span` and some don't is an important detail, but I might be missing out on something!

**Comment by ratmice** [comment]
It doesn't seem like it is going to work though because of object safety, so I'm not sure if it was worth trying to understand.

But mostly my project needs to call span on all the errors that refer to source text, so that errors and the like can be associated with the source text in the editor. And `Spanned` seemed a way to get that while having `add_warning` instead of `add_yacc_warning(warning: YaccGrammarWarning)` and additional functions for each crate that produces a warning.

It seemed like it would be nice if we could just throw them *all* in one `Vec<Box<Spanned>>` and have a single analysis report which covers all crates and their warnings.

**Comment by ltratt** [comment]
Does this mean that my sketch in https://github.com/softdevteam/grmtools/pull/345#issuecomment-1229903259 can't work?

**Comment by ratmice** [comment]
It still can, but we may end up with a `Vec` field for each type of warning/error.
I had been evaluating some options, which are along the same public interface to `CTParserBuilder` but have additional fields.

**Comment by ratmice** [comment]
So, now that I feel like I understand better what you meant in https://github.com/softdevteam/grmtools/pull/345#issuecomment-1229903259 better with `Into<Error>` returning a `Result`, I'm pretty sure that this isn't going to work, at least without `#![feature(try_trait_v2)]` stabilized then we could `impl Try for YaccGrammarAnalysis`, because `Result` doesn't actually implement `Error` it implements `Try`, but since that isn't stabilized we can't actually implement the trait ourselves.

I believe the closest thing we're going to get here is changing `build()` from returning `Result<CTParser, Box<Error>>` to `Result<CTParser, CTParserBuildError>`, adding `warnings: Vec<YaccGrammarWarning>` to *both* `CTParserBuildError` *and* `CTParser`

**Comment by ratmice** [comment]
It has occurred to me that `CTParserBuilder` takes a `grammar_path(p: AsRef<Path>)`, and so the caller doesn't actually have the `String` to associate with the spans, without doing something racey like reading the path themselves or returning the grammar from both the `Ok` and `Err` result.

I'm beginning to think that just keeping builds return value as is and adding a loop for warnings like the one for [`YaccGrammarError`](https://softdevteam.github.io/grmtools/master/api/src/lrpar/ctbuilder.rs.html#365-378) might be the right thing to do really.

**Comment by ltratt** [comment]
> It has occurred to me that CTParserBuilder takes a grammar_path(p: AsRef<Path>), and so the caller doesn't actually have the String to associate with the spans, without doing something racey like reading the path themselves or returning the grammar src from both the Ok and Err result.

Yes, this is inevitable, and I don't think it's worth worrying about. [It affects all sorts of programs inc. debuggers and the like!]

Here's an alternative idea. We make `CTParserBuilder::build` be:

```rust
pub fn build(self) -> Result<CTParser<StorageT>, CTParser<StorageT>>
```

Then `CTParser` stores both errors and warnings. If "errors by default" is on (what's currently called `error_on_conflicts`) then you get an `Err` otherwise an `Ok`. In either case you can inspect `CTParser` to see what warnings/errors you got.

ratmice:
> Here's an alternative idea. We make `CTParserBuilder::build` be:
> 
> ```rust
> pub fn build(self) -> Result<CTParser<StorageT>, CTParser<StorageT>>
> ```

There are a few ways we could do this, one way is by changing `token_map` to return a result (Though I note that unlike conflicts it is not marked unstable/subject to further change). But for the rule_ids field if we encounter errors, we're going to end up with a `rule_ids: Result<HashMap<String, StorageT>, YaccGrammarError>`.

Instead of changing the return type of `token_map` it could also return an empty `HashMap` when errors occurred instead.

FWIW, on the [playground prototype](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=5a795df77171acd3c8bdc434150bf489) is some unstable rust using the `Try` trait, to allow a combined warning/result to use `?`. Simplicity just a single warning/result though, I was mostly curious if there might be something here we could do today, and eventually once it's stablized adding the `Try` impl to.

But this seems as close as I think we can get to https://github.com/softdevteam/grmtools/pull/345#issuecomment-1229903259 I I should credit scottmcm here, as the design he ended up with is much nicer than what I originally prototyped... 

But there really is a form of distributive property in here which makes flattening this down to a single type difficult in all cases.
e.g. we have `warning and (value or error) ↔ (warning and value) or (warning and error)`
where the disjunction makes it difficult for things like the `rule_ids` field which is currently valid by construction.

ltratt:
There are a few ways we could do this, one way is by changing token_map to return a result (Though I note that unlike conflicts it is not marked unstable/subject to further change).

Since we've got a semver change coming anyway, we could break this if necessary (I'd prefer not to, obviously, but I'd rather get the API right for the long term).

> Instead of changing the return type of token_map it could also return an empty HashMap when errors occurred instead.

I'm not super keen on this, as it might stop one doing potentially valid (if odd) things like empty lexers/grammars?

> FWIW, on the [playground prototype](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=5a795df77171acd3c8bdc434150bf489) is some unstable rust using the Try trait, to allow a combined warning/result to use ?. Simplicity just a single warning/result though, I was mostly curious if there might be something here we could do today, and eventually once it's stablized adding the Try impl to.

Wow, that is impressive -- I'm not even going to pretend that I was imagining that level of complexity! Unfortunately, I think we can say that idea isn't a runner, at least until that API stabilises.

OK, so having hurt my little brain, I think I can see where you're going overall. I can immediately think of two potential APIs here. API 2 is less conventional, but seems to me like it *might* be a bit easier to use. Basically people would just have to change `CTParserBuilder::build()?` into `CTParserBuilder::build().into_result()?`. But there may be problems here I haven't thought about!

### API 1

1. `CTParserBuilder::build` returns `Result<CTParser, CTParser>`.
2. `CTParser` then needs a new function called something like `valid_parser() -> bool`.
3. `CTParser::token_map` needs to return `Option`.

### API 2

1. `CTParserBuilder::build` returns `CTParser`.
2. `CTParser` has a new method `into_result(self) -> Result<Self, Box<dyn Error>>` which, if the `CTParser` represents an invalid parse, turns it into an `Err`.
3. `CTParser` would still need `valid_parser` and `token_map` as in API 1.

ltratt:
> I did want to see if an API3 was possible, https://github.com/ratmice/grmtools/commit/ce2321379ed439704a01c7fcd4833d75917f3949 (really ctbuilder.rs changes in there)
This unfortunately has a bunch of other simplifications to remove warnings_are_errors

That is not an approach that had occurred to me! Though given there's an `assert!(warnings.is_empty())` it seems morally equivalent to returning a `Vec`?

> if we added a CTParserBuilder::prepare_sources()? type of function which read the sources, handling all the filesystem errors etc.

Interesting idea. I claim (based on 2 minutes thought!) that it's somewhat similar to the `into_result` idea. Indeed, it makes me realise a refinement which is something like:

```
struct CTParserBuilder { ... }
impl CTParserBuilder {
  fn build(self) -> Result<CTParser, Box<dyn Error>> { ... }
}

struct CTParser { ... }
impl CTParser {
  fn into_result(self) -> Result<Self, Box<dyn Error>> {
    if self.errors.is_empty() { Ok(self) } else { Err(<some nice string about the error>.into()) }
  }
}
```

In other words, if `CTParserBuilder::build()` returns `Err` it means "I couldn't build the grammar". You then get a `CTParser` which you can either inspect for warnings / errors or convert into a `Result` which if it's an `Err` will tell you what errors there in the grammar. If we want, we can even repurpose `Builder::error_on_conflicts` to mean "`build` returns an `Err` if there are any errors when building the grammar." This does have the virtue of being source-compatible for the grmtools common case (we would deprecate `error_on_conflicts` and tell people "use the new function `grammar_errors_return_err`" (that's a terrible name but you get the idea!)).

Either way, I can't think of a way that doesn't involve changing the return type of `tokens_map`. I think we might as well just bite that bullet.

ratmice:
> > I did want to see if an API3 was possible, [ratmice@ce23213](https://github.com/ratmice/grmtools/commit/ce2321379ed439704a01c7fcd4833d75917f3949) (really ctbuilder.rs changes in there)
> > This unfortunately has a bunch of other simplifications to remove warnings_are_errors
> 
> That is not an approach that had occurred to me! Though given there's an `assert!(warnings.is_empty())` it seems morally equivalent to returning a `Vec`?

The basic savings here is the caller gets warnings for *both* the `Err` and `Ok` branches without needing it in the types.
i.e. I guess it is technically equivalent to `Result<(YaccGrammar, Vec<YaccGrammarWarning>), (Vec<YaccGrammarWarning>, Vec<YaccGrammarError>)>`, minus a move.

It *does* look though like we could move `warnings` into `YaccGrammar`, simplifying it to:
`new_with_storaget((yacc_kind: YaccKind, s: &str) -> Result<YaccGrammar, (Vec<YaccGrammarWarning>, Vec<YaccGrammarError>)>`

ratmice:
> It _does_ look though like we could move `warnings` into `YaccGrammar`, simplifying it to:
> `new_with_storaget((yacc_kind: YaccKind, s: &str) -> Result<YaccGrammar, (Vec<YaccGrammarWarning>, Vec<YaccGrammarError>)>`

So, tugging on this thread seems to have been a nice cleanup I remember why now I passed them as `&mut` parameters, So I could use them from the lrpar crate, with this it seems unnecessary.

1. Move Warnings `YaccGrammar`, `CTParser` now has warnings but it's behind an option  https://github.com/ratmice/grmtools/commit/bec66aa1eb32ab4c96215c8cf738424f64da0ceb
2. Lift that `YaccGrammar` out of the conflicts option. https://github.com/ratmice/grmtools/commit/ab2914865891d7e7fdef49b627ef49a181888e06

These rely on it being a `pub` field of `YaccGrammar` so they can take ownership of the vector from the other crate.
alternately we could do something like `pub fn warnings_owned(self) -> Vec<...>`.

It is worth mentioning, adding this to `YaccGrammar` makes `Box<[YaccGrammarWarning]>` beneficial to the size of expanded `YaccGrammar`.  So we might revisit that.

ltratt:
Isn't https://github.com/ratmice/grmtools/commit/bec66aa1eb32ab4c96215c8cf738424f64da0ceb making warnings return an `Err` with no way of having it return `Ok`?

ratmice:
> Isn't [ratmice@bec66aa](https://github.com/ratmice/grmtools/commit/bec66aa1eb32ab4c96215c8cf738424f64da0ceb) making warnings return an `Err` with no way of having it return `Ok`?

Sorry about that, I had temporarily removed the `warnings_are_errors` behavior, while fighting the borrow checker, but forgot to add it back. In that commit, It would return `Ok` even though there were warnings.

I re-added warnings_are_errors stuff in https://github.com/ratmice/grmtools/commit/beffac6af6f9f44662bb5353599e80c63a032687

In https://github.com/ratmice/grmtools/commit/5e566dd616e3ab40f8dc0563740dfabbf479d803 I tried to make some enum out of `YaccParserError` to get a better idea of what information exists at various points we emit errors within `build`.

ltratt:
But now, I think, in https://github.com/ratmice/grmtools/commit/beffac6af6f9f44662bb5353599e80c63a032687 if you get warnings and an `Ok`, you can't access the warnings?

ratmice:
You actually can, is the borrow checker fight that was happening, they are within the `grm`, I added a `CTParser::warnings()` to access them but wasn't brave enough to just give direct access to grm which is something we could consider.

ltratt:
I must admit, my little brain is exploding from all the possibilities! Is there some way to put what you think is the best idea into a tidy package?

ratmice:
Indeed, I think it is overdue, that I rebase/reorder remove unnecessary stuff from this series, and try to make some tidy commits.
I'll try to clean it up and come back with a new branch.

ltratt:
Thanks!

PR #346, opened 2022-08-29
Boxed slice err value

ratmice:
It occurred to me that there are some potentially good reasons to use `into_boxed_slice()` instead of returning `Vec`.
I'm not sure if it actually shrinks the `Result` type (probably depends on how big the `Ok` value is I suppose.

The biggest impact seemed to be that it isn't implicitly converted into an iterator, and it requires some boiler plate before we can pattern match on it.  But it seemed worth considering.

### DIFF
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -118,7 +118,7 @@ where
                 let mut ast = yp.ast();
                 let r = ast.complete_and_validate();
                 if r.is_err() {
-                    return Err(vec![r.unwrap_err()]);
+                    return Err(vec![r.unwrap_err()].into_boxed_slice());
                 }
 
                 ast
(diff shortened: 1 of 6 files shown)


ltratt:
I didn't know about `into_boxed_slice`! I'm neutral on this. On the up side it does shrink the `Vec` down to the minimum necessary, but on the down side it will probabyl unfamiliar to many users.

Can we maybe check what `size_of` says about this before and after? If it's smaller in size, I think we can consider merging this.

ratmice:
It doesn't help for `new_with_storage_t(...) -> YaccGrammarResult<YaccGrammar>` where the generic `T` in `YaccGrammarResult` exceeds the `size_of::<Vec<YaccGrammarError>()` (416 in both cases)

 but does for `YaccParser::parse(...) -> YaccGrammarResult<usize>` it shrinks from `32` to `24`, where the size of `Vec` dominates the size of `T`.

ltratt:
Thanks for the data! I think you're right that this is borderline, and because the API is (at least to me...) a little unfamiliar (I haven't seen `Box<[T]>` before!) I think it's probably best to stick with what we have. I'm glad to know about this Rust API though!

PR #347, opened 2022-09-03
Unused symbol 2nd

ratmice:
For me at least, this has been the implementation that has ticked the most boxes.
Here are some of the reasons I like this version

* Relatively source compatible, a few error return types are changed, parameters largely stay the same.
* While there are currently no Spanned error/warning types in lrtable or lrpar, this paves the way for being
   able to return them along side yacc errors/warnings.  E.g. we could consider turning the `CTConflictsError`.
   into a spanned error, adding it to the `CTParserError` type, and the dyn iterators.  Once that is in place, it seems
   possible we could deprecate `conflicts()` without replacement.
* I like that in the Ok branch of `new_with_storaget` the warnings are hidden within the `YaccGrammar`, and we don't have to deal with tuple return types.  But it isn't really that important they could easily be returned along with the `YaccGrammar`.

I'm less thrilled with the whole `CTParserError::{PreBuild, Build}` and `BuildErrorData`, but seemed like the least terrible
thing I could come up with that allowed us to keep e.g. `token_map` the same.

Anyhow sorry if in fact I am a bit too noisy. :(

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -10,7 +10,7 @@ use std::{
     fmt::{self, Debug, Write as fmtWrite},
     fs::{self, create_dir_all, read_to_string, File},
     hash::Hash,
-    io::{self, Write},
+    io::{self, Read, Write},
     marker::PhantomData,
     path::{Path, PathBuf},
     sync::Mutex,
@@ -137,6 +137,7 @@ where
     error_on_conflicts: bool,
     warnings_are_errors: bool,
     visibility: Visibility,
+    src_buf: Option<&'a mut String>,
     phantom: PhantomData<(LexemeT, StorageT)>,
 }
 
@@ -176,6 +177,7 @@ where
             error_on_conflicts: true,
             warnings_are_errors: true,
             visibility: Visibility::Private,
+            src_buf: None,
             phantom: PhantomData,
         }
     }
@@ -292,6 +294,15 @@ where
         self
     }
 
+    /// If set, [CTParserBuilder::build] will use buffer for reading the source text, any
+    /// contents will be overwritten. the user can specify this if they require access to
+    /// the sources after build returns, any spans returned from build will be relative
+    /// to the contents at exit.
+    pub fn set_source_buffer(mut self, src_buf: &'a mut String) -> Self {
+        self.src_buf = Some(src_buf);
+        self
+    }
+
     /// Statically compile the Yacc file specified by [CTParserBuilder::grammar_path()] into Rust,
     /// placing the output into the file spec [CTParserBuilder::output_path()]. Note that three
     /// additional files will be created with the same name as specified in [self.output_path] but
@@ -344,7 +355,7 @@ where
     /// # Panics
     ///
     /// If `StorageT` is not big enough to index the grammar's tokens, rules, or productions.
-    pub fn build(self) -> Result<CTParser<StorageT>, CTParserError> {
+    pub fn build(mut self) -> Result<CTParser<StorageT>, CTParserError> {
         let grmp = self
             .grammar_path
             .as_ref()
@@ -368,8 +379,18 @@ where
             lk.insert(outp.clone());
         }
 
-        let inc = read_to_string(grmp).unwrap();
-        let grm = YaccGrammar::<StorageT>::new_with_storaget(yk, &inc);
+        let mut f = File::open(grmp).unwrap();
+        let mut buf = String::new();
+
+        let inc = if let Some(src_buf) = self.src_buf.as_deref_mut() {
+            src_buf.clear();
+            f.read_to_string(src_buf).unwrap();
+            src_buf
+        } else {
+            f.read_to_string(&mut buf).unwrap();
+            &buf
+        };
+        let grm = YaccGrammar::<StorageT>::new_with_storaget(yk, inc);
 
         let mut build_err = BuildErrorData {
             src: inc.clone(),
@@ -409,7 +430,6 @@ where
                         if outc.contains(&cache) {
                             return Ok(CTParser {
                                 grm,
-                                src: inc,
                                 regenerated: false,
                                 rule_ids,
                                 conflicts: None,
@@ -488,7 +508,6 @@ where
         } else {
             Ok(CTParser {
                 grm,
-                src: inc,
                 regenerated: true,
                 rule_ids,
                 conflicts,
@@ -589,6 +608,7 @@ where
             error_on_conflicts: self.error_on_conflicts,
             warnings_are_errors: self.warnings_are_errors,
             visibility: self.visibility.clone(),
+            src_buf: None,
             phantom: PhantomData,
         };
         Ok(cl.build()?.rule_ids)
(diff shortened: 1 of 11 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -1189,13 +1193,11 @@ pub struct CTParser<StorageT = u32>
 where
     StorageT: Eq + Hash,
 {
+    src: String,
```
ltratt, line 1217: I don't think we want to store this. If the user wants to store it, that's up to them, but we don't even make use of it.

ratmice: The issue is that the user doesn't have it, it is currently local to build, so if we don't return it in some way the user can never get to it.

ratmice: I pushed a proposed fix in [eafabc4](https://github.com/softdevteam/grmtools/pull/347/commits/eafabc42bc04f20f4022b4faa240b39add9fd23a) allowing the user to specify a buffer so they'll own it.

ratmice:
I would like to bring in an example which uses this branch along with https://crates.io/crates/miette
for pretty error reporting from build.rs but there are some issues

1. It is going to fail, so i'm worried if I put it in lrpar/examples/ it'll try and run along with the rest during `cargo test`.
2. `dev-dependencies` cannot be optional, so it would be bringing in another crate I think.

I think for now I'll just upload it to it's own repo and we can figure out if there is a good way to integrate it later.
Maybe we could add a new fancy errors section in the book, and just link to that repository from there?

ltratt:
I've chewed on this and I think we're trying to do too many things in one PR -- too many for my pea brain, at least!

1. In an ideal world I'd like to get something like https://github.com/softdevteam/grmtools/pull/347/commits/75fb27d4902cb1dbc708509488f8413cab3203dc into master straight away (possibly merging https://github.com/softdevteam/grmtools/pull/347/commits/9e787622c76f0ab6fdf438887807df80fb99d345 into it?), modelling it exactly on the other `expect-` rules. Let's not worry about warnings for this. I think this should be its own PR.

2. https://github.com/softdevteam/grmtools/pull/347/commits/90950fd596d0bb962e6ec635ad609b58ca3a63b1 seems fairly uncontentious, even if it's currently a bit orphaned.

3. I'm not totally sure I understand *why* https://github.com/softdevteam/grmtools/pull/347/commits/6d7db43eb64b4b9e6a0eb6885996e453307cf871 is needed (I can believe it is needed, but I need a bit more motivation). Maybe that even comes in a PR with 90950fd?

4. Then a PR with the warning API stuff where we can thrash that out. I'm still not sure we've got a good design for this, and since it's so user-facing, I agree with your implication that we need to do our best to get it right (if we are going to break source compatibility, we want to do so for a good reasons, and not to repeat the process if we can avoid doing so!).

Sound like a plan? Sorry to muck you about, but I find the problem with these big PRs is that we end up with "what" but not "why and I'm just averse to not understanding "why" :)

ratmice:
I think it's a fair assessment, I'd rather spend the time to get things right so we're all certain that we're happy with it.

3. I think i'd like to wait on https://github.com/softdevteam/grmtools/commit/6d7db43eb64b4b9e6a0eb6885996e453307cf871
    A question popped into my head around those marker traits, and whether if we stabilize them we'd eventually want/need to add  `Send + Sync` bounds etc.

As for the *why* of 3, currently our `Error` coalescing is done via `Box<Error>`, used in API like [lrpar_config()](https://softdevteam.github.io/grmtools/master/api/lrlex/struct.CTLexerBuilder.html#method.lrpar_config).
If we want to pass errors to an external renderer like [miette](https://crates.io/crates/miette) or [codespan](https://github.com/brendanzab/codespan) from build.rs for more rust-compiler like error messages which annotate the sources, We'll have to pass these renderers the string and the span.

So the general idea for 3 is to build a mechanism which we can replace Box<Error> retaining the ability to coalesce errors across e.g. `CTLexerBuilder`/`CTParserBuilder` while adding mechanisms to get at the information needed to pass these on to an external error renderer.

There is definitely a valid question 'why not just pass in a `render` function which renders it writing the result of that into a `Box<Error>` as part of the `CT*Builder` and just continue returning `Box<Error>` as before.  The reason this gives me pause is just that it would be nice if my lsp work could eventually generate rust sources, allowing you to use rust-analyzer to typecheck the actions, it isn't currently doable because `CTParserBuilder` expects `Path` while the LSP works with files in memory, at that point working with `Box<Error>` becomes really unappealing, you would need to render by side-effect and there is no way to differentiate `Spanned`, from normal unspanned `Error`.

I'll spend some time as I indicated above working on a build.rs based consumer of this API, based on this prototype
So we can start looking at concrete examples from the top-down, rather than this bottom-up existential consumer work.

ltratt:
> I think i'd like to wait on https://github.com/softdevteam/grmtools/commit/6d7db43eb64b4b9e6a0eb6885996e453307cf871

Ah, yeah, I sort-of understand where you're coming from based on your description. There's no problem waiting and making sure we've got a good design for that!

ratmice:
Closing now, I'll make individual pull-requests as noted above after i've had a chance to get some working examples using all the pub API this series introduces so we can see them in use.

lrpar/src/lib/ctbuilder.rs
ratmice: Just wanted to note down one thing (not caused by) but brought to my attention due to the removal here, is the lack of any form of `SrcIdx` or anything which can `Eq` test whether a src file belongs to a `Spanned`.  Which will become more apparent if we coalesce errors across lex/yacc.

ltratt: Now that's an interesting question! I guess this is disambiguated if we have `GrammarError` vs `LexerError`, but you're right that we have to guarantee that there is at least *some* way of disambiguating things!

ratmice:
I did manage to add an example here of using these traits, though I'd say it is still fairly rough in dealing with non-spanned errors.
https://github.com/ratmice/grmtools/commit/2bd4522b3c626787b612ec8cd397ddeaf1878d33

ltratt:
Hmm, interesting. As before, I'm not dogmatic about retaining source compatibility, but your `check` idea does make me agree with you that we could retain source compatibility with `build` by adding a second function. Perhaps:

1. `build` is exactly as it is currently.
2. `build_with_analysis` returns (different-than-`build`?) structs that allow you to access warnings (and maybe other analyses on maybe-broken grammars).

Most people would call `build` as they do now, so it could retain its nice simple interface. Those few, those happy few, who want warnings could call `build_with_analysis` and get access to lots more information.

If we do go down such a route, I'd rather not bake *just* warnings in (as I think the `Iterator` idea does). `build_with_analysis` could also maybe return other analyses that most users wouldn't need (and that might be moderately expensive to compute).

ratmice:
Just one thing to note is that it is possible that `build` could still emit warnings, just not providing any direct access to their spans.

I'll have to think about what a generic `Analysis` trait like this would require, it isn't clear to me if it can get away with *just* associated types for the information returned, or if it would require GATs for https://rust-lang.github.io/generic-associated-types-initiative/design_patterns/many_modes.html

PR #348, opened 2022-09-05
Unused symbol 3rd

ratmice:
From https://github.com/softdevteam/grmtools/pull/347#issuecomment-1236191561

> 1. In an ideal world I'd like to get something like https://github.com/softdevteam/grmtools/commit/75fb27d4902cb1dbc708509488f8413cab3203dc into master straight away (possibly merging https://github.com/softdevteam/grmtools/commit/9e787622c76f0ab6fdf438887807df80fb99d345 into it?), modelling it exactly on the other expect- rules. Let's not worry about warnings for this. I think this should be its own PR.

This brings in the 2 commits mentioned unmodified from the previous series, on top of that is a patch to remove the dependence on warnings. It runs the check (to avoid warnings, and ensure the testsuite runs it) but doesn't do anything with the result yet.  That allowed a small cleanup to remove a temporary vector, since it no longer would capture `self` in the closure.

As such it will need to be squashed, let me know if you prefer that I do that squash before you want to look at it.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, fmt};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt,
+};
 
 use indexmap::{IndexMap, IndexSet};
 
@@ -55,6 +58,28 @@ pub enum Symbol {
     Token(String, Span),
 }
 
+/// Specifies an index into a `GrammarAst.tokens` or a `GrammarAST.rules`.
+/// Unlike `cfgrammar::Symbol` it is not parameterized by a `StorageT`.
+#[derive(Eq, PartialEq, Debug, Copy, Clone)]
+pub(crate) enum SymbolIdx {
+    Rule(usize),
+    Token(usize),
+}
+
+impl SymbolIdx {
+    pub(crate) fn symbol(self, ast: &GrammarAST) -> Symbol {
+        match self {
+            SymbolIdx::Rule(idx) => {
+                let rule = &ast.rules[idx];
+                Symbol::Rule(rule.name.0.clone(), rule.name.1)
+            }
+            SymbolIdx::Token(idx) => {
+                let tok = &ast.tokens[idx];
+                Symbol::Token(tok.clone(), ast.spans[idx])
+            }
+        }
+    }
+}
 impl fmt::Display for Symbol {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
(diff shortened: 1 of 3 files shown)


cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -28,6 +31,8 @@ pub struct GrammarAST {
     pub expectrr: Option<(usize, Span)>,
     pub parse_param: Option<(String, String)>,
     pub programs: Option<String>,
+    // Unchecked `Symbol` names, with spans pointing into the `%expect-unused` declaration.
```
ltratt, line 34: I'm not quite sure what "unchecked" means here.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -202,8 +230,101 @@ impl GrammarAST {
...
+                        });
+                    }
+                }
+            }
+        }
         Ok(())
     }
+
+    /// Returns tokens, and rules unreachable from the start rule
+    /// by index into the `ast.tokens` and `ast.rules` fields.
```
ltratt, line 258: I can parse line 257 but not line 258. I *think* it means "Return the indices of unexpectedely unused rules (relative to `ast.rules`) and tokens (relative to `ast.tokens`)."?

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -202,8 +230,101 @@ impl GrammarAST {
...
+                            seen_tokens.insert(name);
+                        }
+                    };
+                }
+            }
+        }
+        self.rules
+            .iter()
+            .enumerate()
+            .filter_map(move |(rule_id, (rule_name, _))| {
```
ltratt, line 308: Were we able to use `difference` here or did it not work?

cfgrammar/src/lib/yacc/ast.rs
ratmice: I think it should probably say "Potentially invalid", e.g. the caller must check them for validity, before using them and potentially return an `YaccGrammarError::UnknownRuleRef` or `UnknownToken`, as is done on lines 218 and 226 of https://github.com/softdevteam/grmtools/pull/348/commits/4fea2b669b97966165c5131963dba30156472d9a

ltratt: Maybe we can use the idiom from https://github.com/softdevteam/grmtools/pull/348#discussion_r962587727 of "unexpectedly unused"?

cfgrammar/src/lib/yacc/ast.rs
ratmice: Yes, exactly I should probably add "as `SymbolIdx`s"

cfgrammar/src/lib/yacc/ast.rs
ratmice: Yeah, I never could get it to work due to the type differences between `IndexMap`/`IndexSet` of `self.rules`, `self.tokens` and the `HashMap`s of the `expected` and `seen` tables.

ltratt: OK.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Hopefully I didn't go overboard in clarifying this in [e81fa9d](https://github.com/softdevteam/grmtools/pull/348/commits/e81fa9dfc66d6953acdf9d96c5bcd871848074e7) I didn't have much luck trimming it down, while still getting all the intricacies of it conveyed.

ltratt: I think the 2nd and 3rd sentences are repetitions of the 1st :) So just "If a `Symbol` is listed here but unused in the grammar, the fact that it is unused will be ignored. The `Symbol`s span points to the name given within a `%expect-unused` declaration." would be fine!

ratmice: I just want to triple check: 
"A symbol listed here that is not actually present in `rules` or `tokens` should be an error.",
is actually different than `unused`, which are present in the rules/tokens but unreachable from the `start_rule()`.

e.g. 
```
%expect-unused nonexistent
%%
Start: ;
```

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -254,8 +259,8 @@ impl GrammarAST {
         Ok(())
     }
 
-    /// Returns tokens, and rules unreachable from the start rule
-    /// by index into the `ast.tokens` and `ast.rules` fields.
+    /// Return the indices of unexpectedely unused rules (relative to ast.rules)
```
ltratt, line 262: One 'e' too many in 'unexpectedly'?

ratmice: Fixed in [26d1b6f](https://github.com/softdevteam/grmtools/pull/348/commits/26d1b6f026e96488b5c909739cc8b5f97fb60037)

cfgrammar/src/lib/yacc/ast.rs
ratmice: Anyhow I pushed the less verbose version in https://github.com/softdevteam/grmtools/pull/348/commits/a3cea98a5cb65598ef282f9e8ebf0be3eea88f79

ltratt: > I just want to triple check:
"A symbol listed here that is not actually present in rules or tokens should be an error.",
is actually different than unused, which are present in the rules/tokens but unreachable from the start_rule().

I'm not sure I understand the distinction.

ratmice: Above `NonexistentRule` is neither a rule or a token, it produces a `UnknownRuleRef` error, and `'NonexistentToken'` an `UnknownToken`.  While both `ExistsRule`, and `ExistsToken` are unused, so the distinction I was trying to highlight was one of  `unknown` vs `unused`.

```
%expect-unused NonexistentRule ExistsRule 'NonexistentToken' 'ExistsToken' 
%token ExistsToken
%%
Start: ;
ExistsRule: ;
```

ltratt: Ah, duh, I see what you mean now. Sorry!

How about something simple like "The set of symbol names that, if unused in a grammar, will not cause a warning or error"?

ratmice: Sounds good to me, I just wanted to ensure it didn't sneak through. [3b546bf](https://github.com/softdevteam/grmtools/pull/348/commits/3b546bf3c6f754490d309ee96e325c3285fcc835)

ltratt:
I think we're ready to squash?

ratmice:
Squashed, but let me know if you prefer just one commit.

PR #350, opened 2022-09-07
Nicer printing of string errors.

ratmice:
I've noticed that build.rs output since we started adding newlines for individual errors will end up printing
>Error: "Error 1\nError 2"

Because it is using the `Debug` implementation of `String`, rather than `Display`.
This just implements a wrapper around string to use the `Display` impl.

I noticed that rustfmt turns this (and an upcoming patch) into a larger diff than is necessary,
So, I was wondering if it helps if I rustfmt in a separate patch for review? Although this case is
fairly mild, I was more asking for when some particularly in egregious case where this happens.


ltratt:
I'm used to rustfmt's reformatting, so it's easier/better to stick it all in one go IMHO.

ltratt:
Please squash.

ratmice:
Okay, I'll try and remember that

Squashed.

PR #353, opened 2022-09-14
Analysis builder

ratmice:
Headed out the door for a few days, and haven't gone over this rebasing of my branch with a fine tooth comb, 
as such I'm marking it as a draft. But I wanted to try and get this series in some shape which could at least be be looked at.

I've tried to split up the changes of migrating code from `build()` into the various `CTAnalysisBuilder` structs,
over a series of patches.

There is one place that deserves particular scrutiny I've added a note to an existing comment before the call to
`std::fs::remove_file`.  Which details why I believe the changing of timing between returning `ConflictsError` now being before the call to `remove_file` call should be ok since the bug requiring it to be returned after should be fixed, there is a link to the bug.

I should probably say something about the `Analysis` trait, I had tried having something with a closure that returns things.
but usage in that manner broke the builder so you couldn't just chain calls, so I ended up with how it is.


cfgrammar/src/lib/analysis.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl<SourceId: PartialEq + ToOwned + ?Sized> Deref for YaccGrammarWarningAnalysis<SourceId> {
+    type Target = Box<[YaccGrammarWarning]>;
+
+    fn deref(&self) -> &Self::Target {
+        if let Some(warnings) = self.warnings.as_ref() {
+            warnings
+        } else {
+            lazy_static! {
+                static ref EMPTY_WARNINGS: Box<[YaccGrammarWarning]> = vec![].into_boxed_slice();
```
ltratt, line 45: This is... rather evil ;) I'd much rather have `warnings: Option<Vec<YaccGrammarWarning>>` and remove the `lazy_static`.

cfgrammar/src/lib/analysis.rs
```diff
@@ -0,0 +1,73 @@
+use crate::yacc::ast::{GrammarAST, Symbol};
+use crate::yacc::{YaccGrammarWarning, YaccGrammarWarningKind};
+use lazy_static::lazy_static;
+use std::ops::Deref;
+
+/// Performs an analysis on a given `Subject`, the given
+/// subject may be known to have pre-existing errors.
+pub trait Analysis<Subject> {
```
ltratt, line 8: I'm not sure I understand why we need such a generic struct. Any suggestions?

[I'd also prefer `fn analyse` as the spelling is more congruent with `trait Analysis`. But this is perhaps just my inner British spelling nerd coming out...]

lrpar/src/lib/ctbuilder.rs
```diff
@@ -1125,6 +1109,76 @@ where
...
+                grmp: grmp.to_owned(),
+                outp: outp.to_owned(),
+                yk,
+            },
+            pb: self.pb,
+        })
+    }
+}
+
+struct FileMeta {
```
ltratt, line 1114: What's the utility of this struct?

ltratt:
I'm not quite sure what the purpose of https://github.com/softdevteam/grmtools/pull/353/commits/1af9ba262e001a6183f098102437aa6a56ef256a -- it doesn't really seem related to this PR?

cfgrammar/src/lib/analysis.rs
ratmice: The main reason it's generic is so it can be used when the `Subject`s are declared in the various crates.
I couldn't really think of anything else which fit, given the limitations.

1. It's owned/taking self by mutable ref so instead of returning it could be used in the builder call chains.
2. It's generic on the ast/grammar/table because those all live in different crates.
3. Then does the work via side-effect to extract an analysis specific type to avoid imposing a return type on every analysis.

I really couldn't think of anything else that would cover those 3 constraints, the last seemed like the most important, 
The other 2 seemed somewhat necessary to achieving that.

Anyhow that was overall the reasoning behind the trait, I tend to think it is pretty strange trait though.

I hadn't thought of the spelling difference, but am happy to change it.

lrpar/src/lib/ctbuilder.rs
ratmice: Well, before I introduced this you could still get the data in *fmeta*, but it was stuck in one of the builders, so you would end up with `builder.tb.ab.pb` and `builder.tb.ab.grmp`, it seemed instead it would be better to just keep the `pb` and the various data that was still needed rather than keeping all the builder self values.

cfgrammar/src/lib/analysis.rs
ratmice: > I'm not quite sure what the purpose of [1af9ba2](https://github.com/softdevteam/grmtools/commit/1af9ba262e001a6183f098102437aa6a56ef256a) -- it doesn't really seem related to this PR?

The reason it is here, is so that `build_grammar` can return a `YaccGrammarError` or a `Spanned` error.
We have Non-Spanned errors before that, (reading yacc files to build the AST), and non-spanned errors again after `build_grammar`  (`write_parser` here).

cfgrammar/src/lib/analysis.rs
ratmice: So, I think one thing which might work instead of this generic analysis trait, is make `analyse_*` take closures,

```
fn analyse_ast<T>(t: T, f: &dyn Fn(GrammarAST, T)) {
    f(GrammarAST, t)
}

  let mut warnings: Vec<YaccGrammarWarning> = Vec::new();
  analyse_ast(&mut warnings, &|ast, ws| { 
      ws.push(YaccGrammarWarning)
  })
```

At least I believe that could be one way to get rid of the trait.

ratmice:
This comment was posted earlier, but somehow in response to the wrong thing, reposting here:

> I'm not quite sure what the purpose of [1af9ba2](https://github.com/softdevteam/grmtools/commit/1af9ba262e001a6183f098102437aa6a56ef256a) -- it doesn't really seem related to this PR?

The reason it is here, is so that `build_grammar` can return a `YaccGrammarError` or a `Spanned` error.
We have Non-Spanned errors before that, (reading yacc files to build the AST), and non-spanned errors again after `build_grammar`  (`write_parser` here).

Edit:
The one other thing to say about  [1af9ba2](https://github.com/softdevteam/grmtools/commit/1af9ba262e001a6183f098102437aa6a56ef256a) is without it, `build_for_analysis` can't write a grammar.
so it makes that accessible to both `build()` and `build_for_analysis()`

cfgrammar/src/lib/analysis.rs
ratmice: Yeah, I couldn't get myself to expose the Option, because as long as an analysis is run it will be at least `Some(&[])`
So, this is just some weird case where the user allocated an analysis never ran it, and now wants to deref it.

I think the right thing here just not delay the allocation until `analyse` is called, but allocate `warnings` in `new` and drop the option entirely.

ratmice: should be fixed in [c2dd0fa](https://github.com/softdevteam/grmtools/pull/353/commits/c2dd0faef76b415da366c6a13a59fec9de2851b6)

cfgrammar/src/lib/analysis.rs
ltratt: I think part of my confusion is that I think I only see one `impl` of this `trait`?

ratmice: There is a total of 4, throughout the series,
2. [`CTConflictsAnalysis`](https://github.com/softdevteam/grmtools/pull/353/commits/fd683fe9b3e27c4aea87979ce59aac830c8dacb0#diff-f389393ec514f9efce5aaf9e5188b4c3f2628445cd5fefad72e0006fb7334930R1315-R1331)

2 in the example:

3. [`AriadneWarningAnalysis`](https://github.com/softdevteam/grmtools/pull/353/commits/440754e50622276e3f63183375c503464c6b0057#diff-cb061aa1540e6f5a20a8464a3739c334243bf0957ea32b18419e0cc7eafe40b9R100-R104) -- which just wraps the YaccGrammarAnalysis
4. [`EmptyAnalysis`](https://github.com/softdevteam/grmtools/pull/353/commits/440754e50622276e3f63183375c503464c6b0057#diff-cb061aa1540e6f5a20a8464a3739c334243bf0957ea32b18419e0cc7eafe40b9R17-R19) -- used to just show that multiple analysis can be run and the places it can be used

There should be one in my project too which has a mpsc channel field.

ltratt: Ah, sorry, I didn't search for the right string! Let me take a gander.

ratmice: I recall now, that I have one more thing for generating railroad diagrams, which I intend to convert to an analysis,
I could quite easily add it to the example, It would likely mean we should rename the pretty_errors example, and it adds some dependencies though.

Anyhow, let me know if you'd like another.

ltratt:
I've had a couple of goes at reviewing this PR, and I must admit that I've been defeated both times, so it's probably time to admit that out loud! I think I have two problems:

1. I'm lacking motivation/context for why we need the `Analysis` trait. [The commit messages tell me what the commits do, but not why, so I end up guessing... except I mostly confuse myself :)]
2. I'm not really sure what the `CTBuilder` changes are doing. Are they important or ... ?

Pointers/suggestions are most welcome!

ratmice:
Well I totally understand your position, I've had some difficulty trying to split this patch apart into coherent stages.
I'll try and give a general overview of the motivation for the various things:

###### ASTValidation
<p>    This is probably the most important part of the patch and the easiest to split out from the rest, selectively giving access to a potentially invalid `GrammarAST` and `Vec<YaccGrammarError>`s produced, instead of the current `GrammarAST` or Vec<YaccGrammarError>`, while not loosing track of the fact that the specific `GrammarAST` may or may not actually be valid.

We could perhaps stop here and make `unused_symbol_warnings` public, and forgo adding the `Analysis` trait, but this doesn't make warnings accessible from the Builder API.</p>

###### CTBuilder
Trying to make it accessible from the builder API we run into quite a few distinct issues, but they are all somewhat tangled together. The common underlying factor is that `CTBuilder::build()` hides everything:
1. It hides types: for instance it produces `Box<dyn Error>` instead of `Spanned` error types.
2. It keeps the source `String`, and the `GrammarAST` as local values which never escape `build()`, on success it writes out the parser sources never revealing the `GrammarAST`, on failure it produces `Box<dyn Error>` also never revealing the `GrammarAst` or `Spanned` errors.

The builder changes, make these local values like `GrammarAST`, and `Spanned` errors available from a parallel builder API using a funny closure which is the `Analysis` trait.  It also takes buffers things like the source text to write into so the caller can access the sources to associate with `Spanned` errors that are returned.

`build()` then becomes equivalent to 
```
build_for_analysis()
   .build_ast()?
   .build_grammar()?
   .build_tables()?
   .source_generator()
   .write_parser()?
```
where `Spanned` Errors have been converted into `Box<dyn Error>`, and between each `build_` there is an optional analysis that can be run, which reveals to the analysis the local `GrammarAST` or `YaccGrammar`, etc value which was previously entirely local to build.

In between each of the `build_{ast, grammar, tables}` the trait just acts like a callback essentially takes the form of:

```
build_ast().analyze_ast(&mut wa: WarningAnalysis, |warning_analysis, &grm: GrammarAST| {
   /// Pull some values out of grm, and mutate them into self which the caller owns.
}).analyze_ast(&mut rda: RailroadDiagramAnalysis, |diagram_analysis, &grm:GrammarAST| {
   /// Generate an image of the AST in the form of a railroad diagram.
}).analyze_ast(&mut cra: RuleNamesShouldBeCapitalizedAnalysis, |capitalization_analysis, &grm: GrammarAST| {
   /// Check that each rule starts with a capital letter.
}).build_grammar()
```

Except traits with their self types allow you to collapse the mutable parameter and closure down into one self parameter.

So one of the key things here also is the ability to run multiple distinct analysis, while only calling the builder one time.
While leaving the old *simple* `build()` intact merely as the composition of this new verbose one which hides less information.

So that is kind of where this is coming from, wanting to do things which if we did them in grmtools proper would pull in dependencies like image or svg crates, using existing nice error formatting crates, so that we don't *have* to add these types of things as a dependency/feature of grmtools itself.  I.e. instead you could just bring in an analysis crate which is all set up to do that and add it as part of your build.rs script.

I should also stress, that I'm not enamored with how this patch is, but it has managed to fit those criteria while being flexible for future expansion.  I hope this helps unravel some of the motivation for this.

ratmice:
Well, I went ahead and added the railroad diagram example, this is mostly here to provide motivation,
because the example was originally intended to show errorrs, the yacc file errors out before actually generating the diagram.
But if you swap in a working one, it should generate the diagram on stdout.

The thing to note is that this is pulled out of a project which is using the `RTBuilder`, but there is currently no way to do this kind of thing when using `CTBuilder`, because of the hiding of values I mentioned above.

lrpar/examples/pretty_errors/src/main.rs
```diff
@@ -239,6 +239,9 @@ fn main() -> ExitCode {
                     .write_parser()
                     .unwrap();
             } else {
+                // Here if running under build.rs/cargo you should use `write` instead of `eprint`
```
ratmice, line 240: So one problem with this example has a catch-22 style issue, in build.rs it is giving an example where build.rs fails.
If we include that in the workspace, there is no way to signal to cargo that this is the expected result.
If we do not include it in the workspace, it won't even be compiled though, and could easily bit-rot.

One thing we _could_ do is just include warnings but not errors and include this in build.rs,
but that means that a top-level compile will actually emit warnings I think, but it would resolve some of the issues
with this code trying to act like a build.rs but inside a main.rs, lacking the cargo environment.

Maybe it is best to link from this example a separate repository or a chapter in the book with an _actual_ example as it is intended to be used under cargo?

ratmice:
I went ahead and created a simpler PR with dustbins the Analysis and extra builder stuff, we could close this one if you'd rather start from that?

ltratt:
It does feel like closing this one now makes

ratmice:
Indeed

PR #354, opened 2022-09-26
Ctbuilder warnings

ratmice:
Here is an alternate minimalist version of PR #353 which removes the Analysis trait, and the extra builder path.
With the `CTBuilder::build()` function is pretty much left as-is except now it directly emits warnings into the `Box<dyn Error>`.

The difference from the previous patch is this one doesn't allow you to get spanned errors/warnings out of the build process,
or interact with `GrammarAST`, `YaccGrammar`, `State*`.  That makes it quite a bit simpler than the previous patch, but lacks some of the flexibility of not being able to run an arbitrary user defined analysis.

Overall, I do struggle to find anything which allows that flexibility which doesn't entail all the complexity therein.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -5,9 +5,54 @@ use std::{
 
 use indexmap::{IndexMap, IndexSet};
 
-use super::{Precedence, YaccGrammarError, YaccGrammarErrorKind};
+use super::{parser::YaccParser, Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccKind};
 
 use crate::Span;
+/// Contains a `GrammarAST` structure produced from a grammar source file.
+/// As well as any errors which occurred during the construction of the AST.
+pub struct ASTWithValidityInfo {
+    pub(crate) ast: GrammarAST,
+    pub(crate) errs: Vec<YaccGrammarError>,
+}
+
+impl ASTWithValidityInfo {
+    /// Parses a source file into an AST, returning an ast and any errors that were
+    /// encountered during the construction of it.  The `ASTWithValidityInfo` can be
+    /// then unused to construct a `YaccGrammar`, which will either produce an
+    /// `Ok(YaccGrammar)` or an `Err` which includes these errors.
+    pub fn new(yacc_kind: YaccKind, s: &str) -> Self {
+        let mut errs = Vec::new();
+        let ast = match yacc_kind {
+            YaccKind::Original(_) | YaccKind::Grmtools | YaccKind::Eco => {
+                let mut yp = YaccParser::new(yacc_kind, s.to_string());
+                yp.parse().map_err(|e| errs.extend(e)).ok();
+                let mut ast = yp.ast();
+                ast.complete_and_validate().map_err(|e| errs.push(e)).ok();
+                ast
+            }
+        };
+        ASTWithValidityInfo { ast, errs }
+    }
+
+    /// Returns a `GrammarAST` constructed as the result of parsing a source file.
+    /// When errors have occurred and `is_valid` returns false, this AST is the
+    /// subset of the source file which parsed correctly while not encountering
+    /// any errors. As such even when an AST is not valid, it will return an AST.
+    pub fn ast(&self) -> &GrammarAST {
+        &self.ast
+    }
+
+    /// Returns whether any errors where encountered during the
+    /// parsing and validation of the AST during it's construction.
+    pub fn is_valid(&self) -> bool {
+        self.errs.is_empty()
+    }
+
+    /// Returns all errors which were encountered during AST construction.
+    pub fn errors(&self) -> &[YaccGrammarError] {
+        self.errs.as_slice()
+    }
+}
 
 /// An AST representing a grammar. This is built up gradually: when it is finished, the
 /// `complete_and_validate` must be called exactly once in order to finish the set-up. At that
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -5,7 +5,10 @@ use std::{
 
 use indexmap::{IndexMap, IndexSet};
 
-use super::{parser::YaccParser, Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccKind};
+use super::{
+    parser::YaccParser, Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccGrammarWarning,
+    YaccGrammarWarningKind, YaccKind,
+};
 
 use crate::Span;
 /// Contains a `GrammarAST` structure produced from a grammar source file.
@@ -298,6 +301,21 @@ impl GrammarAST {
             }
         }
         Ok(())
+    }
+
+    pub fn warnings(&self) -> Vec<YaccGrammarWarning> {
+        self.unused_symbols()
+            .map(|symidx| {
+                let (kind, span) = match symidx.symbol(self) {
+                    Symbol::Rule(_, span) => (YaccGrammarWarningKind::UnusedRule, span),
+                    Symbol::Token(_, span) => (YaccGrammarWarningKind::UnusedToken, span),
+                };
+                YaccGrammarWarning {
+                    kind,
+                    spans: vec![span],
+                }
+            })
+            .collect()
     }
 
     /// Return the indices of unexpectedly unused rules (relative to ast.rules)
--- a/lrpar/cttests/build.rs
+++ b/lrpar/cttests/build.rs
@@ -34,6 +34,50 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 }
                 s => panic!("YaccKind '{}' not supported", s),
             };
+            let (negative_yacc_flags, positive_yacc_flags) = &docs[0]["yacc_flags"]
+                .as_vec()
+                .map(|flags_vec| {
+                    flags_vec
+                        .iter()
+                        .partition(|flag| flag.as_str().unwrap().starts_with('!'))
+                })
+                .unwrap_or_else(|| (Vec::new(), Vec::new()));
+            let positive_yacc_flags = positive_yacc_flags
+                .iter()
+                .map(|flag| flag.as_str().unwrap())
+                .collect::<Vec<_>>();
+            let negative_yacc_flags = negative_yacc_flags
+                .iter()
+                .map(|flag| {
+                    let flag = flag.as_str().unwrap();
+                    flag.strip_prefix('!').unwrap()
+                })
+                .collect::<Vec<_>>();
+            let yacc_flags = (&positive_yacc_flags, &negative_yacc_flags);
+            let (negative_lex_flags, positive_lex_flags) = &docs[0]["lex_flags"]
+                .as_vec()
+                .map(|flags_vec| {
+                    flags_vec
+                        .iter()
+                        .partition(|flag| !flag.as_str().unwrap().starts_with('!'))
+                })
+                .unwrap_or_else(|| (Vec::new(), Vec::new()));
+            let negative_lex_flags = negative_lex_flags
+                .iter()
+                .map(|flag| {
+                    let flag = flag.as_str().unwrap();
+                    flag.strip_prefix('!').unwrap()
+                })
+                .collect::<Vec<_>>();
+            let positive_lex_flags = positive_lex_flags
+                .iter()
+                .map(|flag| flag.as_str().unwrap())
+                .collect::<Vec<_>>();
+            let negative_lex_flags = negative_lex_flags
+                .iter()
+                .map(|flag| flag.strip_prefix('!').unwrap())
+                .collect::<Vec<_>>();
+            let lex_flags = (&positive_lex_flags, &negative_lex_flags);
 
             // The code below, in essence, replicates lrlex and lrpar's internal / undocumented
             // filename conventions. If those change, this code will also have to change.
@@ -51,21 +95,62 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             let mut outp = PathBuf::from(&out_dir);
             outp.push(format!("{}.y.rs", base));
             outp.set_extension("rs");
-            let cp = CTParserBuilder::<DefaultLexeme<u32>, _>::new()
+            let cp_build = CTParserBuilder::<DefaultLexeme<u32>, _>::new()
                 .yacckind(yacckind)
                 .grammar_path(pg.to_str().unwrap())
-                .output_path(&outp)
-                .build()?;
+                .output_path(&outp);
+            let cp_build = if let Some(flag) = check_flag(yacc_flags, "error_on_conflicts") {
+                cp_build.error_on_conflicts(flag)
+            } else {
+                cp_build
+            };
+            let cp_build = if let Some(flag) = check_flag(yacc_flags, "warnings_are_errors") {
+                cp_build.warnings_are_errors(flag)
+            } else {
+                cp_build
+            };
+            let cp_build = if let Some(flag) = check_flag(yacc_flags, "show_warnings") {
+                cp_build.show_warnings(flag)
+            } else {
+                cp_build
+            };
+            let cp = cp_build.build()?;
 
             let mut outl = PathBuf::from(&out_dir);
             outl.push(format!("{}.l.rs", base));
             outl.set_extension("rs");
-            CTLexerBuilder::new()
+            let cl_build = CTLexerBuilder::new()
                 .rule_ids_map(cp.token_map())
                 .lexer_path(pl.to_str().unwrap())
-                .output_path(&outl)
-                .build()?;
+                .output_path(&outl);
+            let cl_build = if let Some(flag) = check_flag(lex_flags, "allow_missing_terms_in_lexer")
+            {
+                cl_build.allow_missing_terms_in_lexer(flag)
+            } else {
+                cl_build
+            };
+            let cl_build =
+                if let Some(flag) = check_flag(lex_flags, "allow_missing_tokens_in_parser") {
+                    cl_build.allow_missing_tokens_in_parser(flag)
+                } else {
+                    cl_build
+                };
+            cl_build.build()?;
         }
     }
     Ok(())
+}
+
+fn check_flag((positive, negative): (&Vec<&str>, &Vec<&str>), flag: &str) -> Option<bool> {
+    assert_eq!(
+        positive.contains(&flag) | negative.contains(&flag),
+        positive.contains(&flag) ^ negative.contains(&flag)
+    );
+    if positive.contains(&flag) {
+        Some(true)
+    } else if negative.contains(&flag) {
+        Some(false)
+    } else {
+        None
+    }
 }
--- a/lrpar/cttests/build.rs
+++ b/lrpar/cttests/build.rs
@@ -15,6 +15,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     for entry in glob("src/*.test")? {
         let path = entry.unwrap();
         if path.is_file() {
+            println!("cargo:rerun-if-changed={}", path.display());
             // Parse test file
             let s = fs::read_to_string(&path).unwrap();
             let docs = YamlLoader::load_from_str(&s).unwrap();
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -19,7 +19,7 @@ use std::{
 use bincode::{deserialize, serialize_into};
 use cfgrammar::{
     newlinecache::NewlineCache,
-    yacc::{YaccGrammar, YaccKind, YaccOriginalActionKind},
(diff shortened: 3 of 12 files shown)


ratmice:
I had forgotten to do nimbleparse, while doing that I noticed i'd forgotten to implement `Spanned` for `LexBuildError`.
So this does those 2 things, I didn't implement anything like `warnings_are_errors` for nimbleparse just let me know if you think we should do that.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -5,10 +5,56 @@ use std::{
...
+use super::{parser::YaccParser, Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccKind};
 
 use crate::Span;
 
+pub struct ASTValidation {
+    pub ast: GrammarAST,
+    pub errs: Vec<YaccGrammarError>,
+}
+
+/// Provides access to an ast which may or may not be valid,
```
ltratt, line 20: I can't quite parse the second sentence relative to the first (e.g. is a valid AST not kept?).

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -5,10 +5,56 @@ use std::{
...
+/// An AST which is invalid is provided for further analysis.
+impl ASTValidation {
+    /// The AST will be invalid if errors were encountered during
+    /// construction in which case `errors()` will be non-empty.
+    pub fn new(yacc_kind: YaccKind, s: &str) -> Self {
+        let mut errs = Vec::new();
+        let ast = match yacc_kind {
+            YaccKind::Original(_) | YaccKind::Grmtools | YaccKind::Eco => {
+                let mut yp = YaccParser::new(yacc_kind, s.to_string());
+                let _ = yp.parse().map_err(|e| errs.extend(e));
```
ltratt, line 30: Is this equivalent to `yp.parse().map_err(|e| errs.extend(e)).ok()` or similar?

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -5,10 +5,56 @@ use std::{
...
+                ast
+            }
+        };
+        ASTValidation { ast, errs }
+    }
+
+    /// Returns a `GrammarAST` which may contain errors and should be validated.
+    ///
+    /// You may use an ast with errors to perform further analysis.
+    /// Consumers of `ASTValidation` such as [YaccGrammar::new_with_storaget](crate::yacc::YaccGrammar::new_with_storaget]
```
ltratt, line 42: I think a bit of line wrapping is needed :)

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -5,10 +5,56 @@ use std::{
...
+    /// should propagate any errors within it along with any additional errors that they produce.
+    ///
+    /// User specified consumers should not drop the errors without propagating them upwards.
```
ltratt, line 45: I'm not quite sure I understand this sentence: they can't drop the errors can they?

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -301,6 +304,19 @@ impl GrammarAST {
         Ok(())
     }
 
+    pub fn warnings(&self) -> impl Iterator<Item = YaccGrammarWarning> + '_ {
```
ltratt, line 307: I wonder if having this be an iterator is useful in the sense that callers will always `collect` it? [I don't have a strong opinion here -- this is a genuine question!]

lrpar/src/lib/ctbuilder.rs
```diff
@@ -297,6 +300,12 @@ where
         self
     }
 
+    /// If set to true, [CTParserBuilder::build] will return an error if the grammar contains warnings.
```
ltratt, line 303: It would be good to document the default case (I think the other methods do).

lrpar/src/lib/ctbuilder.rs
```diff
@@ -374,24 +383,59 @@ where
...
+            {
+                format!("{} at line {line} column {column}", x)
+            } else {
+                format!("{}", x)
+            }
+        };
+
+        let res = YaccGrammar::<StorageT>::new_with_validation(yk, ast_validation);
+        let grm = match res {
+            Ok(grm) if self.warnings_are_errors.not() || warnings.is_empty() => {
```
ltratt, line 400: I haven't seen `not` before. Is this equivalent to `!self.warnings_are_errors`?

ltratt: [I actually wonder if the `if` clause might be more obvious as `if !(self.warnings_are_errors && !warnings.is_empty())` but it's perhaps a wash either way!]

ltratt:
Thanks for this PR -- I now feel I understand what's going on! [And sorry for my tardiness in the other PR -- my brain kept exploding.]

I'm happy with this overall direction. I think one question is whether `ASTValidation` is the right name. I originally thought this was a "validator" but it's really a "container of an AST and warnings/errors". My first thought was `ValidatedAST` but of course the AST might be invalid! Any thoughts on how we can clarify what this struct is doing?

lrpar/src/lib/ctbuilder.rs
ratmice: > I haven't seen `not` before. Is this equivalent to `!self.warnings_are_errors`?

Yeah it is totally equivalent, I just find the binding strength of `.not()` slightly more obvious than `!`, but not steadfastly resolved in any one way.

ltratt: I agree that `!` can be easily missed, but since it's more common (I just looked up `not` which says something :)), I'd prefer to go with that.

ratmice:
> Thanks for this PR -- I now feel I understand what's going on! [And sorry for my tardiness in the other PR -- my brain kept exploding.]
> 
> I'm happy with this overall direction. I think one question is whether `ASTValidation` is the right name. I originally thought this was a "validator" but it's really a "container of an AST and warnings/errors". My first thought was `ValidatedAST` but of course the AST might be invalid! Any thoughts on how we can clarify what this struct is doing?

Yeah, i'm in agreement about `ASTValidation`, one of the names I had used was `ASTValidity`, maybe that was better? 
I also really don't think `YaccGrammar::new_with_validation` is a very descriptive name, but naming is *definitely* not my strong suit.

maybe something like `ASTValidity` and `YaccGrammar::new_validate_ast`?

ltratt:
The best I can currently think of is `ASTWithValidityInfo`.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Yeah, I don't know if it helps here, I think that just ends up having to `collect` it twice, in the case where `YaccGrammar::new_with_validation`  takes ownership, and these generally need to outlive that until the `chain` call.

I also find the iterator here sketchy too though, and much preferred the way that the `Analysis` could take ownership while avoiding making any iterator public.

ltratt: >I also find the iterator here sketchy too though, and much preferred the way that the Analysis could take ownership while avoiding making any iterator public.

Could you expand this thought?

cfgrammar/src/lib/yacc/ast.rs
ratmice: Well, they kind of can because the caller *owns* an `ASTValidation`, and so they can drop that.
I should probably just reformulate the sentance to say `should make certain to propagate the errors upwords` which is the important point.

ltratt: I half get it, but what's a "user specified consumer"?

cfgrammar/src/lib/yacc/ast.rs
ratmice: Yeah, that seems clearer.

cfgrammar/src/lib/yacc/ast.rs
ratmice: It would be something that calls `ASTWithValidityInfo`, that never goes on to construct a `YaccGrammar` from it.
Once you construct a `YaccGrammar`, the errors get collated with the `YaccGrammarResult`, 
But if the caller never proceeds that far, it is their own burden to propagate those errors somehow.
`user specified consumer`, should probably be replaced by `caller`, *on the other hand* if the user does construct a `YaccGrammar`, they *don't* want to do anything with the errors at risk of double reporting them.

Double reporting should be obvious once it occurs, I think I should just say here:
> `YaccGrammar` will consume the errors upon construction. If not constructing
>  a `YaccGrammar` it is the callers responsibility to report errors.

or something to that effect.

ltratt: Nope, I still don't get it :) The warning seems to suggest there's something very unusual about `ASTWithValidityInfo` that requires careful thought from the user in order for bad things not to happen, but I'm not sure I see that link. Maybe I'm missing out on something!

cfgrammar/src/lib/yacc/ast.rs
ratmice: I find it sketchy mostly because of that lifetime, what if another type of warning comes along and it is a pain to chain it to the `unused_symbol_warnings().map(...)` due to some lifetime requirement. I would much prefer as you indicated above to `collect` and have an owned `Vec<YaccGrammarWarning>` somewhere.  The complication is the lifetimes of it all.

With `YaccGrammar` taking ownership of `ASTWithValidityInfo` and essentially dropping it before return,
when we want `warnings()` beyond that (when we want to report both warnings and errors).

 `Analysis` was nice in that it was owned by the caller, and consumed `warnings()` via a `&mut self`. So `Analysis` could outlive `ASTWithValidityInfo` after the latter was consumed by a `YaccGrammar`.

ltratt: `GrammarAST` could have a `warnings: Vec<...>` field or similar perhaps? Or we could make the `warning`s function `crate` only and external users have to find this info out via `ASTWithValidityInfo`?

ratmice: It'd have to be `YaccGrammar` having the warnings field, because `GrammarAST` doesn't live long enough either.
But that doesn't actually work in the case where `YaccGrammarResult` returns an `Err`. 

One thing we could perhaps do though is make it `fn get_warnings(&mut Vec<YaccGrammarWarning>)`?

cfgrammar/src/lib/yacc/ast.rs
ratmice: I think you might get it, It certainly is unusual, I would prefer not having something like this public. But it has to be because the callers that need it are in the `lrpar` crate rather than `cfgrammar`.  The only real bright side to it is that passing it to `YaccGrammar` is one of the few reasonable things to do with it.

ltratt: Let me ask the question in the other way: what happens if you fling `ASTWithValidityInfo` away and then create a `YaccGrammar`?

ratmice: You would just not see or report any of the errors produced when creating the first `GrammarAST`, and rebuild a new `GrammarAST` when you create the `YaccGrammar`.  The worst that can really happen is that `drop(AstWithValidityInfo)` is an implicit `.ok()`.

cfgrammar/src/lib/yacc/ast.rs
ltratt: Maybe there's something implicit here: I think you want the warnings identified at AST-time to live after the AST has been consumed and a `YaccGrammar` created?

ratmice: Yes, exactly the AST only lives until `new_with_validation()`, while all the `warnings.is_empty()` and iteration, are done
matching on the result of that call.

cfgrammar/src/lib/yacc/ast.rs
ratmice: I tried to clarify this entire structures documentation in [1e6c7d3](https://github.com/softdevteam/grmtools/pull/354/commits/1e6c7d3556741f4c9dbad09dce4b5bef2f28e5b8)

cfgrammar/src/lib/yacc/ast.rs
ratmice: Should be fixed in [1e6c7d3](https://github.com/softdevteam/grmtools/pull/354/commits/1e6c7d3556741f4c9dbad09dce4b5bef2f28e5b8)

lrpar/src/lib/ctbuilder.rs
ratmice: Should be fixed in [1e6c7d3](https://github.com/softdevteam/grmtools/pull/354/commits/1e6c7d3556741f4c9dbad09dce4b5bef2f28e5b8)

lrpar/src/lib/ctbuilder.rs
ratmice: I ended up swapping this if condition to the other case then using
 `self.warnings_are_errors && !warnings.is_empty()` there since it removed one layer of `!`

cfgrammar/src/lib/yacc/ast.rs
ratmice: I think this should also be fixed in [1e6c7d3](https://github.com/softdevteam/grmtools/pull/354/commits/1e6c7d3556741f4c9dbad09dce4b5bef2f28e5b8)

ratmice:
So there are still some things unsettled by my last patch, primarily around the signature of the `warnings` function/iterator, and ownership in general, but I believe that covered (or at least tried to) update the docs, and various minor changes requested.

cfgrammar/src/lib/yacc/ast.rs
ratmice: I'm not sure what I was thinking here, because we don't store `warnings` as a field in `ASTWithValidityInfo`,
But it is a value derived from `GrammarAST`, we don't end up having to clone the vector, to extend it's lifetime, e.g.
such if we were trying to keep `errors` field around after the ast has been moved.

So it seems fine to just collect in the `warnings()` function and return/move it?

I went ahead and commited that in [8530c8c](https://github.com/softdevteam/grmtools/pull/354/commits/8530c8c05371999fcc155e1f108bdee9cbf1d1b2)

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -11,48 +11,50 @@ use super::{
...
 
-/// Provides access to an ast which may or may not be valid,
-/// An AST which is invalid is provided for further analysis.
-impl ASTValidation {
-    /// The AST will be invalid if errors were encountered during
-    /// construction in which case `errors()` will be non-empty.
+impl ASTWithValidityInfo {
+    /// Parses a source file into an AST, returning an ast and any errors that were
+    /// encountered during the construction of it.  The `ASTWithValidityInfo` can be
+    /// then usued to construct a `YaccGrammar`, which will either produce an
```
ltratt, line 24: s/usued/used/

cfgrammar/src/lib/yacc/ast.rs
ltratt: OK, so I *think* we should remove the "User specified consumers should not drop the errors without propagating them upwards" comment because there's nothing special about this struct?

ltratt:
I tend to think we're getting there.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -11,48 +11,50 @@ use super::{
 };
 
 use crate::Span;
-
-pub struct ASTValidation {
+/// Contains a `GrammarAST` structure produced from a grammar source file.
+/// As well as any errors which occurred during the construction of the AST.
+pub struct ASTWithValidityInfo {
```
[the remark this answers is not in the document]
ratmice: They needed to be `pub(crate)`, so `YaccGrammar` can take ownership of them, but yes they didn't need to be `pub`.
should be fixed in [bd55d40](https://github.com/softdevteam/grmtools/pull/354/commits/bd55d40d754ea9625b105a89de79774920ae9de7)

cfgrammar/src/lib/yacc/ast.rs
ratmice: should be fixed in [9947d5e](https://github.com/softdevteam/grmtools/pull/354/commits/9947d5e559cd265de3f6265ff777db7db15a0c5f)

cfgrammar/src/lib/yacc/ast.rs
ltratt: Right, so I think the sentence at https://github.com/softdevteam/grmtools/pull/354/commits/1e6c7d3556741f4c9dbad09dce4b5bef2f28e5b8#diff-8aeec91eb9693494ba63059be659b7c45094f8747d2b54ac288f5ec62ec56aafR27 should be deleted?

ratmice: Removed it in [4311b96](https://github.com/softdevteam/grmtools/pull/354/commits/4311b961a1ef081c7c3136786a5e3108ebc0e851)

lrpar/src/lib/ctbuilder.rs
```diff
@@ -374,24 +383,59 @@ where
...
-            })
-            .map_err(ErrorString)?;
+                    }
+                }
+                grm
+            }
+            Err(errs) => {
+                let mut line_cache = NewlineCache::new();
+                line_cache.feed(&inc);
+                return Err(ErrorString(
```
ltratt, line 429: I just realised that I don't know what `ErrorString` is! Any clues?

ratmice: It's just a `Debug` impl for string which uses `Display` rather than `Debug`, so that it doesn't emit `"`, at the beginning/end of the string or "\n" literally, at the end of each line, etc.

ltratt: Oh, I see it's defined here https://github.com/softdevteam/grmtools/blob/4311b961a1ef081c7c3136786a5e3108ebc0e851/lrpar/src/lib/ctbuilder.rs#L94.

ltratt:
OK, I'm now OK with the API in this PR: but we do need some tests that exercise it. Once that's done, we'll be ready to merge.

ratmice:
Yeah, I'll need to look into how to do tests involving build.rs, in particular *lrpar/cttests* currently only seems to test positive cases which work correctly.

I assume that the thing we're looking for here is like the tests for `CTConflictsError` using `downcast_ref` https://github.com/softdevteam/grmtools/blob/master/lrpar/src/lib/ctbuilder.rs#L1343-L1374  checking that the `ErrorString` contains an `YaccGrammarWarning::UnusedToken(..).to_string()`, etc?

ltratt:
We should, in the medium term, look at negative `cttests` but that's not something we should worry about in this PR: positive tests are good enough!

ratmice:
So, I went ahead and implemented some flag setting for cttests, I ran into an issue which means I think this approach probably isn't working, but pushed it despite that (because the flag parsing code might still be useful?)

I am basically running into the exact same issue that I was running into with the other PR #353 which caused me to use a example/bin instead of a cttest/build.rs:

when doing a top-level build these cttests now emit warnings, I'm not sure there is anything we can do to suppress the output during build short of perhaps throwing the test in a separate workspace, and depending upon it. 

So, I'm of the opinion that the previous patch is probably not acceptable, unless we can find a way to suppress the output.

lrpar/cttests/build.rs
```diff
@@ -33,6 +34,61 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                .collect::<Vec<_>>();
+            let positive_lex_flags = positive_lex_flags
+                .iter()
+                .map(|flag| {
+                    let flag = flag.as_str().unwrap();
+                    flag.strip_prefix('+').unwrap_or(flag)
+                })
+                .collect::<Vec<_>>();
+            let positive_lex_flags = positive_lex_flags
+                .iter()
```
ratmice: This is a duplication of the statement above...
Fixed in a subsequent patch.

lrpar/cttests/build.rs
```diff
@@ -50,21 +106,59 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            let cp_build = if let Some(flag) = check_flag(yacc_flags, "error_on_conflicts") {
+                cp_build.error_on_conflicts(flag)
+            } else {
+                cp_build
+            };
+            let cp_build = if let Some(flag) = check_flag(yacc_flags, "warnings_are_errors") {
+                println!("Warnings some {}", flag);
+                cp_build.warnings_are_errors(flag)
+            } else {
+                println!("Warnings none");
```
ratmice: Debugging code leftover, fixed in a subsequent patch.

ltratt:
Please squash.

ratmice:
Squashed.

This at least seemed like a logical sequence of commits to me, though perhaps we could combine 342169e & 7098fcb, 
Or even pull f7dd0e2 out for another PR, since it is just something that I noticed was missing while editing test files when cargo build didn't rerun.

ltratt:
It's fine as-is!

bors r+

PR #355, opened 2022-10-04
Improve error alignment when multiple errors occur.

ratmice:
When rust prefixed the error text with "Error:", and multiple errors occur, indent them on their own line.
Would it be better to "\n\t", also in the single error cases as well?

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -409,13 +409,19 @@ where
             Ok(_) if self.warnings_are_errors && !warnings.is_empty() => {
                 let mut line_cache = NewlineCache::new();
                 line_cache.feed(&inc);
-                return Err(ErrorString(
-                    warnings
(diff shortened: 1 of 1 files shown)


ltratt:
I must admit I'm not sure I can remember what outcome this PR refers to! What triggers this code execution path?

ratmice:
This is the *build.rs* one for `cargo build`/`CTBuilder::build()`, the other paths all print errors themselves so they don't need the same treatment, since they aren't interspersed with IO errors and print errors themselves.

The following is the master output vs this branch output (last), from commenting out the yacc_flags in cttests/warnings.test

```
  --- stderr
  Error: Unused rule at line 5 column 1
  Unused token at line 2 column 8
```

```
  --- stderr
  Error:
        Unused rule at line 5 column 1
        Unused token at line 2 column 8
```

ltratt:
Got it!

bors r+

PR #356, opened 2022-10-07
Merge deletes when pretty printing repair sequences

ltratt:
After various setup, the main bulk of this PR merges together `Delete` repairs when they directly follow on from one another. For example for an input such as:

```
abc
```

where `bc` should be deleted we previously printed:

```
Delete "b", Delete "c"
```

With this commit we print:

```
Delete "bc"
```

For long token names this can be a real win. For example in pizauth's configuration file, which has long token names, I sometimes found myself typing `auth_notify_intervl` instead of `auth_notify_interval` and would then be greeted with this horror:

```
Parsing error at line 5 column 1. Repair sequences found:
   1: Insert AUTH_NOTIFY_INTERVAL, Delete a, Delete u, Delete t, Delete h, Delete _, Delete n, Delete o, Delete t, Delete i, Delete f, Delete y, Delete _, Delete i, Delete n, Delete t, Delete e, Delete r, Delete v, Delete l
   2: Insert REFRESH_RETRY_INTERVAL, Delete a, Delete u, Delete t, Delete h, Delete _, Delete n, Delete o, Delete t, Delete i, Delete f, Delete y, Delete _, Delete i, Delete n, Delete t, Delete e, Delete r, Delete v, Delete l
```

With this commit I get the rather nicer:

```
Parsing error at line 5 column 1. Repair sequences found:
   1: Insert AUTH_NOTIFY_INTERVAL, Delete auth_notify_intervl
   2: Insert REFRESH_RETRY_INTERVAL, Delete auth_notify_intervl
```

Note that this only merges together `Delete`s that directly follow each other: even a single blank space between them will cause them not to be
merged.

### DIFF
--- a/lrpar/src/lib/parser.rs
+++ b/lrpar/src/lib/parser.rs
@@ -931,7 +931,7 @@ pub(crate) mod test {
             .collect();
         let lexer_rules = small_lexer(lexs, rule_ids);
         let lexemes = small_lex(lexer_rules, input);
-        let lexer = SmallLexer { lexemes };
+        let lexer = SmallLexer { lexemes, s: input };
         let costs_tidx = costs
             .iter()
             .map(|(k, v)| (grm.token_idx(k).unwrap(), v))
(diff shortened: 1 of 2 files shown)


lrpar/src/lib/parser.rs
```diff
@@ -653,21 +653,41 @@ impl<LexemeT: Lexeme<StorageT>, StorageT: Hash + PrimInt + Unsigned>
                             + 1;
                         write!(out, "\n  {}{}: ", " ".repeat(padding), i + 1).ok();
                         let mut rs_out = Vec::new();
-                        for r in rs {
-                            match r {
+
+                        let mut i = 0;
+                        while i < rs.len() {
+                            match rs[i] {
+                                ParseRepair::Delete(l) => {
```
ratmice, line 660: I think it would be helpful to have a comment here, something perhaps like,
```
// Combine adjoining `Delete` repairs".
```

ltratt: Fixed in https://github.com/softdevteam/grmtools/pull/356/commits/0b93ba106c84260820d9a0160c713669e02cce8e.

ratmice:
Please squash.

ltratt:
Squashed.

Issue #323, opened 2022-07-21
Remove debug formatting in non-debug locations


ltratt:
In a couple of places (e.g. https://github.com/softdevteam/grmtools/blob/master/lrlex/src/lib/ctbuilder.rs#L419) we use debug formatting in a non-debug location. This feels somewhat unsatisfactory, particularly as there are fewer guarantees about stability.

ratmice:
I had a look through these, and it appears to me that these just want to be turned into `\"{}\"`.
I.e. they are all names, etc that we wouldn't expect to contain any escape sequences. So I don't
think we need to go as far as something like `\"{}\", x.escape_default()`, to try mirroring that debug
also escapes.

ltratt:
> I think actually '\' and '"' may be valid lex names that need escaping here.

Good point! Yes, this needs checking.

PR #358, opened 2022-10-09
Clean up issue #323 by using the quote crate.

ratmice:
Here is an attempt at fixing #323, this uses the quote crate (lrlex already depends upon it non-optionally it seems via serde).

The only other quoting mechanism I could find that seemed like it would work is the `escape_debug` function,
which is not very well described except that it "escapes like the str debug implementation".

After this, there are a couple of uses "{:?}", they are all for StorageT, we could perhaps add a bounds on `T: quote::ToTokens`
and implement `impl ToTokens for IdxNewtype!(T)`?  We can't quite manage it with just num_traits bounds alone.

quote (well proc_macro2 which it returns) also isn't very explicit in it's documentation for it's `Display` impl.
https://docs.rs/proc-macro2/latest/proc_macro2/struct.TokenStream.html#impl-Display-for-TokenStream

It is a bit unfortunate that rust really doesn't seem to have anything quite like python's `repr` function.

### DIFF
--- a/lrpar/cttests/src/quoting.test
+++ b/lrpar/cttests/src/quoting.test
@@ -1,0 +1,15 @@
+name: Test NoAction using the calculator grammar
+yacckind: Original(YaccOriginalActionKind::NoAction)
+grammar: |
+    %start S 
+    %%
+    S: '\' | '"' | '<' | '+' '🦀' ;
+
+lexer: |
+    %%
+    " '"'
+    \< '<'
+    \\ '\'
+    \+ '+'
+    🦀 '🦀'
+    [\t ]+ ;
--- a/lrlex/Cargo.toml
+++ b/lrlex/Cargo.toml
@@ -28,5 +28,6 @@ lrpar = { path = "../lrpar", version = "0.12" }
 regex = "1"
 regex-syntax = "0.6.27"
 num-traits = "0.2"
+quote = "1.0"
 serde = "1.0"
 try_from = "0.3"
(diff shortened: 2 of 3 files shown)


lrpar/cttests/src/quoting.test
```diff
@@ -0,0 +1,15 @@
...
+    S: '\' | '"' | '<' | '+' '🦀' ;
+
+lexer: |
+    %%
+    " '"'
+    \< '<'
+    \\ '\'
+    \+ '+'
+    🦀 '🦀'
+    [\t ]+ ;
```
ratmice, line 15: One thing I think would be nice is if we had an `input:` field, for giving example inputs that could run the compiled parser.

ltratt:
A possibly silly question: does `quote!("str")` ever produce different output than `r#""str""#`?

ratmice:
> A possibly silly question: does `quote!("str")` ever produce different output than `r#""str""#`?

I'm assuming str figuratively, and if so there are cases where it returns a different string, for instance in the patch
the r.re_str().replaces get removed, because `quote!` performs the necessary escaping. 

```
+           let regex = &r.re_str;
+           quote!(#regex)
-               r.re_str.replace('\\', "\\\\").replace('"', "\\\""),
```

ltratt:
Ah, I get it! So this way we get proper quoting instead of the hack we had before. Works for me!

bors r+

ratmice:
Fixed the rustfmt issue.

ltratt:
Please squash.

ratmice:
Squashed.

PR #363, opened 2022-10-30
Conflicts spans

ratmice:
Here are the changes we had discussed for making spans accessible from conflicts,
There are ultimately 2 things these patches are trying to achieve:

-  Make the `YaccGrammar` take an AST by reference, rather than owned,
So we can delay dropping the AST until after `StateTable`, and friends have produced conflicts.
- Add some additional information to `StateTableError` for the second `PIdx` in conflict (if any), This tests an edge case
 relating to conflicts occurring in the *Production* `Start` of the implicit start rule `^: Start` which is present in the grammar but absent from the AST.  For that we add the `r_pidx` information to the error, which can sometimes contain a production within the *Rule* `Start` which is in conflict, in the test case this correlates to a production/span which occurs in the AST.

Finally, this latter part only touches Accept/Reduce conflicts. Both Shift/Reduce, and Reduce/Reduce conflicts I believe already
contain all the available PIdx/TIdx info, so I don't think anything extra is needed for those beyond the first bullet point.


cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -106,17 +106,17 @@ where
...
     pub fn new_with_storaget(yacc_kind: YaccKind, s: &str) -> YaccGrammarResult<Self> {
         let ast_validation = ast::ASTWithValidityInfo::new(yacc_kind, s);
-        Self::new_from_ast_with_validity_info(yacc_kind, ast_validation)
+        Self::new_from_ast_with_validity_info(yacc_kind, &ast_validation)
     }
 
     pub fn new_from_ast_with_validity_info(
         yacc_kind: YaccKind,
-        ast_validation: ast::ASTWithValidityInfo,
+        ast_validation: &ast::ASTWithValidityInfo,
```
ltratt, line 114: This might be a silly question, but why make this a reference given the `clone`s below?

ratmice: It is actually a key point, without making this a reference `new_from_ast_with_validity_info` takes ownership of the AST,
The `GrammarAST` however isn't moved into the `YaccGrammar`, only those portions of it which we now have to clone were moved into the `YaccGrammar`.

So, to make `GrammarAST` accessible *after* we have constructed a `YaccGrammar` from it, we need to either keep it around as a field of `YaccGrammar` or some such, alternately (as in this patch) avoid taking ownership of it herein by taking a reference to it. I *think* taking a reference to it is preferred, since that still allows you to drop it if it really is unneeded.

ltratt: > So, to make GrammarAST accessible after we have constructed a YaccGrammar from it

Is this something that e.g. the client you're writing needs to do?

ratmice: > Is this something that e.g. the client you're writing needs to do?

Yeah, but the tests added here too,
>        let grm = YaccGrammar::<u32>::new_from_ast_with_validity_info(yk, &ast_validity).unwrap();
later when we go:
>                let prod = &ast_validity.ast().prods[usize::from(pidx)];

Without taking it by reference here, The ast will have moved. Only then can we get spans from the ast.

ltratt: I wonder if rather than passing just `&` we pass a lifetime tied to `self`? So that might make it something like:

```
struct <'a> YaccGrammar {
   ast: &'a ASTWithValidityInfo
}
```

This would avoid the `clone`s (it's not so much that the `clone`s are expensive, but that it feels potentially easy to forget which `YaccGrammar` is related to which `ASTWithValidityInfo`, since it's not enforced by the static type system).

ratmice: Yeah, we could perhaps do that, but I wasn't sure if freeing the rest of the memory associated with the `GrammarAST` while keeping the `YaccGrammar` was desirable.  I could try it out though, if having these lifetimes entangled is better.

ratmice: It actually looks like entangling these lifetimes will be fairly difficult, because of the `Deserialize` trait on `YaccGrammar`.
So, you can currently serialize/deserialize a `YaccGrammar`, without a reference to an AST being involved.  I think in those cases it makes sense that we deserialize to owned values?

ltratt: Hmm. I *think* this is `Deserialize` so that `YaccGrammar` can be used by error recovery, but my memory is a bit fuzzy.

I must admit that I am slightly nervous that we're going to make the API easy to misuse -- but I don't have a quick answer for how to do any better :/

ratmice: I may very well be missing something, but I don't actually see the potential for misuse:
The biggest concern I see is that owning an AST might give the user the ability to mutate said AST.
But that has been avoided by `ASTWithValidityInfo` which only ever returns immutable references to an `&GrammarAST`.
So I don't really think this conveys anything above and beyond what can already be done besides controlling when
the AST gets dropped.

I personally, kind of think that it makes sense for the `YaccGrammar` to live beyond the length of the `GrammarAST`.
E.g. while I need to extend the lifetime of a `GrammarAST` in my program to be live until after a `StateTable` is received,
for some analysis. I don't actually need it to live beyond that, for example by the time i'm passing input strings to the grammar to be parsed I don't have any use for the `GrammarAST`.

An an alternative option (which I believe we shot down earlier) here is deriving e.g. `Clone` for `Production`, so that
you can `ast_validity.ast().prods.clone()` before `new_from_ast_with_validity_info()` takes ownership.

ltratt: > I personally, kind of think that it makes sense for the YaccGrammar to live beyond the length of the GrammarAST.

That is a fair point!

ratmice:
Perhaps, there is some changes to comments in the second patch I would rather not be there, but it is so minor I'm not sure if it is worth the trouble!

ltratt:
Feel free to push a commit -- it's probably easier to do it now rather than later!

ratmice:
Force pushed to just move that comment change into the first patch.

PR #366, opened 2022-11-02
Doc upates

ratmice:
This tries to use a uniform *calc.l* between *docs/\*.md*, *lrpar/examples/*, and *lrlex/examples*

Running through the error recovery example snippets, it showed some changes, and updates them accordingly.
I'm not certain whether these are algorithmic changes, or perhaps just lexical ordering (PLUS/MUL being alphabetically ordered differently than +, * maybe?).   We might need to revisit that as in at least one place this leads to a more complicated example being described due to "Integer from thin air" being introduced earlier.

In one place (errorrecovery.md) I had to reintroduce 'PLUS', 'MUL', as it motivated the example of `%epp`.


ltratt:
The error recovery algorithm has some non-determinism in it, so there can be changes in the order (there's no rhyme or reason to the order -- it's genuinely non-deterministic).

doc/src/errorrecovery.md
```diff
@@ -253,7 +253,21 @@ few lines of code.
 ## Making use of `%epp` for easier to read repair sequences
 
 By default, pretty-printing lexeme types prints out their identifier in the
-grammar. These rarely match what the user would expect:
+grammar. To this point these have matched the output the user would expect,
```
ltratt, line 256: Do we need to repeat this? I think this file has the lexer definition earlier on in it?

doc/src/quickstart.md
```diff
@@ -235,11 +235,11 @@ We can now `cargo run` our project and evaluate simple expressions:
 
 ```
 >>> 2 + 3
-Result: 5
+Result: Ok(5)
```
ltratt, line 238: Oops! If that does happen, IMHO we should `unwrap` the `Ok` somehow so that it just prints out `5` not `Ok(5)`.

ratmice:
> The error recovery algorithm has some non-determinism in it, so there can be changes in the order (there's no rhyme or reason to the order -- it's genuinely non-deterministic).

Okay, in that case, I'd like to fudge the output to match the existing text rather than ensure that the output matches what I saw when running through the snippets.

doc/src/quickstart.md
```diff
@@ -251,42 +251,42 @@ errors and have the system recover automatically:
 ```
 >>> 2 + + 3
 Parsing error at line 1 column 5. Repair sequences found:
-   1: Delete +
-   2: Insert INT
-Result: 5
+   1: Insert INT
```
ltratt, line 254: For this one, either order of repair sequences is correct, but I feel that explaining to users the `Err` case is more confusing than the `Delete +` case, so I'd be inclined to keep the original ordering in this one case.

doc/src/errorrecovery.md
ratmice: The lexer definitions all refer to the initial `calc.l` definition from the quickstart.md,
which now specify '*' instead of 'MUL', but this section explicitly needs 'MUL', etc.
So that it can be renamed '*' by `%epp MUL "*"`, substituting textual names for symbols during error recovery.

So this section kind of needs some textual name to motivate the example yes.

doc/src/quickstart.md
ratmice: Fixed by reverting back to the original repair sequence in [892cc71](https://github.com/softdevteam/grmtools/pull/366/commits/892cc71bbbae9eb24f9da69a547b80167198ef1d)

doc/src/quickstart.md
ratmice: I went with changing the pattern match from to `Ok(r)`, as that was used in errorrecovery.md previously.
That portion of errorrecovery required reordering when we make changes to main.rs to cope.
 [892cc71](https://github.com/softdevteam/grmtools/pull/366/commits/892cc71bbbae9eb24f9da69a547b80167198ef1d)

doc/src/errorrecovery.md
ratmice: I tried to clarify this text in [892cc71](https://github.com/softdevteam/grmtools/pull/366/commits/892cc71bbbae9eb24f9da69a547b80167198ef1d) to explain the motivation for this extra lex definition here.

doc/src/errorrecovery.md
```diff
@@ -253,9 +254,9 @@ few lines of code.
 ## Making use of `%epp` for easier to read repair sequences
 
 By default, pretty-printing lexeme types prints out their identifier in the
-grammar. To this point these have matched the output the user would expect,
-this is rarely always the case though.  If we used word identifiers instead of symbols
-as given by the following calc.l:
+grammar. Up until this point lexeme identifiers and the output has matched the output
```
ltratt, line 257: I can't really parse the "up until" sentence. Maybe "Up to now, we have used lexeme types when showing output to the user. While the lexeme types are sometimes adequate for this purpose, this is not always the case. Consider this lex file:"?

ratmice: Fixed that in [9e8b8e9](https://github.com/softdevteam/grmtools/pull/366/commits/9e8b8e9142500a82f4f9ebf5369619ec40778ae4) along another small change to the next sentence.

ltratt:
Please squash.

ratmice:
Squashed.

PR #367, opened 2022-11-03
Make examples and generated code work with rust_2018_idioms lint.

ratmice:
This is an update to my previous pull request #365, towards the end of writing this patch
my understanding of `rust_2018_idioms` changed, e.g. for some reason I thought it was a 2018 -> 2021 thing,
but it's actually a 2015 -> 2018 change.  It lints against code which was previously accepted by 2015 compilers,
was deprecated in the 2018 edition.

But the lints which deprecated have never been completed to the satisfaction of being enabled by default.
tracking issue: https://github.com/rust-lang/rust/issues/54910

As such i'm really not sure about the conditional code generation being necessary for this.
the new code generated for > 2015 was accepted by 2015 edition, and expect the old code to have lints enabled for them
in the future.

But perhaps this edition stuff would be useful to have everything in place for future edition changes?
As such I figure i'll just make the PR with what I have, and we can trim it down if we want...


lrlex/src/lib/ctbuilder.rs
```diff
@@ -68,6 +68,15 @@ impl Visibility {
     }
 }
 
+/// Specifies the [^Rust Edition] that will be emitted during code generation.
```
ltratt, line 71: Is the `^` intentional?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -315,6 +326,13 @@ where
         self
     }
 
+    /// Sets the rust edition to be used for generated code.  Defaults to the latest edition of
```
ltratt, line 329: Double space after the first `.`.

ratmice: I thought I was mirroring the surrounding text style, but it turns out the other double space was added by me,
I removed it too.

fixed in [bb1e970](https://github.com/softdevteam/grmtools/pull/367/commits/bb1e9702e37f638c0d70545a33661bc8aa47dbd9)

lrlex/src/lib/ctbuilder.rs
ratmice: It was part of the footnote syntax of cargo doc, but looking at the generated doc (oops),
it looked terrible with the footnote style.  I managed to get it working so the link text shows up inline.
while not making line-length terrible asymmetric.

ltratt: There are still a couple of other `[^`s in the PR -- should they disappear too?

ratmice: Sorry still :sleeping: :coffee:, they probably should.

ratmice: Fixed the one in lrlex now in [931202f](https://github.com/softdevteam/grmtools/pull/367/commits/931202ffb74712dc6251ec8cebc9fa25adbcec56)

ltratt:
Please squash.

PR #369, opened 2022-11-08
`Span` hasn't been a tuple for a while.

ltratt:


ratmice:
bors r+

Well spotted

PR #370, opened 2022-11-21
Enable forbid(unsafe_code) except in parser action code

ratmice:
I was reading through https://gist.github.com/repi/d98bf9c202ec567fd67ef9e31152f43f,
point 3.1 seemed like it should be easy to do since there isn't any usage of unsafe code.
That said I totally understand if it isn't something you want to merge.

In the generated code, this allows unsafe blocks within user actions, and adds a test for that,
I manually tested by removing the `#[allow(unsafe_code)]` that the new cttest fails.


ltratt:
Somehow I failed to notice this PR -- sorry!

This is a good idea -- thanks!

bors r+

PR #371, opened 2022-12-02
Generic `LexError`

ltratt:
This PR tries to make `LexError` a generic trait, so that we can (re)separate lrlex and lrpar. @ratmice I *think* this PR succeeds in its aims, but I'd be grateful if you could check it carefully, including whether it breaks "normal" use cases (e.g. `build.rs` files based on the quick start guide). I don't think we should blindly merge this PR until we're convinced that it doesn't impose an extra burden on users.

After two simple warmup commits, the main commit in this trio is https://github.com/softdevteam/grmtools/commit/9e38e9a2433adcb5a468212684431fce8eb5be14, whose commit message hopefully outlines what's been done and why.

### DIFF
--- a/doc/src/errorrecovery.md
+++ b/doc/src/errorrecovery.md
@@ -537,11 +537,15 @@ the first parsing error, with the `recoverer` method in `CTParserBuilder` or
 `RTParserBuilder`. For example, we can change `calc`'s `build.rs` file to:
 
 ```rust,noplaypen
-    let lex_rule_ids_map = CTParserBuilder::new()
-        .yacckind(YaccKind::Grmtools)
-        .recoverer(lrpar::RecoveryKind::None)
-        .grammar_path_in_src("calc.y")?
-        .process()?;
+CTLexerBuilder::new()
+    .lrpar_config(|ctp| {
+        ctp.yacckind(YaccKind::Grmtools)
+            .recoverer(lrpar::RecoveryKind::None)
+            .grammar_in_src_dir("calc.y")
+            .unwrap()
+    })
+    .lexer_in_src_dir("calc.l")?
+    .build()?;
 ```
 
 and then no matter how many syntax errors we make, only one is reported:
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -12,11 +12,11 @@ use num_traits::{PrimInt, Unsigned};
 use regex::{self, Regex, RegexBuilder};
 use try_from::TryFrom;
 
-use lrpar::{lex_api::StartStateId, LexError, Lexeme, Lexer, NonStreamingLexer};
+use lrpar::{Lexeme, Lexer, NonStreamingLexer};
 
 use crate::{
     parser::{LexParser, StartState, StartStateOperation},
-    LexBuildResult,
+    LRLexError, LexBuildResult, StartStateId,
 };
 
 #[derive(Debug)]
@@ -249,7 +249,7 @@ impl<
         let mut state_stack: Vec<(usize, &StartState)> = Vec::new();
         let initial_state = match self.get_start_state_by_id(0) {
             None => {
-                lexemes.push(Err(LexError::new(Span::new(i, i), None)));
+                lexemes.push(Err(LRLexError::new(Span::new(i, i))));
                 return LRNonStreamingLexer::new(s, lexemes, NewlineCache::from_str(s).unwrap());
             }
             Some(state) => state,
@@ -262,7 +262,7 @@ impl<
             let mut longest_ridx = 0; // This is only valid iff longest != 0
             let current_state = match state_stack.last() {
                 None => {
-                    lexemes.push(Err(LexError::new(Span::new(i, i), None)));
+                    lexemes.push(Err(LRLexError::new(Span::new(i, i))));
                     return LRNonStreamingLexer::new(
                         s,
                         lexemes,
@@ -293,7 +293,7 @@ impl<
                             lexemes.push(Ok(Lexeme::new(tok_id, old_i, longest)));
                         }
                         None => {
-                            lexemes.push(Err(LexError::new(Span::new(old_i, old_i), None)));
+                            lexemes.push(Err(LRLexError::new(Span::new(old_i, old_i))));
                             break;
                         }
                     }
@@ -302,7 +302,7 @@ impl<
                     let state = match self.get_start_state_by_id(*target_state_id) {
                         None => {
                             // TODO: I can see an argument for lexing state to be either `None` or `Some(target_state_id)` here
-                            lexemes.push(Err(LexError::new(Span::new(old_i, old_i), None)));
+                            lexemes.push(Err(LRLexError::new(Span::new(old_i, old_i))));
                             break;
                         }
                         Some(state) => state,
@@ -328,7 +328,7 @@ impl<
                                 }
                             }
                             None => {
-                                lexemes.push(Err(LexError::new(Span::new(old_i, old_i), None)));
+                                lexemes.push(Err(LRLexError::new(Span::new(old_i, old_i))));
                                 break;
                             }
                         },
@@ -336,11 +336,9 @@ impl<
                 }
                 i += longest;
             } else {
-                let current_state_id: Option<StartStateId> =
-                    StartStateId::try_from(current_state.id).ok();
-                lexemes.push(Err(LexError::new(
+                lexemes.push(Err(LRLexError::new_with_lexing_state(
                     Span::new(old_i, old_i),
-                    current_state_id,
+                    StartStateId::new(current_state.id),
                 )));
                 break;
             }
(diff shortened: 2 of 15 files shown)


ratmice:
I've had a read through the patch, and while just reading it seems nice.  I will have to do a more careful check tomorrow.
Just one minor comment on the commit message in that read through.

>  The trait solely defines a `span` method which seems about the only generic thing we can reasonably expect a lexer to
provide.

I would say something to the effect of "The trait extends the `Error` trait with a `span` method" etc, or something to that effect...

lrlex/src/lib/lexer.rs
```diff
@@ -337,8 +337,8 @@ impl<
                 i += longest;
             } else {
                 let current_state_id: Option<StartStateId> =
-                    StartStateId::try_from(current_state.id).ok();
-                lexemes.push(Err(LexError::new(
+                    Some(StartStateId::from(current_state.id));
```
ratmice, line 340: Fairly certain that we can change this to `StartStateId { _id: current_state.id }`, and remove the `From<usize>` impl,
now that this call-site is in the same crate as the structure.

ltratt: I did change this... and then realised that I marginally preferred `from` just because it abstracts away the `_` bit of `_id` (which we currently don't use). So I'm inclined to keep this as-is.

ltratt:
@ratmice Reviewing this again I realised that we'd made `LRLexError` a little difficult for people who want to use lrlex for manual lexing. https://github.com/softdevteam/grmtools/pull/371/commits/616f51f8d75743b81b729d757498bb64f43d5744 is my attempt to fix this. See what you think -- if you like it, it'll certainly need squashing into the previous commit.

lrlex/src/lib/lexer.rs
ratmice: I hope you don't mind if I make one last effort to dissuade this, because the `From` trait is `pub`, and `impl` has implicit visibility qualifiers, this makes it possible for users to construct a `StartStateId`, when really there is no benefit for having that public.

Maybe a compromise would be to move it into a non-pub `fn new` like below?

```
impl StartStateId {
    fn new(_id: usize) -> Self {
          Self { _id }
    }
}
```

ltratt: Very good point! Fixed in dcc5838.

ratmice (commented):
The [turning off error recovery](https://github.com/ltratt/grmtools/blob/generic_lex_error/doc/src/errorrecovery.md#turning-off-error-recovery) section also will need an update either by changing it to use lrpar_config like:

```
CTLexerBuilder:::new()
    .lrpar_config(|ctp| {
        ctp.yacckind(YaccKind::Grmtools)
            .recoverer(lrpar::RecoveryKind::None)
            .grammar_in_src_dir("calc.y")
            .unwrap()
    })
    .lexer_in_src_dir("calc.l")?
    .build()?;
```
Alternately if we want to keep the same style, I believe we'd need to add type annotations:
```
use lrlex::LRLexError;
let lex_rule_ids_map = CTParserBuilder::<DefaultLexeme, u32, LRLexError>new()
        .yacckind(YaccKind::Grmtools)
        .recoverer(lrpar::RecoveryKind::None)
        .grammar_in_src_dir("calc.y")?
        .build()?
        .token_map();
```

(finally just to write this down, so I remember to make a separate patch for it), that example uses a function `process()` in place of `build()`, which I believe was a combination of `build()` and `token_map()`.  But I don't actually see this function appearing in any release.

It is referred to in some more doc strings:
* [new()](https://docs.rs/lrpar/latest/lrpar/struct.CTParserBuilder.html#method.new) should be .build()
* [process_file()](https://docs.rs/lrpar/latest/lrpar/struct.CTParserBuilder.html#method.process_file) should be ".build(), and .token_map()'

ltratt:
> The [turning off error recovery](https://github.com/ltratt/grmtools/blob/generic_lex_error/doc/src/errorrecovery.md#turning-off-error-recovery) section also will need an update either by changing it to use lrpar_config like:

That was quite out of date anyway! Fixed in 9c972b5.

I think I've fixed the others in 0064bd9.

Good spots!

ratmice:
So, beyond that I also tried updating a few random projects found at, https://github.com/softdevteam/grmtools/network/dependents
the only things I noticed one of them using the old build.rs builder style with separately constructed `CTParserBuilder`/`CTLexerBuilder` i.e not `lrpar_config` which were would require a bit of effort to switch to using `lrpar_config`.

In that case I had to add type annotations.
In one other case it was updating from 0.9, I was momentarily confused because `NonStreamingLexer<StorageT>` -> `NonStreamingLexer<LexemeT, StorageT>` -> `NonStreamingLexer<LexemeT, StorageT, LexErrorT>`,

So when upgrading from 0.9, the existing type parameter ends up in the middle.
This is kind of unfortunate, but I think `LexErrorT` probably does belong at the end.

ltratt:
> So, beyond that I also tried updating a few random projects

Cool, thanks for trying that! It sounds like "mostly there's no change, and where there is a change, we're probably about as good (or unbad!) as we can be"?

doc/src/errorrecovery.md
```diff
@@ -537,11 +537,15 @@ the first parsing error, with the `recoverer` method in `CTParserBuilder` or
 `RTParserBuilder`. For example, we can change `calc`'s `build.rs` file to:
 
 ```rust,noplaypen
-    let lex_rule_ids_map = CTParserBuilder::new()
-        .yacckind(YaccKind::Grmtools)
-        .recoverer(lrpar::RecoveryKind::None)
-        .grammar_path_in_src("calc.y")?
-        .process()?;
+CTLexerBuilder:::new()
```
ratmice, line 540: Oops,  ':::new()', here should be '::', twas my bad.

ratmice:
> > So, beyond that I also tried updating a few random projects
> 
> Cool, thanks for trying that! It sounds like "mostly there's no change, and where there is a change, we're probably about as good (or unbad!) as we can be"?

Yeah, I guess the only thing I wondered about is the removal of `StorageT = u32`, and instead adding `LexErrorT = LRLexError` as well for default type parameters.  Then adding a section for the *manual written lexers`, I presumed you had reasons for removing the existing default.

Anyhow, I believe that I am happy with this if you are as well, so if you feel it is ready then: Please squash

doc/src/errorrecovery.md
ltratt: My bad! Fixed in 8901c94.

ltratt:
> Yeah, I guess the only thing I wondered about is the removal of StorageT = u32, and instead adding LexErrorT = LRLexError as well for default type parameters.

I should have pointed this one out. The problem is that Rust forces default type paramaters to come last so we can't have `StorageT=u32` without `LexErrorT=<something>` but I don't know what `<something>` could be. lrpar shouldn't rely, IMHO, on lrlex, so saying `LexErrorT=LRLexError` would feel wrong.

Fortunately -- and I think you've verified this! -- because of the way people typically use `build.rs` they pick up a `=u32` default type parameter from lrlex, so the removal it from lrpar doesn't matter.

ratmice:
> Fortunately -- and I think you've verified this! -- because of the way people typically use `build.rs` they pick up a `=u32` default type parameter from lrlex, so the removal it from lrpar doesn't matter.

Fair enough, it is only those rare cases where you construct `CTLexerBuilder`, and `CTParserBuilder` separately where it matters,
that is probably rare enough for *new* projects.  I think the only reason I thought of it is that there were some old *build.rs* files in the things that I ported, using the separately constructed Builders.

In those cases of updating from really old versions, it is a little difficult to follow the error messages e.g. if you rename `LexerBuilder` -> `CTLexerBuilder`, you now need to add annotations, whereas in previous versions it seems like such annotations weren't required..

I suppose that is the only reason I bring it up.

ltratt:
If we can think of a nicer way of helping nudge people forward I'd be happy to do so! I don't have any great ideas, though, unfortunately :/

ratmice:
Yeah, I think the only 2 thoughts that I had were:

1. Adding the defaults back so we have `pub struct CTParserBuilder<'a, LexemeT, StorageT=u32, LexErrorT=LRLexError>`
2. Perhaps generic associated types, would work but that would both be a pretty large departure
from the current code and tie us to a new compiler version.  I think it'd be much easier to experiment with that on a smaller scale though before trying to do it in grmtools proper, and it'd probably take a while.

ratmice:
So, I decided to give it one more try, and managed to get what I was trying working...
It doesn't actually help with inference (There are maybe things we could do here by providing a `const` instance of this new type I added and passing it to e.g. `new()`), but it does reduce the number of type parameters.

I thought it was worthwhile to post it before I went further though, here is the commit:
https://github.com/ratmice/grmtools/commit/cfacb5fcf6bdc724c014df024e6b4c70fb7dffc6
It is built on top of this PR, in the following branch.
https://github.com/ratmice/grmtools/tree/generic_lexerror_assoc

ltratt:
I had not thought of using associated types -- interested! I'm not 100% sure how best to handle this, but I can see that we might be able to tame all the type parameters somewhat -- I probably need to experiment with this a bit.

I suggest that we get this PR in (if you're happy with it!) and then in the next PR I can experiment with the associated types idea and see what we can do.

ratmice:
Sounds good to me, please squash.

ltratt:
Squashed.

PR #373, opened 2022-12-22
Keep cargodoc happy.

ltratt:
Without backticks, [`crate::Node<StorageT>`] is now interpreted as HTML (this wasn't previously the case).


ratmice:
bors r+

Its interesting, the rustdoc docs for links seem quite different from the actual behavior now in a number of ways,
but that is for another place :).

ltratt:
Yes, I haven't kept up with rustdoc, and I suspect the codebase needs a good pass, both to bring it upto date, and also to fix things that don't work as they used to.

PR #378, opened 2023-01-07
Less serde

ltratt:
One of the things that sunk my attempts to make nice `trait` containers for type parameters was that `Serialize` crept in too often. Having looked at this afresh, it seems that some of lrpar's `Serialize` use is unnecessary. Even if this doesn't help me `trait`ify things, having fewer unnecessary trait bounds probably isn't a bad thing in its own right!

### DIFF
--- a/lrpar/src/lib/parser.rs
+++ b/lrpar/src/lib/parser.rs
@@ -12,6 +12,7 @@ use cactus::Cactus;
 use cfgrammar::{yacc::YaccGrammar, RIdx, Span, TIdx};
 use lrtable::{Action, StIdx, StateTable};
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
+#[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
 use crate::{cpctplus, LexError, Lexeme, NonStreamingLexer};
@@ -603,7 +604,8 @@ pub(super) trait Recoverer<
 }
 
 /// What recovery algorithm should be used when a syntax error is encountered?
-#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum RecoveryKind {
     /// The CPCT+ algorithm from Diekmann/Tratt "Don't Panic! Better, Fewer, Syntax Errors for LR
     /// Parsers".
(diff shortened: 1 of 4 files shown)


lrpar/src/lib/parser.rs
```diff
@@ -603,7 +602,7 @@ pub(super) trait Recoverer<
 }
 
 /// What recovery algorithm should be used when a syntax error is encountered?
-#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[derive(Clone, Copy, Debug)]
```
ratmice, line 605: It seems that I am actually using this `Deserialize` impl here at https://github.com/ratmice/nimbleparse_lsp/blob/main/toml/src/lib.rs#L21 there is an example of the toml format that library reads here: https://github.com/ratmice/crimson/blob/main/nimbleparse.toml

Though I've generally been setting it to `None`, and considering changing the default_recovery_kind function to `None` there, because it is so much faster during interactive editing with nimbleparse_lsp.

It probably will not be until tomorrow evening before I have a chance to think about what I want to replace that linked code with.  I guess the implication here for removing it is that if grmtools adds a new `RecoveryKind`, currently it will automagically be supported when updating the version.  If we go through with removing it it would require a minor change if/when a new `RecoveryKind` is added.

ltratt: Aha, that's interesting. I think lrpar should grow a `serde` feature to deal with this. Would that help?

ratmice: Certainly wouldn't hurt, I don't think it really needs to be enabled by default, I've never seen another project using that feature, and that would match cfgrammar?

ltratt: Agreed. I'll have a go tomorrow.

ltratt: Does f69679d work for you?

ratmice: Almost, I tested it out in my project the previous patch in the series removed the use `serde::{Deserialize, Serialize}`.
That needs to be readded with `#[cfg(feature = "serde")]` or changed to  use `derive(serde::Serialize, serde::Deserialize)`
But it's fine with either of those changes.

ltratt: Oops! Fixed in 6ad3752 and also added a test of the serde feature to buildbot to stop this one creeping in accidentally in the future.

ratmice:
Please squash.

ltratt:
Squashed.

PR #379, opened 2023-01-09
Bundle the lex API types into a trait/struct.

ltratt:
The aim of this PR is to try and reduce the number of generic parameters that users have to deal with. This is a partial success: we still put more of a burden onto users than we perhaps should. That said, we do at least get rid of the `LexError` types.

I'm raising this for comments: I'm not yet convinced this is good enough to go in (but I might just need to sleep on it). I feel like we can probably do a better job at hiding more ickiness from users, but I'm curious what you think, @ratmice.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -18,14 +18,14 @@ use std::{
 
 use cfgrammar::{newlinecache::NewlineCache, Spanned};
 use lazy_static::lazy_static;
-use lrpar::{CTParserBuilder, Lexeme};
+use lrpar::{CTParserBuilder, LexerTypes};
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
 use quote::quote;
 use regex::Regex;
 use serde::Serialize;
 use try_from::TryFrom;
 
-use crate::{DefaultLexeme, LRLexError, LRNonStreamingLexerDef, LexerDef};
+use crate::{DefaultLexerTypes, LRNonStreamingLexerDef, LexerDef};
 
 const RUST_FILE_EXT: &str = "rs";
 
@@ -80,42 +80,40 @@ pub enum RustEdition {
 
 /// A `CTLexerBuilder` allows one to specify the criteria for building a statically generated
 /// lexer.
-pub struct CTLexerBuilder<'a, LexemeT: Lexeme<StorageT>, StorageT: Debug + Eq + Hash = u32> {
-    lrpar_config: Option<
-        Box<
-            dyn Fn(
-                CTParserBuilder<LexemeT, StorageT, LRLexError>,
-            ) -> CTParserBuilder<LexemeT, StorageT, LRLexError>,
-        >,
-    >,
+pub struct CTLexerBuilder<'a, LexerTypesT: LexerTypes = DefaultLexerTypes<u32>>
+where
+    LexerTypesT::StorageT: Debug + Eq + Hash,
+    usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+{
+    lrpar_config: Option<Box<dyn Fn(CTParserBuilder<LexerTypesT>) -> CTParserBuilder<LexerTypesT>>>,
     lexer_path: Option<PathBuf>,
     output_path: Option<PathBuf>,
     lexerkind: LexerKind,
     mod_name: Option<&'a str>,
     visibility: Visibility,
     rust_edition: RustEdition,
-    rule_ids_map: Option<HashMap<String, StorageT>>,
+    rule_ids_map: Option<HashMap<String, LexerTypesT::StorageT>>,
     allow_missing_terms_in_lexer: bool,
     allow_missing_tokens_in_parser: bool,
 }
 
-impl<'a> CTLexerBuilder<'a, DefaultLexeme<u32>, u32> {
+impl<'a> CTLexerBuilder<'a, DefaultLexerTypes<u32>> {
     /// Create a new [CTLexerBuilder].
     pub fn new() -> Self {
-        CTLexerBuilder::<DefaultLexeme<u32>, u32>::new_with_lexemet()
+        CTLexerBuilder::<DefaultLexerTypes<u32>>::new_with_lexemet()
     }
 }
 
-impl<'a, LexemeT, StorageT> CTLexerBuilder<'a, LexemeT, StorageT>
+impl<'a, LexerTypesT: LexerTypes> CTLexerBuilder<'a, LexerTypesT>
 where
-    LexemeT: Lexeme<StorageT>,
-    StorageT: 'static + Copy + Debug + Eq + Hash + PrimInt + Serialize + TryFrom<usize> + Unsigned,
-    usize: AsPrimitive<StorageT>,
+    LexerTypesT::StorageT:
+        'static + Debug + Eq + Hash + PrimInt + Serialize + TryFrom<usize> + Unsigned,
+    usize: AsPrimitive<LexerTypesT::StorageT>,
 {
     /// Create a new [CTLexerBuilder].
     ///
-    /// `StorageT` must be an unsigned integer type (e.g. `u8`, `u16`) which is big enough to index
-    /// all the tokens, rules, and productions in the lexer and less than or equal in size
+    /// `LexerTypesT::StorageT` must be an unsigned integer type (e.g. `u8`, `u16`) which is big enough
+    /// to index all the tokens, rules, and productions in the lexer and less than or equal in size
     /// to `usize` (e.g. on a 64-bit machine `u128` would be too big). If you are lexing large
     /// files, the additional storage requirements of larger integer types can be noticeable, and
     /// in such cases it can be worth specifying a smaller type. `StorageT` defaults to `u32` if
@@ -124,7 +122,7 @@ where
     /// # Examples
     ///
     /// ```text
-    /// CTLexerBuilder::<DefaultLexeme<u8>, u8>::new_with_lexemet()
+    /// CTLexerBuilder::<DefaultLexerTypes<u8>>::new_with_lexemet()
     ///     .lexer_in_src_dir("grm.l", None)?
     ///     .build()?;
     /// ```
@@ -163,10 +161,7 @@ where
     /// ```
     pub fn lrpar_config<F>(mut self, config_func: F) -> Self
     where
-        F: 'static
-            + Fn(
-                CTParserBuilder<LexemeT, StorageT, LRLexError>,
-            ) -> CTParserBuilder<LexemeT, StorageT, LRLexError>,
+        F: 'static + Fn(CTParserBuilder<LexerTypesT>) -> CTParserBuilder<LexerTypesT>,
     {
         self.lrpar_config = Some(Box::new(config_func));
         self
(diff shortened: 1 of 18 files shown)


ltratt:
bors try

lrpar/src/lib/lex_api.rs
```diff
@@ -3,26 +3,41 @@
...
+    type LexemeT: Lexeme<Self::StorageT>;
+    type StorageT: 'static + fmt::Debug + Hash + PrimInt + Unsigned;
+    type LexErrorT: LexError;
+}
 
 /// The base trait which all lexers which want to interact with `lrpar` must implement.
-pub trait Lexer<LexemeT: Lexeme<StorageT>, StorageT: Hash + PrimInt + Unsigned, LexErrorT: LexError>
+pub trait Lexer<
+    StorageT: 'static + Hash + PrimInt + Unsigned,
+    LexerTypesT: LexerTypes<StorageT = StorageT>,
```
ratmice, line 20: I need to look at this more carefully, but I couldn't remember why we needed both `StorageT`, and `LexerTypesT::StorageT` here. So I tried to combine them being certain that there was a reason and that would remind me.  The attempt seemed to work, but it is quite possible that I am missing out on some detail here.

https://github.com/ratmice/grmtools/commit/62879fac7422028fdffebcac1177a6a50081ef01

ratmice: I went ahead and tried with `CTLexerBuilder` and `CTParserBuilder` to see if similar change would work there too,
and that seemed to work too.  https://github.com/ratmice/grmtools/commit/db9298ddd0a405140f76b3d8eef567d3882624e1

ltratt: Yes at the moment we don't need a separate `StorageT`. My vague thinking -- and I guess I'll actually have to try this a bit to know if it's a good idea or not! -- is that we might want `ParserTypes` to match `LexerTypes`. An advantage of doing this is that we would open up the possibility of one day having separate types for rules, productions, and so on, without sharing the same `StorageT` -- in a sense, `StorageT` is my hack to avoid proliferating type parameters.

ratmice: Aha, now it makes more sense to me that this is intended as an extension point rather than an opportunity for simplification.

While admittedly not quite understanding the plan in detail one thing I worry about is that `LexerTypes<StorageT = StorageT>` is setting up a type equivalence for these 2 types as a part of the public signature, so there is a tinge of worry that this equivalency "leaks out" in a way that could make changing either of the types be a breaking change.

So there is a bit of hesitation that this could end up in a situation where we could ultimately end up with 2 copies of the same type parameter which must be equivalent, for no gain of flexibility. But I don't feel like I have a good enough understanding here to say whether this skepticism is actually warranted or not.

ltratt: On reflection, I think I was reaching too far, too soon (and, possibly, too far ever!). I've cherry picked your commits and added a minor fix I found when porting something using grmtools.

I have made an attempt to remove more `StorageT`s, but have hit a brick wall where `where` clauses want to proliferate. I'll have to start from scratch, but that can wait until tomorrow. I thought it's useful to let you know that I am still chipping away at this in the meantime!

lrlex/src/lib/ctbuilder.rs
```diff
@@ -414,7 +414,7 @@ where
             LexerKind::LRNonStreamingLexer => (
                 "LRNonStreamingLexerDef",
                 format!(
-                    "LRNonStreamingLexerDef<{storaget}, {lexertypest}<{storaget}>>",
+                    "LRNonStreamingLexerDef<{storaget}, {lexertypest}>",
```
ratmice, line 417: Good catch, seems like an obscure one!

lrpar/src/lib/lex_api.rs
ltratt: I have spent some time playing around, and I *think* we can do better but it's going to be a lot of work. I might still try doing that, but it's going to be too much for this PR -- and I think this PR does improve the current situation. So I think we should review this PR as-is, consider it for merging, then I may continue experimenting separately.

[Current problems include the fact that the `Step` trait is marked unstable as it's been "redesigned" -- but the redesign happened in 2017... So we can't do `StorageT::zero()..<blah>` which is frustrating.]

ratmice: I think I see what you mean about the `Step` trait, but I've never actually noticed it before. I wonder if using the [`NumOps`](https://docs.rs/num-traits/latest/num_traits/trait.NumOps.html) instead might be an option (this question may be absurd?).

As it is on the PR i'll have another look through, I kind of view this sort of change as a tradeoff.  It kind of makes the implementation more complex, with more difficult trait bounds and such, but avoids pushing lots of generics onto callers.
I also kind of lean towards thinking it seems like an improvement.

ltratt: > I wonder if using the [NumOps](https://docs.rs/num-traits/latest/num_traits/trait.NumOps.html) instead might be an option (this question may be absurd?).

Good question... dunno?!

> It kind of makes the implementation more complex, with more difficult trait bounds and such, but avoids pushing lots of generics onto callers.

Agreed, for 90% of users, it simplifies things. For advanced users, it complicates things, but that's probably a tolerable trade-off.

lrlex/src/lib/ctbuilder.rs
ratmice: I had misread this as *adding* the `<{storaget}>` rather than *removing* it and figuring that it was previously always generating code using the default. Just so I understand, this is removing `<{storaget}>`, because `type_name::<LexerTypesT>()` actually already contains the `<{storaget}>` part.  So this works now for both the cases where a non-default, and default `StorageT` are provided. The following playground link seems to indicate so.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=47ef81f885c6b78264af6bc88bc3fc42

ltratt: Correct: before this fix we ended up producing things like `LexerTypes<u8><u8>`.

lrpar/src/lib/lex_api.rs
ratmice: > I have spent some time playing around, and I _think_ we can do better but it's going to be a lot of work. I might still try doing that, but it's going to be too much for this PR -- and I think this PR does improve the current situation. So I think we should review this PR as-is, consider it for merging, then I may continue experimenting separately.

I can agree with this, I went ahead and tried looking into the `StorageT` generics that seemed to be left in `LexerDef` and `LRNonStreamingLexer`, akin to the patches I had previously done in this series, and there is quite a bit of fallout from those.  Is that the kind of stuff you were referring to for a separate PR?

Here is that additional patch https://github.com/ratmice/grmtools/commit/d604ca721ade1fb8f015548eb202cf1efa6d5fc4

ltratt: I was trying that + some other deeper changes, but I should have stopped earlier! I've cherry picked your commit into this branch.

ltratt:
@ratmice I think this PR is now as complete as we should try making it. If you're happy with it, we'll need to bring in an external reviewer (@ptersilie will hopefully oblige!). If/when he's happy with things, we can then squash down to a single co-authored commit.

[One thing I did wonder is what knock-on documentation changes we might have missed... If anyone spots any, please shout!]

ratmice:
@ltratt I'm happy with it, I noticed one last thing in the series still using `NonStreamingLexer<StorageT, LexerTypes>`; but feel like i'm already pushing the envelope on this patch. I haven't had a look through the docs yet, but will do a pass over them.

ltratt:
> I noticed one last thing in the series still using NonStreamingLexer<StorageT, LexerTypes>

Ah, where?

lrpar/src/lib/lex_api.rs
```diff
@@ -3,26 +3,39 @@
...
 pub trait NonStreamingLexer<
     'input,
-    LexemeT: Lexeme<StorageT>,
-    StorageT: Hash + PrimInt + Unsigned,
-    LexErrorT: LexError,
->: Lexer<LexemeT, StorageT, LexErrorT>
+    StorageT: 'static + Hash + PrimInt + Unsigned,
+    LexerTypesT: LexerTypes<StorageT = StorageT>,
+>: Lexer<LexerTypesT> where
+    usize: AsPrimitive<StorageT>,
```
ratmice, line 38: I believe this one is the last remaining instance of `<StorageT, LexerTypesT>`.
I had a look at doing the same for this as I had done for the others, but have run into some troubles with type inference there in the tests.

ltratt: I'll have a look.

ltratt: 6b65f60 should do the trick (we no longer needed all those generics in the test suite, as we hardcode `u16` in various places, hence the bizarre warning saying "expected u16, got u16"!).

ratmice: Awesome, going to resolve this and overall happy with the patch now.

ltratt: Cool! I think it's just the inevitable missing bits in the docs that we need to think about now?

ratmice: Yeah, that sounds right to me.

ltratt:
@ratmice Did you spot any doc issues in the end?

Either way, @ptersilie I think this is ready for code review (perhaps there'll be a minor doc change or two to come, but maybe not!).

ptersilie:
I've only done a shallow review as I realised that this project is starting to go over my head. But I could see that you got rid of some cruft and simplified bits, which is always good. Anything you want to squash or is this ready to go?

ltratt:
Let's see if @ratmice turns up any documentation doo-dahs. If he does, I'll fix those; if he doesn't, then we can probably squash this.

ratmice:
Towards the end of my pass through the docs, I noticed there are places still that refer to `StorageT` e.g. https://softdevteam.github.io/grmtools/master/api/lrlex/struct.CTLexerBuilder.html#method.rule_ids_map
rather than `LexerTypesT::StorageT`, the fully qualified type to the trait is directly above it in the function signature though.
There are perhaps more instances like this floating about.  This is complicated a bit by the fact that some types e.g: https://softdevteam.github.io/grmtools/master/api/lrlex/trait.LexerDef.html still take `StorageT` directly as a parameter.
I wasn't sure how we wanted to handle this, and it would need another pass over them looking for that.

I did find a few code snippets that needed updating, I haven't actually tried compiling the examples with all of setting up cargo to build dependencies from a PR however. Those are at the following

https://github.com/ratmice/grmtools/commit/0e33b0c4b9091b40917c8f4b3c67c967aecb2bfe
on this branch https://github.com/ratmice/grmtools/tree/associated_types_docs

ltratt:
I did wonder about things like `LexerDef`, but wondered if the churn to `LexerTypes` was worth it or not. Perhaps, on reflection, consistency of `LexerTypes` is less confusing for the user?

ratmice:
Perhaps, I really don't have much preference though just noting that it makes the usage of `StorageT` a bit context dependent, so I didn't go replace them all if you preferred I e.g. always use `LexerTypesT::StorageT`.  I'm not sure if there is any future reason we might need additional types in `LexerTypes` within `LexerDef` though it seems unlikely?

ltratt:
@ratmice I've slurped in your doc change commit -- thanks!

>  I'm not sure if there is any future reason we might need additional types in LexerTypes within LexerDef though it seems unlikely?

I agree it's probably unlikely but that's more of an internal detail than an external one. But I think I can do that in another PR, as one's already got quite big and I pity the poor reviewer :)

@ptersilie If you're OK with this PR, I think it's now ready for squashing.

ptersilie:
Excellent. Please squash.

ltratt:
Squashed.

PR #381, opened 2023-01-17
More `LexerType` bits

ltratt:
Overall, to my mind, this is probably worth the consistency in the external API, even though it does make a few of the methods very slightly messier internally. I don't feel strongly about this, though, so please shout if you disagree!


ratmice:
I don't really have a strong opinion here either, but somewhat lean towards this for the consistency reasons mentioned,
ready to go?

ltratt:
Ready as far as I'm concerned!

PR #382, opened 2023-01-17
make generictree action fn public

ratmice:
This is an RFC about making generic_ptree pub, and renaming it actions_generictree.
In particular, I would like to make some parse action interruptable using panic_handler, and an AtomicBool
then call the normal GenericParseTree action, and for this I need the function underlies GenericParseTree action kind.

I had some difficulty trying to add this to the existing `lrpar::Parser::do_tests` testing based workflow,
So for the moment there are no specific tests that e.g. `parse_generictree(...) == parse_actions(&[action_generictree; ...]);`,
in part because the tests are shared between paths that have non-deterministic order to the output with recovery I believe.

### DIFF
--- a/lrpar/src/lib/mod.rs
+++ b/lrpar/src/lib/mod.rs
@@ -207,6 +207,7 @@ pub use crate::{
     parser::{LexParseError, Node, ParseError, ParseRepair, RTParserBuilder, RecoveryKind},
 };
 
+pub use crate::parser::action_generictree;
 /// A convenience macro for including statically compiled `.y` files. A file `src/a/b/c.y`
 /// processed by [CTParserBuilder::grammar_in_src_dir] can then be used in a crate with
 /// `lrpar_mod!("a/b/c.y")`.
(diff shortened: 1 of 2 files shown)


ltratt:
I think I get the general idea, though I'm less clear on the specifics. I think what you're fundamentally asking for is some sort-of generic callback mechanism that can be used for actions, rather than having to either: specify each action manually; get back a generic parse tree and then do something over it?

ratmice:
Sorry, actually we already have such a thing (for `RTParserBuilder`, unsure about `CTParserBuilder`). Which is [`parse_actions`](https://docs.rs/lrpar/latest/lrpar/struct.RTParserBuilder.html#method.parse_actions).

So literally just asking to make the underlying actions which implements the `GenericParseTree` to be made `pub`.
then I can wrap it like:
```
fn action_blah(...) {
   blah();
   action_generictree()
}

pb.parse_actions(..., &[action_blah, ...], ())
```

Thus the callback mechanism is already there, but the specific callback which implements generic trees is private.

ltratt:
I think I see. So is this PR all you need to do that? [I was almost expecting you'd need lots more stuff to do it!]

ratmice:
Nope this is it!

ltratt:
Then I'm definitely fine with doing this!

bors r+

Issue #349, opened 2022-09-05
Adding a %grammar-kind declaration?


ratmice:
Before I try and come up with a patch, I figured it would be good to discuss this in an issue,
I was considering potentially adding a declaration `%grammar-kind Original(NoAction)`, etc

One of the problems with this is that it is likely that we want to parse the value by just using `Deserialize` on the `YaccGrammarKind`,
this would at least be the *easiest* way. But it brings about a few issues 

1. `cfgrammar` has Optional deserialization support, so if we deserialized that way `%grammar-kind` would only work with serde feature enabled.  Alternately we could just implement this by hand instead of serde?
2. Some declarations depend upon a specific `%grammar-kind`, we may have to move some checks from `parse_declarations` to `ast.complete_and_validate`.

But it could potentially reduce the number of places that `YaccGrammarKind` needs to be specified (build.rs, nimbleparse command line, etc).  So it seemed like it might be worth considering.

ltratt:
Oops, I missed this one entirely! I think I'm broadly in favour of this, though we'd have to define what happens if a user specifies `YaccGrammarKind` *and* the grammar specifices `%grammar-kind`. Probably "specifying `YaccGrammarKind` overrides `%grammar-kind`" is reasonable?

ratmice:
Yeah, that seems like the logical/reasonable choice to me, I'll look into an exploratory patch for it.

ratmice:
So I have also been thinking about this as related to #490 and `RegexOptions` 
We can see an example in [cttests/src/regex_opt.test](https://github.com/softdevteam/grmtools/blob/master/lrpar/cttests/src/regex_opt.test) of this embedded in yaml.

One of the difficulties I ran into with `%grammar-kind` is that it doesn't fit well with the parsing of other directives, it is a new form of directive in that it actually affects other directives, I think that it should probably have the following properties.

1. Because it affects other directives it needs to come first.
2. Because the affect on directives needs to be consistent throughout the file it should be unique.

Things like `RegexOptions` being multi-key brings about a slightly different need, the following is an idea I have been considering including a `%header` directive.

for yacc:
```
%header {
   grammar-kind Original(YaccOriginalActionKind::NoAction)
}
```

for lex:

```
%header {
  RegexOptions octal !dot_matches_new_line posix_escapes
}
```

in @ltratt comment above he mentions
[...]

PR #384, opened 2023-01-18
Better document the changes users might see.

ltratt:
I (re)migrated a project using grmtools-0.12 to see what breaks and realised that we'd missed some things out in `CHANGES.md`. This commit adds in all the things that I know users will encounter (there might be others, but hopefully not!).

@ratmice I'm starting to think that we're also getting to the right point to make a 0.13 release. Are there any other parts of the API that you can see should be considered for change before we do so?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 This release contains a number of new features and breaking changes. Most of
 the breaking changes are in advanced/niche parts of the API, which few users
-will notice. However, two breaking changes might affect a more substantial
+will notice. However, four breaking changes might affect a more substantial
 subset of users: those are highlighted in the "breaking changes (major)"
 section below.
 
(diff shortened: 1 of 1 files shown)


ratmice:
> @ratmice I'm starting to think that we're also getting to the right point to make a 0.13 release. Are there any other parts of the API that you can see should be considered for change before we do so?

Forgot to respond to this.  There isn't anything critical, I had looked at #349 and it is going to take some thinking since it delays the time at which we may first have a `YaccGrammarKind` until much later than we currently do, and there are various assumptions that we have one.  That being such a larger change relative to the minor enhancement it provides I don't think it may not be worth delaying, or trying to rush getting that in the release.

ltratt:
Hmm, yes, #349 might not be worth waiting for. OK, my plan is to make a 0.13 release tomorrow PM GMT/UTC, unless anyone thinks of something that should cause us to pause.

Issue #129, opened 2019-12-15
Allow leading "|" in first alternative of a production


pablosichert:
It would be nice if it was allowed to place a leading `|` in the first alternative of a production.

E.g.:

```rust
Production -> [...]:
 |  Foo { [...] }
 |  Bar { [...] }
 |  Baz { [...] }
;
```

As far as I'm aware this wouldn't change any semantics, as also the `ε`-alternative is required to have a body `{}`.

ltratt:
For Yacc and Bison compatibility, we can't do this, because grammars don't have to have actions (e.g. for https://docs.rs/lrpar/0.4.3/lrpar/struct.RTParserBuilder.html#method.parse_generictree). If we were to do this, these two grammars would be impossible to parse unambiguously:

```
A: B | C;
```

vs.

```
A: | B | C;
```

Does `A` have an epsilon rule or not?

benblank:
Apologies for necroing this issue, but it describes what I want exactly, so I wasn't sure creating a new one would be appropriate. 🙂

I'm fairly new to grmtools (and, in fact, Yacc in general) and find the "inconsistent" leading whitespace on this type of rule to be somewhat irritating, particularly when rearranging productions. Having to format the first production specially (no vertical bar) is much the same kind of annoyance as having to format the last element of an array/vector/list specially in languages which do not allow trailing commas. And for grmtools, it even feels like a bit of a backwards step when compared to the roughly-equivalent Rust syntax (match branches, which *do* allow a vertical bar in front of the first pattern).

I had a few thoughts on how this might be accomplished without creating ambiguity:

1. **Introduce an `%explicit-empty` (or similar) declaration**
  Using this declaration would then require empty branches to use an [`%empty` declaration](https://www.gnu.org/software/bison/manual/html_node/Empty-Rules.html), allowing `A: | B | C;` to be easily distinguished from `A: %empty | B | C;`. Anyone wanting to use the "ignored initial vertical bar" syntax described in this issue would have to opt in to it and explicitly accept responsibility for keeping the grammar unambiguous.
2. **Allow (and merge) "duplicate" rules**
  I have found a few instances ([1](https://docs.oracle.com/cd/E19504-01/802-5880/6i9k05dgt/index.html#6i9k05dh0), [2](https://www.gnu.org/software/bison/manual/bison.html#Explanation-of-exp)) of the vertical bar being described as syntactic sugar for repeating the rule's LHS multiple times. In other words, the following would be considered equivalent:

[...]

PR #389, opened 2023-01-23
Allow `%empty` to be used to highlight empty productions.

ltratt:
This is a Bison-ism, but it seems reasonable to allow it even in "Yacc" mode.

Addresses one suggestion from https://github.com/softdevteam/grmtools/issues/129.


ratmice:
This looks good to me, and now I can get rid of a bunch of Empty:; rules :+1: 
bors r+

PR #390, opened 2023-01-25
Implement repeated rules discussed in #129

ratmice:
This implements the other part of the suggestion, in #129 by @benblank to remove the `DuplicateRules` error and combine the productions of duplicate rules together, a lot of this is just testsuite fallout, where `DuplicateRule` was one of the default choices for errors that contain multiple spans.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -28,7 +28,6 @@ pub enum YaccGrammarErrorKind {
     IllegalName,
     IllegalString,
     IncompleteRule,
-    DuplicateRule,
     IncompleteComment,
     IncompleteAction,
     MissingColon,
@@ -85,7 +84,6 @@ impl fmt::Display for YaccGrammarErrorKind {
             YaccGrammarErrorKind::IllegalName => "Illegal name",
             YaccGrammarErrorKind::IllegalString => "Illegal string",
             YaccGrammarErrorKind::IncompleteRule => "Incomplete rule",
-            YaccGrammarErrorKind::DuplicateRule => "Duplicated rule",
             YaccGrammarErrorKind::IncompleteComment => "Incomplete comment",
             YaccGrammarErrorKind::IncompleteAction => "Incomplete action",
             YaccGrammarErrorKind::MissingColon => "Missing ':'",
@@ -241,8 +239,7 @@ impl Spanned for YaccGrammarError {
             | YaccGrammarErrorKind::UnknownToken(_)
             | YaccGrammarErrorKind::NoPrecForToken(_)
             | YaccGrammarErrorKind::UnknownEPP(_) => SpansKind::Error,
-            YaccGrammarErrorKind::DuplicateRule
-            | YaccGrammarErrorKind::DuplicatePrecedence
+            YaccGrammarErrorKind::DuplicatePrecedence
             | YaccGrammarErrorKind::DuplicateAvoidInsertDeclaration
             | YaccGrammarErrorKind::DuplicateExpectDeclaration
             | YaccGrammarErrorKind::DuplicateExpectRRDeclaration
@@ -307,16 +304,13 @@ impl YaccParser {
         // this points to the beginning of a UTF-8 character (since multibyte characters exist, not
         // every byte within the string is also a valid character).
         let mut result = self.parse_declarations(0, &mut errs);
-        result = self.parse_rules(
-            match result {
-                Ok(i) => i,
-                Err(e) => {
-                    errs.push(e);
-                    return Err(errs);
-                }
-            },
-            &mut errs,
-        );
+        result = self.parse_rules(match result {
+            Ok(i) => i,
+            Err(e) => {
+                errs.push(e);
+                return Err(errs);
+            }
+        });
         result = self.parse_programs(
             match result {
                 Ok(i) => i,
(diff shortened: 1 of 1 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -28,7 +28,6 @@ pub enum YaccGrammarErrorKind {
     IllegalName,
     IllegalString,
     IncompleteRule,
-    DuplicateRule,
```
ratmice, line 31: After thinking about it I think we should keep this, but only produce it from `add_rule`, which won't be exercised by the parser.  Since add_rule is only called on the first instance of the rule.

ratmice: nevermind, `add_rule` doesn't actually return anything currently (or check for duplicates), so we'd need to make it return `Result<(), YaccGrammarError>` and do double `self.rules.get(rule_name).is_none()` in `add_rule` and outside it to avoid the error.  So it kind of seems like it is fine as-is?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -2425,43 +2414,69 @@ x"
         let e = parse(
             YaccKind::Grmtools,
             "%start A
+                %start A
+                %start A
```
ltratt, line 2418: Are the multiple `%start`s intentional?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -2425,43 +2414,69 @@ x"
...
-    fn test_duplicate_rules_and_missing_arrow() {
-        let src = "
-        %%
-        A -> () : 'a1';
-        A -> () : 'a2';
-        B -> () : 'b1';
-        A -> () : 'a3';
-        B -> () : 'b2';
-        B";
+    fn test_duplicate_rule() -> Result<(), Vec<YaccGrammarError>> {
```
ltratt, line 2434: I didn't know test functions could have a non-void return type! Are there any advantages to doing so here?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -2425,43 +2414,69 @@ x"
...
+        let mut prod_names = HashSet::new();
+        for pidx in &expr_rule.pidxs {
+            for sym in &ast.prods[*pidx].symbols {
+                let name = match sym {
+                    Symbol::Token(name, _) | Symbol::Rule(name, _) => name.clone(),
+                };
+                prod_names.insert(name);
+            }
+        }
+        let mut expected_prod_names = HashSet::new();
```
ltratt, line 2454: Could this be something like `assert_eq!(prod_names, HashSet::from(["A", ..., "D"]));` ? [I don't know if `HashSet` supports such a constructor, but you never know! I suppose `collect` works at a push?]

cfgrammar/src/lib/yacc/parser.rs
ltratt: I tend to agree that it's best not to produce this from `add_rule`.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Primarily that you can use `?` (used on line 2443).

ltratt: Is that clearer than `unwrap`? I wonder if it shows you which `?` went wrong?

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yeah, this is just a general testing for errors that produce multiple spans, and errors with multiple spans and multiple error kinds.
Previously we were using `DuplicateRule` as the 'multiple spans' error, I replaced that with `DuplicateStartRule`.

ltratt: Since we have another test with multiple `%start`s below, maybe it's not worth checking that here?

ratmice: Sure, I could just remove this one, and leave the other in place.

ltratt: Works for me!

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yeah it implements `FromIterator`, so yeah collect should work!

cfgrammar/src/lib/yacc/parser.rs
ratmice: Not sure, I mostly just did it out of habit, i'll have a look at whether there is a difference!

ratmice: It is a little more concise and has less information/no backtrace if I introduce a syntax error.
Result:
```
---- yacc::parser::test::test_duplicate_rule stdout ----
Error: [YaccGrammarError { kind: UnknownDeclaration, spans: [Span { start: 0, end: 0 }] }]
```
unwrap:
```
---- yacc::parser::test::test_duplicate_rule stdout ----
thread 'yacc::parser::test::test_duplicate_rule' panicked at 'called `Result::unwrap()` on an `Err` value: [YaccGrammarError { kind: UnknownDeclaration, spans: [Span { start: 0, end: 0 }] }]', cfgrammar/src/lib/yacc/parser.rs:2423:11
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

ltratt: Interesting! I think `unwrap` "wins" in the sense that if there are two `?`s (not true in this case, I know!), then we can't easily work out which caused the error?

ratmice:
Updated, will need a squash.

ratmice:
Squashed but added some text to the commit message as it was pretty lacking.

ltratt:
Thanks!

bors r+

ratmice:
Forgot to rustfmt, will need a squash again.

ltratt:
Please squash.

ratmice:
Squashed

PR #394, opened 2023-02-10
Fix unwrap in Quickstart chapter

Benjscho:
Fixed unwrap usage in the quickstart doc. Use of `?` for unwrapping was giving an error when used in a function that didn't give have an option return type.


ltratt:
Oops, thanks!

bors r+

ltratt:
Looks like a transitory network error.

bors r+

PR #395, opened 2023-02-19
Enable the unreachable_pub lint in mod.rs.

ratmice:
This is a relatively minor and non-essential thing.  I've noticed on occasion that it can be difficult to know if things are pub but not exported from a module, due to usage within a test module for example or whether they are actually pub and part of the API.  This enables the `unreachable_pub` lint, so you should be able to tell from the visibility modifier whether it is a part of the API.

With this lint enabled items labeled `pub` must be reachable from the top-level module.  Items unreachable from top-level modules should be labeled `pub(crate)`.

### DIFF
--- a/cfgrammar/src/lib/mod.rs
+++ b/cfgrammar/src/lib/mod.rs
@@ -4,6 +4,7 @@
 #![allow(clippy::unnecessary_wraps)]
 #![allow(clippy::upper_case_acronyms)]
 #![forbid(unsafe_code)]
+#![deny(unreachable_pub)]
 
 //! A library for manipulating Context Free Grammars (CFG). It is impractical to fully homogenise
 //! all the types of grammars out there, so the aim is for different grammar types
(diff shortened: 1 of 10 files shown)


cfgrammar/src/lib/mod.rs
```diff
@@ -4,6 +4,7 @@
 #![allow(clippy::unnecessary_wraps)]
 #![allow(clippy::upper_case_acronyms)]
 #![forbid(unsafe_code)]
+#![deny(unreachable_pub)]
```
ratmice, line 7: perhaps these should use `#![deny(rustc::unreachable_pub)]` in accordance with the `clippy::lint_name` usage above?

ltratt: I have no idea! I'm fine either way, so let's do whichever you feel most comfortable with.

ratmice: I had a look through https://doc.rust-lang.org/rustc/lints/index.html and I couldn't find using the `rustc::` prefix actually documented anywhere, so I'm kind of inclined to leave it as is.

ltratt:
Thanks for this -- good find!

bors r+

Issue #399, opened 2023-03-11
error when using fake UMINUS token in %prec directive


yuanweixin:
In yacc, it is possible to declare a fake token with the highest precedence, 
```
/* ... above are other precedence decls with lower prec */
%left UMINUS
```
then in the grammar, use it to give highest precedence to a unary minus expression: 
```
"-" exp %prec UMINUS
```

When I try this with grmtools, it complains that `%prec not followed by token name`. If I try to fix it by introducing the token with `%token UMINUS`, I would get 
```
Error: these tokens are referenced in the grammar but not defined in the lexer:
  UMINUS
```
So it seems the intent here is to force a 1-1 correspondence btw the lexer and parser's space of tokens. But that leaves no leeway for this feature. 

I checked the [test cases](https://github.com/softdevteam/grmtools/blob/3da61f82a56086327b682d921f682e8d796eeeac/cfgrammar/src/lib/yacc/parser.rs#L1725) for the prec overrides, and it seems to not cover this usage. 

Since this is not documented in the yacc-compatibility, I assume this is a bug. And, I'd argue it should be a feature, because without it, I would have to restructure my grammar. The yacc-compatibility is one of the reasons that drew me to use grmtools. 

Complete example: 
.l file
```
%x COMMENT

%%
[\n] ;
[ \t\r\n] ;
import "IMPORT"
primitive "PRIMITIVE"
class "CLASS"
extends "EXTENDS"
method "METHOD"
new "NEW"
type "TYPE"
var "VAR"
function "FUNCTION"
break "BREAK"
of "OF"
end "END"
in "IN"
nil "NIL"
let "LET"
do "DO"
to "TO"
for "FOR"
while "WHILE"
else "ELSE"
then "THEN"
if "IF"
array "ARRAY"
:= "ASSIGN"
\| "OR"
& "AND"
>= "GE"
\<= "LE"
> "GT"
\< "LT"
= "EQ"
\<> "NEQ"
/ "DIVIDE"
\* "TIMES"
- "MINUS"
\+ "PLUS"
\. "DOT"
\} "RBRACE"
\{ "LBRACE"
\] "RBRACK"
\[ "LBRACK"
\) "RPAREN"
\( "LPAREN"
: "COLON"
; "SEMICOLON"
, "COMMA"
[...]

PR #400, opened 2023-03-12
Add `CTBuilder` options for setting behavior of regex.

ratmice:
This is a patch that I discussed in #399
It is still missing some odds and ends, e.g. it only currently works with `CTLexerBuilder` and is missing support within the runtime builders.  But it seemed a worthwhile place to stop and ponder if this is a viable approach.  I tried to do it using a mechanism that allows us to avoid adding a new builder function for each regex setting, It seemed at least worth trying however i'm not sure how I feel about how the actual calls to `RegexBuilder` now look.

This isn't something that I really need, or seems to be blocking anything to my knowledge, so that alone should maybe induce some skepticism about the idea.

- [X] works with CTLexerBuilder
- [ ] runtime builders?
- [ ] Should it use the `quote::ToTokens` trait instead of `Debug`?


ltratt:
Like you, I think, I'm not 100% sure what the right approach here is. One thing that occurs to me looking at the code is that the `enum` variants are one more thing for people to remember, because they have to map those over to the regex crate functions (https://docs.rs/regex/1.7.1/regex/struct.RegexBuilder.html). Which makes me wonder: should we just expose function names that are identical to those in `RegexBuilder` (e.g. `pub fn case_insensitive(&mut self, yes: bool) -> ...`) ? Warning: this might be a terrible idea!

ratmice:
Yeah, I guess the main benefit (which I entirely forgot to implement) of the approach that I took is that we can implement `Serialize`/`Deserialize` for the enum which can be helpful if I would expose these options in nimbleparse_lsp, so I think that is the underlying reason I tried that approach first.  I can have a look at the normal builder approach tomorrow though.

ratmice:
FWIW I put the attempt at using normal builder style here: https://github.com/ratmice/grmtools/tree/regex_options_take_2

ratmice:
Think i'll just close this one for now.

ltratt:
> FWIW I put the attempt at using normal builder style here: https://github.com/ratmice/grmtools/tree/regex_options_take_2

Do you think this route can work? I'm not quite sure if it suffers from the problem you mentioned above or not.

ratmice:
Yeah, it still does suffer from that problem.  The main missing piece of it is that there is no way to call the `LexParser::new_with_regex_options` function I added passing in options from `CTLexerBuilder` with regex_options being routed through `LexerDef::from_str` or `LRNonStreamingLexerDef::from_str`.  So it's mostly just public API decisions.  We can probably keep `from_str` letting it use the default options. But i'm mostly unsure what to do with from_str being a trait function.

ltratt:
Hmm, yes, I don't have an immediately good answer to that :/

ratmice:
Yeah, i'm pretty hesitant to make such changes API without a real motivating use case, so I don't think there is any immediate need. But I do plan on tinkering with a few potential things to see if I can come up with anything that seems worthwhile.

PR #402, opened 2023-03-21
Regex options take 2

ratmice:
I've been a bit slow in getting this finished, but this is basically the finished patch of the second attempt at regex options,
Overall exporting the constructor was simple, I didn't modify the trait `from_str` function at all, just added a new concrete function for  `LRNonStreamingLexerDef`.

I note that `from_str` does not take `self`, so there appear to be no issues involving things like `new_with_options(s,opts).from_str(s)` returning a lexer with default options.

While testing this, I noted some cleanups and bugs in `lrpar/cttests/build.rs`, when I added the lex_flags previously it was flipped,
there was a duplicate call `strip_prefix`, and some if statements could be simplified.  These are within the last patch adding tests.

Edit: By setting the `octal` flag to `'!octal'`, you can induce a regex/build error here.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -24,7 +24,9 @@ use quote::quote;
 use regex::Regex;
 use serde::Serialize;
 
-use crate::{DefaultLexerTypes, LRNonStreamingLexerDef, LexerDef};
+use crate::{
+    DefaultLexerTypes, LRNonStreamingLexerDef, LexerDef, RegexOptions, DEFAULT_REGEX_OPTIONS,
+};
 
 const RUST_FILE_EXT: &str = "rs";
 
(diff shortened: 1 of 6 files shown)


ltratt:
This looks promising! Silly question: I think this is 100% backwards compatible?

ratmice:
The only incompatibility should be the additional parameter to `Rule::new` which is documented as an unstable API which should only be called by the crates builders.

ltratt:
Aha, I see! Yep, I'm fine with adding that argument to the unstable interface. OK to merge?

ratmice:
I should probably squash the first two commits, and rewrite the first commit message to something such as:
"Add CTLexerBuilder options for configuring regex behavior.", but am otherwise happy with it.

ltratt:
Please squash.

ratmice:
Squashed.

PR #406, opened 2023-08-01
Upgrade dependencies

rneswold:
This pull request updates the version of two dependent crates. A project I'm working on now has duplicate crates in it because `lrlex` and `lrpar` are using older instances. It would reduce compile time and executable size if I only had to build one copy of these crates.

I ran `cargo test` and every test passed, so this pull request shouldn't introduce any regressions.

### DIFF
--- a/cfgrammar/Cargo.toml
+++ b/cfgrammar/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/lib/mod.rs"
 
 [dependencies]
 lazy_static = "1.0"
-indexmap = "1"
+indexmap = "2"
 num-traits = "0.2"
 regex = "1"
 serde = { version="1.0", features=["derive"], optional=true }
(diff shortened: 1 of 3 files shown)


rneswold:
The `vergen` dependency is also out of date, but that has some API changes so it'll take some thought to upgrade it.

ratmice:
The `regex-syntax` bump LGTM, the only function lrlex uses is `is_meta_chararacter` which hasn't changed since the previous release. :+1:

ltratt:
Thanks!

bors r+

ltratt:
Thanks!

bors r+

ltratt:
Ah, I think I need to do so boilerplate stuff due to a buildbot change in our infrastructure. I'll do that in another PR then come back to this one. Sorry for the noise!

ltratt:
OK, I've hopefully done the boilerplate. Let's find out!

bors r+

PR #407, opened 2023-08-01
Update to the latest version of `vergen`.

ltratt:


ltratt:
bors try

lrlex/build.rs
```diff
@@ -1,8 +1,5 @@
-use vergen::{vergen, Config};
+use vergen::EmitBuilder;
 
 fn main() {
-    let mut config = Config::default();
-    *config.build_mut().timestamp_mut() = true;
-    *config.build_mut().semver_mut() = false;
```
ratmice, line 6: Never looked into vergen before now (So i'm only vaguely aware of what it is doing).

The only thing I notice here is that we no longer have the `semver_mut()` call.
I had some trouble tracking this down in the vergen repo, it seems like it there used to be an environment variable
`VERGEN_BUILD_SEMVER` that has now been replaced by `CARGO_PKG_VERSION`, however there doesn't seem to be a toggle any longer for it's use.

I think these old removed functions were perhaps macro generated making them difficult to find.
Anyhow not knowing what these did or why they were disabled originally, I'm finding it a bit difficult to review.

Is the place where this could have some effect such as different formatting `/* CACHE INFORMATION` comment in the generated `.y.rs` sources?
has some extra information?

ratmice: Looking at it some more IIUC `semver_mut() = false` it *seems* like it was just avoiding emitting an environment variable `VERGEN_BUILD_SEMVER` which we never actually used.  Perhaps to just to avoid having it showing up in logs or some such. Now that was rendered duplicate information via `CARGO_PKG_VERSION` another variable we don't use, but likely already had via cargo.

So I think I see why dropping this is fine.

ltratt: Honestly, I couldn't work out why `semver_mut()` was there, so I guessed it probably wasn't serving a purpose (and, perhaps, never did). Certainly, things seem to build OK for me. It's not the best explanation, I will admit... but if you're OK with this PR I think I am too!

PR #409, opened 2023-08-13
Various suggestions from Clippy.

ltratt:


lrpar/src/lib/dijkstra.rs
```diff
@@ -66,8 +66,7 @@ where
         .drain(usize::from(c)..usize::from(c) + 1)
         .next()
         .unwrap();
-    while !scs_todo.is_empty() {
-        let n = scs_todo.pop().unwrap().1;
+    while let Some((_, n)) = scs_todo.pop() {
```
ratmice, line 69: Nice, Similar to this I was hoping to post a patch soon which looks at cases where tuple `.0` and `.1` etc are used and see if there are any cases where pattern matching could improve clarity.

ratmice:
Was waiting for bors, but I'm no longer sure if it is perhaps stuck in a queue, or whether it's actually waiting for me to review before it runs. Either way this looked good to me.

bors r+

lrpar/src/lib/dijkstra.rs
ltratt: This is the only one I did by hand (but I wouldn't have thought to look for it without Clippy's other suggestions).

PR #410, opened 2023-08-14
Replace tuple field access with pattern matching

ratmice:
Uncertain how I feel about most of these changes, I am pretty indifferent to the majority of them.

Where I think it improved things is perhaps hoisting the `start_name` variable out of the loop possible because `ast` is a shared reference.  The other one I thought was nice was the `ast::Rule` *until* `cargo fmt` split it up into 4 very short lines but after that happened... :man_shrugging:

Edit: It may in fact be easiest to just note those which you feel like keeping?

### DIFF
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -175,8 +175,15 @@ where
             }
         };
 
-        for (k, rule) in &ast.rules {
-            rule_names.push((k.clone(), rule.name.1));
+        for (
+            k,
+            ast::Rule {
+                name: (_, name_span),
+                ..
+            },
+        ) in &ast.rules
+        {
+            rule_names.push((k.clone(), *name_span));
         }
         let mut rules_prods: Vec<Vec<PIdx<StorageT>>> = Vec::with_capacity(rule_names.len());
         let mut rule_map = HashMap::<String, RIdx<StorageT>>::new();
@@ -214,6 +221,7 @@ where
         let mut prods_rules = vec![None; ast.prods.len()];
         let mut actions = vec![None; ast.prods.len()];
         let mut actiontypes = vec![None; rule_names.len()];
+        let (start_name, _) = ast.start.as_ref().unwrap();
         for (astrulename, _) in &rule_names {
             let ridx = rule_map[astrulename];
             if astrulename == &start_rule {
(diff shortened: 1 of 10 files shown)


ltratt:
Oops, I missed this one (I only check PRs I'm "assigned" to). Sorry!

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -176,7 +176,11 @@ where
         };
 
         for (k, rule) in &ast.rules {
-            rule_names.push((k.clone(), rule.name.1));
+            let ast::Rule {
```
ltratt, line 179: I *think* (but I admit I've never tried!) that this can be collapsed into the `for`? So something like:

```rust
for (k, ast::Rule { name: (_, name_span), .. }) in &ast.rules)
```

?

ltratt:
I'm not sure that every one of these makes things better but, overall, they do make things better, so I'm inclined to accept the lot (that one comment above withstanding).

cfgrammar/src/lib/yacc/grammar.rs
ratmice: Yes, it can!  I had originally had it that way, but it still formats the pattern the same way on a bunch of lines.
That is why I had pulled it out into a let, but happy to pull it up into the for loop.  If after seeing the result we can always reverse course if it's too bad.

ltratt: Good idea to give it a go and see if it's better/worse!

ratmice: I've pushed that, let me know what you think

ltratt: Hmm, very much "in some ways better, in some ways worse"! Let's keep it: we can always change our minds again if we want.

ltratt:
Please squash.

ratmice:
Squashed & rebased onto current master.

ltratt:
Thanks -- and sorry for the delay!

bors r+

ratmice:
I was going to next have a look through uses of `is_some`/`is_none`, not today but sometime soon when I get a chance if that would be welcome.

ltratt:
If it improves the code, I'm all for it! [Please add me with the "assigned" button (rather than "reviewer") and that way I'll notice the PR more reliably.]

ratmice:
> If it improves the code, I'm all for it! [Please add me with the "assigned" button (rather than "reviewer") and that way I'll notice the PR more reliably.]

Will do, (and sorry about that!)

PR #411, opened 2023-08-23
Raise an error if rules start with whitespace.

ratmice:
Here is a patch which turns rules with leading whitespace into errors as discussed.

Posix lex treats these as verbatim copied into the generated code BSD/System V lex copy these into unreachable areas of code. Which are still used as the primary mechanism for adding comments to lex files.

Before this patch we trimmed leading spaces, and parsed them as normal rules. Afterwords we now throw a `VerbatimUnsupported` error.

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -24,6 +24,7 @@ lazy_static! {
         Regex::new(r"^(([xuU][[:xdigit:]])|[[:digit:]]|[afnrtv\\]|[pP]|[dDsSwW]|[AbBz])").unwrap();
     // Vertical line separators.
     static ref RE_LINE_SEP: Regex = Regex::new(r"[\p{Pattern_White_Space}&&[\p{Zl}\p{Zp}\n\r\v]]").unwrap();
+    static ref RE_LEADING_LINE_SEPS: Regex = Regex::new(r"^[\p{Pattern_White_Space}&&[\p{Zl}\p{Zp}\n\r\v]]*").unwrap();
     // Horizontal space separators
     static ref RE_SPACE_SEP: Regex = Regex::new(r#"[\p{Pattern_White_Space}&&[\p{Zs}\t]]"#).unwrap();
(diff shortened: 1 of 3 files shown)


ratmice:
One thing I should mention probably is this does allow entirely empty lines, that is `\n\n\n` is silently skipped, this is covered by the testcase `exceed_tok_id_capacity`, where the `writeln` ends up with double `\n`s at `writeln!(src, "x 'x{}'\n", i).ok();`

ltratt:
This looks mostly good! Possibly silly question: I'm not sure I see a test which checks that `VerbatimNotSupported` is returned?

lrlex/src/lib/parser.rs
```diff
@@ -332,7 +333,39 @@ where
...
+            // > shall be written to lex.yy.c
+            //
+            // > The reason for the undefined condition associated with text beginning with a <blank> or within "%{" and "%}"
+            // > delimiter lines appearing in the Rules section is historical practice.
+            //
+            // > Both the BSD and System V lex copy the indented (or enclosed) input in the Rules section (except at the beginning)
+            // > to unreachable areas of the yylex() function (the code is written directly after a break statement).
+            // > In some cases, the System V lex generates an error message or a syntax error, depending on the form of indented input.
+            //
+            // Previously we allowed these, and trimmed leading spaces, parsing any rules after
```
ltratt, line 352: Can probably merge this line with the next?

ratmice:
> This looks mostly good! Possibly silly question: I'm not sure I see a test which checks that `VerbatimNotSupported` is returned?

I meant to add one, totally forgot it after fixing up the existing tests.  Will have to add it in the morning.

lrlex/src/lib/parser.rs
ratmice: I did this and also evened out the line lengths on one of the paragraphs above it.

ltratt:
Thanks! Please squash.

ratmice:
Squashed.

PR #412, opened 2023-08-24
Document lex differences for verbatim code entries in the rules section

ratmice:
I didn't think about it until the previous commit had landed, but figured it would be good to add a short blurb to the `lex compatibility` section about verbatim code in the rules section.


ltratt:
Good idea!

bors r+

PR #413, opened 2023-08-28
Cleanup some use sites of option types.

ratmice:
No behavioral changes are expected or intended.

* Use if let Some(x) = x { } instead of if x.is_some() { ... x.unwrap(); }
* Use the Option PartialOrd implementation in a few places to avoid some unwraps.
* Hoist loops beginning with if _.is_some() { break; } into the while condition.


ratmice:
I understand changes of this nature may be a stylistic matter of opinion, so if there is anything here for which the original is
preferred I would happily drop any parts of the patch.  I felt these were largely pretty nice though.

ltratt:
These are all good -- thanks!

bors r+

PR #415, opened 2023-08-30
report both terms & tokens missing from parser/lexer from CTLexerBuilder

ratmice:
The first patch coalesces the missing token/term for lexer/parser errors, so they will both be emitted.
The second patch, just moves some code in cttests/build.rs into it's own function in prep for adding cttests which result in codegen failure (the third patch).

The third patch may well require some work, it would probably ideally be within a `#[should_panic]`, `#[test]` but I wanted to reuse the ctbuilder helpers within cttests/build.rs.  Perhaps this could be put into a module to be shared between lib.rs and build.rs.

I'm not sure (but I assume) that e.g. `OUT_DIR` env var is available within `#[test]`.
Anyhow it currently isn't ideal, i.e no output on success see `cargo -vvv` for output, and doesn't re-run when cargo test is run without build.rs or .test file changes.  I haven't looked into what the situation is with the stablization of `#![test_runner]` but that would probably be ideal in allowing one to treat each glob entry as a test.

Edit: One thing this doesn't capture yet is errors that don't panic but return errors from `build()`. Figured these could be done as we add a test that returns an error.

### DIFF
--- a/lrpar/cttests/build.rs
+++ b/lrpar/cttests/build.rs
@@ -2,11 +2,15 @@ use cfgrammar::yacc::{YaccKind, YaccOriginalActionKind};
 use glob::glob;
 use lrlex::{CTLexerBuilder, DefaultLexerTypes};
 use lrpar::CTParserBuilder;
-use std::{env, fs, path::PathBuf};
+use std::{
+    env, fs,
+    path::{Path, PathBuf},
+};
 use yaml_rust::YamlLoader;
 
-fn run_test_path(path: PathBuf) -> Result<(), Box<dyn std::error::Error>> {
+fn run_test_path<P: AsRef<Path>>(path: P) -> Result<(), Box<dyn std::error::Error>> {
     let out_dir = env::var("OUT_DIR").unwrap();
+    let path = path.as_ref();
     if path.is_file() {
         println!("cargo:rerun-if-changed={}", path.display());
         // Parse test file
@@ -146,6 +150,18 @@ fn run_test_path(path: PathBuf) -> Result<(), Box<dyn std::error::Error>> {
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     for entry in glob("src/*.test")? {
         run_test_path(entry.unwrap())?;
+    }
+    for entry in glob("src/ctfails/*.test")? {
+        let path = entry.unwrap();
+        let result =
+            std::panic::catch_unwind(|| std::panic::AssertUnwindSafe(run_test_path(&path).is_ok()));
+        if !result.is_err() {
+            panic!(
+                "ctfails/{}: succeded unexpectedly with result {:?}",
+                path.display(),
+                result
+            );
+        }
     }
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


ratmice:
So, AFAICT `test_runner` isn't stable and still requires the `custom_test_framework` feature from nightly.
I *think* I could probably hack something together though by running a proc-macro which does the `glob`, and generates a `#[test]` function for each entry of the glob that wraps a `run_test_path(...)`.

Does that sound worth the effort/complexity?

lrpar/cttests/build.rs
```diff
@@ -5,143 +5,147 @@ use lrpar::CTParserBuilder;
 use std::{env, fs, path::PathBuf};
 use yaml_rust::YamlLoader;
 
+fn run_test_path(path: PathBuf) -> Result<(), Box<dyn std::error::Error>> {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    if path.is_file() {
```
ltratt, line 14: Are there situations where the path wouldn't be a file?

ltratt:
> I think I could probably hack something together though by running a proc-macro which does the glob, and generates a #[test] function for each entry of the glob that wraps a run_test_path(...).
>
> Does that sound worth the effort/complexity?

Honestly, it doesn't sound worth the effort to me. Hopefully one day this feature will finally stabilise!

FWIW, I think we can merge this PR unless you want to do more to it?

ltratt:
> I would guess they don't affect bors which I would imagine does a clean build?

It does.

That said:

> Anyhow it currently isn't ideal, i.e no output on success see cargo -vvv for output, and doesn't re-run when cargo test is run without build.rs or .test file changes.

Ah, that's because we've put "do the test" code in `build.rs` I guess? This PR doesn't make that original sin any worse but, I agree, it would be nice to fix that.

One possibility (for a future PR) might be to move these parts of grmtools' testing to https://crates.io/crates/lang_tester?

ratmice:
> > I would guess they don't affect bors which I would imagine does a clean build?
> 
> It does.
> 
> That said:
> 
> > Anyhow it currently isn't ideal, i.e no output on success see cargo -vvv for output, and doesn't re-run when cargo test is run without build.rs or .test file changes.
> 
> Ah, that's because we've put "do the test" code in `build.rs` I guess? This PR doesn't make that original sin any worse but, I agree, it would be nice to fix that.

Yeah, the current tests still get some output though because they all have additional `#[test]` entries in `cttests/src/lib.rs`.
I.e. they are expecting both codegen to succeed, and also subsequently that the generated code compiles/modules load.

So those others at least get some output. But these ones where we are expecting codegen to fail don't really follow the current pattern.

> 
> One possibility (for a future PR) might be to move these parts of grmtools' testing to https://crates.io/crates/lang_tester?

I haven't heard of that one, will have a look

ltratt:
OK, so I suggest we merge this PR and then in another PR think about improving the whole testing framework. Sound like a plan?

ratmice:
Sounds good.

Issue #425, opened 2023-12-10
Detailed debug info for reduce/reduce shift/reduce errors? 


DavidHVernon:
I was wondering if there is a way to get more detailed info on CTConflictsError errors (reduce/reduce and shift/reduce). In particular it would be handy to know which production rules are clashing.

ratmice:
have a look at the [`conflicts`](https://docs.rs/lrpar/latest/lrpar/struct.CTParser.html#method.conflicts) function of lrpar,
You may need to `error_on_conflicts(false)`, and you should be able to get more detailed info from there.

From my lsp here it gets the prod idx then converts that to spans & lsp errors.  It is a bit messy though.
https://github.com/ratmice/nimbleparse_lsp/blob/main/server/src/parse_thread.rs#L505-L571

DavidHVernon:
Thank you very much.

ratmice:
It seems worth noting that code uses the `RTParserBuilder`, but I think `CTParserBuilder` should behave basically the same in this regard.  Also worth mentioning I don't think that code has made it to a published version of the lsp yet, but should basically work.

DavidHVernon:
I'm having trouble seeing how to get access to an lrpar, and where to use conflicts. I don't know much about these libraries (because they work well enough that I haven't needed to), but I get that this needs to be in the build phase. 

I have copied the boilerplate build.rs code. It looks like the parser is being built by the lexer builder via lrpar_config(). That is where I get lost.

ratmice:
I see what you mean, I *think* there are 2 ways you can go about it.
In the old style, you build the `CTLexerBuilder`, `CTParserBuilder` separately, and combine them with a
token map, much like is described in https://softdevteam.github.io/grmtools/master/book/manuallexer.html
So you can build the `CTParser` yourself, outside of `CTLexerBuilder`, the example I can find from the repository that behaves this way is

https://github.com/softdevteam/grmtools/blob/4ed68d6feeff01738c073eced73a3192daad51c7/lrpar/cttests/src/cgen_helper.rs#L109-L112
https://github.com/softdevteam/grmtools/blob/4ed68d6feeff01738c073eced73a3192daad51c7/lrpar/cttests/src/cgen_helper.rs#L91-L94

There is some conflict testing code here also (build calls on lines 104-140 respectively)
[...]

PR #426, opened 2023-12-10
Add error callbacks to CTBuilders.

ratmice:
Here is an initial draft of the patch I was thinking in #425 
I still need to add tests, and an example with pretty errors.

But in case @DavidHVernon wanted to test it out, and see if it helps solve his problem in a cleaner fashion, I wanted to get it out as a draft.

Edit: This is still not catch a few things, 

"Missing from lexer", "Missing from parser", aren't grammar errors, or lexer errors, but some kind of error relation between the two.
It also doesn't get called for warnings.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -25,7 +25,8 @@ use regex::Regex;
 use serde::Serialize;
 
 use crate::{
-    DefaultLexerTypes, LRNonStreamingLexerDef, LexerDef, RegexOptions, DEFAULT_REGEX_OPTIONS,
+    DefaultLexerTypes, LRNonStreamingLexerDef, LexBuildError, LexerDef, RegexOptions,
+    DEFAULT_REGEX_OPTIONS,
 };
 
 const RUST_FILE_EXT: &str = "rs";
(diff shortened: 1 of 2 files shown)


ratmice:
So, looking forward to other things I am inclined to try implementing 2 changes to this basic API, despite the fact that I like how simple it is, these changes described are in-order (so building upon one another).

1. If we pass a trait to the builder, and the trait implements all the functions
2. adding an arbitrary user-supplied argument to it kind of the equivalent of `%parse_param`.
3. add an extra call `to_error(T) -> Box<dyn Error>`
4. `CTParserBuilder` then gets a `fn error_handler(GrammarErrorHandler)` instead of separate `on_*` functions.

So it would look something to the effect of the following (which probably doesn't compile):

```
trait<T> GrammarErrorHandler {
   fn on_grammar_error() -> Option<for<'a> Fn (&mut T, YaccGrammarError)>
   fn on_unexpected_conflicts() -> Option<for<'a> Fn(&mut T, ...)>
   fn to_error(T) -> Box<dyn Error>
}
```

The purpose of this is to open the door for a separate call from `build`, which i'll call `check` for now.
This would basically do what `build` does, except it wouldn't write the generated sources to disk, and instead of returning an error as a result, it could return `T`.

The motivation for this extra parameter and new `check` phase, is that `T` could be something which generates reports such as
those in the `SARIF` static analysis results interchange format. In addition to the use case here for customizing error output.

I think from a `build.rs` perspective at least, settings a single field on the builder, rather than one for each callback may be an improvement, in that it is much easier to provide a reusable libraries which take care of these things. I think having to supply a `T` is probably negative, my hope is we can default to `T = ()`.

ratmice:
Reminder to self, I probably should consider passing in the lex/yacc source string to their various callbacks.

ratmice:
I put up an alternative trait-ized implementation in #428, which is perhaps worth looking at instead?

ratmice:
It seems to me like this one can be closed for now, that we're at least going to try and go for the trait based impl instead.

PR #427, opened 2023-12-12
Clippy fallout & overzealous #[deny(unsafe_code)]

ratmice:
The first patch is just fixing normal clippy stuff.

In the second patch clippy was suggesting that we remove the unsafe code because it (being just a test) doesn't actually do anything. Rather than `#[allow(...)]` I tried defining an unsafe function in the program section and call that.

Turns out `#[deny(unsafe_code)]` is still enabled in the program section, since my patch at https://github.com/softdevteam/grmtools/pull/370

Subsequently the second patch writes the combined user actions + program section into a single module, and `#![allow(unsafe_code)]` within that module.  Previously the `program()` section could not specify inner attributes since those must be specified at the top of a module, so they couldn't easily opt in using `#![allow(unsafe_code)]`. With this reorganization it is now possible that the program section is at the top of the module, and can specify an inner attribute.

user actions + program code, need to be in the same module, because we currently don't require `pub` visibility modifiers between them.

Edit: The other option being we could just revert the `#[deny(unsafe_code)]` part from the generated sources, WDYT?

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -676,12 +676,19 @@ where
     ) -> Result<(), Box<dyn Error>> {
         let mut outs = String::new();
         writeln!(outs, "{} mod {} {{", self.visibility.cow_str(), mod_name).ok();
+        // Emit user program section, and actions at the top so they may specify inner attributes.
+        if let Some(YaccKind::Original(YaccOriginalActionKind::UserAction) | YaccKind::Grmtools) =
+            self.yacckind
+        {
+            outs.push_str(&self.gen_user_actions(grm));
+        }
+        outs.push_str("    mod _parser_ {\n");
         outs.push_str(
-            "    #![allow(clippy::type_complexity)]
-    #![allow(clippy::unnecessary_wraps)]
-    #![deny(unsafe_code)]
-    #[allow(unused_imports)]
-    use ::lrpar::Lexeme;
+            "        #![allow(clippy::type_complexity)]
+        #![allow(clippy::unnecessary_wraps)]
+        #![deny(unsafe_code)]
+        #[allow(unused_imports)]
+        use super::*;
 ",
         );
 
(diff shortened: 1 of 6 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -1145,7 +1151,7 @@ where
                 "    // {rulename}
     #[allow(clippy::too_many_arguments)]
     #[allow(unsafe_code)] // Allow an action to embed unsafe blocks within it.
-    fn {prefix}action_{}<'lexer, 'input: 'lexer>({prefix}ridx: ::cfgrammar::RIdx<{storaget}>,
+    pub fn {prefix}action_{}<'lexer, 'input: 'lexer>({prefix}ridx: ::cfgrammar::RIdx<{storaget}>,
```
ltratt, line 1154: Does this need to be `pub`? I'd ideally prefer not to expose it to the user (but if we have to, we have to).

ratmice: I'm pretty sure it needs some visibility, you're right though `pub` is probably too much I think it can just be `pub(super)`.
In reality though even though it is `pub`, because the `mod` isn't pub, and we aren't `pub use` it should only be visible to `super`.

But it's better to spell these things out with `pub(super)` I think.

ltratt:
Please squash.

ltratt:
> Did we lose the rustfmt check?

It's still part of `.buildbot.sh` but we now use github's merge queues to kick buildbot off, rather than bors.

ratmice:
Squashed.

ltratt:
@ratmice If you click on "details" then "buildbot/buildbot-script" you can see the CI log and what went wrong. Hopefully it's not too annoying to fix!

ratmice:
I *think* I just need to readd the `pub` to `pub use _user_section_::*`, however I'm very curious why I don't seem to be able to reproduce locally.

ratmice:
So that fixed one of the issues, but the other might be insurmountable.  `calc_actions/src/calc.y` doesn't re-export it's `use std::error::Error` without changing it to `pub use std::error::Error`.

I'll have to have a closer look, but it definitely might sink this approach.

ratmice:
So, I basically had to flip it inside out, moving the stuff that had `#![deny(unsafe_code)]` to it's own module instead.

ltratt:
Excellent! Please squash.

ltratt:
Thanks!

PR #428, opened 2023-12-15
Error handler trait

ratmice:
Here is the best attempt I think I can muster for implementing the patch in #426 as a trait instead of plain callback functions,
I figured it might be better to push it to a separate branch since @DavidHVernon has been doing some test on that, and this switches the API quite a bit.

This allows for the `ErrorHandler` to be owned by the outer thing, so values in it's `Self` type can escape. So that is the benefit over the plain callbacks approach which can only return via the `Box<dyn Error>`.

I've pushed a branch of `error_callbacks_test` repo which implement this trait
First a very quick/dirty implementation using `String` as a backing store.
https://github.com/ratmice/error_callbacks_test/blob/implement_as_trait/build.rs#L273-L287
The second one is much nicer using the `ariadne` crate to produce nice messages.
https://github.com/ratmice/error_callbacks_test/tree/ariadne_trait_impl

In the second (the work in progress ariadne errors) these errors look something like:
```
     ╭─[src/erroneous.y:4:1]
     │
   1 │ %token a b
     │        ┬
     │        ╰── Shifted
     │
   4 │ B -> (): a | a a {};
     │ ┬        ┬
     │ ╰─────────── Reduced rule
     │          │
     │          ╰── Reduced production
  ───╯
```

It is outside the repo because it's a failing build.rs, and ctfails tests can't quite handle it.
I was hoping that this would make calling it easier than having a bunch of individual functions for each callback.
It sort of succeeds from the `CT*Builder` perspective, but the lifetimes semantics require a few differences from the normal build.rs e.g. a `move |...|` closure to `lrpar_config`. This is why `LexErrorHandler` can be a `&mut`, while `GrammarErrorHandler` has to be the `Rc<RefCell<_>>`.

I believe that I am happy with this approach in general, I think it can both be used by tools like `nimbleparse` to cut down some code duplication internally (But I don't try to do that here), and useful for outside projects (Like custom error printing, and my future project of producing SARIF output).  As a trait these things should be easier to put into a crates and used.

With the holidays coming up, I'm not certain how quickly I will be able to respond to any requested changes.
If you don't happen to get to it before then either, I wish you happy holidays.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -25,7 +25,8 @@ use regex::Regex;
 use serde::Serialize;
 
 use crate::{
-    DefaultLexerTypes, LRNonStreamingLexerDef, LexerDef, RegexOptions, DEFAULT_REGEX_OPTIONS,
+    DefaultLexerTypes, LRNonStreamingLexerDef, LexBuildError, LexerDef, RegexOptions,
+    DEFAULT_REGEX_OPTIONS,
 };
 
 const RUST_FILE_EXT: &str = "rs";
@@ -79,6 +80,26 @@ pub enum RustEdition {
     Rust2021,
 }
 
+pub trait LexErrorHandler<LexerTypesT>
+where
+    LexerTypesT: LexerTypes,
+    usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+{
+    /// Called with the lexers filename
+    fn lexer_path(&mut self, filename: &Path);
+    /// Called with the lexers source contents.
+    fn lexer_src(&mut self, src: &str);
+    fn on_lex_build_error(&mut self, errors: Box<[LexBuildError]>);
+    fn missing_in_lexer(&mut self, missing: &HashSet<String>);
+    fn missing_in_parser(&mut self, missing: &HashSet<String>);
+    /// This function must return an `Err` variant if any of the following are true:
+    ///
+    /// - missing_in_lexer is called.
+    /// - missing_in_parser is called.
+    /// - on_lex_build_error is called.
+    fn results(&self) -> Result<(), Box<dyn Error>>;
+}
+
 /// A `CTLexerBuilder` allows one to specify the criteria for building a statically generated
 /// lexer.
 pub struct CTLexerBuilder<'a, LexerTypesT: LexerTypes = DefaultLexerTypes<u32>>
@@ -97,6 +118,7 @@ where
     allow_missing_terms_in_lexer: bool,
     allow_missing_tokens_in_parser: bool,
     regex_options: RegexOptions,
+    error_handler: Option<&'a mut dyn LexErrorHandler<LexerTypesT>>,
 }
 
 impl<'a> CTLexerBuilder<'a, DefaultLexerTypes<u32>> {
@@ -141,7 +163,16 @@ where
             allow_missing_terms_in_lexer: false,
             allow_missing_tokens_in_parser: true,
             regex_options: DEFAULT_REGEX_OPTIONS,
+            error_handler: None,
         }
+    }
+
+    pub fn error_handler(
+        mut self,
+        error_handler: &'a mut dyn LexErrorHandler<LexerTypesT>,
+    ) -> Self {
+        self.error_handler = Some(error_handler);
+        self
     }
 
     /// An optional convenience function to make it easier to create an (lrlex) lexer and (lrpar)
@@ -327,7 +358,10 @@ where
         }
 
         let lex_src = read_to_string(lexerp)?;
-        let line_cache = NewlineCache::from_str(&lex_src).unwrap();
+        if let Some(error_handler) = self.error_handler.as_mut() {
+            error_handler.lexer_path(lexerp.as_path());
+            error_handler.lexer_src(lex_src.as_str());
+        }
         let mut lexerdef: Box<dyn LexerDef<LexerTypesT>> = match self.lexerkind {
             LexerKind::LRNonStreamingLexer => Box::new(
                 LRNonStreamingLexerDef::<LexerTypesT>::new_with_options(
@@ -335,6 +369,7 @@ where
                     self.regex_options.clone(),
                 )
                 .map_err(|errs| {
+                    let line_cache = NewlineCache::from_str(&lex_src).unwrap();
                     errs.iter()
                         .map(|e| {
                             if let Some((line, column)) = line_cache.byte_to_line_num_and_col_num(
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -89,7 +89,7 @@ where
     fn lexer_path(&mut self, filename: &Path);
     /// Called with the lexers source contents.
     fn lexer_src(&mut self, src: &str);
-    fn on_lex_build_error(&mut self, errors: Box<[LexBuildError]>);
+    fn on_lex_build_error(&mut self, errors: &[LexBuildError]);
     fn missing_in_lexer(&mut self, missing: &HashSet<String>);
     fn missing_in_parser(&mut self, missing: &HashSet<String>);
     /// This function must return an `Err` variant if any of the following are true:
@@ -371,7 +371,7 @@ where
 
                 Box::new(if let Some(error_handler) = self.error_handler.as_mut() {
                     lexerdef.map_err(|errs| {
-                        error_handler.on_lex_build_error(errs.into_boxed_slice());
+                        error_handler.on_lex_build_error(errs.as_slice());
                         error_handler
                             .results()
                             .expect_err("Expected an error from error_handler.")
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -417,25 +417,39 @@ where
         let mut has_unallowed_missing = false;
         if !self.allow_missing_terms_in_lexer {
             if let Some(ref mfl) = missing_from_lexer {
-                eprintln!("Error: the following tokens are used in the grammar but are not defined in the lexer:");
-                for n in mfl {
-                    eprintln!("    {}", n);
+                if let Some(error_handler) = self.error_handler.as_deref_mut() {
+                    error_handler.missing_in_lexer(mfl);
+                } else {
+                    eprintln!("Error: the following tokens are used in the grammar but are not defined in the lexer:");
+                    for n in mfl {
+                        eprintln!("    {}", n);
+                    }
                 }
                 has_unallowed_missing = true;
             }
         }
         if !self.allow_missing_tokens_in_parser {
             if let Some(ref mfp) = missing_from_parser {
-                eprintln!("Error: the following tokens are defined in the lexer but not used in the grammar:");
-                for n in mfp {
-                    eprintln!("    {}", n);
+                if let Some(error_handler) = self.error_handler.as_deref_mut() {
+                    error_handler.missing_in_parser(mfp);
+                } else {
+                    eprintln!("Error: the following tokens are defined in the lexer but not used in the grammar:");
+                    for n in mfp {
+                        eprintln!("    {}", n);
+                    }
                 }
                 has_unallowed_missing = true;
             }
         }
         if has_unallowed_missing {
             fs::remove_file(outp).ok();
-            panic!();
+            if let Some(error_handler) = self.error_handler.as_deref_mut() {
+                return Err(error_handler
+                    .results()
+                    .expect_err("Expected missing tokens in lexer or parser"));
+            } else {
+                panic!();
+            }
         }
 
         let mod_name = match self.mod_name {
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -85,12 +85,19 @@ where
     LexerTypesT: LexerTypes,
     usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
 {
-    /// Called with the lexers filename
+    /// Will be called with the path to the `.l` file
+    /// before `fn on_*` or `fn missing_*`.
     fn lexer_path(&mut self, filename: &Path);
-    /// Called with the lexers source contents.
+    /// Will be called with the `.y` file sources as `src`
+    /// before any call to `fn on_*`.
     fn lexer_src(&mut self, src: &str);
+    /// Lends `self` a slice containing `LexBuildError`s
     fn on_lex_build_error(&mut self, errors: &[LexBuildError]);
+    /// Lends `self` a set of `String`s denoting tokens
+    /// present in the parser, but missing from the lexer.
     fn missing_in_lexer(&mut self, missing: &HashSet<String>);
+    /// Lends `self` a set of `String`s denoting tokens
+    /// present in the lexer, but missing from the parser.
     fn missing_in_parser(&mut self, missing: &HashSet<String>);
     /// This function must return an `Err` variant if any of the following are true:
     ///
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -363,28 +363,40 @@ where
             error_handler.lexer_src(lex_src.as_str());
         }
         let mut lexerdef: Box<dyn LexerDef<LexerTypesT>> = match self.lexerkind {
-            LexerKind::LRNonStreamingLexer => Box::new(
-                LRNonStreamingLexerDef::<LexerTypesT>::new_with_options(
+            LexerKind::LRNonStreamingLexer => {
+                let lexerdef = LRNonStreamingLexerDef::<LexerTypesT>::new_with_options(
                     &lex_src,
                     self.regex_options.clone(),
-                )
-                .map_err(|errs| {
+                );
+
+                Box::new(if let Some(error_handler) = self.error_handler.as_mut() {
+                    lexerdef.map_err(|errs| {
+                        error_handler.on_lex_build_error(errs.into_boxed_slice());
+                        error_handler
+                            .results()
+                            .expect_err("Expected an error from error_handler.")
+                    })?
+                } else {
                     let line_cache = NewlineCache::from_str(&lex_src).unwrap();
-                    errs.iter()
-                        .map(|e| {
-                            if let Some((line, column)) = line_cache.byte_to_line_num_and_col_num(
-                                &lex_src,
-                                e.spans().first().unwrap().start(),
-                            ) {
-                                format!("{} at line {line} column {column}", e)
-                            } else {
-                                format!("{}", e)
-                            }
-                        })
-                        .collect::<Vec<_>>()
-                        .join("\n")
-                })?,
-            ),
+                    lexerdef.map_err(|errs| {
+                        errs.iter()
+                            .map(|e| {
+                                if let Some((line, column)) = line_cache
+                                    .byte_to_line_num_and_col_num(
+                                        &lex_src,
+                                        e.spans().first().unwrap().start(),
+                                    )
+                                {
+                                    format!("{} at line {line} column {column}", e)
+                                } else {
+                                    format!("{}", e)
+                                }
+                            })
+                            .collect::<Vec<_>>()
+                            .join("\n")
+                    })?
+                })
+            }
         };
         let (missing_from_lexer, missing_from_parser) = match self.rule_ids_map {
             Some(ref rim) => {
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -3,6 +3,7 @@
 use std::{
     any::type_name,
     borrow::Cow,
+    cell::RefCell,
     collections::{HashMap, HashSet},
     convert::AsRef,
     env::{current_dir, var},
@@ -13,13 +14,17 @@ use std::{
     io::{self, Write},
     marker::PhantomData,
     path::{Path, PathBuf},
+    rc::Rc,
     sync::Mutex,
 };
 
 use bincode::{deserialize, serialize_into};
 use cfgrammar::{
     newlinecache::NewlineCache,
-    yacc::{ast::ASTWithValidityInfo, YaccGrammar, YaccKind, YaccOriginalActionKind},
+    yacc::{
+        ast::{ASTWithValidityInfo, GrammarAST},
+        YaccGrammar, YaccGrammarError, YaccGrammarWarning, YaccKind, YaccOriginalActionKind,
+    },
     RIdx, Spanned, Symbol,
 };
 use filetime::FileTime;
@@ -146,6 +151,32 @@ impl Visibility {
     }
 }
 
+pub trait GrammarErrorHandler<LexerTypesT>
+where
+    LexerTypesT: LexerTypes,
+    usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+{
+    fn grammar_src(&mut self, src: &str);
+    fn grammar_path(&mut self, path: &Path);
+    fn warnings_are_errors(&mut self, flag: bool);
+    fn on_grammar_warning(&mut self, ws: Box<[YaccGrammarWarning]>);
+    fn on_grammar_error(&mut self, errs: Box<[YaccGrammarError]>);
+    fn on_unexpected_conflicts(
+        &mut self,
+        ast: &GrammarAST,
+        grm: &YaccGrammar<LexerTypesT::StorageT>,
+        sgraph: &StateGraph<LexerTypesT::StorageT>,
+        stable: &StateTable<LexerTypesT::StorageT>,
+        conflicts: &Conflicts<LexerTypesT::StorageT>,
+    );
+    /// This function must return an `Err` if any of the following are true:
+    ///
+    /// - `on_grammar_error` has been called.
+    /// - `on_unexpected_conflicts` has been called,
+    /// - `on_grammar_warning` was called and `warnings_are_errors` is true.
+    fn results(&self) -> Result<(), Box<dyn Error>>;
+}
+
 /// A `CTParserBuilder` allows one to specify the criteria for building a statically generated
 /// parser.
 pub struct CTParserBuilder<'a, LexerTypesT: LexerTypes>
@@ -166,6 +197,7 @@ where
     show_warnings: bool,
     visibility: Visibility,
     rust_edition: RustEdition,
+    error_handler: Option<Rc<RefCell<dyn GrammarErrorHandler<LexerTypesT>>>>,
     phantom: PhantomData<LexerTypesT>,
 }
 
@@ -209,6 +241,7 @@ where
             show_warnings: true,
             visibility: Visibility::Private,
             rust_edition: RustEdition::Rust2021,
+            error_handler: None,
             phantom: PhantomData,
         }
     }
@@ -260,6 +293,14 @@ where
         write!(leaf, ".{}", RUST_FILE_EXT).ok();
         outp.push(leaf);
         Ok(self.output_path(outp))
+    }
+
+    pub fn error_handler(
+        mut self,
+        error_handler: Rc<RefCell<dyn GrammarErrorHandler<LexerTypesT>>>,
+    ) -> Self {
+        self.error_handler = Some(error_handler);
+        self
     }
 
     /// Set the input grammar path to `inp`. If specified, you must also call
@@ -429,6 +470,12 @@ where
         };
 
         let res = YaccGrammar::<StorageT>::new_from_ast_with_validity_info(yk, &ast_validation);
+        if let Some(error_handler) = self.error_handler.as_ref() {
+            let mut error_handler = error_handler.borrow_mut();
+            error_handler.grammar_path(grmp);
+            error_handler.grammar_src(inc.as_str());
+            error_handler.warnings_are_errors(self.warnings_are_errors);
+        }
         let grm = match res {
             Ok(_) if self.warnings_are_errors && !warnings.is_empty() => {
                 let mut line_cache = NewlineCache::new();
@@ -661,6 +708,7 @@ where
             show_warnings: self.show_warnings,
             visibility: self.visibility.clone(),
             rust_edition: self.rust_edition,
+            error_handler: None,
             phantom: PhantomData,
         };
         Ok(cl.build()?.rule_ids)
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -156,11 +156,20 @@ where
     LexerTypesT: LexerTypes,
     usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
 {
+    /// Will be called with the `.y` file sources
+    /// as `src`, before any any call to `fn on_*`.
     fn grammar_src(&mut self, src: &str);
+    /// Will be called with the path to the `.y`
+    /// file, before any any call to `fn on_*`.
     fn grammar_path(&mut self, path: &Path);
+    /// Will be called with a flag denoting whether warnings
+    /// should be treated as errors before any call to `fn on_*`.
     fn warnings_are_errors(&mut self, flag: bool);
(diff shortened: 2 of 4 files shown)


lrlex/src/lib/ctbuilder.rs
```diff
@@ -79,6 +80,26 @@ pub enum RustEdition {
...
 
+pub trait LexErrorHandler<LexerTypesT>
+where
+    LexerTypesT: LexerTypes,
+    usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+{
+    /// Called with the lexers filename
+    fn lexer_path(&mut self, filename: &Path);
+    /// Called with the lexers source contents.
+    fn lexer_src(&mut self, src: &str);
```
ratmice, line 91: one thing that just came to mind is we could consider adding here is `fn generated_src(&mut self, src:&str)`?

lrlex/src/lib/ctbuilder.rs
```diff
@@ -79,6 +80,26 @@ pub enum RustEdition {
...
+    fn on_lex_build_error(&mut self, errors: Box<[LexBuildError]>);
```
ratmice, line 92: These should probably just be `&[LexBuildError]` slices I think.

ratmice: Done in 122dba3

lrlex/src/lib/ctbuilder.rs
```diff
@@ -79,6 +80,26 @@ pub enum RustEdition {
...
+    fn missing_in_lexer(&mut self, missing: &HashSet<String>);
+    fn missing_in_parser(&mut self, missing: &HashSet<String>);
```
ratmice, line 94: It is probably worth thinking about whether we can do better than just `HashSet<String>` for `missing_in_*`.

I.e. for `missing_in_lexer`, we can perhaps give a `PIdx` to the parsers production referencing the token.
It would probably need to be given a `YaccGrammar`, and a `YaccGrammarAST` to go from there to spans.
I *think* this is okay to `lrlex`, because it depends on `lrpar`, and overall is where this check occurs currently.

And for `missing_in_parser`, we have `LRNonStreamingLexerDef::get_rule_by_name`, but we aren't passing in a `LRNonStreamingLexerDef` here.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -79,6 +80,26 @@ pub enum RustEdition {
...
+    /// This function must return an `Err` variant if any of the following are true:
+    ///
+    /// - missing_in_lexer is called.
+    /// - missing_in_parser is called.
+    /// - on_lex_build_error is called.
+    fn results(&self) -> Result<(), Box<dyn Error>>;
```
ratmice, line 107: The `Result<(), Box<dyn Error>>` return value isn't currently very useful,
It's only called in cases it *must* return an `Err`, and since the `on_*` is never called when it is allowed to return `Ok`
It never has anything to do further analysis on, so it can't currently introduce it's own errors above and beyond the current error checking.  Perhaps we could consider adding a `fn analyse(...)`, It could be a method provided by the trait with an empty implementation even, which trait implementations can override.

Finally a call to `results()` at the end even if nothing `lrpar` considers an error has occurred.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -146,6 +151,32 @@ impl Visibility {
...
 }
 
+pub trait GrammarErrorHandler<LexerTypesT>
+where
+    LexerTypesT: LexerTypes,
+    usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+{
+    fn grammar_src(&mut self, src: &str);
+    fn grammar_path(&mut self, path: &Path);
+    fn warnings_are_errors(&mut self, flag: bool);
```
ratmice, line 167: One thing missing here is that warnings are no longer printed, but there is no mechanism to print them from the trait.
there is also the `show_warnings` flag of `CTParserBuilder` we need to add if traits should print warnings.

One thing to consider is that we could change the `Ok(())` value of `results()` to return formatted warnings,
`CTParserBuilder` then just prints them out to `stderr`, or perhaps some `emit_warnings()` call?

lrlex/src/lib/ctbuilder.rs
ltratt: > It is probably worth thinking about whether we can do better than just HashSet<String> for missing_in_*.

Yes I tend to agree.

> Not sure the above is currently actionable with the current signature of set_rule_ids though.

I also think I agree with that. I don't immediately have a good idea as to whether we can come up with a better API for `set_rule_ids` or not, but I don't object to changing it, because it's not a function that we expect many people to use directly.

lrlex/src/lib/ctbuilder.rs
ltratt: I'm not certain that I understand what this function does and why it does something distinct in the API. Is it because the other methods are expected to do some sort of mutation?

ltratt:
I am broadly in favour of this line of attack: thanks! This part of grmtools is currently undercooked, and I must admit that I had never really worked out in my mind what a better way of doing things might be. I think the `trait` approach pioneered here is broadly the right one. Of course, there are many minor details about the API that we can think about, but at the very least this feels like we're making a substantial move in the right direction.

Here are some half-formed general questions I have:

1. Is/should the error API intended to be incremental? i.e. it calls the functions as soon as problems are detected? Or are they bundled up at the end (and/or in well-defined stages, in the sense of rustc, where type checking happens first and only when fully complete is the borrow checker run)?
2. Should warnings be differentiated from errors? If so, should they be part of one trait or ... ?
3. Should "analysis" be a separate thing than errors/warnings?

My immediate reaction is that I *think* I like the sound of: incremental; warnings are separate but can be part of one trait; and "analysis" doesn't need separating from errors/warnings. But my immediate reactions rarely stand up to long-term scrutiny, so don't read too much into that :)

lrlex/src/lib/ctbuilder.rs
ratmice: Yes exactly, `self` is expected to basically collect the previous calls into some type `T` fully contained within self.
The trait exists to basically hide that `T` from lrpar.

In the `ariadne` trait implementation this is a vector of `ariadne::Report`s, in the simpler test impl, it is a `String` and the `result()` function calls `into()` on it.  for Sarif i'm expecting it to be a type that will be serializable to that format.

It is IMO not the prettiest thing, but I was running short on nicer ideas.

ltratt: My first thought was that `results` might want to take `self` (rather than `&self`) to make it clear that it's the "last" thing. But I'm not quite sure that's possible with Rust's type system in this case.

But then I wondered: do we even need such a "final" function? Can't whoever implements this trait summarise things on their own, without this? In other words, if they want to track whether there are any warnings/errors they can have (say) a `bool=false` and set it to `true` at the start of each implementation of a method in this trait?

ratmice:
Being fully incremental might require some work,

We'd need to make a trait for e.g. `cfgrammar`, to replace the vector here:
https://github.com/softdevteam/grmtools/blob/master/cfgrammar/src/lib/yacc/parser.rs#L306

The tension being that cfgrammar wants to parse produce as many errors as it can, even after previous errors have occurred.
So currently it stores those in a `Vec` and returns them all at once.

lrlex/src/lib/ctbuilder.rs
ratmice: Part of the key to things like `Safir` output is that self is not consumed by this call.
my hope was that we could introduce a call like `dry_run`, which mirrors `build()` function but doesn't perform the filesystem operations, from which can pick up the results.  If you look at the last line in this test, it printlns some stuff out of the `ErrorHandler`, *after* calls to `build()`.

https://github.com/ratmice/error_callbacks_test/blob/implement_as_trait/build.rs#L287

We could definitely get away without `result`, but it means things like `build.rs` need to have some code which reaches into the type and prints warnings and errors, rather than only needing to do so in special cases like where you want to convert it to a specific format.  That was my reasoning for adding `results()`, and having it take a reference.

ltratt:
> Being fully incremental might require some work,

Ah, sorry, I wasn't very clear. So we can be "incremental" (in the sense I intended, but didn't clarify, oops!) API without really being very incremental in the way we use the API (at least yet). But I've realised my question wasn't very well formed. The API you're proposing is inherently incremental (in the sense that it's not a return type for `build` or similar), which I think is good. So we can ignore that high-level question :)

lrlex/src/lib/ctbuilder.rs
ltratt: > We could definitely get away without result, but it means things like build.rs need to have some code which reaches into the type and prints warnings and errors

I think I see where you're going with this. Can we unpack it into two questions?

1. What do we do by default as warnings/errors are discovered as things are processed?
2. What do we do by default at the final compile-time-`build` stage of a lexer/parser?

Those are two subtly different things, and I hadn't really thought about that before.

I think we still want the compile-time-`build` (for lexers/parsers) to produce an `Err` if there are errors discovered. That would remove a normal `build.rs`-using user from having to explicitly add checks for errors/warnings.

That then makes me wonder if we want (at least in part) the `ErrorHandler` traits to be outside the `ctbuilder` modules? In other words, some errors/warnings can be experienced by, say, an LSP, even though it won't have used `ctbuilder` --- and I *think* that there are some additional errors that only make sense in the "build lexer + parser together at compile-time" stage? Maybe one way of doing this is super/sub traits? e.g. for parsers maybe we'd have something like:

```rust
trait GrammarErrorHandler {
  // All the errors/warnings that a grammar can have whether it's compile-time or run-time
}

trait CompileTimeGrammarErrorHandler: GrammarErrorHandler {
  // Additional errors/warnings that make sense only at compile-time
}
```

and then `CTParserBuilder` would take an instance of `CompileTimeGrammarErrorHandler` and then (I think) `YaccGrammar` would take an instance of `GrammarErrorHandler`? I appreciate this implies a fairly deep change to how we handle all this stuff!

ratmice: That is an interesting question, I had debated whether I should actually put the current `GrammarErrorHandler` in `ctbuilder`, as in it's current form it *could* easily be used by `nimbeparse` outside `ctbuilder`, if do similar things for `RTParserBuilder`.

If i might add another layer here:

1. Errors/Warnings that happen during construction.
2. Errors/Warnings that can only be checked after construction of a (valid or invalid) `GrammarAST`  or successful construction of a `YaccGrammar` but are shared by `CT` and `RT`.
3. Errors that are `CT` specific. I think that this class is actually just filesystem errors, for reading the `.y` and writing the `.y.rs` (Which this trait isn't actually currently handling)

So I *believe* the current `GrammarErrorHandler` is the same as the compile-time or run-time one you mention above.

I'm not super well versed in this kind of trait magic, I actually tried to do something like this previously and ran into:
https://rust-lang.github.io/rfcs/3324-dyn-upcasting.html

That is being stablized just now though https://github.com/rust-lang/rust/pull/118133

>  I appreciate this implies a fairly deep change to how we handle all this stuff!

I'll have to remind myself how the construction of `Duplicate*Error` is produced, whether that is by buffering errors, or based on rules etc.  How deep of a change it is seems like it depends on that.

ratmice:
Sorry I haven't answered all of your questions, I want to think about them and don't feel I have clear answers without tinkering with things.

ratmice:
> 3. Should "analysis" be a separate thing than errors/warnings?

The feeling I'm getting is that, `*ErrorHandler` has is in 1 to 1 correspondence to a `RTParserBuilder` or `CTParserBuilder`.
While `analysis` is in many to one, one of the things keeping `missing_from_parser/lexer`, and `conflicts` from being just forms of analysis is that 1 to 1 correspondence.  So it might be good to try to replace those with analysis passes.

ratmice:
> 2. Should warnings be differentiated from errors? If so, should they be part of one trait or ... ?

I kind of feel like it is fine to keep them in one trait,

One thing to note with warnings is that the printing of them from within cargo is a bit strange.
https://github.com/softdevteam/grmtools/blob/master/lrpar/src/lib/ctbuilder.rs#L455C44-L460

Because cargo generally wants to suppress stderr. So `CTParserBuilder` uses an environment variable to tell if
we are running under cargo, and then print them specially, so they get shown regardless of panic.
This seems to me like logic we probably don't want everywhere.

I think I'd like to try the simplest thing I can imagine for warnings, changing the `results()` signature to something like:

```
/// Under the following conditions this function must return an `Err`
/// ....
/// Otherwise an `Ok(vec!["Warnings", ...])` 
fn results(&self) -> Result<Vec<String>, Box<dyn Error>>
```

That way we can keep when/how to actually print them specific to `CTParserBuilder` and each specific tool,
`GrammarErrorHandler` then doesn't have to worry about gaining a `show_warnings` flag, printing or not is up to the tool in question.
I am already imagining though that between Cargo and Ariadne, we might get double "Warning:" text, because both are prefixing their output.

Do you think that is generally worth trying?

ratmice:
So I would like to focus for a minute on ways to integrate this in an 'imperfect' fashion without hamstringing ourselves to an imperfect trait.

- Include the trait as a [`Sealed trait`](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/) trait
- Implement 'DefaultGrammarErrorHandler` and `DefaultLexerErrorHandler`s in lrpar and lrlex
- Migrate tools like nimbleparse/lrlex/lrpar, `CTParserBuilder`, and `RTParserBuilder` to use those `Default*Handler`s internally.
- Because the above cross crate boundries the trait methods must be callable by external crates so this would be a `supertrait sealed` trait in the sealed trait guide above.  The methods would be documented unstable, and `#[doc(hidden)]`
- We could even add a `#[cfg(feature="unstable_api")] struct UnstableAPI()` argument to the methods, to indicate their instability in code.  enabling or disabling the `unstable_api` feature, then turns these from `supertrait sealed trait` to `method signature sealed trait`.
- Once we feel like we have the trait method signatures in good working order and are happy with everything, we could unseal it and either add or unhide the `error_handler` in the `Builder`s

I think a plan of this sort would allow us to migrate the code to the new trait piecemeal, without having to try getting it perfect on the first try, and getting some real experience using it internally within the codebase, before having to worry about stablizing it.

Curious if you think this sounds like a viable plan?

lrlex/src/lib/ctbuilder.rs
ratmice: About `results()`, and whether we need it,

> But then I wondered: do we even need such a "final" function?

I was thinking about this, and one thing that we could do is turn the `on_*` functions to returning `Box<dyn Err>` (back as they were in the callbacks branch).  We could still pass `&mut self`, but that mutation would only be needed if the trait impl wants to store errors.

The caller would be responsible for merging errors in perhaps a slightly awkward fashion though,
The simplest form of which could even be something like: 

```
    use std::error::Error;
    let x: Box<dyn Error> = "foo\n".into();
    let y: Box<dyn Error> = "bar\n".into();
    let z: Box<dyn Error> =  [x.to_string(), y.to_string()].join("Error:\n").into();
```

This sort of removes some of the strangeness about mutation, but maybe moves it elsewhere to anywhere we want to merge individual errors.

ltratt: >  The caller would be responsible for merging errors in perhaps a slightly awkward fashion

I've been chewing on this. I think it goes back to what I was asking in https://github.com/softdevteam/grmtools/pull/428#issuecomment-1858760041 about incremental/batch. The more I think about it, the more I think the API you're proposing is (correctly IMHO!) enabling incremental --- but there's nothing stopping a user "downgrading" it to batch. In other words, we're (potentially) telling the API user "as soon" as issues are discovered and they can then report those on immediately or later as they wish. In that sense, `results` is us trying to make the API support both incremental and batch natively --- but a batch user already has all the bits they need to make a batch API without `results`. So I'm inclined to keep the API minimal-ish and not have `results`.

lrpar/src/lib/ctbuilder.rs
ltratt: > One thing missing here is that warnings are no longer printed, but there is no mechanism to print them from the trait.

The trait doesn't mind if you print or don't :) I have a couple of possible ideas here:

1. The default implementation of the trait provided by the `CTBuilder`s is that they print warnings and/or panic pretty much as they do now.
2. We can provide an implementation of the trait which behaves as it does now but *before* it prints/panics also calls a user implementation of the trait.

Of these, 1) is the easiest to implement and to get right, so I'm inclined to start there, assuming we think it makes sense!

lrlex/src/lib/ctbuilder.rs
ratmice: I tend to agree, I also think it simplifies the contract between caller and trait impl down to basically the method signature,
making the trait impl's more obviously correct.

The method which doesn't quite fit this mold of returning `Box<dyn Error>` is `on_grammar_warning()` I'd be inclined to just have it return `String` -- for which there is `From<String> for Box<dyn Error>`, so callers can still turn them into errors if they desire, we could even drop `warnings_are_errors` from the trait I believe, unless we want this for switching between "Warning" and "Error" in the text.

ltratt: Maybe a silly question -- but why do any of these questions have to return anything? AIUI we're feeding an external implementation with information about what's wrong, and I don't think we care about them filtering out anything? So in that sense (assuming we drop `results`) can't we just have each function take arguments and return nothing?

ratmice: I think it is pretty necessary to return something, because e.g. `build.rs` scripts are going to end in `.build()?`, and if we don't return an error variant, it seems wrong to return `Ok` even if we cannot proceed within the `build` process, due to e.g. the lack of a `YaccGrammar` value.

ltratt: I wonder if I've misunderstood. I thought the trait was something that *we* call and someone *outside* implements? In the case of (say) `CTBuilder` it can track (by implementing the trait) whether there were errors and then cause `build` to fail. Maybe I've got this backwards!

ratmice: Hmm, you've got it right, we call it and a `CTBuilder` calls it, and there can be both an *outside* implementation, and *default* implementation within the crate.

The main thing it is doing is delegating the responsibility of turning concrete e.g. `YaccGrammarError` types,
into the abstract `Box<dyn Error>`, the return type of `build()`  (Currently this is handled with custom formatting internally).  It still definitely tracks and knows which errors are fatal to to the build process.  It just acts as a shim for configuring the formatting, and if need be collecting the errors that occurred into values that escape `build()`.

If understanding your confusion, I do think it isn't totally clear whether this type should be driving any form of `ControlFlow` for error handling. I think part of this is a naming thing `ErrorHandler` implies it should, perhaps
`ErrorFormatter` would be better.  `warnings_are_errors` also seems to imply it should, but it is currently just treated
as a flag for whether we should return a `Err()` or `Ok()` from `results()` in the case where there are no errors but there are warnings.

Putting the `Box<dyn Error>` return value `on_grammar_error` and friends return values, puts control-flow squarely in the hands of the callers, and steps away from trying to mirror the built-in control-flow within the trait.

While leaving the `results()` call in kind of behaves like the trait is actively participating in the control-flow with the caller e.g. `CTBuilder`.  I do think it's definitely an issue we want to deliberately decide which side of the fence we stand on.
(and one that my current trait isn't particularly doing well at in the slightest!).

ltratt: > The main thing it is doing is delegating the responsibility of turning concrete e.g. YaccGrammarError types, into the abstract Box<dyn Error>, the return type of build()

Ah, so I got the wrong end of the stick (as is not uncommon!). So is what you're proposing a sort-of `ErrorMap` trait in the sense of "we give you, the external implementer an error, you `map` (in the sense of `iter().map(|| ...)` it into whatever you want, then we bundle all of those mapped things up and return them"?

ratmice: Yeah, that is a good way of putting it.  I lean towards that being the extent of what it should be doing since trying to do more seems just seems to make things complicated.

ltratt: Right, got it!

So what I'll do is explain what I *thought* you were proposing, then I'll go away and chew over which I think might be best!

What I thought was going to happen is that we provide this trait and as we find a warning/error we call the appropriate function. The user can then do various things: they can, for example, print things out, collate them to do something with later, or whatever. But as soon as we've called that function, our responsibility for that particular warning/error is over (i.e. we don't collate them).

So, for example, `CTBuilder` might choose to collate everything, and then print it out at when `build` is called. In contrast, an editor might try displaying popups immediately to a user while analysis (etc.) continues in a background thread.

ratmice: Yeah, originally it was kind of trying to do *both* things, e.g. it has to sort of provide the `map` because `build()` has to return *something*, in theory it could just return a token indicating `ErrorsOccurred`, and beyond that expect the trait to do what it does.  But also it has to dectect `cargo` and print things using a cargo in-band [cargo:warning](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargo-warning).  So if we were to delegate all that to the trait, it is a bit of build script environment variable testing and special casing that each trait needs to implement when called from `build()`.

The above is what probably kept me from further going down that path.

ratmice: One last thing about e.g. pop-up alerts, I would note that this is still possible with the `map` centric approach, but it must act in *batches*, and require the *outside* to implement the pop-up behavior.
```
struct MyErrorThing {
   errors: Vec<YaccGrammarError>,
   warnings: Vec<YaccGrammarWarning>,
}

impl ErrorMapper for MyErrors {
     fn on_grammar_error(&mut self, error: YaccGrammarError) -> Box<dyn Error> {
         self.errors.push(error);
         // this perhaps would need to Box::new()?
         error.into() 
     }
}

then from build.rs:
```
let _ = ct.error_mapper(myerrors).build();
for e in my_errors {
    popup_alert(e);
}
```

Something like that anyways.

**Comment by ltratt** [comment]
So I've been chewing for a bit, and my current thinking is that the "map" approach does, by its nature, lock in a fairly specific kind of behaviour. In many cases it's a very reasonable style of behaviour, but it seems to me that we have a more general kind of behaviour at our fingertips, and might as well see if we can work out exactly what that is. [Part of the reason I'm inclined towards the more general approach is that whatever we do now is an API break. Since grmtools current approach to reporting errors was always short-termist, that break is not just OK, but I've always been expecting it to come. That said, ideally, we'd like to do fewer breaks in the future to the error reporting API because we hope that we're able to soon iterate towards an approach we can more-or-less stick to for the long term.] Of course, maybe the "map" approach will end up being the best --- we can always come back to it if so!

Bringing everything together, I think we can see a possible high-level design principle:

* We want to report errors / warnings / analysis stuff to the user as we discover it. [What I previously called "incremental" though that word might be a bit loaded!]

The obvious question then becomes: how does that play out in practise? I think we can see something else emerge from our discussions: there are various layers of errors / warnings / analysis. Fortunately (AFAICS at the moment, at least!) these can compose. So, for example we might have a couple of traits (as you can see, I don't know what good names are!):

```rust
trait RTParserBuilderErrorWarningAnalysisTraitThing {
  /// Report a syntax error in the input grammar.
  fn syntax_error(...);
}

trait CTParserBuilderErrorWarningAnalysisTraitThing: RTParserBuilderErrorWarningAnalysisTraitThing {
  /// The grammar is referencing tokens that aren't in the lexer.
  fn missing_in_lexer(...);
}
```

Both `RTParserBuilder` and `CTParserBuilder` can (if we want!) provide default implementations.

Because these things compose, `CTParserBuilder` can, internally, provide its own implementation that wraps the user's version, so `build` can still return `Err` if there are errors. So `CTParserBuilder` might do something like:

```rust
struct ErrorWarningWrapper {
  user_impl: Box<dyn RTParserBuilderErrorWarningAnalysisTraitThing>,
  errors_seen: Vec<Box<dyn Error>>
}

impl RTParserBuilderErrorWarningAnalysisTraitThing for ErrorWarningWrapper {
  fn syntax_error(...) {
    self.errors_seen.push(...);
    self.user_impl.syntax_error(...);
  }
}
```

and then `CTParserBuilder` can check `error_wrapper.errors_seen` in `build` and report them back to the user (or just return `Err("Some errors were seen")` or whatever it chooses). In other words, the "standard" case would feel pretty much just like using grmtools today --- but you also have the ability to do fine-grained things with errors / warnings if you want (whether that's printing them to stderr, or popups, or feeding them to a later analysis or whatever).

Please feel free to tell me that I've lost the plot :) You've thought about this more than I have!

ratmice:
> my current thinking is that the "map" approach does, by its nature, lock in a fairly specific kind of behaviour. In many cases it's a very reasonable style of behaviour, but it seems to me that we have a more general kind of behaviour at our fingertips, and might as well see if we can work out exactly what that is.

That seems totally reasonable to want to explore other options, I would say the "map" approach is my best attempt so far with a focus on compatibility. That is it merely adds new optional behavior, and allows for some internal reorganization that didn't affect API but does allow us to remove some duplicated code across a few crates.  I'm happy to look at implementations that focus on things like incremental reporting of errors that will require incompatible changes.

> Part of the reason I'm inclined towards the more general approach is that whatever we do now is an API break. Since grmtools current approach to reporting errors was always short-termist, that break is not just OK, but I've always been expecting it to come. That said, ideally, we'd like to do fewer breaks in the future to the error reporting API because we hope that we're able to soon iterate towards an approach we can more-or-less stick to for the long term.
>
> Bringing everything together, I think we can see a possible high-level design principle:
>
>*  We want to report errors / warnings / analysis stuff to the user as we discover it. [What I previously called "incremental" though that word might be a bit loaded!]

I *think* another term for this could be an error `stream` or `channel` perhaps.  
I'm going to enumerate what I consider to be the main areas of difficulty implementing these (in no particular order)

1. `RTParserBuilder`/`CTParserBuilder` currently receive buffered errors in batches and the user e.g. as a `Vec<*Error>`.
So I think the streaming must go at a lower level, e.g. One of the main sources of errors is currently ending up here: https://docs.rs/cfgrammar/latest/src/cfgrammar/yacc/ast.rs.html#18 Such that we could consider that vec becoming the write end of a channel.
2. At the origin where most errors occur I don't believe we have anything such as the yacc `filename.y`, so we need a way to tack this on in the builders, or otherwise pass it (or even just any form `FileId` ).
3. `add_duplicate_occurrence` is in tension with streaming, because it is collating duplicate rules/names etc which require caching them and reporting them at the end to squash all duplicates to a single error. 
https://github.com/softdevteam/grmtools/blob/0f30331f14bbb0d353d533f16e8b2bebeb485a95/cfgrammar/src/lib/yacc/parser.rs#L272-L291C2

I believe that is going to be all the issues we should expect to encounter (maybe an instance of each across the crates).
It is definitely possible I've overlooked something though. One and two definitely have the most user visible impact on API (I think in this case `ASTWithValidityInfo::new` wants a channel and a some `File Id` or *filename*). 
I'm guessing we'd want to remove 3 or move it out so it is something the read end of an error channel would do compacting duplicate errors?

ltratt:
> That is it merely adds new optional behavior, and allows for some internal reorganization that didn't affect API but does allow us to remove some duplicated code across a few crates.

That's a good point and something I might have underplayed. Though this PR does add new `pub` traits so I think it does also extend the current user-visible API? That said, one thing that might be worth considering is using this PR (or something like it) as a stepping-stone for a minor release while we chug away on a bigger longer-term change for the next major release. I'm open to your thoughts on that!

> I think another term for this could be an error stream or channel perhaps.

"Streaming errors/warnings/etc" (is there a term that covers "errors/warnings/etc"? I can't think of one off-hand!) is certainly a better term than "incremental".

>    RTParserBuilder/CTParserBuilder currently receive buffered errors in batches and the user e.g. as a Vec<*Error>.
>    So I think the streaming must go at a lower level, e.g. One of the main sources of errors is currently ending up here: https://docs.rs/cfgrammar/latest/src/cfgrammar/yacc/ast.rs.html#18 Such that we could consider that vec becoming the write end of a channel.

I think there's a good implicit question here which is: how far should we push streaming errors? We could say "we only really care about RTBuilder/CTBuilder", and as long as they present a nice API we don't mind too much if other parts of grmtools have a slightly cruder error/warnings API. This would probably be a good intermediate step, full stop, and perhaps might be the most pragmatic option anyway?

>  At the origin where most errors occur I don't believe we have anything such as the yacc filename.y, so we need a way to tack this on in the builders, or otherwise pass it (or even just any form FileId ).

Right, so the streaming error API would probably need some sort of helper methods (or sometimes maybe parameters) that allow it to be told such information. I think some implementations will largely ignore this information, but some will want every last bit of it!

>    add_duplicate_occurrence is in tension with streaming, because it is collating duplicate rules/names etc which require caching them and reporting them at the end to squash all duplicates to a single error.
>    https://github.com/softdevteam/grmtools/blob/0f30331f14bbb0d353d533f16e8b2bebeb485a95/cfgrammar/src/lib/yacc/parser.rs#L272-L291C2

I *think* this an instance of the first problem in disguise? Maybe I'm not noticing an important detail though!

Thanks for bearing with me on this. I know it's challenging when good work receives a barrage of questions! But your PRs really have made me think and I'm very grateful that you're willing to bear with me while we work out what to do next!

ratmice:
> > add_duplicate_occurrence is in tension with streaming, because it is collating duplicate rules/names etc which require caching them and reporting them at the end to squash all duplicates to a single error.
> > https://github.com/softdevteam/grmtools/blob/0f30331f14bbb0d353d533f16e8b2bebeb485a95/cfgrammar/src/lib/yacc/parser.rs#L272-L291C2
> 
> I _think_ this an instance of the first problem in disguise? Maybe I'm not noticing an important detail though!

It's definitely dependent on the first problem, I think the important question here is what do we prioritize between reducing error spam, and streaming errors upon discovery.

* It is the only place we do lookups in the vector of errors.
* We currently prioritize an incompatible property, reducing error spam by combining repeated errors and emitting them once with multiple locations.

In this case the errors aren't created in a single pass e.g. given the following:
```
A: 'a';
A: 'a2';
B: 'b';
A: 'a3';
```

we `errs.push(duplicate A)`, parse `B` successfully, then search through errs and add the span for the `a3` rule to the error previously pushed, so it isn't like we do a single pass here over all the duplicate names and emit an error.

> Thanks for bearing with me on this. I know it's challenging when good work receives a barrage of questions! But your PRs really have made me think and I'm very grateful that you're willing to bear with me while we work out what to do next!

ltratt:
>     It is the only place we do lookups in the vector of errors.
>    We currently prioritize an incompatible property, reducing error spam by combining repeated errors and emitting them once with multiple locations.

Good point! I think I'm OK with shoving off to the implementer of the error trait (we could, if we wanted, even provide a de-duper composition trait, but that's a strictly optional extra). That would simplify what we do internally sufficiently to unblock this general idea I think?

>     If it helps, I actually think we aren't too far off from e.g. a fully streaming errors code-wise as the crow flies, but to me the important question is whether it is reasonable from an API compatibility perspective, but I think it is worth making a good faith effort.

Thank you -- and I agree!

ratmice:
So, today I focused on whether or not it is possible to implement this in any roughly compatible way, and it seems like it is possible I changed one (rarely used method) from returning a `slice` to a `Vec`) .  The basic idea is to mirror the current ownership structure.  `ASTWithValidationInfo`, goes from owning a `Vec<YaccGrammarError>` to owning a `(error_sender, Option<error_receiver>)` pair.  We can then reimplement the functions expecting to return a `Vec<YaccGrammarError>` result in terms of an interator over the `error_receiver` endpoint.  Alternately there is a new method which returns the `error_receiver` (basically) `Option::take`, in which case they can receive errors as they are produced.

I seem to have broken quite a bit in my tinkering, so haven't pushed anything but it seemed worth trying.  Because it would either allow us to go through a deprecation cycle, or even just keep the `buffered` error vecs, around for cases where we want the simpler API?

I think there are a few rough edges, revolving around mixing the channel approach with `Vec` result approach, e.g. when the user has taken the `error_receiver`, then subsequently asks for you to convert it into a `Vec<YaccGrammarError>`, I don't believe we currently have the right return values to indicate this nicely.  Anyway, curious about your thoughts on the matter.

PR #432, opened 2023-12-28
Cleanup test helpers

ratmice:
This is a cleanup for the `ErrorsHelper` test functions, it is primarily intended to improve the output.
I.e. it should now print both the error, and the expected, when they don't match now.
Previously this was difficult to do because the expected value was used in the if guard of an `Err(e) if ...` it could no longer be used in plain old the `Err(e)` branch of the match.  As such, I just reorganized the implementation to make that branch unneeded.

Doing the same thing in `lrlex` required we add a `Debug` derive for `LrNonStreamingLexerDef`.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -1060,6 +1060,7 @@ mod test {
     }
 
     impl ErrorsHelper for Result<GrammarAST, Vec<YaccGrammarError>> {
+        #[track_caller]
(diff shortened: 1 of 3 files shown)


ratmice:
It occurred to me that it would be good to add `#[track_caller]` to these helper functions, so that panics in the helper refer to the location in the test where the helper is called, rather than some line within the helper itself.  Pretty sure that is everything for this series.

ltratt:
> It occurred to me that it would be good to add #[track_caller] to these helper functions

I did not know about `#[track_caller]`! Good idea.

PR #434, opened 2023-12-31
Make pub(crate) fields of `ASTWithValidityInfo` private.

ratmice:
This makes some fields which were `pub (crate)` private, and replaces their usage outside the `ast` module with the existing `pub` functions.  Which makes it a tiny bit easier to refactor these fields.


ratmice:
Also noticed in that last commit (even though the private fields are visible at that point), that can be implemented in terms of the `pub` api.

ltratt:
Please squash.

ratmice:
Squashed.

PR #436, opened 2024-01-03
Error reporting

ratmice:
Sorry for sending another large PR.  I was hoping I could just send some smaller ones, but it seems fairly difficult to break up.
In theory I could break it up at patches [04ba0c1](https://github.com/softdevteam/grmtools/pull/436/commits/04ba0c15bb5aad1e1d76b446dbfb8252cff3e92d),  [c7bc81f](https://github.com/softdevteam/grmtools/pull/436/commits/c7bc81fac4a97b1c46cbce5fc439cb32f22b6e3c) and last [050d207](https://github.com/softdevteam/grmtools/pull/436/commits/050d2075f012abe334da79df1a46fe9410f25609).

But [6d41bbb](https://github.com/softdevteam/grmtools/pull/436/commits/6d41bbb8c0a09ef83649195b03cae10b883b4f45) modifies a trait. I wanted the addition of the trait method to be close to its usage, because the specific method was a key source of bugs during the development (when failing to call it) one way to avoid  it would be to call `finish` from the `ErrorMap` trait.  But we'd have to document that `finish` can be called multiple times, so using `drain` would be necessary. In a sense that would just swapping API awkwardness elsewhere.

Overall this branch has aspects of all three of the patches (callbacks, trait, and channel series minus the channels).
I'm not soo terribly certain how well actually trying to do `incremental` or prompt receiving of errors at the time they occurr.
In particular the combination of `warnings` and `errors` means that one of warnings or errors could block the other.
unless you spawn a thread for each, or use a channel which is sync on the grmtools side, and async on the outside (like my lsp does).

I'm not very smitten on some of the names, but have lacked any good alternatives like `ErrorReport` is kind of like `ErrorEscrow` or `ErrorCache` and using the `NewlineCache` style names for methods like `feed_error()`, `feed_warning()`.
But I tried not to get bogged down on naming and just get it working from end to end.

There isn't anything in the branch which calls it from build.rs, follows is an example, but FWIW, the parts hooking it up to `CTParserBuilder` could probably use some work, but at least as a proof of concept show it is possible here to get the source + the path to an error formatter without plumbing it all the way through the parser, or repeatedly building `NewlineCache`.

```
struct Formatter {}

impl Formatter {
    fn formatter(_src: &str, _path: &std::path::Path) -> Box<dyn ErrorFormatter> {
        Box::new(Self {}) as Box<dyn ErrorFormatter>
    }
}

impl ErrorFormatter for Formatter {
    fn format_error(&self, _: &YaccGrammarError) -> Box<dyn Error> {
        "fancy error".into()
    }

    fn format_warning(&self, _: &YaccGrammarWarning) -> String {
        "fancy warning".to_string()
    }
}
```
...
.lrpar_config(|ctp| {
                ctp.error_formatter(Box::new(Formatter::formatter)).grammar_in_src_dir(...).unwrap()
})
```

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -7,40 +7,70 @@ use indexmap::{IndexMap, IndexSet};
 
 use super::{
     parser::YaccParser,
-    reporting::{ErrorReport, SimpleReport},
-    Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccGrammarWarning, YaccGrammarWarningKind,
-    YaccKind,
+    reporting::{ASTBuilderError, DedupReport, ErrorReport, ReportHandler, SimpleReport},
+    Precedence, YaccGrammar, YaccGrammarError, YaccGrammarErrorKind, YaccGrammarWarning,
+    YaccGrammarWarningKind, YaccKind,
 };
 
 use crate::Span;
+
+pub struct ASTBuilder<'a, R> {
+    source: &'a str,
+    yacckind: Option<YaccKind>,
+    report_handler: Option<ReportHandler<'a, R>>,
+}
+
+pub struct GrammarBuilder {
+    yacckind: YaccKind,
+    ast: GrammarAST,
+    any_errors_found: bool,
+}
+
 /// Contains a `GrammarAST` structure produced from a grammar source file.
 /// As well as any errors which occurred during the construction of the AST.
 pub struct ASTWithValidityInfo {
     ast: GrammarAST,
-    report: SimpleReport,
+    report: DedupReport<SimpleReport>,
 }
 
-impl ASTWithValidityInfo {
-    /// Parses a source file into an AST, returning an ast and any errors that were
-    /// encountered during the construction of it.  The `ASTWithValidityInfo` can be
-    /// then unused to construct a `YaccGrammar`, which will either produce an
-    /// `Ok(YaccGrammar)` or an `Err` which includes these errors.
-    pub fn new(yacc_kind: YaccKind, s: &str) -> Self {
-        let mut report = SimpleReport::new();
-        let ast = match yacc_kind {
+impl<'a, R> ASTBuilder<'a, R>
+where
+    R: ErrorReport,
+{
+    /// Returns a new `ASTBuilder` with `source`, and an optional `path`
+    /// The path may be used for error reporting, but is not otherwise used for any
+    /// filesystem operations or checked to be a valid path.
+    pub fn with_source(source: &'a str) -> Self {
+        Self {
+            source,
+            yacckind: None,
+            report_handler: None,
+        }
+    }
+    pub fn yacckind(mut self, yk: YaccKind) -> Self {
+        self.yacckind = Some(yk);
+        self
+    }
+    pub fn error_report(mut self, report: &'a mut R) -> Self {
+        self.report_handler = Some(ReportHandler::new(report));
+        self
+    }
+
+    /// Builds an `ast`, and returns a `GrammarBuilder`
+    pub fn grammar_builder(mut self) -> Result<GrammarBuilder, ASTBuilderError> {
+        let report_handler = self
+            .report_handler
+            .as_mut()
+            .ok_or(ASTBuilderError::MissingErrorReport)?;
+        let yacckind = self.yacckind.ok_or(ASTBuilderError::MissingYaccKind)?;
+
+        let ast = match yacckind {
             YaccKind::Original(_) | YaccKind::Grmtools | YaccKind::Eco => {
-                let mut yp = YaccParser::new(yacc_kind, s.to_string());
-                yp.parse()
-                    .map_err(|es| {
-                        // Interim change until `ReportHandler`.
-                        for e in es {
-                            report.grammar_error(e)
-                        }
-                    })
-                    .ok();
+                let mut yp = YaccParser::new(yacckind, self.source.to_string());
+                yp.parse_with_report(report_handler).ok();
                 let mut ast = yp.ast();
                 ast.complete_and_validate()
-                    .map_err(|e| report.grammar_error(e))
+                    .map_err(|e| report_handler.non_fatal_error(e))
                     .ok();
                 ast
             }
@@ -50,12 +80,53 @@ impl ASTWithValidityInfo {
                 Symbol::Rule(_, span) => (YaccGrammarWarningKind::UnusedRule, span),
                 Symbol::Token(_, span) => (YaccGrammarWarningKind::UnusedToken, span),
             };
-            report.grammar_warning(YaccGrammarWarning {
+            report_handler.warning(YaccGrammarWarning {
                 kind,
                 spans: vec![span],
             })
         });
-        ASTWithValidityInfo { ast, report }
+        report_handler.finish();
+        let any_errors_found = report_handler.any_errors_found();
+        Ok(GrammarBuilder {
+            yacckind,
+            ast,
+            any_errors_found,
+        })
+    }
+}
+
+impl GrammarBuilder {
+    pub fn build<StorageT>(&self) -> Result<YaccGrammar<StorageT>, ASTBuilderError>
+    where
+        StorageT: num_traits::PrimInt + num_traits::Unsigned + 'static,
+        usize: num_traits::AsPrimitive<StorageT>,
+    {
+        YaccGrammar::new_from_grammar_builder(self.yacckind, self)
+    }
+    pub fn any_errors_found(&self) -> bool {
+        self.any_errors_found
+    }
+    pub fn ast(&self) -> &GrammarAST {
+        &self.ast
+    }
+}
+
+impl ASTWithValidityInfo {
+    /// Parses a source file into an AST, returning an ast and any errors that were
+    /// encountered during the construction of it.  The `ASTWithValidityInfo` can be
+    /// then unused to construct a `YaccGrammar`, which will either produce an
+    /// `Ok(YaccGrammar)` or an `Err` which includes these errors.
+    pub fn new(yacc_kind: YaccKind, s: &str) -> Self {
+        let mut report = DedupReport::new(SimpleReport::new());
+        ASTWithValidityInfo {
+            ast: ASTBuilder::with_source(s)
+                .yacckind(yacc_kind)
+                .error_report(&mut report)
+                .grammar_builder()
+                .unwrap()
+                .ast,
+            report,
+        }
     }
 
     /// Returns a `GrammarAST` constructed as the result of parsing a source file.
@@ -74,12 +145,12 @@ impl ASTWithValidityInfo {
 
     /// Returns all errors which were encountered during AST construction.
     pub fn errors(&self) -> &[YaccGrammarError] {
-        self.report.errors()
+        self.report.child().errors()
     }
 
     /// Returns all errors which were encountered during AST construction.
     pub fn warnings(&self) -> &[YaccGrammarWarning] {
-        self.report.warnings()
+        self.report.child().warnings()
     }
 }
 
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -6,8 +6,10 @@ use std::{
 use indexmap::{IndexMap, IndexSet};
 
 use super::{
-    parser::YaccParser, Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccGrammarWarning,
-    YaccGrammarWarningKind, YaccKind,
+    parser::YaccParser,
+    reporting::{ErrorReport, SimpleReport},
+    Precedence, YaccGrammarError, YaccGrammarErrorKind, YaccGrammarWarning, YaccGrammarWarningKind,
+    YaccKind,
 };
 
 use crate::Span;
@@ -15,7 +17,7 @@ use crate::Span;
 /// As well as any errors which occurred during the construction of the AST.
 pub struct ASTWithValidityInfo {
     ast: GrammarAST,
-    errs: Vec<YaccGrammarError>,
+    report: SimpleReport,
 }
 
 impl ASTWithValidityInfo {
@@ -24,17 +26,36 @@ impl ASTWithValidityInfo {
     /// then unused to construct a `YaccGrammar`, which will either produce an
     /// `Ok(YaccGrammar)` or an `Err` which includes these errors.
     pub fn new(yacc_kind: YaccKind, s: &str) -> Self {
-        let mut errs = Vec::new();
+        let mut report = SimpleReport::new();
         let ast = match yacc_kind {
             YaccKind::Original(_) | YaccKind::Grmtools | YaccKind::Eco => {
                 let mut yp = YaccParser::new(yacc_kind, s.to_string());
-                yp.parse().map_err(|e| errs.extend(e)).ok();
+                yp.parse()
+                    .map_err(|es| {
+                        // Interim change until `ReportHandler`.
+                        for e in es {
+                            report.grammar_error(e)
+                        }
+                    })
+                    .ok();
                 let mut ast = yp.ast();
-                ast.complete_and_validate().map_err(|e| errs.push(e)).ok();
+                ast.complete_and_validate()
+                    .map_err(|e| report.grammar_error(e))
+                    .ok();
                 ast
             }
         };
-        ASTWithValidityInfo { ast, errs }
+        ast.unused_symbols().for_each(|symidx| {
+            let (kind, span) = match symidx.symbol(&ast) {
+                Symbol::Rule(_, span) => (YaccGrammarWarningKind::UnusedRule, span),
+                Symbol::Token(_, span) => (YaccGrammarWarningKind::UnusedToken, span),
+            };
+            report.grammar_warning(YaccGrammarWarning {
+                kind,
+                spans: vec![span],
+            })
+        });
+        ASTWithValidityInfo { ast, report }
     }
 
     /// Returns a `GrammarAST` constructed as the result of parsing a source file.
@@ -53,7 +74,12 @@ impl ASTWithValidityInfo {
 
     /// Returns all errors which were encountered during AST construction.
     pub fn errors(&self) -> &[YaccGrammarError] {
-        self.errs.as_slice()
+        self.report.errors()
+    }
+
+    /// Returns all errors which were encountered during AST construction.
+    pub fn warnings(&self) -> &[YaccGrammarWarning] {
+        self.report.warnings()
     }
 }
 
--- a/cfgrammar/src/lib/yacc/reporting.rs
+++ b/cfgrammar/src/lib/yacc/reporting.rs
@@ -1,4 +1,8 @@
 use super::{YaccGrammarError, YaccGrammarWarning};
+use crate::{
+    yacc::{parser::SpansKind, YaccGrammarErrorKind},
+    Span, Spanned,
+};
 use std::fmt;
 
 /// Caller gives ownership of errors and warnings to the impl.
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: cfgrammar/src/lib/yacc/reporting.rs [hunk 1]
```diff
@@ -44,6 +45,11 @@ pub trait ErrorReport {
     fn finish(&mut self) {}
 }
 
+pub trait ErrorFormatter {
+    fn format_error(&self, e: &YaccGrammarError) -> Box<dyn Error>;
```
ratmice, line 49: One thing we could consider (Which I haven't tried so it could be a bad idea) is if something like `Into<Box<dyn Error>>` would be better?

cfgrammar/src/lib/yacc/reporting.rs
```diff
@@ -50,6 +50,23 @@ pub trait ErrorFormatter {
...
 }
 
+/// Applies a `map` operation with the function `f` over
+/// the error values owned by self.
+pub trait ErrorMap {
+    /// Applies `f` to all errors owned by self where `f` is a function
+    /// returning a type that implements `Error`.
+    fn format_errors<'a, F: ErrorFormatter + ?Sized>(
+        &'a self,
+        f: &'a F,
```
ratmice, line 60: This could instead be: 
```
f: &'a (impl ErrorFormatter + ?Sized)
```

Not really sure if that is any better though

ratmice:
I just cleaned up some by removing some lifetimes, these were holdovers from experiments regarding where path/sources should be fed into formatters from (Specifically it was in the `ErrorReport` at some time, but it turned out to simplify things immensely to pass it directly to the `ErrorFormatter` instead).  Anyhow with that stuff no longer the case it seems we can elide some lifetimes that I wasn't able to before.

cfgrammar/src/lib/yacc/reporting.rs
```diff
@@ -7,6 +11,38 @@ pub trait ErrorReport {
     fn grammar_error(&mut self, err: YaccGrammarError);
     /// Gives ownership of a warning to self.
     fn grammar_warning(&mut self, err: YaccGrammarWarning);
+
+    /// Default implementation does nothing. If you override this
```
ltratt, line 15: Do we need a function for this -- can't the user do it with a `drop` method?

[The idea of a `Stage` below is interesting. I'm mildly nervous about it, though, because it does somewhat bake in some aspects of the implementation that aren't fundamental. I also wonder if a user who cares about these distinctions can already infer it from the error kinds they see?]

cfgrammar/src/lib/yacc/reporting.rs
```diff
@@ -198,6 +198,9 @@ pub struct SimpleErrorFormatter<'a> {
 
 impl<'a> SimpleErrorFormatter<'a> {
     #[allow(clippy::result_unit_err)]
+    // We could be more efficient by calling `feed(src[0..max_span_offset])`
+    // but we don't have the max span offset here, it may be worth considering
+    // now though because this is pub.
     pub fn new(src: &'a str, path: &'a path::Path) -> Result<Self, ()> {
```
ratmice, line 195: I considering doing that in `format_{warnings,errors}(&self, ...` using interior mutability, generally though these are reliant on the shared reference, because of the `move` in the `impl ErrorMap` of `SimpleReport`

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -6,16 +6,109 @@ use std::{
...
+
+impl GrammarBuilder {
+    pub fn build<StorageT>(&self) -> Result<YaccGrammar<StorageT>, ASTBuilderError>
+    where
+        StorageT: num_traits::PrimInt + num_traits::Unsigned + 'static,
+        usize: num_traits::AsPrimitive<StorageT>,
+    {
+        YaccGrammar::new_from_grammar_builder(self.yacckind, self)
+    }
+    pub fn any_errors_found(&self) -> bool {
```
ltratt, line 106: Dumb question: what is the scope of the "any errors"? With a builder, I think a normal Rust expectation is that nothing much happens until `build` is called, at which point `self` tends to be consumed and one gets out another thing entirely. Here we have `build(&self)` so I'm not quite sure what the expected chain is.

cfgrammar/src/lib/yacc/reporting.rs
ratmice: We actually can't usefully implement it from `Drop`, because of the following sequence:

```
let foo = DedupReport(ErrorReport::new());
gb.build(); // Assume errors are piling up in `DedupReport` but don't call finish();
assert_eq!(foo.format_errors(...).collect<Vec<_>>(), vec![]);
drop(foo); // Too late, it's the call to `foo` that depends upon the call to finish(), since errors are buffering in `DedupReport`. But `format_errors` delegates to `child_report`. 
```

If were to then say call finish within `format_errors`, we just need to be careful that we never feed an error multiple times in the case where `finish` was already called (i.e. drain the errors as it does, so in theory we could maybe drain the). 

I *think* perhaps we could drop `finish` if we just drain to `child` in `format_errors` without a call to `finish`.
The consequences of that is that if you call `format_errors` while errors are still arriving you could be getting duplicates.

I think that is the best way I can explain it.

cfgrammar/src/lib/yacc/ast.rs
ratmice: It is basically the negation of `ASTWithValidationInfo::is_valid`, so the scope is during `GrammarAST` construction,
I think you had used the term in one of the previous PRs where we were discussing `ASTBuilder` so we could keep it the same across those two if you prefer?

ratmice: But yeah, I believe the reasons it is taking `&self` were historical, and no longer relevant.
It seems fine to take `self` here to me.

ratmice: Actually nevermind, the real reason it doesn't take self is because the user may want to call `grammar_builder.ast()` in response to an error e.g. to format conflicts.  But there is nothing here exercising that yet.

ltratt: I'm inclined to think that we should bite the bullet on this sort of thing and work out what the "best" API would be and worry about backwards compatibility afterwards. I suspect that in most cases the breakage will probably be minimal, and perhaps non-existent.

ratmice: Interestingly `any_errors_found()` isn't needed for backwards compatibility, where it uses `self.errors().is_empty()`.
It is the movement of ownership of `errors` field into an `ErrorReport` that causes us to need that boolean (for `YaccGrammar::new_from_grammar_builder`.

My lsp is very likely the only thing actually using these APIs though. I think we could make `pub(crate) any_errors_found()` instead of `pub`.

But having the `ast` even if we can avoid it entirely for error formatting of things like conflicts in the future, still seems pretty crucial for many lsp needs in particular this would mostly surface in things like *jump to rule*.  Anyhow i'd be happy to remove e.g. `ASTWithValidationInfo` and `YaccGrammar::new_from_ast_with_validation_info` as I think I'm the only one currently using *both* the *ast* and the *grammar* outside the libraries.

ratmice: So, if we made `any_errors_found` private, that just leaves us with `ast()` and `build()`, we could rename `ast()` `ast_ref()` perhaps, and replace `build()` with `TryFrom`.  Which to me is starting to sound something like an `GrammarAstRef`.  I think the key weirdness here is that we have an *Owned* (currently) `GrammarBuilder` value, which only gives references to it's inner value `&GrammarAST`.  Otherwise `GrammarAST` could be mutated and we would no longer know if it is valid.

I think these are some of the key points about why this API seems a little awkward?

ltratt: > I think these are some of the key points about why this API seems a little awkward?

Yes, I think so.

Can we reimagine the API at an even higher level perhaps? [I don't have immediate ideas on this I must admit!]

ratmice: Or maybe even at a lower level, like we could have a trait or another type that gives an owned variation of `GrammarAST` but without any `pub` fields or `&mut` functions, then it would be pretty natural to use/pass references to it?

If we did it as a trait we wouldn't necessarily have to add a new type. For example we could return something like `impl ReadonlyGrammarAST`. That leaves open the question of validity, would it make sense to have a `is_valid()` function which would only allow us to return them after `complete_and_validate`?

ratmice: The other thing that strikes me is that perhaps just a terminology change might help, such as we could rename `grammar_builder()` to `validate()`, which could return perhaps a `ASTReceipt` instead of the nested `GrammarBuilder`.
from there we could have `TryFrom` or `receipt.grammar()`

ltratt: > The other thing that strikes me is that perhaps just a terminology change might help, such as we could rename grammar_builder() to validate()

Yes, I think that's probably a good idea! That could easily a PR in its own right I guess?

ratmice: It kind of depends I think, if we do that first and then decide whether we want to change the ownership situation of errors later?  Then it is essentially a renaming and just changing where one function is `pub`.

If we view `ASTBuilder` (e.g. as in this PR) a rename of `ASTWithValidationInfo`, I think we have a few things:
* Rename, [easy]
* Move where `YaccGrammar::new_from_ast_with_validation_info` is pub into the chain. [easy]
* Deprecation and removal of `errors()` and `warnings()` which is the primary difficulty that causes all the 7 or so patches of new code before I introduce it in this series.

ratmice: So, I thought about your notion of needing to perhaps think about how we can fix things at a higher level.
What I think a higher level notion we might be missing here is a `Driver`.

I also am kind of doubting the entire builder-esque pattern is a good fit for this instance, if we look at the fields in my `ASTBuilder` impl they are actually all required, and we're returning errors when any are none, which seems weird.
I.e. it's just introducing potential errors, so we can use method chaining basically.  

I was looking through the [init-struct-pattern](https://xaeroxe.github.io/init-struct-pattern/) there is another pattern not shown in that link, but is similar since it has an init function, and `Default` implementing field. That allows for required options, and seems nice to me.

I took a stab at a mockup implementation of how a notion of a `Driver` could work using the `Report` trait idea.
I think part of the problem with this builder notion is also that results break the chain, but we have multiple values we may want to return.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df2df54c4578dda35d7bc5d2b6698bdc

But with this driver notion, I think the let binding ends up working pretty well, since we've taken the field construction out of the method chain.  Curious if you have any thoughts?

cfgrammar/src/lib/yacc/reporting.rs
ratmice: I updated the example code snippet in the comment above to hopefully better reflect the problem.

ratmice: One thing though we can do, is call this function from `ReportHandler::drop`, potentially even giving it an sealed argument so it can't be called outside of report handler, but that is exactly the same point I called it from `build()` chain above.

ratmice:
I wanted to break this out a bit, (sorry for all the noise). 

> [The idea of a Stage below is interesting. I'm mildly nervous about it, though, because it does somewhat bake in some aspects
> of the implementation that aren't fundamental.

I went from `finish()` (revealing the least amount of information) to revealing the most amount of information (the actual stage).
In theory there is a middle ground that just notifies when a scope in which we do the checking for specific errors is entered/exited.
I.e. `more than finished()` but less than `stage()`.  

I at least understand that this bakes in a guarantee that is a little bit weird, and not impossible to break if we start producing the same errors across different sections.  But I think the consequences of failure to uphold it are pretty minimal in this usage at least (potential unmerged duplicate errors being reported seperately when they could be merged).  But not really sure about other cases.


>I also wonder if a user who cares about these distinctions can already infer it from 
> the error kinds they see?]

In particular our errors are pretty opaque now with `kind` field being private, and no accessor function. 
So they would either have to parse text output, or map span information to the spans of sections (concrete syntax we don't keep anywhere in the AST).

The important thing that we *don't* get from mapping errors to section information though,
is an indication of progress *absent* errors. Let me just give a list of thoughts on use cases
I imagined for stages (and other random thoughts):

- Reporting progress in steps (possible with just start/finish events not revealing the actual stage if you use an indeterminate number).
- Providing stage information along with errors: "[Error] In the rules section: ..."
- Knowing when an error has *not* occurred, so you can flush buffers or stop listening for them.
- If we did associate span information to `Stage`, that is a way we could map errors to sections, without having to add
section spans to the `AST`, then figuring out the section of an error would be as easy as section spans and error spans to a interval tree.

ltratt:
@ratmice Sorry I'm so slow on this one. It's big enough that every time I sit down to think it through I run out of time / mental capacity! That's not your fault at all, but I'm making things worse by not responding promptly! I wonder if we can start breaking this down into chunks that an idiot like me can understand? Here are some concrete suggestions --- please feel free to tell me these are a bit off!

https://github.com/softdevteam/grmtools/pull/436/commits/04ba0c15bb5aad1e1d76b446dbfb8252cff3e92d (though obviously missing the `SimpleReport` struct ATM?) looks like a simple change we can happily PR against master?

https://github.com/softdevteam/grmtools/pull/436/commits/c7bc81fac4a97b1c46cbce5fc439cb32f22b6e3c is a little trickier, and probably couldn't be PRed as is, but it does suggest a path forward. Most obviously this is a part of grmtools which probably has very few direct users, so API breakage is acceptable, assuming it leads us to a better system.

The observation (I think) underlying c7bc81f is that we've got two completely stages, each of which can have different kinds of errors: parsing the input grammar into an AST; converting the AST to a more thoroughly checked format. [In retrospect "AST" was probably the wrong term here. It's really a sort of "semi-processed parse tree" or "unchecked AST". Perhaps the latter is the least-worst term?] Dumb question: *should* we be separating those two stages out? In other words, are there use cases where someone really needs the `UncheckedAST` but not the `YaccGrammar`? My gut feeling is that users don't really care about the `UncheckedAST` phase per se: they want to know about syntax errors in their input; and they want to know about the semantic errors; but I don't think they really want to separate those two things out so strictly that we have separate APIs for each. So, perhaps, we could merge these two things into a single `GrammarBuilder`? Obviously we'd then have to think about the error API, but that's step 2 :)

ratmice:
Do not worry about taking time, I want to apologize too because while I've tried, I haven't really managed to split it up nicely enough due to all the dependencies across the series.

04ba0c1 I intended to say more to the effect of "Convert ASTWithValidityInfo to use SimpleReport", 
`SimpleReport` is in abc5f0a which itself depends on 2244b6a The whole of  2244b6a .. 04ba0c1 probably has to be squashed into one patch (or two if we want to separate the existing code from the additions).

Let me try and quickly explain the original purpose of returning a "the valid partial fragment of an invalid AST", basically it allows us to fail to parse the grammar, continue parsing the lexer, and do checks like "token present in parser but missing from lexer", check on the lexer still. It was kind of a precursor to my `Analysis` branch which never managed to make it in.  I.e. you could run still an analysis because the AST only ever contains the part which was valid up-to the point of the error which caused it to be invalid.  So part of the thinking was using it as a way to move some of the checks which happen currently inside the main parse function out, or doing additional restrictions like "all tokens should be uppercase" or something.

So, it was part of an attempt to produce more errors within a single run even when the parser fails to build, but for cross-tool ones which depend upon the AST from which we can't produce a full lrtable.

ltratt:
Why don't we do a PR for the (probably-squashed) https://github.com/softdevteam/grmtools/commit/2244b6a6545172699b14f33779eb01601a44f07d .. https://github.com/softdevteam/grmtools/commit/04ba0c15bb5aad1e1d76b446dbfb8252cff3e92d as that seems largely uncontentious to me?

> Let me try and quickly explain the original purpose of returning a "the valid partial fragment of an invalid AST", basically it allows us to fail to parse the grammar, continue parsing the lexer, and do checks like "token present in parser but missing from lexer", check on the lexer still. It was kind of a precursor to my Analysis branch which never managed to make it in. I.e. you could run still an analysis because the AST only ever contains the part which was valid up-to the point of the error which caused it to be invalid. So part of the thinking was using it as a way to move some of the checks which happen currently inside the main parse function out, or doing additional restrictions like "all tokens should be uppercase" or something.

Right, but I *think* we could still give users the same kinds of errors (including those where we've attempted a sort-of error recovery) and only expose 1 stage as `pub` externally? Internally we could have as many stages as we want, of course! But I might be very wrong and have failed to understand the big picture.

cfgrammar/src/lib/yacc/ast.rs
ratmice: FWIW, I did do some experimenting here with the notion of a cross-tool driver:

Which tries to abstract the notion of lrpar/lrlex into a `Tool`, and define a abstraction boundary to run them,
managing source file loading, and error reporting (There are a bunch of renames there `Report` changed to `Diagnostics` and `ReportHander` to `DiagnosticsEmitter`,  then `finished()` to `no_more_data()`).

https://github.com/ratmice/driver/

When I try and ask myself "What problem is this trying to solve", the answer that I have is that it is trying to bridge the `CTParserBuilder` and `CTLexerBuilder` so that the stuff which is currently locked away behind within the `CTParserBuilder::build()` closure, can be used by `CTLexerBuilder`, e.g. in the `missing_from_lexer()` check which cannot happen until *after* the parser is built, we have access to the data about the symbol in the parser that is missing.

Overall the `Driver` implementation turned out a bit more complex than I had hoped, mostly so we aren't baking in choices that we don't have a mechanism to expand upon.  I'm not certain I was entirely successful at baking those choices in.  But overall, it *is* a different way to look at things than this entire series.

In general if we were to go this route `CTLexerBuilder` would own a parser driver, and be able to extract the `options` from the `CTParserBuilder` returned from `lrpar_config`.  Which it could then send to the `Driver` instead of calling `build`.

I really don't know if this is going to be fruitful, but I do think it highlights part of the issue here being abstraction boundaries, and passing mutable state as a way to avoid giving up ownership across those abstraction boundaries.

ratmice:
> Why don't we do a PR for the (probably-squashed) [2244b6a](https://github.com/softdevteam/grmtools/commit/2244b6a6545172699b14f33779eb01601a44f07d) .. [04ba0c1](https://github.com/softdevteam/grmtools/commit/04ba0c15bb5aad1e1d76b446dbfb8252cff3e92d) as that seems largely uncontentious to me?

Sure, I would be happy to do that, it would be good to have a look at driver though, it pulls some of these structure out and some stuff in `cfgrammar` like `Span`/`Spanned`, it's own crates and adds some things like `SourceId` `Error: SourceArtifact` where `SourceArtifact` (Artefact?) being a way to get a `SourceId` from an `Error`.

In other words it contains much of the stuff in *this patch* but pulls it out into it's own crate/interface that both lrlex/lrpar could depend upon.

It would be probably good to see if this has any legs before we continue with this series.

ratmice:
> Why don't we do a PR for the (probably-squashed) [2244b6a](https://github.com/softdevteam/grmtools/commit/2244b6a6545172699b14f33779eb01601a44f07d) .. [04ba0c1](https://github.com/softdevteam/grmtools/commit/04ba0c15bb5aad1e1d76b446dbfb8252cff3e92d) as that seems largely uncontentious to me?
> 
> > Let me try and quickly explain the original purpose of returning a "the valid partial fragment of an invalid AST", basically it allows us to fail to parse the grammar, continue parsing the lexer, and do checks like "token present in parser but missing from lexer", check on the lexer still. It was kind of a precursor to my Analysis branch which never managed to make it in. I.e. you could run still an analysis because the AST only ever contains the part which was valid up-to the point of the error which caused it to be invalid. So part of the thinking was using it as a way to move some of the checks which happen currently inside the main parse function out, or doing additional restrictions like "all tokens should be uppercase" or something.
> 
> Right, but I _think_ we could still give users the same kinds of errors (including those where we've attempted a sort-of error recovery) and only expose 1 stage as `pub` externally? Internally we could have as many stages as we want, of course! But I might be very wrong and have failed to understand the big picture.

The main difficulty here with making a 1 stage `pub` API here, with all this internal to it has been the cross-crate aspects of it, so by the time we have a *both* lex symbols, and the list of parse symbols from the return value, the AST is gone from dropping the closure that returned it, and the sources with it. I *think* the new `unstable_api` stuff might be able to help here, and even hide it from the docs.

ratmice:
I'm going close this, and continue experimenting, because
I think the driver experiment shows that even some of the things within the
2244b6a .. 04ba0c1 changes might be better suited to being in their own
library, where they can be used by both lrpar/lrlex.

But it also seems clear to me that the experiment is over complex, and I don't know
what it should ultimately look like. And I'd like to see whether making user visible
breaking changes (around lrpar_config closure) can reduce the reliance on mutable
state or if it's really inherent to the problem.

Since I guess feel now for the time being that I am not actually confident in any of this
or the other stuff ending up as something I would actually propose for merging.
It seems best for now to do that experimentation in a way that has the least impact for everybody.

PR #437, opened 2024-01-04
Prepare a 0.13.14 release.

ltratt:
@ratmice I'm not sure if I've adequately described the changes -- do they look sensible  to you?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+# grmtools 0.13.3 (2024-01-04)
+
+* Change generated code to avoid errors about `unsafe` action code in input
+  grammars.
+
+* Hide unstable parts of the API behind an `_unstable_api` feature that is off
+  by default.
+
+
 # grmtools 0.13.3 (2023-09-21)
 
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -1,3 +1,11 @@
+# grmtools 0.13.3 (2024-01-04)
+
+* Change generated code to avoid Clippy warnings about `unsafe`.
```
ratmice, line 3: > Change generated code to avoid Clippy warnings about `unsafe`.

It is actually a bit worse than that, because it is a rustc lint that was being turned into errors via `#[deny(rustc::unsafe_code)]`.
I would think something to the effect the following maybe?

* Change generated code to avoid errors about `unsafe` user code in yacc files.

ltratt: Good point! Fixed in 9d554dba84056354bab5314c2fec7df2a548c2ad.

ratmice:
Please squash.

ltratt:
Squashed.

PR #441, opened 2024-03-05
Clippy fixes.

ltratt:


ratmice:
Been a bit enamored with the [`cargo hack`](https://crates.io/crates/cargo-hack) tool lately, so ran this through `cargo hack --feature-powerset`, which came back relatively clean.

It did produce one warning:
```
info: running `cargo clippy --no-default-features` on lrtable (11/20)
warning: field `final_state` is never read
   --> lrtable/src/lib/statetable.rs:144:5
    |
133 | pub struct StateTable<StorageT> {
    |            ---------- field in this struct
...
144 |     final_state: StIdx<StorageT>,
    |     ^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

I *think* that field being unread looks real, but is perhaps being suppressed by reads from the `serde` feature?

ltratt:
> I think that field being unread looks real, but is perhaps being suppressed by reads from the serde feature?

Ah, yes, maybe! Shall we try removing it and see if it goes through CI?

ratmice:
It does actually look like it gets used in one test, but clippy can't see that because of conditional compilation of the test module.
https://github.com/softdevteam/grmtools/blob/b5abf0899535d5a2814e2f05e113230783dad923/lrtable/src/lib/statetable.rs#L645

It seems reasonable to want to keep testing that, I'm not sure how I feel about 
adding the field conditionally during testing, any thoughts?

```
+   #[cfg(test)]
     final_state: StIdx<StorageT>,
```

```
+            #[cfg(test)]
             final_state: final_state.unwrap(),
```

ltratt:
I'm very fine with that sort of conditional testing if you want to raise a PR for it!

PR #442, opened 2024-03-06
(lrtable) Make the final_state field conditional on `cfg(test)`

ratmice:
This conditionally includes a field into `struct StateTable`. Previously `clippy with --no-default-features -p lrtable` would produce a warning about `final_state` never being read.  This was presumably being suppressed by serde derive macros.  Since this field was only used in the test cfg.

It seemed better to make it conditional rather than `#[allow(dead_code)]`.


ltratt:
Thanks!

PR #443, opened 2024-03-22
fix code example in documentation

bendrissou:
Update errorrecovery.md: parse_int function should return Result type.


ltratt:
Thanks!

PR #444, opened 2024-04-18
Centralize crate versions and update `regex-syntax` to v0.8

rneswold:
This pull-request has two commit:

- one puts all the dependency versions in the workspace `Cargo.toml` file. Now you can update the version of a dependent crate in one spot and all sub-crates will use it.
- bump the `regex-syntax` crate to v0.8

Each of these commits was tested with `cargo test` and passed.


ltratt:
Thanks!

Issue #448, opened 2024-05-08
nimbleparse:Suggestions for outputting information


liluyue:
The nimbleparse outputting  info format maybe add "dot" format for Display in the form of a graph using [graphviz](https://graphviz.org/pdf/dotguide.pdf)

For example:

```dot

digraph table{
   label = "Stategraph" 
   node[shape="box"]
   0[label=" 0\n ^ -> . Expr"]
   1[label="1\nTerm -> Term . '/' Factor"]
   2[label="2\nFactor -> '(' . Expr ')'"]
   0 -> 1 [label="Term",decorate=false]
   0 -> 2 [label = "(",decorate =false]
   2 -> 1 [label="Term"]
}

```

ratmice:
This is something i've been curious about doing, but hasn't been a priority for me.

Somewhat related, nimbleparse_lsp can generate railroad diagrams, and display them within vscode,
but there isn't any utility for writing that output to a file.  It generates svg directly (using the railroad crate) rather than relying on an external tool like "dot" to avoid invoking binaries from within the editor, or bringing in the entire libgraphviz image generation stack, that code entails an additional dependency somewhere.  [railroad diagram generation code](https://github.com/ratmice/nimbleparse_lsp/blob/main/server/src/parse_thread.rs#L942).

In theory we wouldn't need the additional dependency if we're just generating "dot" formatted text output though.
But if it is something Laurence does want to include i'd be happy to put some effort towards it.

liluyue:
I got the **nimbleparse_lsp**, should put it in the [document](https://softdevteam.github.io/grmtools/master/book/nimbleparse.html), otherwise it's difficult to realize its existence

ltratt:
I like the idea of having a "grammar -> <output>" tool. I vaguely envisaged nimbleparse as something very simple, but I guess it could grow things like dot output without imposing any complexity on other cases? If so, I'd be all for it!

ratmice:
For railroad diagrams, it would basically be just option parsing affecting the main program, with an additional function where the added complexity can be isolated within it.  I don't think *dot* support would differ from that.

I don't think it is certain that nimbleparse is the right tool/place for this feature though.
The primary thing is nimbleparse *requires* input text, and a lex file, I'm uncertain if the lex file is absolutely necessary (it could perhaps be used to display regex beyond just symbols).

But for outputting diagrams actual input text seems like it should be entirely ignored/superfluous.

[...]

PR #451, opened 2024-05-13
Add a doc section for third party tools

ratmice:
It was recommended in #448 to add a link to the book about nimbleparse_lsp.
I'm not really certain it is a good idea to advertise it yet, as it has fallen behind by a few releases and overall is still fairly
prototype quality.  

Despite that, this adds "third party" section in chapter 6 of "individual libraries and tools".
I had hoped to just link directly to the repository, relying on the external readme.

Unfortunately trying to link to a `[lsp](https://)` within `SUMMARY.md` appends a `.html` extension breaking the URL.
Between adding a subsection, or just embedding a list within the `thirdparty.md` I went with a subsection,
since it more closely mirrored how the current list of tools and libraries is listed.

### DIFF
--- a/doc/src/SUMMARY.md
+++ b/doc/src/SUMMARY.md
@@ -18,4 +18,5 @@
   - [nimbleparse](nimbleparse.md)
   - [cfgrammar](cfgrammar.md)
   - [lrtable](lrtable.md)
+  - [third party](thirdparty.md)
 - [Other Rust parsing tools](othertools.md)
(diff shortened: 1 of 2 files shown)


ltratt:
Can we fold this into a single `thirdparty.md` file? I don't think we need a page per tool, but I agree that linking to nimbleparse_lsp from the book is a good idea.

ratmice:
Sure, I went ahead and did that.  It means we don't get a link directly in the table of contents, but it seems more maintainable to just link directly to each project anyways instead of relying on potentially dated generic summary.

ltratt:
Thanks! Please squash.

ratmice:
Squashed (with a minor grammar fix).

Issue #450, opened 2024-05-13
lrpar:%prec not followed by token name


liluyue:
I use **%nonassoc**,always error
<img width="1179" alt="截屏2024-05-13 14 48 55" src="https://github.com/softdevteam/grmtools/assets/19216288/1ccea5d7-a5eb-4e65-80c9-9202ddbd8846">

ltratt:
If you can include the raw text of the input and the error that's produced, that would be helpful.

liluyue:
> If you can include the raw text of the input and the error that's produced, that would be helpful.
Priority added to the example:
```
%start Expr
%left '+' '-'
%left '*' '/'
%nonassoc 'u'
%%
Expr -> Result<u64, ()>:
      Expr '+' Expr { Ok($1? + $3?) }
    | Expr '-' Expr { Ok($1? - $3?) }
    |  Expr '*' Expr { Ok($1? * $3?) }
    | Expr '/' Expr { Ok($1? / $3?) }
    | Factor { $1 }
    | '-' Factor %prec 'u' { Ok(-$2?) }
    ;


Factor -> Result<u64, ()>:
      '(' Expr ')' { $2 }
    | 'INT'
      {
          let v = $1.map_err(|_| ())?;
          parse_int($lexer.span_str(v.span()))
      }
    ;
%%
// Any functions here are in scope for all the grammar actions above.

fn parse_int(s: &str) -> Result<u64, ()> {
    match s.parse::<u64>() {
        Ok(val) => Ok(val),
        Err(_) => {
            eprintln!("{} cannot be represented as a u64", s);
            Err(())
        }
    }
}

```

ltratt:
Can you also include the error you're seeing please?

liluyue:
```
[Error] in src/calc.y
    12|     | '-' Factor %prec 'u' { Ok(-$2?) }
                                %prec not followed by token name
    
    11|     | Factor { $1 }
              ^^^^^^ Unknown reference to rule 'Factor'
    
```

ltratt:
That *probably* means that your lexer isn't defining `u` is a token type? [It could be a bug in cfgrammar of course, but it has a check `if self.ast.tokens.contains(&sym)` that looks sensible to me, at least at first glance.]

liluyue:
It can be troublesome to resolve the shift/reduce between rules without **%nonassoc“. According to yacc's rules,**% nonassoc** should not be bound to a token

ltratt:
[...]

PR #453, opened 2024-05-15
Allow `%prec` to define a token in the grammar.

ltratt:
Previously if `%prec 'x'` was the first mention of the token `x` then cfgrammar would say "'x' isn't a token" which is clearly incorrect. This commit is maybe best thought of as a quick fix of sorts: it makes `%prec 'x'` define a token `x` if it doesn't already exist.

However -- and my memory is completely fuzzy as to whether this is a Yacc compatibility thing or not -- this means that `%prec 'x'` and `%prec x' (i.e. with and without quote marks) have the semantic meaning because `parse_token` treats both interchangeably. This seems consistent with other parts of the grammar even if a little odd.

I believe this fixes https://github.com/softdevteam/grmtools/issues/450.

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -677,12 +677,11 @@ impl YaccParser {
                 syms.push(Symbol::Token(sym, span));
             } else if let Some(j) = self.lookahead_is("%prec", i) {
                 i = self.parse_ws(j, true)?;
-                let (k, sym, _) = self.parse_token(i)?;
-                if self.ast.tokens.contains(&sym) {
-                    prec = Some(sym);
-                } else {
-                    return Err(self.mk_error(YaccGrammarErrorKind::PrecNotFollowedByToken, i));
+                let (k, sym, span) = self.parse_token(i)?;
+                if self.ast.tokens.insert(sym.clone()) {
+                    self.ast.spans.push(span);
                 }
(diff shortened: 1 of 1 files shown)


ratmice:
I should be honest and say I'm not really familiar with `%prec` or any of the associativity declarations, as I tend towards manually resolving ambiguity even if doing so it isn't terribly convenient.

This seems okay for a quick fix though, in that it seems low impact and isolated to `%prec`, which specifically does not affect `parse_token` which I would be more concerned about.

One other part of grmtools (albeit there isn't an equivalent in yacc) where quotes have semantic meaning and aren't interchangeable is `%expect-unused`, which first tries to parse it as a *name* then failing that tries to parse it as a *token*, which in that path means generally unquoted tokens will be treated as a *name*.

Anyhow i'm happy to approve if you think it is ready.

ltratt:
>  One other part of grmtools (albeit there isn't an equivalent in yacc) where quotes have semantic meaning and aren't interchangeable is %expect-unused, which first tries to parse it as a name then failing that tries to parse it as a token, which in that path means generally unquoted tokens will be treated as a name.

Good point. I wonder if there should be another PR which a) checks exactly what yacc does in these cases (I vaguely remember it allows either case, but I might well be wrong) b) checks the code for precise compatibility with yacc? I'm not sure I'll get to that very soon, so at least this PR nudges us in a better-than-the-status-quo direction...

PR #456, opened 2024-05-26
Tell the user what rule is missing action code.

ltratt:
Before this commit a production with missing action code such as:

```
R -> <...>:
    "a" { todo!() }
  | "b"
  ;
```

would panic on an `unwrap`. Hunting down the offender could be fun!

This commit is a minimal fix for this: it tells the user that rule `R` has a missing production. It would be better to pinpoint the exact production, but at this point in lrpar, I think we've lost the necessary information to do so. On the basis that something is better than nothing, however, this commit still seems somewhat useful.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -679,7 +679,7 @@ where
         if let Some(YaccKind::Original(YaccOriginalActionKind::UserAction) | YaccKind::Grmtools) =
             self.yacckind
         {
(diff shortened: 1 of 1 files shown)


ratmice:
Looks like a nice low-impact fix, agree it is an improvement.
Normally I would probably spend a bit of time poking around this information loss,
but I can't currently commit to that in any reasonable timeframe unfortunately :sneezing_face:.

ltratt:
Totally understood! If you do find time at some point, of course, I'd be super grateful!

PR #457, opened 2024-05-27
Warn the user where an incomplete action started, not finished.

ltratt:
Previously this warning always said "Incomplete action at <end of file>". This change now tells you where the incomplete action started, which is much easier to debug!


ratmice:
Nice catch!

PR #466, opened 2024-08-22
Allow Stdin to be used as input file for lrlex/nimbleparse

ratmice:
Working a small project where I wanted to run lrlex over some compressed data, it seemed like it would be nice
to do so without having to decompress using a temporary file...  This implements it so you can do the the following
```
zcat compressed_stuff| lrlex foo.l -
``` 

Or a nimbleparse example that should work in the repository (echo -n to avoid newline)

```
echo -n "1 + 1" | target/release/nimbleparse lrpar/examples/calc_parsetree/src/calc.{l,y} -
```

### DIFF
--- a/lrlex/src/main.rs
+++ b/lrlex/src/main.rs
@@ -2,7 +2,7 @@ use getopts::Options;
 use std::{
     env,
     fs::File,
-    io::{stderr, Read, Write},
+    io::{stderr, stdin, Read, Write},
     path::Path,
     process,
     str::FromStr,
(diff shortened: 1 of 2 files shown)


lrlex/src/main.rs
```diff
@@ -26,14 +26,18 @@ fn usage(prog: &str, msg: &str) {
 }
 
 fn read_file(path: &str) -> String {
```
ratmice, line 28: It might be worth noting here, that the `read_file` functions between lrlex and nimbleparse,
seem to have diverged in their argument types:

`lrlex` takes an `&str`, which `nimbleparse` takes an `AsRef<Path>`, and the difference effects these two implementations because there is no `Path*` implementation for `PartialEq<&str>`.

This explains why in lrlex I implemented the logic *inside* `read_file`, while in nimbleparse the logic ended up outside.

ltratt: I'd be happy to see the two homogenised in a future PR!

ltratt:
Thanks -- this is good!

ratmice:
Sorry, looks like I failed to rustfmt

ltratt:
Please squash.

ratmice:
Squashed.

PR #470, opened 2024-09-30
Update to latest clippy and compiler

ratmice:
This updates clippy to the latest revision and changes 2024 language edition warning that `proc_macro` will be deprecated in favor of `proc-macro`, The text of the latter warning is:

> `proc_macro` is deprecated in favor of `proc-macro` and will not work in the 2024 edition
(in the `cttests_macro` library target)


ltratt:
Thanks!

PR #474, opened 2024-11-24
nimbleparse: add -d flag to dump state graph

ratmice:
This is a patch I used for testing https://github.com/softdevteam/sparsevec/pull/25
It adds a `-d` flag for dumping the state graph from nimbleparse, then exiting.
It also fixes an issue where the pretty printing function produces non-deterministic output.

I'm not certain it is quite up to par, because of the argument parsing an input file is required even though it is never used.
but making the input file 'required unless the -d flag is specified' is also weird.  Passing in an unused filename like /dev/null was sufficient for my testing purposes though.

### DIFF
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -32,7 +32,7 @@ fn usage(prog: &str, msg: &str) -> ! {
     if !msg.is_empty() {
         eprintln!("{}", msg);
     }
-    eprintln!("Usage: {} [-r <cpctplus|none>] [-y <eco|grmtools|original>] [-q] <lexer.l> <parser.y> <input file>", leaf);
+    eprintln!("Usage: {} [-r <cpctplus|none>] [-y <eco|grmtools|original>] [-dq] <lexer.l> <parser.y> <input file>", leaf);
     process::exit(1);
 }
 
@@ -87,6 +87,7 @@ fn main() {
     let matches = match Options::new()
         .optflag("h", "help", "")
         .optflag("q", "quiet", "Don't print warnings such as conflicts")
+        .optflag("d", "dump-state-graph", "Print the parser state graph")
         .optopt(
             "r",
             "recoverer",
@@ -109,6 +110,7 @@ fn main() {
         usage(prog, "");
     }
 
+    let dump_state_graph = matches.opt_present("d");
     let quiet = matches.opt_present("q");
 
     let recoverykind = match matches.opt_str("r") {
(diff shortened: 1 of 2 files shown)


nimbleparse/src/main.rs
```diff
@@ -187,6 +189,11 @@ fn main() {
         }
     };
 
+    if dump_state_graph {
+        println!("Stategraph:\n{}\n", sgraph.pp_core_states(&grm));
+        process::exit(0);
```
ltratt, line 194: My suggestion is that we don't `exit` here: we print out the stategraph and continue. That simplifies the argument parsing in the sense that users *always* have to provide `.l`, `.y`, and input files. We could be more clever, but documenting that in a digestible way becomes a pain, and I think the simple approach is adequate.

nimbleparse/src/main.rs
```diff
@@ -32,7 +32,7 @@ fn usage(prog: &str, msg: &str) -> ! {
     if !msg.is_empty() {
         eprintln!("{}", msg);
     }
-    eprintln!("Usage: {} [-r <cpctplus|none>] [-y <eco|grmtools|original>] [-q] <lexer.l> <parser.y> <input file>", leaf);
+    eprintln!("Usage: {} [-r <cpctplus|none>] [-y <eco|grmtools|original>] [-q] [-d] <lexer.l> <parser.y> <input file>", leaf);
```
ltratt, line 35: In ye olde unix convention, this would be `[-dq]`.

nimbleparse/src/main.rs
ratmice: Yeah the problem with my testing was that I didn't actually have valid input for any of the github provided parsers I was testing, while it is likely that /dev/null would be a valid for most perhaps even all of them, it was easier to just avoid complications with input/AST production.

Perhaps that is a special case that is no longer necessary, and I suppose it is easy to add the `exit` call if I need something like that again?

ratmice: That could be good, there were a couple of actually the most heavily used libraries depending upon grmtools
which I couldn't test with this approach because they used a manually implemented lexer.

ltratt: I think this PR is useful (without the `exit`) and worth keeping, but an additional `lrtable` binary PR would be cool too!

ratmice: Should be fixed in 2a13f24

There was a little more to it, also needed to avoid duplicate printing of the state graph when there are conflicts.

nimbleparse/src/main.rs
ratmice: Should be fixed in 2a13f24 too.

ltratt:
Please squash.

ratmice:
Squashed.

Issue #477, opened 2024-12-19
Make `Rule::re_str` public again.


pmfirestone:
In commit [c1992b](https://github.com/softdevteam/grmtools/commit/c1992b0f91ca5aad01bb5ff2bd25bdf5ae56f4b1), @ltratt depubbed a number of functions and fields. The commit message includes the following justification:

> grmtools intentionally exposes some details that no-one currently uses, but they might one day

I think that day has arrived. I am implementing an incremental parsing algorithm, from [this paper](https://arxiv.org/abs/2403.01632), page 34. This is part of my project to reimplement the [existing Python version](https://github.com/uiuc-focal-lab/syncode) of the Syncode algorithm in Rust. @shubhamugare is the boss of the project, but I'm working on the Rust version.

In order to integrate cleanly with other logic in the implementation so far, I need to be able to convert from a `Lexeme` struct back to a regular expression (this happens in lines 17, 18, and 21 of Algorithm 4 of the linked paper): I then do some crunching on these regexes by turning them into DFAs and advancing one state at a time; this requires the regex representations of the lexemes in the input (cf. page 10: the algorithm simply assumes that this transformation is trivially possible, and in the current Python implementation, it is).

I worry that this might be an XY problem: I perhaps I could manually track which regexes go with which `TIdx`s. However, it seems to me that making the `re_str` field of the `Rule` struct `pub` again (instead of `pub(super)` as it is now) would solve the problem for me much more gracefully than introducing such logic into my program. I will already have to use [`LexerDef::set_rule_ids`](https://docs.rs/lrlex/latest/lrlex/trait.LexerDef.html#tymethod.set_rule_ids) to synchronize the rules' ids between the parser and the lexer, but this still doesn't allow me direct access to the underlying regexes. I also don't believe this can be gotten out of the `cfgrammar` crate, but maybe an accessor function can be added to the `YaccGrammar` struct to match `token_epp`, `token_name`, and `token_precedence`.

This is a somewhat unusual use case, I admit, but I believe that the particular application we are developing justifies considering the change. On the other hand, do you have any alternative suggestions for getting the regex back from a `TIdx`? It's very possible I've overlooked something! Thanks for the hard work and have a great day :smiley:.

ratmice:
[...]

PR #478, opened 2024-12-21
Public functions for `Rule`'s internals

ltratt:
This PR first fixes https://github.com/softdevteam/grmtools/issues/477 and then goes further and gives the other currently-`pub` fields accessor functions, and marks the fields as deprecated. This will allow us -- one day! -- to remove `pub` from the fields. Right now, this change doesn't require a major release.

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -58,6 +58,7 @@ pub struct Rule<StorageT> {
     pub(super) tok_id: Option<StorageT>,
     /// This rule's name. If None, then text which matches this rule will be skipped (i.e. will not
     /// create a lexeme).
+    #[deprecated(note = "Use the name() function")]
     pub name: Option<String>,
     pub name_span: Span,
     pub(super) re_str: String,
@@ -116,6 +117,7 @@ impl<StorageT: PrimInt> Rule<StorageT> {
         }
 
         let re = re.build()?;
+        #[allow(deprecated)]
         Ok(Rule {
             tok_id,
             name,
@@ -138,6 +140,7 @@ impl<StorageT: PrimInt> Rule<StorageT> {
     /// Return this rule's name. If `None`, then text which matches this rule will be skipped (i.e.
     /// it will not result in the creation of a [Lexeme]).
     pub fn name(&self) -> Option<&str> {
+        #[allow(deprecated)]
         self.name.as_deref()
     }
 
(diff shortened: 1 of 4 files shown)


lrlex/src/lib/lexer.rs
```diff
@@ -125,6 +126,19 @@ impl<StorageT> Rule<StorageT> {
             target_state,
         })
     }
+
+    /// Return this rule's token ID, if any.
+    ///
+    /// If `Some`, this specifies the ID that lexemes resulting from this rule will have. If
+    /// `None`, then this rule specifies lexemes which should not appear in the user's input.
+    pub fn tok_id(&self) -> Option<StorageT> {
+        self.tok_id
```
ratmice, line 140: It is possible I'm missing something obvious, but I'm wondering if there is a reason that this doesn't return a `cfgrammar::TIdx` or some such?

ltratt: I also wondered that, but it's currently storing a `StorageT`. I can't fully remember what I might expect this to be, to be honest...

ratmice: Looks like it is also exposed publicly as `StorageT` via the [`lrpar::Lexeme`](https://docs.rs/lrpar/latest/lrpar/trait.Lexeme.html#tymethod.tok_id) trait [here](https://github.com/softdevteam/grmtools/blob/master/lrlex/src/lib/lexer.rs#L396).  So it doesn't seem to be unprecedented.

ratmice:
Besides that one question I had (which it seems we can disregard) it looked good to me.
I didn't notice anything that appeared to need squashing, but if you would like to, please do.

lrlex/src/lib/lexer.rs
ltratt: Would it be better as `TIdx`? If so, we might as well make the API as unsucky as we can reasonably make it!

ratmice: Looking further into it, I don't think `TIdx` would actually the right thing it should correspond to.  For example, [this](https://github.com/softdevteam/grmtools/blob/master/cfgrammar/src/lib/yacc/grammar.rs#L488-L492) usage in `token_name` isn't referring to one of these `tok_id` values at all, but an index into a `Grammar` specific array.

So, I'm leaning towards it being okay as is, at the very least nothing is ringing any bells for it being any of the other opaque type wrappers, and it certainly doesn't appear that `TIdx` would be right.

ltratt: Excellent detective work! OK I feel adequately comfortable with this PR now then.

PR #483, opened 2025-01-27
Fix backspace lexeme escaping

414owen:
In posix lex, '\b' represents the backspace character. In the rust 'regex' crate, it represents a word boundary assertion.

This patch adds a test that all posix escapes are interpreted correctly, and a fix for the backspace escape incongruity.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -443,6 +443,7 @@ pub fn lexerdef() -> {lexerdef_type} {{
             dot_matches_new_line: {dot_matches_new_line:?},
             multi_line: {multi_line:?},
             octal: {octal:?},
+            posix_escapes: {posix_escapes:?},
             case_insensitive: {case_insensitive:?},
             unicode: {unicode:?},
             swap_greed: {swap_greed:?},
@@ -454,6 +455,7 @@ pub fn lexerdef() -> {lexerdef_type} {{
             dot_matches_new_line = self.regex_options.dot_matches_new_line,
             multi_line = self.regex_options.multi_line,
             octal = self.regex_options.octal,
+            posix_escapes = self.regex_options.posix_escapes,
             case_insensitive = self.regex_options.case_insensitive,
             unicode = self.regex_options.unicode,
             swap_greed = self.regex_options.swap_greed,
@@ -665,6 +667,13 @@ pub fn lexerdef() -> {lexerdef_type} {{
     /// The default value is `true`.
     pub fn multi_line(mut self, flag: bool) -> Self {
         self.regex_options.multi_line = flag;
+        self
+    }
+
+    /// Sets the `regex::RegexBuilder` option of the same name.
+    /// The default value is `false`.
+    pub fn posix_escapes(mut self, flag: bool) -> Self {
+        self.regex_options.posix_escapes = flag;
         self
     }
 
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -25,6 +25,7 @@ pub struct RegexOptions {
     pub dot_matches_new_line: bool,
     pub multi_line: bool,
     pub octal: bool,
+    pub posix_escapes: bool,
     pub case_insensitive: Option<bool>,
     pub swap_greed: Option<bool>,
     pub ignore_whitespace: Option<bool>,
@@ -38,6 +39,7 @@ pub const DEFAULT_REGEX_OPTIONS: RegexOptions = RegexOptions {
     dot_matches_new_line: true,
     multi_line: true,
     octal: true,
+    posix_escapes: false,
     case_insensitive: None,
     ignore_whitespace: None,
     swap_greed: None,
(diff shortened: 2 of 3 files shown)


ratmice:
I guess I have two questions,

Whether people can currently use/rely on the `\b` as word boundary in regex, (i.e. does the \b as word boundary escape sequence code operate correctly such that people could be using it)?

Similarly, when we make `\b` as backspace, currently people who want backspace can match the hex code manually.
But if people want word boundary, there is no hex code for that, (i.e. regardless of whether it currently works, should it be something configurable that we should allow to be used).  

I'm actually a bit surprised that this isn't something configurable by [RegexBuilder](https://docs.rs/regex/latest/regex/struct.RegexBuilder.html)  but perhaps it would be nice if we could just add a flag to [CTLexerBuilder](https://docs.rs/lrlex/latest/lrlex/struct.CTLexerBuilder.html) such as `fn escape_b_as_backspace(...)` and then test that flag in the character replacement code you added?  At least I believe that should work correctly in the fall through case once added to the `RE_LEX_ESC_LITERAL`.

I'm not certain if we actually have a `CTLexerBuilder` at that point so we can access it's fields, or whether making that possible is going to be too painful a task to be worthwhile though.

ltratt:
This PR is getting at a fairly deep point: because I was lazy, lrlex uses the `regex` library and hopes for the best in terms of "it does what you expect" for people who aren't aware of that. That means that there are incompatibilities, and seemingly random extensions, from POSiX Lex. I expected more complaints about this by now, but I suppose *most* lex files only use a small subset of regular expressions (certainly that's the case for me).

I expected that ultimately we would implement a full Lex-compatible matcher. But I'm unlikely to get to any year/decade soon, and we now have a fairly substantial userbase who, whether they intended to or not, are probably now relying on regex compatibility. I would be surprised if someone isn't using `\b` for "word boundary" already. That means that changing this behaviour will cause surprises for existing users. My laziness has come back to bite us!

Arguably what we should do is allow lrlex to support different flavours of regexes. lrpar does this tolerably well with https://docs.rs/lrpar/0.13.8/lrpar/struct.CTParserBuilder.html#method.yacckind. lrlex should probably grow `lexkind` and that could take `PosixLex` or `RustRegex` (etc) as options, or perhaps more honestly `PosixIshLexIsh` if we take the "restrict rustregex approach"!

[Note: lrlex has a `lexerkind` method but that's really quite different, currently meaningless (the accompanying enum has one variant, and arguably not a very good name. For the next major release we should probably remove this method until we work out what it, or some variant, should really be.]

So summary: we can't merge this PR as-is for backwards compatibility reasons, *but* I would love to merge a variant of it! It would need to do something like:

1. Add a `lexkind` method to `CTLexerBuilder`.
2. Add an enum similar to https://docs.rs/cfgrammar/0.13.8/cfgrammar/yacc/enum.YaccKind.html. The default we use should, for backwards compatibility, be `LexKind::RustRegex` but we should allow `LexKind::PosixLex` (or similar).
3. Use information from that enum when parsing the lex file to do different things (similarly to https://docs.rs/cfgrammar/0.13.8/src/cfgrammar/yacc/parser.rs.html#625).
4. [Optional but nice] Allow users of `nimbleparse` to specify the lex variant they're using.

I know that being asked to do more work is never fun, so I apologise for putting up barriers -- I would like to think, though, that this isn't *too* much work, and it would make & keep lots of people happy!

ratmice:
> 4\. [Optional but nice] Allow users of `nimbleparse` to specify the lex variant they're using.

I hadn't really thought of nimbleparse in my earlier reply, it seems like it is also already an issue with the `lrlex::lexer::RegexOptions` settings I had mentioned.

`nimbleparse` will always use the default `dot_matches_newline` behavior (true), there is the fn [`LrNonStreamingLexerDef::new_with_options`](https://docs.rs/lrlex/latest/lrlex/struct.LRNonStreamingLexerDef.html#method.new_with_options) which takes a `RegexOptions` param,
but nimbleparse will always use the default parameters.

I had originally added these for similar reasons as are given by this patch because the default `dot_matches_newline` behavior doesn't match posix, but (currently) there is no code to alter the setting from nimbleparse, though there is some in the cttests code as can be seen in the lex_flags field and it's implementation though.

* https://github.com/softdevteam/grmtools/blob/6967d507e4142a7486c116f71d64a0fc0dd140e5/lrpar/cttests/src/regex_opt.test#L3
* https://github.com/softdevteam/grmtools/blob/6967d507e4142a7486c116f71d64a0fc0dd140e5/lrpar/cttests/src/cgen_helper.rs#L113-L139

So perhaps those code snippets should also make their way towards nimbleparse?
Anyhow apologies if we are giving mixed signals towards how to approach solving the problem.

ltratt:
@ratmice On reflection, you're right: I've muddled two things up. Mea culpa!

Perhaps the right thing is to make this PR use the existing `RegexOptions` facility and then think, in another PR, about how we could do the "bigger" `LexKind` change I'm suggesting. Another way of looking at this is that we could bundle up `RegexOptions` into `LexKind::RustRegex` rather than making it a first-class citizen of lrlex.

ltratt:
@414owen Did our discussion make any sense? I *think* we're proposing a route with a relatively small change to this PR. [@ratmice can correct me if I'm wrong!]

414owen:
@ltratt Sure thing. Introducing this in a non-breaking way makes sense, even though using word break assertions in a lexer is of dubious merit.

I've added the option, defaulted it to off, and added a test case that it's respected. Let me know if I can help any further :)

ltratt:
@414owen This looks great -- thanks! If you can address @ratmice's lifetime suggestion + squash everything down to one commit, we'll be ready to merge!

ltratt:
Thanks!

PR #485, opened 2025-02-04
Prepare a 0.13.9 release.

ltratt:


ratmice:
LGTM

ratmice:
I assume this is just the ci having updated one of rustdoc/clippy since the commits landed, updating to see if I can reproduce locally.

ltratt:
Weirdly that error only seems to crop up with `rustdoc`. c2f6221 should fix it.

ratmice:
Curious I see now I needed to download the build logs manually, as it doesn't scroll up/show the entire log.

Anyhow trying that failing command `RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps` locally, still can't reproduce though that seems to obviously be the solution.

ltratt:
Buildbot doesn't make the logs as obvious as one might hope, but there is a button to display all of them in the browser (including while they're ongoing). CI is hard! But thanks for looking at this -- I really appreciate it!

Issue #446, opened 2024-05-04
lrpar  lose  info about debug


liluyue:
I execute "cargo run"，get info :
```
  called `Result::unwrap()` on an `Err` value: CTConflictsError{0 Reduce/Reduce, 2 Shift/Reduce}
```

It is too vague，I wish can get below info for I fix the issue:

```
sequence.y: warning: 1 shift/reduce conflict [-Wconflicts-sr]
sequence.y: warning: 2 reduce/reduce conflicts [-Wconflicts-rr]
sequence.y: warning: shift/reduce conflict on token "word" [-Wcounterexamples]
  Example: • "word"
  Shift derivation
    sequence
    ↳ 2: maybeword
          ↳ 5: • "word"
  Example: • "word"
  Reduce derivation
    sequence
    ↳ 3: sequence "word"
          ↳ 1: •
```

ratmice:
The most user friendly errors that we currently have can be seen by running the `nimbleparse` tool (unless using a hand-written lexer)... But it has been a bit difficult trying to find the right abstraction boundary to be able to leverage this from within the `CTParserBuilder` used by *cargo*.  However that still doesn't have a counterexample generator, but should at least show a bit more info about source locations of the conflicts.

liluyue:
I doubt   the reliability of  the nimbleparse. I use  test the nimbleparse by  calc exmaple  :

```
markdown_parser git:(master) ✗ nimbleparse src/calc.l src/calc.y a.txt 
src/calc.y: [Error] Missing ':' at line 3 column 6
```

but calc actually is not bad:

```
%start Expr
%%
Expr -> Result<u64, ()>:
      Expr '+' Term { Ok($1? + $3?) }
    | Expr '-' Term { Ok($1? - $3?) }
    | Term { $1 }
    ;

```

so,I hope can find the state table  for analysis the problem

ratmice:
It looks like you need to pass the `-y grmtools` flag to nimbleparse, in this case to set the yacc format.
Because the original format isn't expecting the type `-> Result<u64, ()>`

liluyue:
that is valid.Thanks!

ratmice:
I also notice that the output of nimbleparse in the above error that you quoted doesn't include inline source code.
[...]

PR #491, opened 2025-02-22
Add a `YaccKind::SelfDescribing` style.

### LINKED ISSUE #349: Adding a %grammar-kind declaration?

ratmice:
I belive this should implement what we discussed in #349 for lrpar (It doesn't try to tackle lrlex yet)
I'm still not certain I like the syntax but it seemed not worth worrying about until I got something working at least.

---

This allows the yacckind value to be specified within a grammar. When using the `YaccKind::SelfDescribing` style the grammar must start with an initial `%grmtools` section, as documented in the variant.

For other `YaccKind` variants the `%grmtools` section is allowed (not required) and the `YaccKind` value given to `YaccGrammar::new` overrides one given in the `%grmtools` section.

Fixes #349

### DIFF
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -2815,7 +2815,11 @@ B";
         ];
         for yacc_src in srcs {
             parse(YaccKind::SelfDescribing(None), yacc_src).unwrap();
-            parse(YaccKind::SelfDescribing(Some(Box::new(YaccKind::SelfDescribing(None)))), yacc_src).unwrap();
+            parse(
+                YaccKind::SelfDescribing(Some(Box::new(YaccKind::SelfDescribing(None)))),
+                yacc_src,
+            )
+            .unwrap();
         }
 
         let fallback_srcs = [
@@ -2913,11 +2917,10 @@ B";
             1,
             1,
         );
-        parse(YaccKind::SelfDescribing(Some(Box::new(YaccKind::SelfDescribing(None)))), src).expect_error_at_line_col(
+        parse(
+            YaccKind::SelfDescribing(Some(Box::new(YaccKind::SelfDescribing(None)))),
             src,
-            YaccGrammarErrorKind::InvalidYaccKind,
-            1,
-            1,
-        );
+        )
+        .expect_error_at_line_col(src, YaccGrammarErrorKind::InvalidYaccKind, 1, 1);
     }
 }
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -2779,38 +2779,38 @@ B";
     fn test_self_describing_yacckind() {
         let srcs = [
             r#"%grmtools {
-  yacckind Original(NoAction)
-}
-%%
-S: "()";"#,
+                    yacckind Original(NoAction)
+               }
+               %%
+               S: "()";"#,
             r#"%grmtools { yacckind Original(NoAction) }
-%%
-S: "()";"#,
+               %%
+               S: "()";"#,
             r#"%grmtools {yacckind Original(NoAction)}
-%%
-S: "()";"#,
+               %%
+               S: "()";"#,
             r#"%grmtools
                {yacckind Original(NoAction)}
-%%
-S: "()";"#,
+               %%
+               S: "()";"#,
             r#"
-%grmtools {
-  yacckind Original(UserAction)
-}
-%%
-S: "()";"#,
+                %grmtools {
+                    yacckind Original(UserAction)
+                }
+                %%
+                S: "()";"#,
             r#"%grmtools {
-  yacckind Grmtools
-}
-%start S
-%%
-S -> (): "()" { () };"#,
+                    yacckind Grmtools
+               }
+               %start S
+               %%
+               S -> (): "()" { () };"#,
             r#"%grmtools {
-  yacckind grmtools
-}
-%start S
-%%
-S -> (): "()" { () };"#,
+                    yacckind grmtools
+               }
+               %start S
+               %%
+               S -> (): "()" { () };"#,
         ];
         for yacc_src in srcs {
             parse(YaccKind::SelfDescribing(None), yacc_src).unwrap();
@@ -2840,46 +2840,46 @@ S -> (): "()" { () };"#,
     #[test]
     fn test_self_describing_yacckind_errs() {
         let src = r#"%grmtools {
-invalid value
-}
-%%
-S: "()";"#;
+                                invalid value
+                           }
+                           %%
+                           S: "()";"#;
         parse(YaccKind::SelfDescribing(None), src).expect_error_at_line_col(
             src,
             YaccGrammarErrorKind::InvalidGrmtoolsHeaderKey,
             2,
-            1,
+            33,
         );
 
         let src = r#"%grmtools {
-yacckind Grmtools
-yacckind Grmtools
-}
-%%
-S: "()";"#;
+                                yacckind Grmtools
+                                yacckind Grmtools
+                           }
+                           %%
+                           S: "()";"#;
         parse(YaccKind::SelfDescribing(None), src).expect_error_at_lines_cols(
             src,
             YaccGrammarErrorKind::DuplicateGrmtoolsHeaderKey,
-            &mut [(2, 1), (3, 1)].into_iter(),
+            &mut [(2, 33), (3, 33)].into_iter(),
         );
 
         let src = r#"%grmtools {
-yacckind invalid_yacc_kind
-}
-%%
-S: "()";"#;
+                                yacckind invalid_yacc_kind
+                           }
+                           %%
+                           S: "()";"#;
         parse(YaccKind::SelfDescribing(None), src).expect_error_at_line_col(
             src,
             YaccGrammarErrorKind::InvalidYaccKind,
             2,
-            10,
+            42,
         );
         // Invalid brace after %grmtools [
         let src = r#"%grmtools [
-yacckind Grmtools
-}
-%%
-S: "()";"#;
+                                yacckind Grmtools
+                           }
+                           %%
+                           S: "()";"#;
         parse(YaccKind::SelfDescribing(None), src).expect_error_at_line_col(
             src,
             YaccGrammarErrorKind::MissingGrmtoolsHeader,
@@ -2889,8 +2889,8 @@ S: "()";"#;
 
         // Missing the %grmtools header entirely.
         let src = r#"%start S
-                     %%
-                     S: "()";"#;
+                           %%
+                           S: "()";"#;
 
         parse(YaccKind::SelfDescribing(None), src).expect_error_at_line_col(
             src,
@@ -2900,10 +2900,10 @@ S: "()";"#;
         );
         // Empty %grmtools header with no yacckind.
         let src = r#"%grmtools {
-                     }
-                     %start S
-                     %%
-                     S: "()";"#;
+                           }
+                           %start S
+                           %%
+                           S: "()";"#;
 
         parse(YaccKind::SelfDescribing(None), src).expect_error_at_line_col(
             src,
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -2783,6 +2783,12 @@ B";
 }
 %%
 S: "()";"#,
+            r#"%grmtools { yacckind Original(NoAction) }
+%%
+S: "()";"#,
+            r#"%grmtools {yacckind Original(NoAction)}
+%%
+S: "()";"#,
             r#"
 %grmtools {
   yacckind Original(UserAction)
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -14,7 +14,7 @@ use std::{
 
 pub type YaccGrammarResult<T> = Result<T, Vec<YaccGrammarError>>;
 
-use crate::{Span, Spanned};
+use crate::{yacc::YaccOriginalActionKind, Span, Spanned};
 
 use super::{
     ast::{GrammarAST, Symbol},
@@ -33,6 +33,7 @@ pub enum YaccGrammarErrorKind {
     MissingColon,
     MissingRightArrow,
     MismatchedBrace,
+    MissingGrmtoolsHeader,
     NonEmptyProduction,
     PrematureEnd,
     ProductionNotTerminated,
@@ -44,6 +45,7 @@ pub enum YaccGrammarErrorKind {
     DuplicateImplicitTokensDeclaration,
     DuplicateExpectDeclaration,
     DuplicateExpectRRDeclaration,
+    DuplicateGrmtoolsHeaderKey,
     DuplicateStartDeclaration,
     DuplicateActiontypeDeclaration,
     DuplicateEPP,
@@ -51,6 +53,8 @@ pub enum YaccGrammarErrorKind {
     InvalidString,
     NoStartRule,
     UnknownSymbol,
+    InvalidYaccKind,
+    InvalidGrmtoolsHeaderKey,
     InvalidStartRule(String),
     UnknownRuleRef(String),
     UnknownToken(String),
@@ -88,6 +92,7 @@ impl fmt::Display for YaccGrammarErrorKind {
             YaccGrammarErrorKind::IncompleteComment => "Incomplete comment",
             YaccGrammarErrorKind::IncompleteAction => "Incomplete action",
             YaccGrammarErrorKind::MissingColon => "Missing ':'",
+            YaccGrammarErrorKind::MissingGrmtoolsHeader => "Missing '%grmtools' header section",
             YaccGrammarErrorKind::MissingRightArrow => "Missing '->'",
             YaccGrammarErrorKind::MismatchedBrace => "Mismatched brace",
             YaccGrammarErrorKind::NonEmptyProduction => "%empty used in non-empty production",
@@ -104,6 +109,9 @@ impl fmt::Display for YaccGrammarErrorKind {
             YaccGrammarErrorKind::DuplicateExpectRRDeclaration => {
                 "Duplicate %expect-rr declaration"
             }
+            YaccGrammarErrorKind::DuplicateGrmtoolsHeaderKey => {
+                "Duplicate header key in %grmtools section"
+            }
             YaccGrammarErrorKind::DuplicateImplicitTokensDeclaration => {
                 "Duplicated %implicit_tokens declaration"
             }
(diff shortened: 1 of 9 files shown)


ratmice:
One thing I meant to say is that the parsing code is most likely whitespace sensitive,
e.g. I believe it pretty much expects the format where `{` is followed by a newline and `}` is on it's own line like

```
%grmtools {
  ...
}
```

And would fail with something like `%grmtools { yacckind Grmtools}`,
Similarly keys/values are all case sensitive, these seemed like things we could consider relaxing.
but It didn't seem absolutely necessary for an initial prototype.

ltratt:
I think we would want the eventual parser to be whitespace insensitive, but it shouldn't be *too* hard? We just need to define how we want to map the human-writeable text to the internal format. We don't have to make it too "serde like" or even "Rust like".

ratmice:
Yeah, I don't think it should be too hard either -- it just isn't how our current `parse_ws` helpers behave (which *expect* newline, or expect *no* newline) and I didn't want to strike out on making separate *relaxed* function on the first approach.

Indeed I didn't do much or anything in the way of internal format, the key parsing just writes directly to the `self.yacc_kind` when needed in a ad-hoc way.  A form of internal format might make sharing between lex/yacc easier, but also seemed like something that would need to be decided after this initial prototype stage

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -339,6 +374,95 @@ impl YaccParser {
...
+        for (kind_name, kind) in YACC_KINDS {
+            if s.starts_with(kind_name) {
+                if update_yacc_kind {
+                    self.yacc_kind = kind;
+                }
+                let end_pos = j + kind_name.len();
+                return Ok(end_pos);
+            }
+        }
+        // This strikes me as an interesting case where we know the start pos of the
```
ltratt, line 405: Agreed. In general, I tend to think that the user is helped most by pointing out the earliest possible place that an error is detected, so we're doing the right thing here I think.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -339,6 +374,95 @@ impl YaccParser {
...
+        // invalid yacc kind value, but do not know the end, we could `self.parse_to_eol()`.
+        Err(self.mk_error(YaccGrammarErrorKind::InvalidYaccKind, i))
+    }
+    fn parse_grmtools_header(&mut self, mut i: usize) -> Result<usize, YaccGrammarError> {
+        let mut yacc_kind_key_span = None;
+        let mut _yacc_kind_val_span = None;
```
ltratt, line 411: If we're not using this, we should probably remove the variable?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -1008,6 +1135,7 @@ mod test {
     };
     use std::collections::HashSet;
 
+    #[track_caller]
```
ltratt, line 1138: Not sure we want to leave this in? Though I admit I've never even thought about using it before, so maybe this is a good idea?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -339,6 +374,95 @@ impl YaccParser {
...
+                }
+                let end_pos = j + kind_name.len();
+                return Ok(end_pos);
+            }
+        }
+        // This strikes me as an interesting case where we know the start pos of the
+        // invalid yacc kind value, but do not know the end, we could `self.parse_to_eol()`.
+        Err(self.mk_error(YaccGrammarErrorKind::InvalidYaccKind, i))
+    }
+    fn parse_grmtools_header(&mut self, mut i: usize) -> Result<usize, YaccGrammarError> {
```
ltratt, line 409: Missing newline between functions.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -2635,4 +2763,120 @@ B";
...
+           %start S
+           %%
+           S: "()";"#
+      ];
+      for yacc_src in fallback_srcs {
+            parse(YaccKind::SelfDescribing(Some(Box::new(YaccKind::Original(YaccOriginalActionKind::NoAction)))), yacc_src).unwrap();
+      }
+
+    }
+    #[test]
```
ltratt, line 2807: Missing newline between functions.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -832,7 +837,10 @@ where
             RecoveryKind::CPCTPlus => "CPCTPlus",
             RecoveryKind::None => "None",
         };
-        match self.yacckind.unwrap() {
+        match self.yacckind.clone().unwrap() {
+            YaccKind::SelfDescribing(_) => {
+                unimplemented!("Concrete YaccKind should be known at this point")
```
ltratt, line 842: I wonder if this is a "should" or perhaps more a "most" situation?

ltratt: I notice that in the current code we prefer "must":

```rust
        let yk = match self.yacckind {
            None => panic!("yacckind must be specified before processing."),
```

which feels clearer about what the user should do IMHO.

ltratt:
I've dumped in some detailed, but mostly not deep, comments.One high-level thing that I haven't quite wrapped my head around is how we expect the user to use the `SelfDescribing(default)` case. For example, if I was a naive user, I wonder if I might expect to say I've got `YaccKind::Original(...)` case but if I add a `%grmtools` initial header then I expect that to override things and I might be a bit surprised if I'm told I should use `SelfDescribing` and then particularly if I find I probably want to use `SelfDescribing(YaccKind::Orignal(...))` in case I later don't want the `%grmtools` bit?

ratmice:
I'm still waking up, I haven't managed to parse your confusion above so I don't quite understand the exact scenario given yet. I definitely agree that the number of "degrees of freedom" in the overriding can be confusing because it can be overridden "bottom up" where the `%grmtools` section takes precedence by passing `YaccKind::SelfDescribing(...)` to the constructor, and "top-down" where the `YaccKind` value passed to the constructor does for other values.

Overall my thoughts have been that the typical user shouldn't need to be aware of the `YaccKind::SelfDescribing(...)`,
One of the primary motivations I had in doing this is that the errors when you run `nimbleparse foo.{y,l} input` and `foo.y` uses the `Grmtools` variant but the parser defaults to the `Original` the errors can get wonky like [this](https://github.com/softdevteam/grmtools/issues/446#issuecomment-2094066691) because it is parsing with the default format, which is wrong for that calc example.

Unfortunately that isn't actually fixed by the patch, because for backwards compatibility it uses `SelfDescribing(Original)` still specifying the default.  Rather than `None`.  So we don't actually produce the "nice" error where it says `I don't know the yacc kind`.

My general thinking is that *new code* should typically use `SelfDescribing(None)`, to encourage `%grmtools` sections to be added, and an error produced when missing. Then after they get added you no longer need to specify e.g. `nimbleparse -y grmtools` every time it just uses the format specified by the file. 

There are a few reasons to want to specify values *other* than `SelfDescribing(None)`, I tend to think they fall in two major cateogies (historical behavior, and flexibility).  To give more exact reasons:

1. They want to parse using some exact format regardless of what is specified in the file, such as `YaccKind::Original` specify that kind.
2. They want to parse a grammar using any/all formats, but default to a specific format `SelfDescribing(default)`.
3. They want to parse files which don't have the `%grmtools` section (either of the above).
4. They want to produce an error if `%grmtools` section is missing and only use value provided by the section `SelfDescribing(None)`

So, I think from the perspective of a user, `SelfDescribing(None)` should have the ability to produce the best error in the sense that it can produce `MissingYaccKind`, where defaulting to a format can produce confusing ones.  Because it can unintentionally parse a format using the wrong kind.

That might not answer your specific scenario, I've got to go make some :coffee: and read it again.

lrpar/src/lib/ctbuilder.rs
ratmice: Yeah, must is better!

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -339,6 +370,100 @@ impl YaccParser {
...
+            ),
+            (
+                "Original(UserAction)",
+                YaccKind::Original(YaccOriginalActionKind::UserAction),
+            ),
+            (
+                "Original(GenericParseTree)",
+                YaccKind::Original(YaccOriginalActionKind::GenericParseTree),
+            ),
+            ("Eco", YaccKind::Eco),
```
ratmice, line 392: I noticed that nimblepase uses lowercase here, and also has `original` with no action kind specified,
> [-y <eco|grmtools|original>]

Should we try and make this case insensitive?
Not sure about the action kind though.

cfgrammar/src/lib/yacc/parser.rs
ratmice: I went ahead and removed it in [a307980](https://github.com/softdevteam/grmtools/pull/491/commits/a307980095753e5fa028e6de465eab90a7fd7e72) I was reading the diff context wrong anyways and don't see where the panic in the function even is.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Should be fixed in [a307980](https://github.com/softdevteam/grmtools/pull/491/commits/a307980095753e5fa028e6de465eab90a7fd7e72)

cfgrammar/src/lib/yacc/parser.rs
ratmice: Should be fixed in [a307980](https://github.com/softdevteam/grmtools/pull/491/commits/a307980095753e5fa028e6de465eab90a7fd7e72)

lrpar/src/lib/ctbuilder.rs
ratmice: Should be fixed in [a307980](https://github.com/softdevteam/grmtools/pull/491/commits/a307980095753e5fa028e6de465eab90a7fd7e72)

cfgrammar/src/lib/yacc/parser.rs
ratmice: Should be fixed in [a307980](https://github.com/softdevteam/grmtools/pull/491/commits/a307980095753e5fa028e6de465eab90a7fd7e72)

cfgrammar/src/lib/yacc/parser.rs
ltratt: Yes, case insensitive seems probably the right thing? I don't feel strongly about it though.

ltratt:
> My general thinking is that new code should typically use SelfDescribing(None), to encourage %grmtools sections to be added, and an error produced when missing.

I agree with this and I tend to think this should become the default in the next major release i.e. we would no longer force API users to specify `yacckind`.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -323,10 +323,14 @@ impl YaccParser {
         let mut errs: Vec<YaccGrammarError> = Vec::new();
         let mut result = self.parse_grmtools_header(0);
 
-        if let YaccKind::SelfDescribing = self.yacc_kind {
+        if let YaccKind::SelfDescribing(default) = &self.yacc_kind {
```
ratmice, line 326: One minor thing occurred to me,

This default/fallback setting code should probably be done in a loop, 
otherwise silly values can cause this to reach later panics with
`SelfDescribing(SelfDescribing(None))`, strangely making it a while causes an infinite loop. 
I don't yet understand why since I thought it should be decreasing/well founded.

ratmice: Fixed in 19b6629

ratmice:
Cool, well I *think* I've dealt with all the issues that I had brought up after my initial commits, and feel like I'm satisfied with how it has turned out, unless anything comes up in another round of review.

ratmice:
> > My general thinking is that new code should typically use SelfDescribing(None), to encourage %grmtools sections to be added, and an error produced when missing.
> 
> I agree with this and I tend to think this should become the default in the next major release i.e. we would no longer force API users to specify `yacckind`.

So after my last comment I got to thinking I'm unsure how I should take this, I initially got to thinking this should mean the next time we consider making a breaking change we should make the `yacckind` fields that are required `Option<YaccKind>`.

But I later considered that I could take it to mean that e.g. instead of adding a new variant, we could do much the same,
e.g. `None` becomes `SelfDescribing(None)` and `SelfDescribing(Some(default))` becomes `Some(yacckind)`?

I do think skipping the `add a new variant` and moving right into the breaking change does have the benefit of getting rid of the needed `Box`, because it no longer needs to be recursive, so perhaps it is better?  I don't feel like I have a strong opinion on this.

ltratt:
Long and short: whereas in `build.rs` we force users to do something like `ctp.yacckind(YaccKind::Grmtools)` I'd like them to be able to avoid that call entirely, assuming they have a `%grmtools` declaration in the grammar itself.

ratmice:
Ahh yeah I think I see,

That isn't too hard of a task (e.g. the builder code already works with `Option<YaccKind>` internally) it's that they call `YaccGrammar::new`  which requires a `YaccKind`.

So I think the simplest change is for that is turning a Builder `None` value into `SelfDescribing(None)`.
I just wasn't sure if we didn't want to take the opportunity to thread `Option<YaccKind>` all throughout `YaccGrammar::new` instead.

ratmice:
I went ahead and made that change relaxing the builder requirements.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -401,9 +401,8 @@ where
             .as_ref()
             .expect("output_path must be specified before processing.");
         let yk = match self.yacckind.clone() {
-            Some(YaccKind::SelfDescribing(_)) | None => {
-                panic!("yacckind must be specified before processing.")
-            }
+            None => YaccKind::SelfDescribing(None),
+            Some(YaccKind::SelfDescribing(default)) => YaccKind::SelfDescribing(default),
```
ratmice, line 405: This can be simplified to the following, but I suppose the current code is more robust to not potentially doing the wrong thing when adding new variants?

```
             None => YaccKind::SelfDescribing(None),
             Some(YaccKind::Eco) => panic!("Eco compile-time grammar generation not supported."),
             Some(x) => x,
```

So I didn't want to just simplify it without asking first.

ltratt: I think the simplification is the right thing to do here.

[TBH, we should drop the `Eco` support at some point. That tool has bitrotted, and carrying stuff around for it probably isn't worthwhile anymore.]

nimbleparse/src/main.rs
```diff
@@ -152,7 +152,7 @@ fn main() {
 
     let yacc_y_path = PathBuf::from(&matches.free[1]);
     let yacc_src = read_file(&yacc_y_path);
-    let ast_validation = ASTWithValidityInfo::new(yacckind, &yacc_src);
+    let ast_validation = ASTWithValidityInfo::new(yacckind.clone(), &yacc_src);
```
ratmice, line 157: So, working on the exploratory PR for the other approach, it became apparent that there is actually an issue here,
It's actually quite a major issue!

Basically the clone calls are keeping around a `YaccKind::SelfDescribing`, and because `YaccParser.yacc_kind` isn't `pub`, it never gets read back out after parser initialization.  The testsuite isn't actually picking it up yet,

We don't seem to have tests for nimbleparse, and I haven't implemented the `cttests` since modifying `CTParserBuilder` yet.

`echo "1+1" | ../../../../target/debug/nimbleparse calc.{l,y} -`

```
thread 'main' panicked at cfgrammar/src/lib/yacc/grammar.rs:154:17:
not implemented: Concrete YaccKind must be known by this point
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So, that is definitely a hurdle that this PR needs to overcome.

ratmice: This should be fixed in 98b6ee2

ratmice: cttests added, and fixed similarly in [23232f8](https://github.com/softdevteam/grmtools/pull/491/commits/23232f8d4ad680d7441a18092f3e235027e721e0)

ltratt: Good point about nimbleparse. We could at least test it minimally in `.buildbot.sh`?

ltratt: I don't think we want to `pub(crate)` the `yacc_kind: YaccKind` field?

ratmice: Would a `pub(crate) fn yacc_kind(&self) -> &YaccKind` be okay?
We need it to escape, along side this [YaccParser::ast()](https://docs.rs/cfgrammar/latest/src/cfgrammar/yacc/parser.rs.html#342) function which is `pub(crate)`?

ratmice: I went ahead and tried that later option plus renaming the `ast` function which seemed like the cleanest to me.

ratmice:
Here is the initial results of that other exploratory branch,
I didn't use `Option<Either<YaccKind,YaccKind>>` but made an Ad-hoc enum with no generics that is roughly equivalent.
I haven't added all of the parsing code from this branch yet (so it doesn't pass the testsuite)

But given that it is a much more breaking change, I wanted to post it for feedback on whether you think that approach is beneficial enough to justify the breakage, before actually going through the effort of making it fully functional.

https://github.com/ratmice/grmtools/tree/grmtools_directive_2

nimbleparse/src/main.rs
ratmice: For the record, I had a crisis of conscience in regard to that, e.g. whether what I did was right vs adding a `yacc_kind` field to `GrammarAST`. So I also tried adding it to `GrammarAST`, but the takeaway was that currently `GrammarAST` is completely devoid of anything related syntax.  You can even construct one by hand with no syntax at all.  Where `YaccKind` is kind of `YaccParser` specific thing. So my takeaway from that experiment was that I still think prefer that `finish` function.

ltratt: > Would a pub(crate) fn yacc_kind(&self) -> &YaccKind be okay?

Yep!

ratmice:
> Obvious question: what would the default be? `NoDefault` or ... ?

Well, I've used `NoDefault` for `CTParserBuilder` but because *nimbleparse* currently defaults to `YaccKind::Original(GenericParseTree)`, I switched it from having the behavior of `Force(YaccKind::Original(GenericParseTree))` to `Default(YaccKind::Original(GenericParseTree))`.

For nimbleparse though I just did that because that is what is most compatible with the current behavior.
Making the behavior `NoDefault` across the board leads to the best error messages.
So I would lean towards that if changing the behavior of nimbleparse is in play or we the behavior to be the same across `CTParserBuilder` and *nimbleparse*

ltratt:
> Making the behavior NoDefault across the board leads to the best error messages.
> So I would lean towards that if changing the behavior of nimbleparse is in play or we want the behavior to be the same across CTParserBuilder and nimbleparse

I am inclined to agree. Yes, it's a breaking change, but it's a pretty sensible one. Even better, we can presumably make it so that lrpar users specifying `yacckind(...)` have a zero-change upgrade path. Basically the only people really affected are nimbleparse users, and I think changing that is more acceptable than the library API.

Go ahead!

ratmice:
Sounds good, I'll go ahead and close this PR now then and open a new one once I've finished merging all the parsing code from this one, and updated it based on the things we've discussed.

PR #492, opened 2025-02-27
Change YaccKind parameters to a new `YaccKindResolver` type

ratmice:
Here is take two on my previous PR #491 

Had to slightly rush to get this out the door, as there is a possibility I may not be able to work on this or respond for a few days but hopefully nothing has been overlooked.

I did slightly relax the documentation string for `YaccKindResolver::Force` that you had given, changing it from being an error if they differ, to an error unless they are *compatible*.  This is primarily to allow for differing `YaccOriginalActionKind` behaviors when used with the `Original` kind.

I've added `%grmtools` sections to *some* of the files in the repo while testing, but certainly not *all* this should probably
done though.

--

This is a breaking change which adds a `YaccKindResolver` type. `YaccKindResolver` can allow a yacc grammar to be read from a `%grmtools` section of a yacc grammar source file.

CTParserBuilder which previously required a YaccKind, has been relaxed to allow the YaccKind to be unspecified in which case it will now attempt to find the `yacckind` in the `%grmtools` section.

nimbleparse previously would default to using `YaccKind::Original`, now attempts to read the yacc kind from the `%grmtools` section and otherwise throws an error. This should improve error messages and avoid defaulting to the wrong yacc kind.

### DIFF
--- a/cfgrammar/src/lib/yacc/mod.rs
+++ b/cfgrammar/src/lib/yacc/mod.rs
@@ -14,6 +14,17 @@ pub use self::{
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum YaccKindResolver {
+    /// The user can specify `%grmtools` in their grammar but unless it is compatible with this `YaccKind`, it's an error
+    Force(YaccKind),
+    /// Use `YaccKind` if the user doesn't specify `%grmtools` in their grammar
+    Default(YaccKind),
+    /// The user must specify `%grmtools` in their grammars or we throw an error
+    NoDefault,
+}
+
 /// The particular Yacc variant this grammar makes use of.
 #[derive(Clone, Copy, Debug)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
--- a/cfgrammar/src/lib/yacc/parser.rs
+++ b/cfgrammar/src/lib/yacc/parser.rs
@@ -18,7 +18,7 @@ use crate::{Span, Spanned};
 
 use super::{
     ast::{GrammarAST, Symbol},
-    AssocKind, Precedence, YaccKind,
+    AssocKind, Precedence, YaccKind, YaccKindResolver, YaccOriginalActionKind,
 };
 
 /// The various different possible Yacc parser errors.
@@ -56,6 +56,10 @@ pub enum YaccGrammarErrorKind {
     UnknownToken(String),
     NoPrecForToken(String),
     UnknownEPP(String),
+    InvalidYaccKind,
+    InvalidGrmtoolsHeaderKey,
+    DuplicateGrmtoolsHeaderKey,
+    MissingGrmtoolsHeader,
 }
 
 /// Any error from the Yacc parser returns an instance of this struct.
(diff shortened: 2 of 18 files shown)


cfgrammar/src/lib/yacc/mod.rs
```diff
@@ -14,6 +14,27 @@ pub use self::{
...
+    /// The user can specify `%grmtools` in their grammar but unless it is compatible with this `YaccKind`, it's an error
+    Force(YaccKind),
+    /// Use `YaccKind` if the user doesn't specify `%grmtools` in their grammar
+    Default(YaccKind),
+    /// The user must specify `%grmtools` in their grammars or we throw an error
+    NoDefault
+}
+
+impl YaccKindResolver {
+    pub(crate) fn self_describing(self) -> bool {
```
ratmice, line 29: Hmm, there is probably a better name for the like: `uses_grmtools_section`

ratmice: Nevermind, once clippy was done with these, they became one liners, and since they were only used in one place.
It is much easier to just inline them and remove the methods entirely.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -339,8 +375,99 @@ impl YaccParser {
...
+            } else if require_yacc_kind {
+                return Err(self.mk_error(YaccGrammarErrorKind::MissingGrmtoolsHeader, i));
+            }
+        } else if require_yacc_kind {
+            return Err(self.mk_error(YaccGrammarErrorKind::MissingGrmtoolsHeader, i));
+        }
+        Ok(i)
+    }
+
+    pub(crate) fn finish(self) -> (Option<YaccKind>, GrammarAST) {
```
ltratt, line 469: I wonder if we should call this `build` -- it's a more common name for this sort of Rust thing? But perhaps `build` isn't quite right, and `finish` (or maybe `to_ast`?) is better?

ltratt:
This looks close to being ready to merge to me. Since it will imply a breaking release I have two "not exactly related, but worth noting here" thoughts:

1. We should do a final 0.13 release now before merging this PR.
2. If we're going to do a breaking release, what other breaking things do we want to consider?

cfgrammar/src/lib/yacc/parser.rs
ratmice: `build` seems good to me, in the why didn't I think of that way.

ratmice: should be fixed in [c84c3f2](https://github.com/softdevteam/grmtools/pull/492/commits/c84c3f28530f09b49df6b9dff8089b3ef35b368f)

ltratt:
I think we're ready to squash? If so, please go ahead.

ratmice:
Squashed.

ratmice:
So the first thing that came to mind (could potentially be some kind of breaking change, uncertain about that though).

Has to do with source code generation, currently we have like nimbleparse, and build.rs code typically uses `CTParserBuilder`.  I would *like* to get to the point where *nimbleparse_lsp* can lint actions code, by generating sources and running clippy/etc on the generated source and translating.

It isn't really clear that `CTParserBuilder` is the right thing that an lsp should be running.
E.g. it writes directly to the file system, where a LSP should probably generate code to a user provided string.
I suppose we could just add an `output_string(self: CTParserBuilder, &mut String)` and that would be a compatible change?

Then we also have issues like #448 (This is only tangentially related to the above in that one could consider graphviz or railroad diagrams as *output*) and one could imagine some kind of `YaccOutputKind` type and everything that encompasses both source generation and diagram generation and the like, and have the things that operate directly on filesystem paths internally use the string based APIs.

```
enum YaccOutputKind {
    RailroadDiagramSvg,
    GraphvizDiagram,
    RustParserModuleSource,
}
```

I'm not certain I currently have any real specific/concrete API in mind, and I'm not sure it's necessary to do so in a breaking fashion, but there is this seed of an idea that both diagram and source generation have this generation/output step in common, so we could consider that sort of thing from first principles if you'd like.

Perhaps I should make a separate issue for this stuff?

PR #494, opened 2025-03-01
Remove debug formatting for CTLexerBuilder src gen

### LINKED ISSUE #323: Remove debug formatting in non-debug locations

ratmice:
I was looking through issues, trying to see if there was anything therein that would require a breaking change,
I couldn't remember why I didn't finish fully migrating away from debug formats for CTLexerBuilder leaving the issue open.  There were two issues, one was encoding `Option` types, there was a quote issue I've referenced/borrowed some code from for this.

The other one is (technically?) a breaking change technically because it required us to add a new bounds for `StorageT`,
though I think it should always be satisfied given the `AsPrimitive` bounds.

I went head and also did CTParserBuilder, so this one should fix #323 entirely.

This adds a dependency on the `proc_macro2` crate, but technically `quote` already depends upon it, so I don't think it would actually introduce any additional overhead.


lrlex/Cargo.toml
```diff
@@ -27,6 +27,7 @@ vergen = { version = "8", default-features = false, features = ["build"] }
 [dependencies]
 cfgrammar = { path = "../cfgrammar", version = "0.13" }
 lrpar = { path = "../lrpar", version = "0.13" }
+proc-macro2 = "1.0"
```
ratmice, line 30: This should probably be in ../Cargo.toml and use `proc-macro2.workspace = true` here.

ratmice: Fixed in [acf3134](https://github.com/softdevteam/grmtools/pull/494/commits/acf3134e53ab0675d5c7a093340b32b52df2af3f)

lrpar/src/lib/ctbuilder.rs
```diff
@@ -920,7 +927,7 @@ where
             if !grm.rule_to_prods(ridx).contains(&grm.start_prod()) {
                 write!(
                     outs,
-                    "    #[allow(dead_code)]\n    pub const R_{}: {} = {:?};\n",
+                    "    #[allow(dead_code)]\n    pub const R_{}: {} = {};\n",
```
ratmice, line 930: two things to note here,

1. There are other places we are encoding usize with plain old `{}` 
2. quote does the wrong thing here, as it will encode type specifies e.g. `0usize` or whatever
and we want this to be something of `type_name::<StorageT>()` below.

As such, I believe it is okay to just remove this debug format

lrpar/src/lib/ctbuilder.rs
```diff
@@ -733,17 +734,23 @@ where
...
+        let recoverer = self.recoverer;
+        let yacckind = self.yacckind;
+        let visibility = self.visibility.cow_str();
+        let error_on_conflicts = self.error_on_conflicts;
+        writeln!(cache, "   Build time: {}", quote!(#build_time)).ok();
+        writeln!(cache, "   Grammar path: {}", quote!(#grammar_path)).ok();
+        writeln!(cache, "   Mod name: {}", quote!(#mod_name)).ok();
+        writeln!(cache, "   Recoverer: {}", quote!(#recoverer)).ok();
+        writeln!(cache, "   YaccKind: {}", quote!(#yacckind)).ok();
+        writeln!(cache, "   Visibility: {}", quote!(#visibility)).ok();
```
ratmice, line 749: All these cache lines will now include fully qualified symbols where before they just used e.g.
```RecoveryKind::CPCTPlus` If I understand correctly the only effect changing it will cause is there will be an extra complete rebuild since the cache changed.

**Comment by ratmice** [comment]
FWIW, there are some more cleanups that can be done here, there are a few places we are encoding Options that could be migrated to `QuoteOption`, then they won't have to match 

A couple of these cases can be found with ripgrep, `rg 'format!\("Some'`

File: lrpar/src/lib/ctbuilder.rs [hunk 4]
```diff
@@ -1249,11 +1264,6 @@ where
     }
 }
 
-/// Return a version of the string `s` which is safe to embed in source code as a string.
-fn str_escape(s: &str) -> String {
```
ratmice, line 1246: This function was added here:
https://github.com/softdevteam/grmtools/commit/87a16197f54d9bdba130991d4ba2bba5333105f9

But I believe that the tests that were added in that patch now live here:
https://github.com/softdevteam/grmtools/blob/master/lrpar/cttests/src/epp.test

So I believe that `quote!` does the right thing here.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -78,11 +78,29 @@ pub enum RustEdition {
...
+        });
+    }
+}
+
 /// A `CTLexerBuilder` allows one to specify the criteria for building a statically generated
 /// lexer.
 pub struct CTLexerBuilder<'a, LexerTypesT: LexerTypes = DefaultLexerTypes<u32>>
 where
-    LexerTypesT::StorageT: Debug + Eq + Hash,
+    LexerTypesT::StorageT: Debug + Eq + Hash + ToTokens,
```
ratmice, line 103: To me this here, and the related bounds immediately below  this are the most controversial aspects to this patch,
and the part that we need to consider most carefully.

The other thing we could do here is change this `ToTokens` to `Display`,

Changing it to `ToTokens` kind of just enables one minor cleanup, the `QuoteOption(tok_id)` cleanup in [18f1d9d](https://github.com/softdevteam/grmtools/pull/494/commits/18f1d9d6238f998edc673c8c5361fbd140885b38)

I think it would definitely be fine to consider using `Display` here instead of `ToTokens` and dropping that last cleanup?

ratmice: One more thing to note here, which I had mentioned in a previous comment on usize, 
using `ToTokens` here also has the effect of including the type in the value e.g. this patch is currently encoding `StorageT` values as `Some (5u32))`, a change from `Some(5)` with `Debug` and `Display`.

Looking at the sources generated by the testsuite these `StorageT` values i'm seeing are *all* using `u32`.
So if we do go with `ToTokens` it would be good to try and exercise this in the testsuite in a way that uses another type here `u8`.

ratmice:
One more thing to consider here, which we aren't currently doing, somewhat to the detriment of the code generated is
using [prettyplease](https://crates.io/crates/prettyplease) to format the `quote!` generated code, as described here in the [quote!](https://docs.rs/quote/latest/quote/#non-macro-code-generators) docs

lrlex/src/lib/ctbuilder.rs
```diff
@@ -525,31 +543,18 @@ pub fn lexerdef() -> {lexerdef_type} {{
...
-                n,
-                n_span,
-                quote!(#regex),
-                quote!([#(#start_states),*]),
-                target_state,
+                "{}\n",
+                quote! { Rule::new(::lrlex::unstable_api::InternalPublicApi, #tok_id, #n,
+                                          #n_span, #regex, vec![#(#start_states),*], #target_state,
+                                          &regex_options).unwrap(),
+                }
```
ratmice, line 557: I'm not really sure if or how much we want to migrate this code into the `quote!` style,
It is a little bit beyond just fixing the removal of `Debug` format stuff.

It isn't totally apparent in the  apparent in the diff context, the trailing comma in `unwrap(),` shows it a little
that this is outputting a `Vec<Rule>`, I think leaning as far as we could in direction we would end up with something like 
```
let rules_iter = lexerdef.iter_rules();
format!("{}", quote!{vec![#(#rules_iter),*]})
```

Implementing `ToTokens` for `Rule` itself, with `QuoteOption` and friends being used inside the `ToTokens` implementation for `Rule`.  It isn't totally clear to me that this leads to nicer source gen, and/or the best practices for generating source code with `quote` isn't something I am familiar enough with that I feel like I can know/expect what is the nicest/clearest way to approach it.

At this point I just feel that it is worthwhile experimenting with trying to utilize `quote` more here.
Anyhow, I'll probably wait for some feedback before pushing any further in this direction.

lrlex/src/lib/ctbuilder.rs
ratmice: I don't have a very strong opinion here, but I'm somewhat leaning towards liking the `1_u8` form of `ToTokens`.

I tested the codegen of `ToTokens` here, by changing the `lrpar/examples/calc_actions/build.rs` to use
`CTLexerBuilder::<DefaultLexerTypes<u8>>::new_with_lexemet()` and it makes sense to me that in theory it should be fine for `StorageT` to include the concrete type in literals, since kind of the point is that it has a concrete size known to the caller.

Other places where we rely on the *can be converted into a usize losslessly* we could potentially use a *newtype* around known integer types  to print literals without the type suffix.

Either way, since we didn't remove the `Debug` suffix, at worst we can go back to using `Debug`?

lrpar/src/lib/ctbuilder.rs
ratmice: I went ahead and made a newtype wrapper in [484abd1](https://github.com/softdevteam/grmtools/pull/494/commits/484abd1f6a4196453beccfac96a875c4f9286ce0) that does this unsuffixed printing, or at least shows how it would be achieved with `ToTokens` instead of format printing If we wanted to go that route.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -78,11 +78,47 @@ pub enum RustEdition {
     Rust2021,
 }
 
+use proc_macro2::TokenStream;
+use quote::{ToTokens, TokenStreamExt};
+/* Quote by default quotes option values as an empty string for `None`,
```
ltratt, line 83: I think we can use `//` here :)

lrlex/src/lib/ctbuilder.rs
```diff
@@ -78,11 +78,47 @@ pub enum RustEdition {
     Rust2021,
 }
 
+use proc_macro2::TokenStream;
+use quote::{ToTokens, TokenStreamExt};
```
ltratt, line 82: These should probably go with the other `use`s at the top of the file.

lrlex/src/lib/ctbuilder.rs
ltratt: I'm fine with having the type suffix for literals.

One thing I haven't fully understood is: what types is `ToTokens` implemented on? I assume it's implemented on all primitive integers by `quote`? In that sense, this isn't going to be much of a breaking change unless a user does something odd like pass a newtype here: that would force them to know about the `ToTokens` trait and `quote`. In that sense, `Display` would be preferable, but I take your point in https://github.com/softdevteam/grmtools/pull/494/commits/18f1d9d6238f998edc673c8c5361fbd140885b38 that `Display` means we end up having to do our own string escaping, which is a bit nasty. On that basis, I'm fine using `ToTokens`.

lrlex/src/lib/ctbuilder.rs
ltratt: My suggestion would be: in this PR, do the minimum-ish necessary; but after that I do think it would be worth considering using `quote` (or similar) to generate code in a more "modern Rust" fashion.

When I started grmtools, the code generation ecosystem was small, the "eventual winners" didn't yet exist or weren't clear, and so I just resorted to doing it the "old fashioned" way. But the code is, because of that, pretty nasty: whenever I have to fiddle with it, it makes my head hurt!

It would be nice to give this aspect of grmtools an update: it may also help us take advantage, in the future (or maybe even now?) of things like "embed source location" more easily, as and when the various libraries start supporting it?

lrpar/src/lib/ctbuilder.rs
ltratt: That's fine: a one-off rebuild isn't going to trouble anyone.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -52,6 +54,30 @@ struct CTConflictsError<StorageT: Eq + Hash> {
...
+impl<T: ToTokens> ToTokens for QuoteOption<T> {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        tokens.append_all(match self.0 {
+            Some(ref t) => quote! { ::std::option::Option::Some(#t) },
+            None => quote! { ::std::option::Option::None },
+        });
+    }
+}
+
+struct UnsuffixedUsize(usize);
```
ltratt, line 73: I like the idea behind this: it'll need a motivational docstring to remind us *why* it's here though!

lrpar/src/lib/ctbuilder.rs
ltratt: Works for me!

ltratt:
I have a few stylistic comments, and a couple of "deeper but I tend to agree with the approach taken in the PR" comments. I think we can tidy up things and get this merged fairly quickly.

lrlex/src/lib/ctbuilder.rs
ltratt: So long as the integer types are there, I think we'll have covered nearly every sensible use that I can conceive of.

ratmice: The one thing I guess I would say is that for the specific case of `StorageT` we don't have to worry about string escaping at all, I think it is also probably the most stable of all format printing we could likely expect format printing to be.  So I don't think it is too far fetched to use `ToTokens` internally, but not add the bounds for `StorageT` and use `Debug` or `Display` just for that.

So to me `ToTokens` over `Debug` or `Display` for `StorageT` mostly revolves around whether we want to be able to include the type suffix in integer literals. It isn't a huge thing but maybe adds a little bit of type checking we don't currently get (for better or worse!).

ltratt: If our plan is to use `quote` (and friends) for all code generation inside grmtools, I think the `ToTokens` bound makes sense: we'll have bought into that ecosystem, and might as well take advantage of it, and make clear that we've done so to users. If we don't plan on going on all the way with `quote` then perhaps we should just fall back on `Display`.

lrlex/src/lib/ctbuilder.rs
ratmice: Ok, I went ahead and just reverted the whole string formatting change that was done there,
In favor of the minimal change of just quoting the values passed to the string formatter.
Since it's likely this would just change again in a future series.

So I think that should be fixed in [f97f05a](https://github.com/softdevteam/grmtools/pull/494/commits/f97f05a6266571cc160143ba49509db8d6e16b6a)

lrpar/src/lib/ctbuilder.rs
ratmice: Should be fixed in [f6d33f4](https://github.com/softdevteam/grmtools/pull/494/commits/f6d33f49b5d685aa9952122f6640f80433926fdc)

lrlex/src/lib/ctbuilder.rs
ratmice: Should be fixed in [f6d33f4](https://github.com/softdevteam/grmtools/pull/494/commits/f6d33f49b5d685aa9952122f6640f80433926fdc)

lrlex/src/lib/ctbuilder.rs
ratmice: I just rewrote all this as docstrings, and adding one for all those wrapper types.
So should be fixed in [f6d33f4](https://github.com/softdevteam/grmtools/pull/494/commits/f6d33f49b5d685aa9952122f6640f80433926fdc)

ltratt:
Let me summarise where I think we are:

1. This PR is complete and ready for squashing/merging.
2. By adding the `ToTokens` bound we are implicitly saying "we think we should move the code generation part of grmtools from its current stringy approach to a `quote`-and-friends approach".

Does that sound correct?

ratmice:
I believe so to both points, however one thing to note on the second point.  I didn't actually remove the `Debug` bounds, which allows us to revert back entirely to string printing for `StorageT`, so we could fall back to that for any reason.
In theory `Debug` printing of `StorageT` is still going to be useful for debugging, so it is not being kept around merely just-in-case.  

But yeah, I'd say the intent is there to migrate, as it seems to me like it should lead to cleaner code, but requires some experimentation to get the appropriate code organization.

ltratt:
Looking forward to the migration!

Please squash.

ratmice:
Squashed.

ratmice:
Should be fixed, will need another squash.

ltratt:
Please squash.

ratmice:
Squashed.

PR #495, opened 2025-03-08
Fix VerbatimNotSupported error spans and corner case.

ratmice:
In addition to the things this fixes, it adds a test for one corner case (but doesn't change the behavior).
Where a line contains whitespace only, I believe (and recall testing) that lex does output these lines to generated sources.
But it is kind of a questionable corner case

(Note spaces on the blank line)
```
%%
   
. "dot"
```
Before
```
[Error] in ugh.l
    2|
       ^^^
    3| . "dot"
       ^^^ Verbatim code not supported
```

After
```
[Error] in ugh.l
    2|
       ^^^ Verbatim code not supported
```

The testsuite of lex isn't really set up to catch this case yet because it only tests the start of spans.

----

In the first case, the `end()` of VerbatimNotSupported's span was continuing to the _next_ line because `parse_ws` matches newlines.

VerbatimNotSupported was triggering in one case where spaces occur between `%%` and a newline. It should only trigger when spaces occur at the beginning of a line.

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -27,6 +27,7 @@ lazy_static! {
     static ref RE_LEADING_LINE_SEPS: Regex = Regex::new(r"^[\p{Pattern_White_Space}&&[\p{Zl}\p{Zp}\n\r\v]]*").unwrap();
     // Horizontal space separators
     static ref RE_SPACE_SEP: Regex = Regex::new(r"[\p{Pattern_White_Space}&&[\p{Zs}\t]]").unwrap();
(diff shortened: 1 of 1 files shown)


lrlex/src/lib/parser.rs
```diff
@@ -365,15 +366,15 @@ where
             let j = self.parse_ws(i)?;
```
ratmice, line 366: So the main case is where this is hitting A string like "   \n" it includes the newline in the return value `j`.
Perhaps this should also consider changing to use the `parse_spaces` that I added for the other issue to avoid matching that newline.

ltratt:
Please squash.

ratmice:
Squashed.

PR #496, opened 2025-03-09
First take at grmtools section for lrlex

ratmice:
This mainly just is a prototype implementation for the parsing of the `%grmtools` section.
This doesn't change `RegexOptions`, and implements the behavior that `%grmtools` section in the file
overrides `RegexOptions` that were passed in from the `CTLexerBuilder`, or via `new_with_regex_options`.
Similarly, if none were specified by the `CTLexerBuilder`, it the `%grmtools` section overrides `DEFAULT_REGEX_OPTIONS`.

So, as currently implemented in this PR the `%grmtools` section in the `.l` file always takes precedence.

I figure follow up steps could include:
* Renaming `RegexOptions` to something like `LexFlags`
* Deciding whether we need to implement the more complex fallback mechanisms like those that were required for lrpar.

Regarding fallback and the comparison to `YaccKindResolver`, I'm not certain I see the need for all the complexity of that for *lrlex*.  In particular `YaccKind`s overlap such that a single file can be parse with multiple kinds (the simple case being just different `YaccOriginalActionKinds`).  There is maybe some things that don't really affect parsing, like the `size` limits.    

Where with lrlex, I don't really see how we can implement anything resembling `YaccKindResolver::NoDefault`,
Seems like there is just too many flags, and we'll never want to require people to specify all of them?
So I think the question are whether we want an equivalent to the `Force` and `Default` variants of `YaccKindResolver`

Edit:
One feeling I get is that if we *do* want a `Resolver` type, it would be best to change the `RegexOptions` values that are currently `bool` types to be `Option<bool>` so that the `Default` variant can avoid overriding flags set in the `%grmtools` section, and not specify all of the flags.

The main things that I think seem reasonable to want to override from the `Builder` because they don't affect the parsing validity is things like the `nest_limit`, `size_limit`, `dfa_size_limit`

### DIFF
--- a/lrlex/src/lib/lexer.rs
+++ b/lrlex/src/lib/lexer.rs
@@ -20,12 +20,16 @@ use crate::{
 
 #[doc(hidden)]
 /// Corresponds to the options for `regex::RegexBuilder`.
-#[derive(Clone)]
-pub struct RegexOptions {
-    pub dot_matches_new_line: bool,
-    pub multi_line: bool,
-    pub octal: bool,
-    pub posix_escapes: bool,
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct LexFlags {
+    // The following values when `None` grmtools provides default values for in `DEFAULT_LEX_FLAGS`
+    pub dot_matches_new_line: Option<bool>,
+    pub multi_line: Option<bool>,
+    pub octal: Option<bool>,
+    pub posix_escapes: Option<bool>,
+
+    // All the following values when `None` default to the `regex` crate's default value.
     pub case_insensitive: Option<bool>,
     pub swap_greed: Option<bool>,
     pub ignore_whitespace: Option<bool>,
@@ -35,11 +39,67 @@ pub struct RegexOptions {
     pub nest_limit: Option<u32>,
 }
 
-pub const DEFAULT_REGEX_OPTIONS: RegexOptions = RegexOptions {
-    dot_matches_new_line: true,
-    multi_line: true,
-    octal: true,
-    posix_escapes: false,
+impl LexFlags {
+    /// Merges flags from `other` into `self`
+    /// Flags which are `Some` in `other` overriding flags in self.
+    pub fn merge_from(&mut self, other: &Self) {
+        if other.dot_matches_new_line.is_some() {
+            self.dot_matches_new_line = other.dot_matches_new_line;
+        }
+        if other.multi_line.is_some() {
+            self.multi_line = other.multi_line;
+        }
+        if other.octal.is_some() {
+            self.octal = other.octal;
+        }
+        if other.posix_escapes.is_some() {
+            self.posix_escapes = other.posix_escapes;
+        }
+        if other.case_insensitive.is_some() {
+            self.case_insensitive = other.case_insensitive;
+        }
+        if other.swap_greed.is_some() {
+            self.swap_greed = other.swap_greed;
+        }
+        if other.ignore_whitespace.is_some() {
+            self.ignore_whitespace = other.ignore_whitespace;
+        }
+        if other.unicode.is_some() {
+            self.unicode = other.unicode;
+        }
+        if other.size_limit.is_some() {
+            self.size_limit = other.size_limit;
+        }
+        if other.dfa_size_limit.is_some() {
+            self.dfa_size_limit = other.dfa_size_limit;
+        }
+        if other.nest_limit.is_some() {
+            self.nest_limit = other.nest_limit;
+        }
+    }
+}
+
+/// LexFlags with flags set to default values.
+pub const DEFAULT_LEX_FLAGS: LexFlags = LexFlags {
+    dot_matches_new_line: Some(true),
+    multi_line: Some(true),
+    octal: Some(true),
+    posix_escapes: Some(false),
+    case_insensitive: None,
+    ignore_whitespace: None,
+    swap_greed: None,
+    unicode: None,
+    size_limit: None,
+    dfa_size_limit: None,
+    nest_limit: None,
+};
+
+/// LexFlags with all of the values `None`.
+pub const UNSPECIFIED_LEX_FLAGS: LexFlags = LexFlags {
+    dot_matches_new_line: None,
+    multi_line: None,
+    octal: None,
+    posix_escapes: None,
     case_insensitive: None,
     ignore_whitespace: None,
     swap_greed: None,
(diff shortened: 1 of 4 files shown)


ratmice:
This one also uses some functions that I added in #495 so these two PR's will conflict

lrlex/src/lib/parser.rs
```diff
@@ -221,11 +224,144 @@ where
...
+                                self.regex_options.dot_matches_new_line = flag
+                            }
+                            "multi_line" => self.regex_options.multi_line = flag,
+                            "octal" => self.regex_options.octal = flag,
+                            "posix_escapes" => self.regex_options.posix_escapes = flag,
+                            _ => unreachable!(),
+                        }
+                        i += opt.len();
+                        span_map.insert(opt, Span::new(start_pos, i));
+                        return self.parse_nl(i);
```
ltratt, line 275: Is the `return` here intentional? It surprised me a bit.

ratmice: Not really, does seem like we should just let fall though to the `parse_ws` call below doesn't it.

ltratt: I *think* so. Indeed, I kind of expected all of these clauses to end in the same way, so the differences did jump out at me a bit.

ratmice:
FWIW, the implementation of `LexFlags` I'm inclined to attempt instead of using variants for `Force` and `Default`
uses values `force: LexFlags`, `default: LexFlags`, which select whether `%grmtools` overrides or gets overridden.

Perhaps it is also straight forward to implement a flag `LexFlags.strict_non_overlapping_flags` to indicate that if a value is `Some` in `force` it should be `None` in `%grmtools` and vice versa.  Then remaining `None` values get populated by the `DEFAULT_REGEX_OPTIONS` constant?

That at least seems like it is a flexible/future proof way to extend the whole `force/default` paradigm to multiple flags...
It's probably overkill, but I don't feel like it's particularly difficult to implement if we add some kind of `merge(self, other: Self) -> Self`...

Anyhow I might play around with that if it doesn't sound too terrible

ltratt:
Right, so the trick is probably to be clear for `merge` which side can get overridden. I *think* it's probably `merge_into` (i.e. `other` overrides `self`)?

ratmice:
Yeah either way, but `merge_into` seems fine with me.

lrlex/src/lib/parser.rs
ratmice: Should be fixed in [509296c](https://github.com/softdevteam/grmtools/pull/496/commits/509296c9958edc3bc2cef2d29a61aa5b049df528)

ltratt:
Do we think this one is potentially mergeable now, or do you want to continue pushing to this PR for a bit?

ratmice:
I'd kind of like to see how that option resolving stuff turns out first, I think that is going to make it not modify self.regex_options directly, but take an `&mut` one as a parameter.

as well as just general hesitation to merge something on the day it was written

ratmice:
This needs a bit better testing, I did test `lrlex`, `nimbleparse`

```
%grmtools {
  !unicode
}
%%
. "dot"
```

Ensuring that e.g. it routes flags all the way through to `regex`.

```
[Error] in bad.l
    5| . "dot"
        Invalid regular expression
```
I noticed that the error message could be improved a little, e.g. there is some more info here, that perhaps we might be able to route through the error variant.
I'll leave that for another PR!

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024

lrlex/src/lib/lexer.rs
```diff
@@ -20,12 +20,15 @@ use crate::{
...
-pub struct RegexOptions {
-    pub dot_matches_new_line: bool,
-    pub multi_line: bool,
-    pub octal: bool,
-    pub posix_escapes: bool,
+#[derive(Clone, Debug)]
+pub struct LexFlags {
+    // The following values when `None` grmtools provides default values for in `DEFAULT_LEX_FLAGS`
+    pub dot_matches_new_line: Option<bool>,
+    pub multi_line: Option<bool>,
```
ratmice, line 27: One thing we could consider here is using the `Default` trait,

We kind of have the same thing with the `const UNSPECIFIED_LEX_FLAGS` though.
the way construct this in a way that allows us to add fields in the future is similar to how the Default trait is often used.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=028e54b786184a9d7f4db48a82aabd88

ratmice: Anyhow, I think we should try and make this  `#[non_exhaustive] struct LexFlags`,
It seems that precludes functional record update, but I think currently works/forces
the pattern i've been using of.

```
let mut x = UNSPECIFIED_LEX_FLAGS;
x.flag = Some(...)
````

ratmice: Should be fixed in [d5bfe8a](https://github.com/softdevteam/grmtools/pull/496/commits/d5bfe8a411d1591577d482f579f78352c5ba904a)

ltratt: `non_exhaustive` is a new one on me, but having look it up, I agree that it seems like a good idea here. [I wonder if there are other places we should use this in the public(ish) API?]

ltratt:
I think it's less important with lrlex to allow the degree of configuration that lrpar/cfgrammar have. So if it's awkward to do, at least for now, let's punt on it. Perhaps a cunning plan will occur to us, perhaps it won't!

ratmice:
Alright, well in that case I think this branch at the moment accomplishes all that it had set out to.

Despite my reservations that e.g. different `LexerDef` implementations might want different sets of flags.
I think that is largely a question due to ignorance (both in how that would work, and whether that is how it *should* work)
E.g. I could imagine that it is the case that e.g. we want `LrStreamingLexerDef` to basically take the exact same flags and just generate a different module signature/entry point that works on discontiguous strings with all the same flags.

So while we currently only have *one* `LexerDef` currently, so I do find it difficult to anticipate whether they would have the same or unique configuration choices.

I think part of my panic is just that it doesn't *perfectly* match general trajectory of the API how `LexerDef` is a trait and so-on it seems like it is trying to be much more decoupled, while this is very much a concrete type shared across all those trait implementations.

At the very least, by the time we need that we should have a second implementations of the traits, and that itself is likely to be a fairly disruptive change which seems difficult to anticipate without actually doing all that work.

ratmice:
I'll go ahead and unmark this as no-longer-draft.
I don't mind though if you want me to break this up into smaller patches.
I *think* that the patches would turn out much smaller if I basically did things entirely in reverse.

`RegexOptions -> LexFlags`, `bool -> Option<bool>`, with adding the parsing code for the grmtools section at the end.
So, if you would like me to approach breaking the patch up that way it might be more review friendly.

ltratt:
Please squash as you see fit.

ratmice:
Squashed and reordered down to two commits, the first refactoring and the second adding the new grmtools section.

lrlex/src/lib/lexer.rs
ratmice: I think `LexErrorKind` and `YaccGrammarErrorKind` are both probably good candidates for `non_exhaustive`.

PR #499, opened 2025-03-10
Document grmtools section

ratmice:
This adds a pair of sections for extensions to lex/yacc, as kind of companion chapters to their respective compatibility chapters. I apologize in advance for being terrible at writing.

This also fixes a couple of issues I noticed when documenting the new `%grmtools` stuff in the book.
* I  had missed adding the `CTLexerBuilder` function for `allow_wholeline_comments`
* the documentation for `parse_escapes` was listed as a `regex` flag, but it is grmtools specific.


doc/src/lexextensions.md
```diff
@@ -0,0 +1,59 @@
...
+    allow_wholeline_comments
+    !octal
+    size_limit 1024
+}
+%%
+. "rule"
+```
+## List of flags:
+
+* `posix_escapes: bool` <p>
```
ratmice, line 25: Perhaps this would be better as a table, so it could have the flag name, and arg in separate cells.

doc/src/yaccextensions.md
```diff
@@ -0,0 +1,14 @@
+# Yacc Extensions
+
+At the beginning of a `.y` file is a `%grmtools{}` section, by default this section is required.
+But a default can be set or forced by using a `YaccKindResolver`.
+
+## Required flags
+    * yacckind
```
ratmice, line 7: Missing argument type, and this should list valid values.

doc/src/lexextensions.md
ratmice: Should be fixed in [7c21a9d](https://github.com/softdevteam/grmtools/pull/499/commits/7c21a9df013bb73600cdd8c7bd647c943db92cf6)

doc/src/yaccextensions.md
ratmice: Should be fixed in [7c21a9d](https://github.com/softdevteam/grmtools/pull/499/commits/7c21a9df013bb73600cdd8c7bd647c943db92cf6)

doc/src/lexextensions.md
ltratt: Formatting this sort of thing, especially in Markdown, is always a challenge. I think a table is not a bad idea but the "extra description rows" confused me at first. One way around this might be something like:

```
| `allow_wholeline_comment` [*] | bool  | &cross;  | &cross;       |
...

[*] Enables c++ style `// comments` at the start of lines. Note: The `/` character may require escaping when used in a regex.
```

Maybe using † instead of `[*]` or whatnot?

doc/src/yaccextensions.md
```diff
@@ -3,8 +3,15 @@
...
 But a default can be set or forced by using a `YaccKindResolver`.
 
-## Required flags
-    * yacckind
+| Flag                      | Value     | Required     |
+|---------------------------|-----------|--------------|
+| `yacckind`                | YaccKind  | &checkmark;  |
+
+`YaccKind` Values
+* Grmtools
```
ltratt, line 11: Perhaps we should just link to the complete `YaccKind` descriptions in https://softdevteam.github.io/grmtools/master/book/yacccompatibility.html#yacckinds -- or maybe move them here or ... ?

ratmice: I think the main reason I was including it directly here, is because of the parsing code,
It doesn't take the `YaccKind::` or the `YaccOriginalActionKind::` rust fully qualified module paths
and stuff (at least at the moment!).

In the same way that *nimbleparse* takes a `YaccKind` and converts it to a string.
So this documentation was kind of trying to be like the `nimbleparse` help text which also duplicates the names of the kinds.

I should probably make it also accept `YaccKind` qualified paths, and link to that?

ltratt: Works for me!

doc/src/lexextensions.md
```diff
@@ -0,0 +1,59 @@
...
+```
+%grmtools {
+    allow_wholeline_comments
+    !octal
+    size_limit 1024
+}
+%%
+. "rule"
+```
+## List of flags:
```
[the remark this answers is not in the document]
ratmice: Should be fixed in [62b616f](https://github.com/softdevteam/grmtools/pull/499/commits/62b616f090e397c60f57c54a1d3802c8fc2a0e90)

doc/src/lexextensions.md
ratmice: mdbook doesn't actually seem to preserve the footnote character unfortunately, but should be fixed in [62b616f](https://github.com/softdevteam/grmtools/pull/499/commits/62b616f090e397c60f57c54a1d3802c8fc2a0e90)

doc/src/yaccextensions.md
ltratt: Do we need to add test(s) for the code in be067d9?

ratmice: Probably a good idea.

ratmice: Added in [91794aa](https://github.com/softdevteam/grmtools/pull/499/commits/91794aa99e7c94c7d0558dfc0b474f1272df7d91)

ltratt:
Please squash.

ratmice:
rustfmt + Squashed.

PR #500, opened 2025-03-11
Add grmtools section to `.y` files

ratmice:
During development I'd added the `%grmtools` section to a few of the examples, to test it.
But forgot to add them to all the rest, this adds those, and removes the `.yacckind` calls in `build.rs`. 

Tested using the following shell script, which never specifies a `-y` flag and also performing `cargo run` in the examples,
```
echo -n "/* /* */ */" | ./target/debug/nimbleparse lrpar/examples/start_states/src/comment.{l,y} -
echo -n 5+++ | ./target/debug/nimbleparse ./lrpar/examples/clone_param/src/param.{l,y} -
echo -n 1+1 | ./target/debug/nimbleparse ./lrpar/examples/calc_storaget/src/calc.{l,y} -
echo -n 1+1 | ./target/debug/nimbleparse ./lrpar/examples/calc_parsetree/src/calc.{l,y} -
echo -n 1+1 | ./target/debug/nimbleparse ./lrpar/examples/calc_ast/src/calc.{l,y} -
echo -n 1+1 | ./target/debug/nimbleparse ./lrpar/examples/calc_actions/src/calc.{l,y} -
```

All looked good, the only question I have is about `calc_parsetree` which when I `cargo run` the compile time version,
I don't seem to see a parse tree emitted (just the calc computation), which is not what I had expected, but I'm now uncertain of whether that behavior has changed with these recent patches.

I will need to build a prior release and check the behavior against that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members=[
     "lrpar/examples/calc_actions",
     "lrpar/examples/calc_ast",
     "lrpar/examples/calc_parsetree",
+    "lrpar/examples/calc_storaget",
     "lrpar/examples/start_states",
     "lrpar/examples/clone_param",
     "lrtable",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ members=[
     "lrpar/examples/calc_actions",
     "lrpar/examples/calc_ast",
     "lrpar/examples/calc_parsetree",
-    "lrpar/examples/calc_storaget",
     "lrpar/examples/start_states",
     "lrpar/examples/clone_param",
     "lrtable",
--- a/README.md
+++ b/README.md
@@ -14,14 +14,12 @@ grammar and lexing definitions). First we need to create a file `build.rs` in
 the root of our project with the following content:
 
 ```rust
-use cfgrammar::yacc::YaccKind;
 use lrlex::CTLexerBuilder;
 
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -10,6 +10,7 @@ members=[
     "lrpar/examples/calc_actions",
     "lrpar/examples/calc_ast",
     "lrpar/examples/calc_parsetree",
+    "lrpar/examples/calc_storaget",
```
ratmice, line 13: Whoops, this one doesn't actually exist in the repository, and adding it belongs to a separate PR most likely.

ratmice:
So, I did test every `0.x.0` release from `0.13.0` to `0.5.0` and all had the same output for `cargo run` in `calc_parsetree`.
So I believe that this is behaving correctly, and perhaps just me misunderstanding the expected behavior.

ratmice:
I *think* that should cover all of the examples/documentation, assuming we want to go so far as removing the places where it was set but could be inferred.

ltratt:
> the only question I have is about calc_parsetree which when I cargo run the compile time version, I don't seem to see a parse tree emitted

The idea of `calc_parsetree` is to use grmtools' generic parse tree internally: in essence the `calc_*` examples are mostly intended to be "let's show you different ways of doing the same thing".

PR #502, opened 2025-03-11
Add regex error source to `RegexError`

ratmice:
Marking as a draft, because this is perhaps questionable.

Here is the example source producing the error.

```
%grmtools {
  !unicode
}
%%
. "dot"
```

There are two questionable things here, the first being the `Eq` implementation relying on `regex::Error`'s `to_string()` impl,
because `regex::Error` is only `PartialEq`.

The second being when we run the example through *nimbleparse* we see the `regex::Error` includes lrlex's wrapping of the regular expression e.g. `\A(:?`  unfortunately, regex doesn't really give us any kind of Span we otherwise might be able to use to translate this back to the lex source locations.

Previously this just produced "Invalid regular expression", with this change it now produces a reason.
So, I'm a bit torn on this because I do feel like the extra information is helpful.

```
[Error] in bad.l
    5| . "dot"
        Invalid regular expression: regex parse error:
        \A(?:.)
             ^
    error: pattern can match invalid UTF-8
```

### DIFF
--- a/lrlex/src/lib/mod.rs
+++ b/lrlex/src/lib/mod.rs
@@ -46,7 +46,7 @@ pub struct LexBuildError {
 impl Error for LexBuildError {}
 
 /// The various different possible Lex parser errors.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
 pub enum LexErrorKind {
     PrematureEnd,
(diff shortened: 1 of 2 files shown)


ltratt:
This might be a dumb question but do we *need* `PartialEq` on `LexErrorKind`? I can't, off the top of my head, immediately think why it's there.

ratmice:
It looks like it is used for merging spans of duplicate occurrences so that you get an error at a bunch of places,
rather than a bunch of errors each having a place, I think I added it, and for that reason.

ratmice:
It was also used in the testsuite, so I went with the more generic `is_same_kind()`.

ltratt:
Please squash (and rebase to remove the conflict).

ratmice:
Squashed.

PR #503, opened 2025-03-11
Refactor lrlex codegen using quote

ratmice:
Here is my first attempt to do a more holistic refactoring of the code gen of lrlex using the Quote library.
When using the string formatting, we generate the code from top-down.

When using quote it takes a more inside-out approach generating the module contents, or a functions block, then plugging that in place.

In particular I tried to organize it such that we reduce the number of variables in scope.

So the `lexerdef` function impl block is generated over few blocks of code with only those variables in scope when they are needed. We could reduce that down to e.g. generating the function impl over a single block, but it increases the number of variables in scope and there *is* quite a few because of the `lex_flags`.

I don't really have a preference both choices seemed like a trade-off to me.


ltratt:
This looks like an improvement to me, so am happy to merge!

PR #504, opened 2025-03-12
feedback on `mod _cache_information_` instead of `CACHE_INFORMATION` comment.

ratmice:
This moves away from a `/* CACHE INFORMATION` comment. Instead it uses a module and places all the data into constant values. Because the `quote` crate relies heavily on proc-macro2 token types which lacks a comment token. So it doesn't actually have any representation for comments.

One trick is to the attribute for a doc comment but hide the doc comment in some inner scope.
This is fairly hacky, and inconvenient because rustdoc still will want it to be properly formatted documentation, there is also the possibility that it the tools become smart enough to figure out that these doc strings are unreachable.

```
fn foo() {
	#[doc = "we can use a doc comment via an attribute in place of a normal comment"]
	mod _ignore_this_module_{}
}
```

Instead this patch migrates the cache information to a module with constant values. This module is nested within the `lrpar_mod!` module in order to avoid introducing multiple modules with the `cache_information` name at top-level when generating multiple parsers.

```
mod mod_name_y {

   #[allow(unused)]
   mod cache_information {
      ...

      const ERROR_ON_CONFLICTS: bool = true;
      ...
   }
}
```


lrpar/src/lib/ctbuilder.rs
```diff
@@ -757,43 +771,45 @@ where
...
+        let rule_map_len = rule_map.len();
+        let cache_module = quote! {
+            #[allow(unused)]
+            mod _cache_information_ {
+                use ::lrpar::{RecoveryKind, Visibility};
+                use ::cfgrammar::yacc::YaccKind;
+
+                const BUILD_TIME: &'static str = #build_time;
+                const GRAMMAR_PATH: &'static str = #grammar_path;
+                const MOD_NAME: Option<&'static str> = #mod_name;
```
ratmice, line 804: This is currently returning `None` unless explicitly set, because
this call happens *before* it defaults to the filename with the extension removed.

So currently we can see `mod foo_y { const MOD_NAME = None; ... }`
But that seems to go for the current CACHE comment.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -757,43 +771,45 @@ where
...
+                const RECOVERER: RecoveryKind = #recoverer;
+                const YACC_KIND: YaccKind = #yacckind;
+                const VISIBILITY: &'static str = stringify!(#visibility);
```
ratmice, line 807: I wasn't totally sure the appropriate thing to do here for visibility.
I actually kind of found it insightful.  in the sense that here we actually probably want 

`const VISIBILITY: Visibility`, but elsewhere we want `Visibility` to convert to the respective `pub` syntactical visibility modifier.

I ended up using `stringify!` here, so e.g. `PublicCrate` turns into `"pub(crate)"`.
But I think it'd be better to maybe have an `impl Visibility` function `to_token_variant`, which would convert to `Visibility::PublicCrate` instead.

Probably overthinking it but it is an interesting case where we ourselves may want to tokenize the same type in multiple ways.

ratmice: We can kind of get away with out a newtype, see [f425a4e](https://github.com/softdevteam/grmtools/pull/504/commits/f425a4eb9029d639b56c25482df26d7f0be81681)

lrpar/src/lib/ctbuilder.rs
```diff
@@ -190,6 +190,20 @@ impl ToTokens for Visibility {
...
 
+impl Visibility {
+    fn to_variant_tokens(&self, tokens: &mut TokenStream) {
+        tokens.extend(match self {
+                Visibility::Private => quote!(::lrpar::Visibility::Private),
+                Visibility::Public => quote!(::lrpar::Visibility::Public),
+                Visibility::PublicSuper => quote!(::lrpar::Visibility::PublicSuper),
+                Visibility::PublicSelf => quote!(::lrpar::Visibility::PublicSelf),
+                Visibility::PublicCrate => quote!(::lrpar::Visibility::PublicCrate),
+                Visibility::PublicIn(data) => quote!(::lrpar::Visibility::PublicIn(#data)),
```
ratmice, line 201: This needs the `QuoteToString` type from `lrlex::ctbuilder`.
But even then that won't work because `to_string()` cannot be used in a const context.

ratmice: So really the only way I know how to fix that is to either include a lifetime `Visibility<'a>` so we can say `Visibility::PublicIn(&'a str)`  or change it to `Visibility::PublicIn(&'static str)`.

Maybe It is best to just go back to `VISIBILITY: &'static str = stringify(#visibility)`.

ratmice:
FWIW, been doing local testing using the following patch, this PR is now in a state where `PublicIn` still works.
But does change the lifetime of that struct.  However `mod_name` is also an `&'a str`, so it is thankfully not a huge task to add that lifetime.

diff --git a/lrpar/examples/calc_actions/build.rs b/lrpar/examples/calc_actions/build.rs
index c441cb01..e40c6d3b 100644
--- a/lrpar/examples/calc_actions/build.rs
+++ b/lrpar/examples/calc_actions/build.rs
@@ -8,6 +8,7 @@ fn main() {
         .rust_edition(lrlex::RustEdition::Rust2021)
         .lrpar_config(|ctp| {
             ctp.rust_edition(lrpar::RustEdition::Rust2021)
+                .visibility(lrpar::Visibility::PublicIn("crate"))
                 .grammar_in_src_dir("calc.y")
                 .unwrap()
         })

lrpar/src/lib/ctbuilder.rs
ltratt: Good point. In retrospect, it *probably* would be better to initialise the field with the default name and allow it to be overridden. The only issue with that is that we can't catch the user overriding it multiple times but I think I was probably worrying unduly about that.

Summary: I think we should consider changing `CTBuilder`'s behaviour, but that could be in a separate PR.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -757,43 +771,45 @@ where
...
+            })
+            .collect::<Vec<_>>();
+        let rule_map_len = rule_map.len();
+        let cache_module = quote! {
+            #[allow(unused)]
+            mod _cache_information_ {
+                use ::lrpar::{RecoveryKind, Visibility};
+                use ::cfgrammar::yacc::YaccKind;
+
+                const BUILD_TIME: &'static str = #build_time;
```
ratmice, line 817: Both the old cache code, and this PR are missing the `RustEdition` variable,
I'll do that in a follow up patch if that is okay.

lrpar/src/lib/ctbuilder.rs
ltratt: Hmm, the `'a` lifetime here could make the `enum` a pain to use in some situations for external users?

lrpar/src/lib/ctbuilder.rs
ltratt: Works for me.

lrpar/src/lib/ctbuilder.rs
ltratt: I agree that `'a` and `'static` feel like they're pushing an internal detail (how quote works) to external users, and that if we can avoid burdening external users it would be better.

ratmice: Just to clarify it isn't exactly a quote implementation detail.
for instance we can reproduce the issue entirely without quote by trying to construct it in a `const` fashion

```
  const V: Visibility = Visibility::PublicIn("crate".to_string());
```

This is basically running into the same error as the following, because string isn't `const` due to heap allocations.
```
const X: String = "y".to_string();
```

It occurs to me that the following is fine:
```
quote!{
    fn visibility() -> Visibility {
       ::lrpar::Visibility::PublicIn(#data.to_string())
    }
}
```

So perhaps an alternative to `stringify!` is to make this case a function.

ltratt: > So perhaps an alternative to stringify! is to make this case a function.

If so, that sounds good!

ratmice: Should be fixed in [7e9a17b](https://github.com/softdevteam/grmtools/pull/504/commits/7e9a17b901930fc96bff9df295f05fcceda22522)

ltratt:
Please squash.

ratmice:
Squashed.

PR #505, opened 2025-03-12
Add `RustEdition` entry to cache info. Simplify `Visibility` entry.

ratmice:
In adding the `rust_edition` entry for the cache, it occurred to me we could also simplify the `to_variant_tokens` impl for `Visibility` a tiny bit.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -184,6 +184,16 @@ pub enum RustEdition {
     Rust2021,
 }
 
+impl RustEdition {
+    fn to_variant_tokens(self) -> TokenStream {
+        match self {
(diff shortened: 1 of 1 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -184,6 +184,16 @@ pub enum RustEdition {
     Rust2021,
 }
 
+impl RustEdition {
+    fn to_variant_tokens(&self) -> TokenStream {
```
ratmice, line 188: Can just be `self` because copy.

ratmice: Fixed in [294ce8e](https://github.com/softdevteam/grmtools/pull/505/commits/294ce8e1df172c1c60e45992c8dfb6ffdb426b07)

ltratt:
Please squash.

ratmice:
Squashed.

PR #507, opened 2025-03-12
Add warnings_are_errors and show_warnings to the cache info.

ratmice:
Also rewrite the cache info variable accessors to be more fragile. When done in the `let Self{ ... } = self;` style, this code should now cause errors and lints as new fields get added.

The comment I added mirrors the one in the struct [decl](https://github.com/softdevteam/grmtools/blob/master/lrpar/src/lib/ctbuilder.rs#L236-L238) but be more local to where the above fragility will happen.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -811,14 +811,27 @@ where
         // Record the time that this version of lrpar was built. If the source code changes and
         // rustc forces a recompile, this will change this value, causing anything which depends on
         // this build of lrpar to be recompiled too.
+        let Self {
+            // All variables except for `output_path` and `phantom` should
+            // be written into the cache.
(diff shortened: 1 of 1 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -811,14 +811,27 @@ where
         // Record the time that this version of lrpar was built. If the source code changes and
         // rustc forces a recompile, this will change this value, causing anything which depends on
         // this build of lrpar to be recompiled too.
+        let Self {
```
ltratt, line 814: That is a very clever idiom!

ratmice:
Should be fixed, by rustfmt but will need a squash.

ltratt:
Please squash.

ratmice:
Squashed.

PR #510, opened 2025-03-13
refactor `gen_parse_function` and `output_file` using quote

ratmice:
I'm going to just make this a draft, and write down some scattered thoughts.

I kind of think that I probably am doing myself a disservice by migrating `output_file` before all the others.
Because that then makes me use the whole `str::parse::<TokenStream>(string)` pattern.

Unlike lrlex, in lrpar we have user actions and user provided code snippets, so we're going to have to 
use the `str::parse<TokenStream>(user_action)?` there, which presumably will end up doing some syntax checking
during this code gen phase.

But here is at least progress on migrating the parser generating code to `quote`.
Just for feedback/etc...

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -8,13 +8,14 @@ use std::{
     fmt::{self, Debug, Write as fmtWrite},
     fs::{self, create_dir_all, read_to_string, File},
     hash::Hash,
-    io::{self, Write},
+    io::Write,
     marker::PhantomData,
     path::{Path, PathBuf},
     sync::Mutex,
 };
 
-use bincode::{deserialize, serialize_into};
+use crate::{LexerTypes, RecoveryKind};
+use bincode::serde::{decode_from_slice, encode_to_vec};
 use cfgrammar::{
     newlinecache::NewlineCache,
     yacc::{
@@ -27,11 +28,9 @@ use lazy_static::lazy_static;
 use lrtable::{from_yacc, statetable::Conflicts, Minimiser, StateGraph, StateTable};
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
 use proc_macro2::{Literal, TokenStream};
-use quote::{quote, ToTokens, TokenStreamExt};
+use quote::{format_ident, quote, ToTokens, TokenStreamExt};
 use regex::Regex;
 use serde::{de::DeserializeOwned, Serialize};
-
-use crate::{LexerTypes, RecoveryKind};
 
 const ACTION_PREFIX: &str = "__gt_";
 const GLOBAL_PREFIX: &str = "__GT_";
@@ -40,9 +39,6 @@ const ACTIONS_KIND_PREFIX: &str = "Ak";
 const ACTIONS_KIND_HIDDEN: &str = "__GtActionsKindHidden";
 
 const RUST_FILE_EXT: &str = "rs";
-
-const GRM_CONST_NAME: &str = "__GRM_DATA";
-const STABLE_CONST_NAME: &str = "__STABLE_DATA";
 
 lazy_static! {
     static ref RE_DOL_NUM: Regex = Regex::new(r"\$([0-9]+)").unwrap();
@@ -583,7 +579,7 @@ where
                     > FileTime::from_last_modification_time(inmd)
                 {
                     if let Ok(outc) = read_to_string(outp) {
-                        if outc.contains(&cache) {
+                        if outc.contains(&cache.to_string()) {
                             return Ok(CTParser {
                                 regenerated: false,
                                 rule_ids,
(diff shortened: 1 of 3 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -862,160 +873,136 @@ where
...
-        let mut outs = String::new();
-
+    ) -> Result<TokenStream, Box<dyn Error>> {
+        let mut consts = String::new();
         // bincode format is serialized into constants which the generated
         // source code.
-        serialize_bin_output(grm, GRM_CONST_NAME, &mut outs)?;
-        serialize_bin_output(stable, STABLE_CONST_NAME, &mut outs)?;
+        serialize_bin_output(grm, GRM_CONST_NAME, &mut consts)?;
+        serialize_bin_output(stable, STABLE_CONST_NAME, &mut consts)?;
```
ratmice, line 889: So, here is one case worth keeping this in draft, because we're kind of jumping through circus hoops.
Generating these consts to source, then parsing that source to tokens, and then running the result through quote.

I can't imagine it is good given that these can be decently sized.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -754,57 +753,68 @@ where
...
+                pub use _parser_::*;
+                #[allow(unused_imports)]
+                use ::lrpar::Lexeme;
+            } // End of `mod {#mod_name}`
+        };
 
         let mut f = File::create(outp_rs)?;
+        let syntax_tree = syn::parse_str(&out_tokens.to_string()).unwrap();
+        // This unfortunately reacts pretty badly to the consts, generating many lines
+        let outs = prettyplease::unparse(&syntax_tree);
```
ratmice, line 809: Filed https://github.com/dtolnay/prettyplease/issues/105 regarding this comment.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -754,57 +753,69 @@ where
...
+
+                mod _parser_ {
+                    #![allow(clippy::type_complexity)]
+                    #![allow(clippy::unnecessary_wraps)]
+                    #![deny(unsafe_code)]
+                    #[allow(unused_imports)]
+                    use super::*;
+                } // End of `mod _parser_`
+
+                #parse_function
```
ratmice, line 794: Doesn't look right, I think `#parse_function` is supposed to be moved up into the `_parser_` module.
This also explains why I had to add `clippy::type_complexity` to the `parse_function`.

lrpar/src/lib/ctbuilder.rs
ltratt: Is this just a case of changing what `serialize_bin_output` does? Ultimately it's just creating a big array/vec, so should be fairly easy to change if performance is a problem.

ltratt:
Broadly speaking, this looks like it's heading in the right direction.

lrpar/src/lib/ctbuilder.rs
ratmice: fixed via [5ce16e3](https://github.com/softdevteam/grmtools/pull/510/commits/5ce16e3613798f284d62b89a06afa541e8c5e649)

lrpar/src/lib/ctbuilder.rs
ratmice: I noticed that upgrading bincode had these newer functions `encode_to_vec` which seemed like exactly
what we want.  Note that I kept the serde dependency, but there is also a `bincode::Encode`/`bincode::Decode` derives, might be a win but for another day.

This did however seem to increase the minimum rust compiler to 1.85.0 which is pretty new.

[3f0a91b](https://github.com/softdevteam/grmtools/pull/510/commits/3f0a91ba5bed9f3cda239697c1a26b186382a20a)

lrpar/src/lib/ctbuilder.rs
```diff
@@ -990,7 +943,50 @@ where
...
+            },
+            YaccKind::Eco => unreachable!(),
+        };
+        let grm_const_name = format_ident!("{}", GRM_CONST_NAME);
+        let stable_const_name = format_ident!("{}", STABLE_CONST_NAME);
+        let grm_data = encode_to_vec(grm, bincode::config::legacy())?;
+        let stable_data = encode_to_vec(stable, bincode::config::legacy())?;
+        Ok(quote! {
+            const #grm_const_name: &[u8] = &[#(#grm_data,)*];
+            const #stable_const_name: &[u8] = &[#(#stable_data,)*];
```
ratmice, line 981: At this point, all the uses of these `__GRM_DATA` AND `__STABLE_DATA` consts,
which the `GRM_CONST_NAME` and `STABLE_CONST_NAME` variables refer to are within a few lines of one another.

Perhaps we should just replace these `#grm_const_name` variables with `__GRM_DATA` and remove the
indirection.

ltratt: > Perhaps we should just replace these #grm_const_name variables with __GRM_DATA directly and remove the indirection?

Works for me!

lrpar/src/lib/ctbuilder.rs
ltratt: Wow that was quick!

lrpar/src/lib/ctbuilder.rs
ltratt: At first I wasn't a big fan of having a dependency on 1.85.0, but there again, we are implicitly working towards a breaking release, at which point I think it's tolerable for us to have such a restriction.

lrpar/src/lib/ctbuilder.rs
ratmice: Should be fixed in [e2243a6](https://github.com/softdevteam/grmtools/pull/510/commits/e2243a6b0d1dd38c024af11126d91b0307489461)

Overall I'm pretty happy with how this has turned out at this point, and I would be happy with either
unmarking as draft for review, *or* pulling out just the `gen_parse_function` changes.

It don't believe the `output_file` changes when moving the rest of the `gen_` function's to returning `Result<TokenStream,..>`,  would be anything more than one line changes, so I'd be happy to proceed in
whichever fashion you prefer.

ltratt:
I think this is ready to squash and merge. Please go ahead.

ratmice:
Squashed

ratmice:
Unfortunately it occurred to me and does seem like the addition of pretty printing broke the cache.
Because pretty printing only runs on the output_file, not on the cache.
Running it on the cache alone would lead to different indentation.
Thus it always fails since the quote token stream output always gets changed by the pretty printing.

I can think of a few options

- moving back to a comment, i.e. `tokens.tostring()` and having a small amount of
string formatting.
- making the cache check whitespace irrelevant so we would strip whitespace from the file and the cache when performing the cache check.
- Disabling the pretty printing? (If you need to read the generated code run a pretty printer manually?)

I know rustfmt can actually affect comments, I'd kind of doubt though that would be the case for whole line/block comments like the cache previously was.  But it can affect cases like below

```
 .foo() // Comments such as this can be moved to their own lines.
```

FWIW, I've also other issues with pretty printing when it affects user action code, because it can change the spans deviating from the spans in the provided text.

ltratt:
> making the cache check whitespace irrelevant

If that fixes things, it seems like an easy win!

ratmice:
> > making the cache check whitespace irrelevant
> 
> If that fixes things, it seems like an easy win!

It does, but perhaps it is too fragile.  The pretty printing algorithm can make non-whitespace changes, like introducing trailing commas on the last element of an array.  However we already emit the trailing comma where it would add one.

Similarly there is the inverse rule in use statements, where we don't have a trailing comma at the end of a use
statement, but if we added it the pretty printer would strip it off.

i.e. both
```
use ::lrpar::{RecoveryKind, Visibility, RustEdition};
```

and 
```
use ::lrpar::{RecoveryKind, Visibility, RustEdition,};
````

turn into 
```
use ::lrpar::{ RecoveryKind, Visibility, RustEdition };
```

**Comment by ltratt** [comment]
The thing about caching is that while we would ideally like to never build unnecessarily, if it does happens sometimes, it's not the end of the world -- it's just inefficient. The nightmare scenario, though, is not to rebuild when we should. So if we're just talking about "we sometimes rebuild too often because commas get added/removed", we could live with that. [We could even have the caching check strip away commas, perhaps, but I doubt that's really necessary.]

**Comment by ratmice** [comment]
Yeah, I'm mostly just hoping to avoid things like, we change the codegen output, and suddenly break it _every_ time
like I did here, but fail to notice for a while.

I *think* we can just avoid that by adding an environment variable:
`GRMTOOLS_VERIFY_CACHE=true` which asserts the result of the cache check is the same as the value in the env var.
Then run the `cttests` and build phase twice the second time setting the env var.

That in theory catches the "not the end of the world case" but in cases like this one where I broke it in such a way that it fails the cache check _every_ time.

**Comment by ratmice** [comment]
It is closer to the "sometimes failed unnecessarily", except it is "always fails unnecessarily".
So in this case, because the pretty printer always makes changes that cause the cache check to fail.
We'll always unnecessarily rebuild unnecessarily.

So I'm hoping to just have a general sanity check that says something like
if you `cargo run && touch build.rs && CACHE_EXPECTED=true cargo run` the env-var causes an error if the code was regenerated during the second call.

So it's fully within the `it's inefficient`, class of problem.

**Comment by ltratt** [comment]
So is the env var just to make the test suite not rebuild each time? We don't expect normal users to know about it / use it?

**Comment by ratmice** [comment]
The thing I'm trying out does:
"If the cache check is negative, and the environment variable is set, panic/abort regeneration of the output file"
Yeah, normal users never need to see or know about it, it is just for sanity testing in the testsuite

**Comment by ltratt** [comment]
So it would be protected by `#[cfg(test)]` or similar?

**Comment by ratmice** [comment]
Not exactly, I was trying to do it with the examples/.buildbot.sh which I don't think runs under `#[cfg(test)]` 
And I'm actually unsure that `build.rs` even gets `#[cfg(test)]` set when running `cargo test`.

```
echo "2 + 3 * 4" | CACHED_EXPECTED=true cargo run | grep "Result: 14"
```

But seems like that isn't enough to get it to actually rerun the build script subsequently triggering the cache check.
So I need to work on figuring out the right way to trigger all of this. (There is also a possibility that I am just confused!)

```
                        if  outc.contains(&cache) {
                            return Ok(CTParser {
                                regenerated: false,
                                rule_ids,
                                conflicts: None,
                            });
                        } else if std::env::var("CACHE_EXPECTED").is_ok() {
                            // For smoke testing that if we don't change anything the cache validation check should succeed
                            panic!("The cache regenerated however, it was expected to match the output_file's fingerprint");
                        }
```

**Comment by ratmice** [comment]
Aha, so the magic incantation seems to be one of the following:
The problem I was having was either not triggering the `build` script executable (touch src/calc.y alone)
triggering the build script executable but *also* rebuilding it thus invalidating the cache with a timestamp, (touch build.rs) alone

So touching main causes it to rerun, but not rebuild the build script itself.

```
touch src/main.rs && echo "2 + 3 * 4" | CACHED_EXPECTED=true cargo run | grep "Result: 14"
touch src/main.rs && touch src/calc.y && echo "2 + 3 * 4" | CACHED_EXPECTED=true cargo run | grep "Result: 14"
```

PR #511, opened 2025-03-14
Fix cache check, add testing mechanism for it

ratmice:
I don't really have buildbot to actually fully test under that however:

   Fix cache check, add testing mechanism for it
    
    In #510 I broke the cache check so it always regenerated the output source.
    Because pretty printing caused changes to the cache module, and the cache
    check uses string equality.
    
    Technically the pretty printer can cause non-whitespace changes, but it is
    stable/deterministic. It isn't going to *randomly* decide to append a
    trailing comma.
    
    So, it is fragile in a way that can be tested, so this adds an environment
    variable to force that the cache check returns true, and runs this from
    the buildbot script.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -4,6 +4,7 @@ set -e
 
 export CARGO_HOME="`pwd`/.cargo"
 export RUSTUP_HOME="`pwd`/.rustup"
+export RUSTFLAGS="--cfg grmtools_extra_checks"
 
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
 sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain stable -y --no-modify-path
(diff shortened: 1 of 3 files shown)


ltratt:
I like the idea of this check in `.buildbot.sh`! We need to make sure users can't see this code, though, but the idea is sound!

.buildbot.sh
```diff
@@ -23,12 +23,17 @@ cargo test --lib lrpar --features serde
 root=`pwd`
 cd $root/lrlex/examples/calc_manual_lex
 echo "2 + 3 * 4" | cargo run | grep "Result: 14"
+# Touching these files shouldn't invalidate the cache.
+touch src/main.rs src/calc.y && echo "2 + 3 * 4" | CACHE_EXPECTED=y cargo run | grep "Result: 14"
```
ratmice, line 27: I think with #512 we can remove this src/main.rs

lrpar/src/lib/ctbuilder.rs
```diff
@@ -578,13 +578,21 @@ where
...
+                        let mut cache = cache.to_string();
+                        outc.retain(|c| !c.is_whitespace());
+                        cache.retain(|c| !c.is_whitespace());
+                        if outc.contains(&cache) {
                             return Ok(CTParser {
                                 regenerated: false,
                                 rule_ids,
                                 conflicts: None,
                             });
+                        } else if std::env::var("CACHE_EXPECTED").is_ok() {
```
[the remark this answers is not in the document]
ratmice: I went ahead and added a `#[cfg(grmtools_extra_checks)]`  for the buildbot to set via --cfg/RUSTFLAGS,
as I don't think we want to really advertise via features in `Cargo.toml` etc.

.buildbot.sh
ratmice: I went ahead and just left this in though, because it doesn't hurt to leave it?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -584,13 +584,17 @@ where
...
+                        #[cfg(not(grmtools_extra_checks))]
+                        const CACHE_HIT_EXPECTED: bool = false;
                         if outc.contains(&cache) {
                             return Ok(CTParser {
                                 regenerated: false,
                                 rule_ids,
                                 conflicts: None,
                             });
-                        } else if std::env::var("CACHE_EXPECTED").is_ok() {
+                        } else if CACHE_HIT_EXPECTED {
```
ltratt, line 597: Might this be simpler as something like:

```rust
if outc.contains(...) { ... }
else {
   #[cfg(grmtools_extra_checks)]
   if std::env::vars("CACHE_EXPECTED").is_ok() { ... }
}
```

?

ratmice: Used that in [c0311e9](https://github.com/softdevteam/grmtools/pull/511/commits/c0311e9d4336bb89ffbce9cd7902fa4b09a80652)

ltratt:
Please squash.

ratmice:
Squashed.

PR #512, opened 2025-03-14
Emit cargo:rerun_if_changed for lex and yacc files

ratmice:
This allows `touch src/calc.{l,y}` alone to rebuild the project/rerun the build script.
It is gated so this only gets emitted when running under cargo.


ltratt:
Good idea, because if these live outside the source directory, they normally wouldn't be picked up.

PR #513, opened 2025-03-14
reintroduce the cache breakage to test the ci changes

ratmice:
This is just testing previous CI changes catch things, do not merge!

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -600,7 +600,7 @@ where
                     > FileTime::from_last_modification_time(inmd)
                 {
                     if let Ok(outc) = read_to_string(outp) {
-                        if outc.contains(&cache.to_string()) {
+                        if outc.contains(&cache.to_string().to_uppercase()) {
                             return Ok(CTParser {
                                 regenerated: false,
                                 rule_ids,


ltratt:
Testing this locally, alas, is a bit tricky. I suppose one could use (say) `sed` to inject a deliberate mistake into the code and check it's caught?

ratmice:
Welp, I'm *definitely* confused, and the way I have been thinking about the cache is slightly backwards from the way it is documented to work. It also doesn't help that in the interim I stumbled upon a known cargo footgun, in emitting `rerun-if-changed`.

rerun-if-changed apparently disables the automatic `src/*.rs` behavior i.e. if you set it once, you also have to set it for the rest of the `src/` dir, so now that we emit that we ensured a catch-22, the cache will never be checked because the only way to run the check is to modify the input grammar, which then causes the cache check to not be run.


a) remove the `rerun-if-changed`
b) my misunderstanding of the caache check involves the flipping of the modification times check [here](https://github.com/softdevteam/grmtools/blob/master/lrpar/src/lib/ctbuilder.rs#L602-L603)
e.g. to 
```
if FileTime::from_last_modification_time(out_rs_md)
                    < FileTime::from_last_modification_time(inmd)
```

I.e. I thought it was 'if the input is newer than the output` but the cache is unmodified we can skip regeneration.

But the comment seems to be describing something else, where the output is newer than the input, and we may still need to regenerate.  But I don't understand because the subsequent branches don't seem to be able to cause it to regenerate, only skip the regeneration.

So, I'm uncertain whether that comparison should be flipped e.g. to `<`, or i'm just too confused.
Sorry :cry: :facepalm:

ratmice:
So above and beyond that for some reason I don't actually see `touch src/main.rs` rerunning `build.rs`
So, I'm also confused about how to even get this executable where the cache check is to run reliably in a way which doesn't invalidate the cache (like  how touching build.rs would modify the vergen timestamp would).

Anyhow I think that there are probably multiple layers to my confusion, and I wouldn't be surprised to learn that
some change in cargo has made these cache check branches more difficult to reach since this code was written.

ratmice:
Well, I'm not sure what more I can try, I feel confident that the .buildbot.sh check *should* be working,
and the change in this series should be triggering it, as is the case running `.buildbot.sh` locally from this branch.

As such I'll probably just close this but do let me know if there are any ideas.

PR #517, opened 2025-03-14
Refactor `gen_wrappers` using the quote crate

ratmice:
This adds a few comments, and some TODO comments which I believe can clean up some clippy workarounds
in this code.  I didn't want to both refactor it to using quote while also making a semantic change.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -780,7 +780,7 @@ where
         let parse_function = self.gen_parse_function(grm, stable)?;
         let action_wrappers = match self.yacckind.unwrap() {
             YaccKind::Original(YaccOriginalActionKind::UserAction) | YaccKind::Grmtools => {
-                Some(str::parse::<TokenStream>(&self.gen_wrappers(grm))?)
+                Some(self.gen_wrappers(grm)?)
             }
             YaccKind::Original(YaccOriginalActionKind::NoAction)
             | YaccKind::Original(YaccOriginalActionKind::GenericParseTree) => None,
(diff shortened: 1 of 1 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -1042,161 +1042,164 @@ where
...
-                quote!(#pidx_num),
-                storaget = type_name::<StorageT>(),
-                lexertypest = type_name::<LexerTypesT>(),
-                prefix = ACTION_PREFIX,
-                parse_paramdef = parse_paramdef,
-                actionskind = ACTIONS_KIND,
-                edition_lifetime = if self.rust_edition != RustEdition::Rust2015 { "'_, " } else { "" },
-            ).ok();
-
+            let wrapper_fn = format_ident!("{}wrapper_{}", ACTION_PREFIX, usize::from(pidx));
```
ratmice, line 1071: When writing all of this, I had kind of been under the notion that
`format_ident!()` didn't support `{variable}` style of format strings.

But that is not actually the case, that type of format string appears to work fine,
I'm not sure what it was that caused me to believe that.

I tend to feel that the main places that needed `{variables}` were where we had lots of `{}` variables
which are now mostly replaced with `quotes`, which have `#variables` instead.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -1042,161 +1042,164 @@ where
...
+            // That should allow us to remove parse_param_unit and the associated branches.
+            wrappers.extend(quote!{
+                fn #wrapper_fn<'lexer, 'input: 'lexer>(
+                    #ridx_var: ::cfgrammar::RIdx<#storaget>,
+                    #lexer_var: &'lexer dyn ::lrpar::NonStreamingLexer<'input, #lexertypest>,
+                    #span_var: ::cfgrammar::Span,
+                    mut #args_var: ::std::vec::Drain<#edition_lifetime ::lrpar::parser::AStackType<<#lexertypest as ::lrpar::LexerTypes>::LexemeT, #actionskind<'input>>>,
+                    #parse_paramdef
+                ) -> #actionskind<'input> {
+
```
ltratt, line 1176: I guess the blank line is unintentional? If so, please force push an update.

ratmice: Indeed, pushed.

PR #519, opened 2025-03-15
Refactor `gen_user_actions` using quote

ratmice:
I *think* this is the last of the code gen, I'll want to mark this one as draft because this is the one that terrifies me,
with the parsing of arbitrary user_action code into `TokenStreams`.

Some of these things, for instance `$` variables we end up pushing those into one big string and 
then parsing to `TokenStreams` at the end...

Now that I am towards the end of this series, I am having this terrible anxiety that I've been throwing the testsuite,
and a bunch of *well formed* actions at it, and I'm suddenly fearful of how it will react once we start throwing arbitrary random blobs of user text at it.

I suppose I should just take consolation that even in the worst case I feel like there should be something left over that can be salvaged, which should be the case for all the non-user provided text, and at worst we can mix that in with a little bit plain old string formatting for the user_actions.

Going to mark this one as draft because I need to look through it with well rested eyes.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -771,7 +771,7 @@ where
             YaccKind::Original(YaccOriginalActionKind::UserAction) | YaccKind::Grmtools,
         ) = self.yacckind
         {
-            Some(str::parse::<TokenStream>(&self.gen_user_actions(grm)?)?)
+            Some(self.gen_user_actions(grm)?)
         } else {
             None
         };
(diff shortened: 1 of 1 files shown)


ltratt:
It looks good to me!

lrpar/src/lib/ctbuilder.rs
```diff
@@ -1209,20 +1209,27 @@ where
...
+        let mut action_fns = TokenStream::new();
         // Convert actions to functions
-        outs.push_str("\n    // User actions\n\n");
-        let (parse_paramname, parse_paramdef) = match grm.parse_param() {
-            Some((name, tyname)) => (name.to_owned(), format!("{}: {}", name, tyname)),
-            None => ("()".to_owned(), "_: ()".to_owned()),
+        let (parse_paramname, parse_paramdef, parse_param_unit);
+        match grm.parse_param() {
+            Some((name, tyname)) => {
+                parse_param_unit = tyname.trim() == "()";
```
ratmice, line 1223: pedantic, but being user provided this can potentially be "( )", with a space in the between parenthesis,
so this probably better stripping all whitespace.  I suppose the existing code suffered the same problem though.

Not much of a problem really, the worst that happens is the generated code emits a warning.
I'm inclined to just leave this for a future cleanup, to avoid introducing changes in this refactor.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -1322,9 +1321,24 @@ where
                 }
             }
 
-            outs.push_str("\n    }\n\n");
+            let action_args = str::parse::<TokenStream>(&outs)?;
```
ratmice, line 1324: I think I'd rename this `action_body`, while it is parsing the `$args` it is really doing so within the body.

ratmice: Fixed in [bc995e0](https://github.com/softdevteam/grmtools/pull/519/commits/bc995e0e415f5f450342424b819d8431c5886319)

lrpar/src/lib/ctbuilder.rs
ltratt: I also thought the same when I reviewed: I wondered if there was a neater way of asking "is this the unit type", but short of "replace all spaces with ''" I couldn't think of much! I agree it can be done in a future PR with additional tests.

ltratt:

ratmice:
Squashed.

ratmice:
So besides outstanding PRs, and the minor followup we discussed here, 
I believe this should largely conclude the code gen changes I had planned.

I think my todo list is mostly filled with smaller tasks now (in no particular order)

- Resume looking for places to add `non_exhaustive` (I think there is an error `StateTableError`  could perhaps use it)
  A more thorough examination through our public structs seems wise.
- We should probably accept `lexkind` in the lrlex grmtools section even if we don't require it.
- I'm thinking it is good to add LexKindResolver for consistency with lrpar even if we keep the default `Default(LRNonStrea...)` while there is only one kind.
- mod_name appearing as `None` in the cache when falling back to filename.
- whitespace "\ (\ )\ " within the user-provided parse param type, and a cleaner way to check if it's unit.

PR #522, opened 2025-03-15
Add a `derived_mod_name` field to the cache.

ratmice:
I'm not certain, but it seems reasonable that under particular circumstances the derived mod_name should be added to the cache.

I don't have a bug/reproducer but under the following circumstances:
- The output file path is given an explicit path.
- The mod_name is derived from an input grammar path.
- The input grammar path changes.

Thus it seems worth adding it to avoid this kind of situation.
Edit -- It is also possible there is no bug, because e.g. all the ways to change the filename change the timestamp included, I'm uncertain.  It just didn't seem harmful to add it.


ltratt:
Good catch: the module name could change without the path changing.

PR #523, opened 2025-03-15
Make various structs and enums non_exhaustive.

ratmice:


ratmice:
For now I've just added those things that seemed like they fit into the categories of:

1. Adding elements is less under our control `Visibility`, `RustEdition`...
2. We might want to add elements to, and it seemed unlikely external crates would be unlikely to match on.
3. We might want to add elements to, and they are kind of built for expansion (like SpansKind), even though external crates likely want to match on them.

There were some things that seemed like they were pretty unlikely to actually have new elements added to them,
like [`AssocKind`](https://docs.rs/cfgrammar/latest/cfgrammar/yacc/grammar/enum.AssocKind.html).
But perhaps didn't fit so well into the above categories.  I didn't try to add those.

PR #525, opened 2025-03-16
Fix cache check

ratmice:
I *think* this should run the cache check,

E.g. at least the following subset of `.buildbot.sh` now caught a bug where the pretty printer
was emitting a non-whitespace change, and we had to add a comma.

```
#! /bin/sh

set -e
export RUSTFLAGS="--cfg grmtools_extra_checks"
root=`pwd`
# Touching these files shouldn't invalidate the cache (via --cfg grmtools_extra_checks)
touch src/main.rs && CACHE_EXPECTED=y cargo build
cd $root/lrpar/examples/calc_actions
echo -n "2 + 3 * 4" | cargo run --package nimbleparse -- src/calc.l src/calc.y -
touch src/main.rs && CACHE_EXPECTED=y cargo build
cd $root/lrpar/examples/calc_ast
echo -n "2 + 3 * 4" | cargo run --package nimbleparse -- src/calc.l src/calc.y -
touch src/main.rs && CACHE_EXPECTED=y cargo build
cd $root/lrpar/examples/calc_parsetree
echo -n "2 + 3 * 4" | cargo run --package nimbleparse -- src/calc.l src/calc.y -
touch src/main.rs && CACHE_EXPECTED=y cargo build
cd $root/lrpar/examples/clone_param
echo -n "1+++" | cargo run --package nimbleparse -- src/param.l src/param.y -
cd $root/lrpar/examples/start_states
echo -n "/* /* commented out */ */ uncommented text /* */" | cargo run --package nimbleparse -- src/comment.l src/comment.y -
cd $root
```

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -599,13 +599,8 @@ where
                 if FileTime::from_last_modification_time(out_rs_md)
                     > FileTime::from_last_modification_time(inmd)
                 {
-                    if let Ok(mut outc) = read_to_string(outp) {
-                        // Strip whitespace from the output file and the cache since the copy of
-                        // the cache in the output file may be affected by pretty printing.
-                        let mut cache = cache.to_string();
-                        outc.retain(|c| !c.is_whitespace());
-                        cache.retain(|c| !c.is_whitespace());
-                        if outc.contains(&cache) {
+                    if let Ok(outc) = read_to_string(outp) {
+                        if outc.contains(&cache.to_string()) {
                             return Ok(CTParser {
                                 regenerated: false,
                                 rule_ids,
@@ -614,6 +609,8 @@ where
                         } else {
                             #[cfg(grmtools_extra_checks)]
                             if std::env::var("CACHE_EXPECTED").is_ok() {
+                                eprintln!("outc: {}", outc);
+                                eprintln!("using cache: {}", cache,);
                                 // Primarily for use in the testsuite.
                                 panic!("The cache regenerated however, it was expected to match");
                             }
@@ -644,7 +641,20 @@ where
             }
         }
 
-        self.output_file(&grm, &stable, &derived_mod_name, outp, &cache)?;
+        self.output_file(
+            &grm,
+            &stable,
+            &derived_mod_name,
+            outp,
+            &quote! {
+                // This declaration can be affected by the pretty printer.
+                // But we would hope the actual cache string is not.
+                //
+                // This is emitted for the purposes of performing the cache check.
+                // on the output source, but is not used by generated parser.
+                const _: &str = #cache;
+            },
+        )?;
         let conflicts = if stable.conflicts().is_some() {
             Some((grm, sgraph, stable))
         } else {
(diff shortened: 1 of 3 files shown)


ratmice:
With these changes, I think that #513 should consistently fail.

cfgrammar/src/lib/yacc/mod.rs
```diff
@@ -47,7 +47,7 @@ impl quote::ToTokens for YaccKind {
         tokens.extend(match *self {
             YaccKind::Grmtools => quote!(::cfgrammar::yacc::YaccKind::Grmtools),
             YaccKind::Original(action_kind) => {
-                quote!(::cfgrammar::yacc::YaccKind::Original(#action_kind))
+                quote!(::cfgrammar::yacc::YaccKind::Original(#action_kind,))
```
ratmice, line 50: I think perhaps it is best to attempt to do something less fragile here.

My thoughts are instead of emitting the cache module, constant string containing the cache module,
in the hopes that the pretty printer will ignore the string.

ratmice:
Might want to sqaush out the [a08fbdb](https://github.com/softdevteam/grmtools/pull/525/commits/a08fbdb6a3aae9b251f080201ff049df56e41ab9) change?

lrpar/src/lib/ctbuilder.rs
```diff
@@ -865,7 +865,7 @@ where
             })
             .collect::<Vec<_>>();
         let rule_map_len = rule_map.len();
-        quote! {
+        let cache_info = quote! {
             #[allow(unused)]
             mod _cache_information_ {
```
ratmice, line 870: We may want to do something else here, as we no longer actually compile this code
as we change it it could perhaps start to contain syntax errors, etc.

We could emit *both* the cache_info module, and the module text within that constant string.
Or we could just emit a formatted string more like the original comment.

ltratt: Don't we just want to output the information in one place? Perhaps I've misunderstood!

ratmice: Yeah, I guess i'm saying:

this emits a strong:
```
const CACHE_INFO: &str = "#[allow(unused)] mod _cache_information_ {...
```

But because we never actually compile `mod _cache_information_` anymore, it is just embedded in that string,
so it is both confusing, and we could make a change that would cause `mod _cache_information_` to not actually compile but not notice.

I basically just did the minimal change to check that embedding it as a string works as a proof of concept.

ltratt:
Shouldn't we remove `mod _cache_information` then?

lrpar/src/lib/ctbuilder.rs
ratmice: Should be fixed in [a687505](https://github.com/softdevteam/grmtools/pull/525/commits/a6875058c2a420a9c6848e6a35ab5c57d9dfd8d7)

lrpar/src/lib/ctbuilder.rs
```diff
@@ -646,7 +641,12 @@ where
             }
         }
 
-        self.output_file(&grm, &stable, &derived_mod_name, outp, &cache)?;
+        self.output_file(&grm, &stable, &derived_mod_name, outp, &quote!{
+            // This declaration can be affected by the pretty printer.
+            // But we would hope the actual cache string is not.
+            #[allow(unused)]
+            const CACHE_INFO: &str = #cache;
```
ratmice, line 648: Looks like we can name this `const _: &str = #cache`.

I thought this wasn't a valid constant declaration, but it looks like it is now.

ratmice: Fixed by [3b860ca](https://github.com/softdevteam/grmtools/pull/525/commits/3b860ca64fb99da92f96dc8317eee3862bb3d119)

cfgrammar/src/lib/yacc/mod.rs
ratmice: Fixed by [8e32c81](https://github.com/softdevteam/grmtools/pull/525/commits/8e32c81807fd0aab74565f33893eb99517e2a36b) and subsequent commits...

ratmice:
So, I think this is much less fragile now than trying to rely on the whitespace hack.
Probably ready for review.

ltratt:
Please squash.

ratmice:
Squashed.

ratmice:
Sorry, I really need to figure out how to configure my editor so it can conditionally automatically run rustfmt or not depending upon which directory I'm editing.

Because I have some projects that don't use it.

Will need a squash

ltratt:
Please squash.

ratmice:
Squashed.

PR #526, opened 2025-03-17
Move cache back inside a comment.

ratmice:
While prettyplease doesn't format string literals, There isn't a guarantee that it won't gain the ability to format string literals future although there are no plans to implement it.

Similarly, while it doesn't preserve comments. For the cache we can format the comment separate from the output source text. Appending the comment to the pretty printer output.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -646,14 +646,7 @@ where
             &stable,
             &derived_mod_name,
             outp,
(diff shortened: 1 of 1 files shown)


ratmice:
I am sorry that this has had so much back and forth changing the way we emit the cache.
It occurred to me that while comments in general are hard with this source formatting scheme.

For the specific case of the cache comment we can just do that whole thing separately from the code gen, and pretty printing and then print them to the same output.

PR #527, opened 2025-03-17
Use the web-time polyfill

### LINKED ISSUE #461: Make generated parsers usable on the web
As discussed in #145, currently the generated parsers are not available on WebAssembly in browsers, due to `Instant::now()` panicking. I have an application (https://cand.li) which requires me to run the generated parser on the browser, so I would like to solve this compatibility problem. I see two ways forward:
- use a poly-fill such as [web-time](https://crates.io/crates/web-time),
- modify grmtools to disable timeout, for example by creating an abstraction around the notion of "enough work has been done", and allow other conditions such as number of trials.

I do not have a strong opinion either way.

Alternatively (or in addition, also mentioned in #145, but never done), we could also slightly refactor [`parser.rs:371`](https://github.com/softdevteam/grmtools/blob/e30e481ea430bd0b05806a74396d947f76de7da1/lrpar/src/lib/parser.rs#L371) to not call `Instant::now()` when recovery is not used. That might be the cheapest way, in term of changes, but of course only brings us half way (works on the web without error recovery, but at least it would allow the traditional lex/yacc use case there). As an additional benefit, it would enable using the parser without error recovery even when `Instant::now()` is unsupported, for example in [worklets](https://developer.mozilla.org/en-US/docs/Web/API/Worklet), which I would say is strictly a good thing.

ratmice:
This adds a poly-fill crate to fix #461

I don't really know anything about wasm etc, but web_time is a complete drop in replacement for `std::time`,
On platforms where std::time is available, it is simply just a [re-export](https://github.com/daxpedda/web-time/blob/main/src/lib.rs#L185-L186) of std::time as shown in the link.

@stephanemagnenat Curious if you are still interested in testing whether this fixes things for you?

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -2,21 +2,27 @@
 
 set -e
 
-export CARGO_HOME="`pwd`/.cargo"
+export CARGO_HOME="`pwd`/.cargo_install"
 export RUSTUP_HOME="`pwd`/.rustup"
 export RUSTFLAGS="--cfg grmtools_extra_checks"
 
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
 sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain stable -y --no-modify-path
 
-export PATH=`pwd`/.cargo/bin/:$PATH
+export PATH=`pwd`/.cargo_install/bin/:$PATH
+rustup target add wasm32-unknown-unknown
+rustup target add wasm32-wasip2
+cargo install wasm-bindgen-cli
 
 cargo fmt --all -- --check
 
 rustup toolchain install stable
 rustup default stable
+
 cargo test
 cargo test --release
+cargo test --target wasm32-unknown-unknown
+cargo test --target wasm32-wasip2
 
 cargo test --lib cfgrammar --features serde
 cargo test --lib lrpar --features serde
--- a/.buildbot_dockerfile_debian
+++ b/.buildbot_dockerfile_debian
@@ -2,7 +2,8 @@ FROM debian:latest
 ARG CI_UID
 RUN useradd -m -u ${CI_UID} ci
 RUN apt-get update && \
-apt-get -y install build-essential curl procps file
+apt-get -y install build-essential curl procps file \
+rust-wasmtime nodejs
 WORKDIR /ci
 RUN chown ${CI_UID}:${CI_UID} .
 COPY --chown=${CI_UID}:${CI_UID} . .
--- a/lrpar/Cargo.toml
+++ b/lrpar/Cargo.toml
@@ -42,5 +42,8 @@ vob.workspace = true
 syn.workspace = true
 prettyplease.workspace = true
 
-[dev-dependencies]
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+web-time = "1.1.0"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 tempfile = "3.0"
--- a/lrpar/cttests/build.rs
+++ b/lrpar/cttests/build.rs
@@ -1,7 +1,7 @@
 use glob::glob;
 #[path = "src/cgen_helper.rs"]
 mod cgen_helper;
-
+use cfg_aliases::cfg_aliases;
 use cgen_helper::run_test_path;
 
 // Compiles the `*.test` files within `src`. Test files are written in Yaml syntax and have 4
@@ -13,5 +13,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     for entry in glob("src/*.test")? {
         run_test_path(entry.unwrap())?;
     }
+
+    cfg_aliases! {
+        // Platforms
+        wasm32_unknown: { all(target_arch = "wasm32", target_os="unknown", target_vendor="unknown") },
+    }
+
     Ok(())
 }
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,6 @@
+[target.wasm32-wasip2]
+runner = "wasmtime run --dir ."
+
+[target.wasm32-unknown-unknown]
+# Provided by the crate wasm-bindgen-cli.
+runner = "wasm-bindgen-test-runner"
--- a/lrpar/cttests/Cargo.toml
+++ b/lrpar/cttests/Cargo.toml
@@ -6,12 +6,16 @@ edition = "2021"
 license = "Apache-2.0/MIT"
 build = "build.rs"
 
+[lib]
+crate-type = ["cdylib"]
+
 [build-dependencies]
 cfgrammar = { path = "../../cfgrammar" }
 lrlex = { path = "../../lrlex" }
 lrpar = { path = "../" }
 glob = "0.3"
 yaml-rust2 = "0.8"
+cfg_aliases = "0.2.1"
 
 [dependencies]
 cfgrammar = { path = "../../cfgrammar" }
@@ -22,3 +26,9 @@ glob = "0.3"
 
 [dev-dependencies]
 cttests_macro = { path = "../cttests_macro" }
+
+[target.'cfg(all(target_arch = "wasm32", target_os="unknown", target_vendor="unknown"))'.dependencies]
+wasm-bindgen = "0.2.100"
+
+[target.'cfg(all(target_arch = "wasm32", target_os="unknown", target_vendor="unknown"))'.dev-dependencies]
+wasm-bindgen-test = "0.3.50"
--- a/lrpar/cttests/src/calc_wasm.rs
+++ b/lrpar/cttests/src/calc_wasm.rs
@@ -1,0 +1,80 @@
+#[cfg(wasm32_unknown)]
+use wasm_bindgen::prelude::*;
+
+use lrlex::lrlex_mod;
+use lrpar::lrpar_mod;
+
+// Using `lrlex_mod!` brings the lexer for `calc.l` into scope. By default the module name will be
+// `calc_l` (i.e. the file name, minus any extensions, with a suffix of `_l`).
+lrlex_mod!("calc_wasm.l");
+// Using `lrpar_mod!` brings the parser for `calc.y` into scope. By default the module name will be
+// `calc_y` (i.e. the file name, minus any extensions, with a suffix of `_y`).
+lrpar_mod!("calc_wasm.y");
+
+#[cfg_attr(wasm32_unknown, wasm_bindgen)]
+#[allow(unused)]
+pub fn calculate(l: &str) -> Result<u64, String> {
+    // Get the `LexerDef` for the `calc` language.
+    let lexerdef = calc_wasm_l::lexerdef();
+    if l.trim().is_empty() {
+        return Err("input is empty".to_string());
+    }
+    // Now we create a lexer with the `lexer` method with which we can lex an input.
+    let lexer = lexerdef.lexer(l);
+    // Pass the lexer to the parser and lex and parse the input.
+    let (res, errs) = calc_wasm_y::parse(&lexer);
+    if !errs.is_empty() {
+        let mut ret = String::new();
+        for e in errs {
+            use lrpar::LexParseError;
+            match e {
+                LexParseError::ParseError(e) => {
+                    let repairs_flag = !e.repairs().is_empty();
+                    ret.push_str(&format!("Error: {}\n Repairs: {}", e, repairs_flag));
+                }
+                e => ret.push_str(&format!("{}\n", e)),
+            };
+        }
+        if let Some(Err(e)) = res {
+            ret.push_str(&format!("{}\n", e));
+        }
+        return Err(ret);
+    }
+    match res {
+        Some(Ok(r)) => Ok(r),
+        Some(Err(e)) => Err(e.to_string()),
+        None => Err("Unable to parse".to_string()),
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::calculate;
+    #[cfg(wasm32_unknown)]
+    use wasm_bindgen_test::*;
+
+    #[cfg_attr(wasm32_unknown, wasm_bindgen_test)]
+    #[test]
+    fn test_calc_14() {
+        assert_eq!(calculate("2 + 3 * 4").unwrap(), 14);
+    }
+
+    #[cfg_attr(wasm32_unknown, wasm_bindgen_test)]
+    #[test]
+    fn test_lex_error() {
+        assert!(calculate("#1 + #2").is_err());
+    }
+
+    #[cfg_attr(wasm32_unknown, wasm_bindgen_test)]
+    #[test]
+    fn test_recovery() {
+        // We really want to test this recovery path, since it contains
+        // calls to `Instant::now()` which panics on `std`
+        // Thus we need to check that the `web_time` crate is working.
+        let x = calculate("1+");
+        match x {
+            Err(e) => assert!(e.contains("Repairs: true")),
+            Ok(e) => panic!("unexpectedly parsed {}", e),
+        }
+    }
+}
--- a/lrpar/cttests/src/calc_wasm.test
+++ b/lrpar/cttests/src/calc_wasm.test
@@ -1,0 +1,54 @@
+name: Test running on wasm targets
+grammar: |
+    %grmtools {yacckind Grmtools}
+    %start Expr
+    %avoid_insert "INT"
+    %expect-unused Unmatched "UNMATCHED"
+    %epp INT "Int"
+    %%
+    Expr -> Result<u64, Box<dyn Error>>:
+        Expr '+' Term {
+            $1?.checked_add($3?)
+                .ok_or_else(|| Box::<dyn Error>::from("Overflow detected."))
+        }
+        | Term { $1 }
+        ;
+
+    Term -> Result<u64, Box<dyn Error>>:
+        Term '*' Factor {
+            $1?.checked_mul($3?)
+                .ok_or_else(|| Box::<dyn Error>::from("Overflow detected."))
+        }
+        | Factor { $1 }
+        ;
+
+    Factor -> Result<u64, Box<dyn Error>>:
+        '(' Expr ')' { $2 }
+        | 'INT' {
+            parse_int($lexer.span_str($1.map_err(|_| "<evaluation aborted>")?.span()))
+        }
+        ;
+    Unmatched -> (): "UNMATCHED" { };
+    %%
+    // Any imports here are in scope for all the grammar actions above.
+
+    use std::error::Error;
+
+    fn parse_int(s: &str) -> Result<u64, Box<dyn Error>> {
+        match s.parse::<u64>() {
+            Ok(val) => Ok(val),
+            Err(_) => {
+                Err(Box::from(format!("{} cannot be represented as a u64", s)))
+            }
+        }
+    }
+lexer: |
+    %%
+    [0-9]+ "INT"
+    \+ "+"
+    \* "*"
+    \( "("
+    \) ")"
+    [\t ]+ ;
+    . "UNMATCHED"
+
--- a/lrpar/cttests/src/lib.rs
+++ b/lrpar/cttests/src/lib.rs
@@ -1,8 +1,12 @@
 #[cfg(test)]
 use cfgrammar::Span;
+
 mod cgen_helper;
 #[allow(unused)]
 use cgen_helper::run_test_path;
+
+mod calc_wasm;
+
 #[cfg(test)]
 use cttests_macro::generate_codegen_fail_tests;
 use lrlex::lrlex_mod;
--- a/lrpar/src/lib/cpctplus.rs
+++ b/lrpar/src/lib/cpctplus.rs
@@ -3,8 +3,13 @@ use std::{
     collections::HashSet,
     fmt::Debug,
     hash::{Hash, Hasher},
-    time::Instant,
 };
+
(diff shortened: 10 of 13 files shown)


ltratt:
This works for me. I do take @stephanemagnenat's point about the dangers of using a timeout: alas, the "perfect" alternative would be to track memory usage and bork if it gets too high, but that's probably tricky for other reasons. It would be good to revisit that at some point, though, because we might be able to do *something* that's a little less sucky than the current timeout.

lrpar/Cargo.toml
```diff
@@ -42,6 +42,7 @@ static_assertions.workspace = true
 vob.workspace = true
 syn.workspace = true
 prettyplease.workspace = true
+web-time = "1.1.0"
```
ratmice, line 45: In theory we could also make this an optional dependency, and only conditionally depend on it if a feature is enabled.

That duplicates the conditional compilation in the crate, but I get that some people might want to minimize dependencies.

ltratt: I must admit, I was thinking exactly the same. *Personally* I do have a mild preference for only compiling it on wasm targets, because then non-wasm users don't have to think about it as something they need to audit. But I appreciate that's extra work.

ratmice: What I was thinking was slightly different actually than compiling it conditionally per-platform
But using a feature like the following.

```
#[cfg(feature = "webpolyfills")]
use web_time::{Instant, Duration}

#[cfg(not(feature = "webpolyfills"))]
use std::{Instant, Duration};
```

While the name doesn't imply it you could enable webpolyfills on non-web platforms.
Then you could use the same build options to target either web/non-web.

I think we need this kind of feature anyways, just to enable the library from cargo.toml.
At the very least, I don't know about using target specific cfg options that involve dependencies,
that may very well be a thing though.

ltratt: I *think* we can say something like:

```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
web-time = "0.1.0"
```

and then:

```rust
#[cfg(target_arch = "wasm32")]
use web_time::{...};
```

But I haven't thought about whether that's nicer than a feature detected in `build.rs` or not TBH. I'm easy either way.

ratmice:
I'm going to mark this one as draft, until I figure out how to test it.

because just doing the naive thing and installing `wasmtime` and trying to `wasmtime run` the .wasm files
does end up giving a bindgen error.

I think the actual main requires some bindgen attributes, and it may well be that more is needed than just this minimal patch.  I don't know.

stephanemagnenat:
@ratmice thank you for looking into this. I am currently not using grmtools any more, as I switched to lalrpop in the mean time.

ratmice:
Strangely, while working on trying to get this tested (on the current master branch instead of this one), I can't actually reproduce the failure that requires us to need this dependency.

This despite adding a call to `std::time::Instant::now()` at the start of my test, and the test still succeeds.

This was running under `wasm-pack test --node`,
anyhow the rustwasm book seems to be really unmaintained for a long time, and `wasm-pack` seems to be very opaque.
As such I don't know if perhaps it is using a different target than `wasm32-unknown-unknown`.

I'm going to try and look at some modern wasm using crates, and see how they are doing their testing rather than trying to read that book.  Before I close this though.

ratmice:
So, I'm pretty uncertain about the `.buildbot.sh` changes, in particular that it is actually going to install everything
that is needed, and isn't just picking up necessary dependencies that I have installed already.

The example is very much not pretty, but from what I can tell it does actually work.
It could probably use some tlc from someone ~~more familiar with~~  who knows anything about the web.

But I'm going to mark as ready for review because it seems like it's working as intended.

lrpar/examples/calc_wasm/LICENSE-APACHE
```diff
@@ -0,0 +1,10 @@
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License.  You may obtain a copy of the
+License at
+
```
ratmice, line 4: `wasm-pack` is pretty adamant that `LICENSE` files exist.

lrpar/examples/calc_wasm/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+    match res {
+        Some(Ok(r)) => Ok(r),
+        Some(Err(e)) => Err(e.to_string()),
+        None => Err("Unable to parse".to_string().into()),
+    }
+}
+
+#[cfg(test)]
+mod test {
```
ratmice, line 65: I don't think i've ever actually used a `test` folder but I don't see any way to specify a crate type for them.
Because of how `wasm32-unknown-unknown` works, this seems to need to be a `cdylib` crate.

the `#[cfg(test)]` module is used for `wasm32-wasip2` and native targets,
while the other section is used only for `wasm32-unknown-unknown`.

lrpar/examples/calc_wasm/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This contains a rudimentary port of the calculator example to web assembly.
+The intent is that this will work on various platforms *including* web assembly.
+
+The platforms this example is intended to run on include:
+
+* wasm32-unknown-unknown using wasm-bindgen
+* wasm32-wasip2 without wasm-bindgen
+* native targets.
+
```
ratmice, line 11: This should probably give instructions, and requirements.

ratmice: Should be fixed in [bfe3346](https://github.com/softdevteam/grmtools/pull/527/commits/bfe3346452c3ab83eef43c7bb46fa0a95a6e400a)

lrpar/examples/calc_wasm/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+
+#[cfg_attr(
+    all(
+        target_arch = "wasm32",
+        target_os = "unknown",
+        target_vendor = "unknown"
+    ),
+    wasm_bindgen
+)]
+pub fn calculate(l: &str) -> Result<u64, String> {
```
ratmice, line 26: A big reason this is as ugly as it is is that on `wasm32-unknown-unknown` the `Err` variant must be able to be converted into a JsonValue, hence the `String` error.

Uncertain about `wasm32-wasip2`.

lrpar/examples/calc_wasm/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[cfg(all(
+    target_arch = "wasm32",
+    target_os = "unknown",
+    target_vendor = "unknown",
+    test
+))]
+mod wasm_bindgen_test {
+    use super::calculate;
```
ratmice, line 94: Making these separate modules seemed more clear than the alternative of:

#[cfg(any(all(target_arch = "wasm32", target_os = "unknown", target_vendor = "unknown", test), test))]
mod test {
   then 
```
#[cfg_attr(not(...),test)]
#[cfg_attr(all(...), wasm_bindgen_test])
fn test() {...}
```


`wasm_bindgen` here kind of wants to take over `#[wasm_bindgen_test(unsupported=test)]` for tests that should also run on other targets.

But that means bringing in wasm_bindgen as a dependency even for targets where it is unnecessary.
I guess it is worth trying to clean this up via a dev-dependency on `cfg_aliases`, it is an absolutely tiny `macro_rules` only crate which seems to be about the same compilation speed as an empty `build.rs` from my testing.

It seems worth attempting anyways.

ratmice: Cleaned up with cfg_aliases in [2a6e491](https://github.com/softdevteam/grmtools/pull/527/commits/2a6e4912271506b38825ae7352b3c6085026125d), I personally think the dep pulls it's weight.
considering the duplication of either modules or attributes that we'd otherwise have.

lrpar/examples/calc_wasm/LICENSE-APACHE
ltratt: If we really have to do this, we'll need to add the LICENSE-MIT too. But is there any way to shut wasm-pack up on this? Or, maybe, we just `symlink` back to the top-level files, to at least avoid bloating the repository?

ltratt:
I have a very dumb question: what is `calc_wasm` aiming to show? My very limited experience of wasm, mostly, is that it's just a backend target and, mostly, I don't need to know anything about. The `calc_wasm` target, though, seems to be doing various special things that don't fit that mental model. I'm sure I'm missing something obvious!

lrpar/examples/calc_wasm/LICENSE-APACHE
ratmice: I'd thought about symlinks, but IIRC git doesn't do symlinks well on windows.
I forget what it does exactly (either it puts the destination path of the symlink in the contents, or just bloats the repository how we'd be trying to avoid anyways) but I don't think it had a helpful behavior.

I'll look and see if I can't somehow get wasm-pack to pipe down.

ratmice:
It isn't a dumb question at all, 

The thing it is really trying to show is that bug #461 is actually fixed, i.e. it is more of a `#[test]` than an example.
*however* `wasm32-unknown-unknown` has a ton of limitations, in that it requires you to use a *specific crate type*, it only seems to support `crate-type = ["cdylib"]`, and requires you to annotate anything that would invoke javascript or be invoked by javascript with special function annotations.

So yeah the entirety of `wasm32-unknown-unknown` is a ton of various special things if I understand correctly, because it is also dealing with lots of javascript nonsense due.

`wasm32-wasip2` however is much more normal and how you would expect.  Like it I believe that it is able to use a `bin` target and normal main function. (I don't have a `start` function here, but in order to add one you have to annotation `#[wasm_bindgen(start)]`.

So you are not wrong, it just seems to be the madness of `wasm32-unknown-unknown` however.
I'm certain that someone who actually knows javascript and browsers and understands it better could
seriously improve what I've done, but I feel it has achieved the goal of testing #461 since it does panic with the appropriate errors if run on main :)

ltratt:
> it is more of a #[test] than an example.

Ah, I see! Right, I think I now understand your comment about a `tests` directory earlier. I think it would be better to dump it in such a directory so that people don't look at the examples repo, see "wasm" and then go "gosh, grmtools requires all sorts of weird hoops" if that makes sense?

ratmice:
I'll try to see if I can get it working by putting it in `lrpar/cttests/src/lib.rs`.

lrpar/examples/calc_wasm/LICENSE-APACHE
ratmice: There doesn't seem to be any workaround, there has been issues for 6 years, and PRs for at least 3.
I'll look and see though if there is anything I can do that is more similar to what we did with wasmtime,
setting a runner, so we can just avoid wasm-pack all together, as we aren't really using it for it's `pack into a tarball for distribution` behavior (where checking licenses makes sense).

I'm not certain this is going to be doable, e.g. the `trunkrs` project implements a lot of wasm-pack's packing behavior, but still doesn't seem to have it's own testing strategy afaict.

lrpar/examples/calc_wasm/src/lib.rs
ratmice: Should be fixed by moving it into cttests in [c5a4efd](https://github.com/softdevteam/grmtools/pull/527/commits/c5a4efdebab5ee4148203551bdfb1baefec64aab)

lrpar/examples/calc_wasm/LICENSE-APACHE
ratmice: Strangely and unexpectedly this appears to have been fixed by moving it into cttests in [c5a4efd](https://github.com/softdevteam/grmtools/pull/527/commits/c5a4efdebab5ee4148203551bdfb1baefec64aab).

There must be something different about the Cargo.toml that makes it not run the check or maybe being one directory closer to the `../Cargo.toml` instead of `../../Cargo.toml`, I would guess that is it.

.buildbot.sh
```diff
@@ -9,11 +9,26 @@ export RUSTFLAGS="--cfg grmtools_extra_checks"
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
 sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain stable -y --no-modify-path
 
-export PATH=`pwd`/.cargo/bin/:$PATH
+# wasm32-wasip2 target and utilities.
+# WASMTIME really wants to modify a file.
+touch .fake_profile
+export WASMTIME_HOME="`pwd`/.wasmtime"
+rustup target add wasm32-wasip2
+PROFILE=.fake_profile curl https://wasmtime.dev/install.sh -sSf | bash
```
ratmice, line 17: This one doesn't seem to check whether the current install is sufficient before
downloading/going through the install process, so it redownloads wasmtime every time.

I guess we could just check `which wasmtime` and if we ever need specific versions we can start comparing the output of `wasmtime -v`

ratmice: Kind of a hack, but [99cfb93](https://github.com/softdevteam/grmtools/pull/527/commits/99cfb935968ab1e492685f5d39463f3203ee3c69) checks the PATH for wasmtime before install.

ltratt: If/when an appropriate wasmtime is available on Debian, we can edit the docker file in the repo (which buildbot uses). I guess it's either not present or too old?

ratmice: I feel it is worth trying, there are versions of rust-wasm and nodejs on debian.
I don't know where this docker file is however.

I'm actually not certain which versions are required, but since `wasm-pack` has been around and stable for a long time, I would hope the version of nodejs installed there is sufficient.
I don't know what the necessary wasmtime version is, because I've only ever run with the latest version.

https://packages.debian.org/search?searchon=sourcenames&keywords=rust-wasmtime
https://packages.debian.org/search?searchon=sourcenames&keywords=nodejs

There doesn't appear to be a package for wasm-pack.

ltratt: The docker file is `.buildbot_dockerfile_debian`. You'd need to add the appropriate package(s) to the `apt-get -y install` line. I suggest we give it a go, remove the wasmtime manual installation stuff, and only add that back in if we find we have to. If the version in Debian *is* good enough, it'll save us some long term maintenance (it also makes CI a bit quicker!).

ratmice: Hmm, I don't see a `.buildbot_dockerfile_debian` could it be just not checked into the repo?

ltratt: Ooooops! PR up which adds one.

ratmice: Thanks if it's alright, I'll rebase onto master once that goes through,
There are some other ci changes like the license check on master which will also help.

ltratt: Agreed, if that PR goes in first, it should make this PR easier (I hope).

ratmice: Changed to use debian packages in [c679aec](https://github.com/softdevteam/grmtools/pull/527/commits/c679aec6563cfec6b4386095d4fb8120c4d2e010) and [52be2ee](https://github.com/softdevteam/grmtools/pull/527/commits/52be2ee81a616020d25635413583948b880f059b)

.buildbot_dockerfile_debian
```diff
@@ -3,6 +3,7 @@ ARG CI_UID
 RUN useradd -m -u ${CI_UID} ci
 RUN apt-get update && \
 apt-get -y install build-essential curl procps file
```
ltratt, line 5: I think you need to end this line with `\` and remove `apt-get -y install` from the following line. [I remember this format being fussy, and I *think* that's what works.]

ratmice: Doh, I see it now.

should be fixed in [949e713](https://github.com/softdevteam/grmtools/pull/527/commits/949e71328c59ca1a1ec9b8d1b5e09e7f2e8762b0)

lrpar/examples/calc_wasm/LICENSE-APACHE
ratmice: I did manage to get rid of wasm-pack fully in [da72a4d](https://github.com/softdevteam/grmtools/pull/527/commits/da72a4d59328363e05cfa750e14eeb4c70b8c285)

.buildbot.sh
```diff
@@ -47,8 +47,7 @@ echo -n "1+++" | cargo run --package nimbleparse -- src/param.l src/param.y -
 cd $root/lrpar/examples/start_states
 echo -n "/* /* commented out */ */ uncommented text /* */" | cargo run --package nimbleparse -- src/comment.l src/comment.y -
 cd $root/lrpar/cttests/
-# Test wasm32-unknown-unknown under nodejs.
-wasm-pack test --node
+cargo test --target wasm32-unknown-unknown
 cargo test --target wasm32-wasip2
 cargo test
```
ratmice, line 52: These would ideally run `cargo test --target ...` just from the root rather than within this subdir.
However that doesn't seem to work, building the `tempfile` crate needed for `lrpar::ctbuilder`.

I don't have a complete understanding of what is going on but it seems like that tries to run the *codegen* process from within wasi on `wasip2`, and that runs into some issue with nightly features in the std library.

Thus currently we can only `build` the targets at top-level but only `test` them on this cttests.

ratmice: I think this was improved in [96f24cb](https://github.com/softdevteam/grmtools/pull/527/commits/96f24cb215859aaac571b7d7d69ceb311e552f65)

By disabling those tests that required tempfile.

ratmice:
Alright, well I think this now ticks all the boxes for me, assuming the debian packaged versions work out.

lrpar/cttests/build.rs
```diff
@@ -13,5 +15,23 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         run_test_path(entry.unwrap())?;
     }
+
+    cfg_aliases! {
+        // Platforms
+        wasm32_unknown: { all(target_arch = "wasm32", target_os="unknown", target_vendor="unknown") },
+    }
+
+    CTLexerBuilder::new()
+        .rust_edition(lrlex::RustEdition::Rust2021)
```
ratmice, line 25: One last thought is that I don't see any reason we can't move this build.rs, calc_wasm.l, and calc_wasm.y into a `calc_wasm.test` file.

ltratt:
Please squash.

ratmice:
Squashed.

ratmice:
I think the debian packages didn't actually work, but somehow it still succeeded.
Maybe the docker script is missing something equivalet to `set -e` for bash?

```
10:12:30 #7 ERROR: process "/bin/sh -c apt-get update && apt-get -y install build-essential curl procps file rust-wasmtime nodejs" did not complete successfully: exit code: 100
```

Eventually producing the message:
```
10:12:30 4.113 E: Unable to locate package rust-wasmtime
```

Looking at the output here:
https://ci.soft-dev.org/#/builders/3/builds/3180/steps/3/logs/stdio

Now that I'm looking at that, I see that the debian packages `rust-wasmtime` is only available on trixie and sid.
So I think we do need to add the wasmtime install phase to `.buildbot.sh`.
https://packages.debian.org/search?searchon=sourcenames&keywords=rust-wasmtime

ratmice:
Also, i'm assuming that because trixie isn't released we actually want to do the install via `.buildbot.sh` locally

ltratt:
Since `.buildbot.sh` does `set -e`, presumably that part of things worked OK?

ratmice:
That is part of the weird part though, I don't actually see the output of that,
instead, it ran some lang_tests as though it were running the wrong `.buildbot.sh`  perhaps
I'm pretty confused!

```
10:30:49 Finished `test` profile [unoptimized + debuginfo] target(s) in 4.78s
10:30:49 Running run.rs (target/debug/deps/lang_tests-eec52b86122b5dd4)
10:30:49 Running tests with interpreter binary: bf_base
10:30:49
10:30:49 running 7 tests
10:30:49 test lang_tests::hello.bf ... ok
10:30:49 test lang_tests::hashes.bf ... ok
10:30:49 test lang_tests::numwarp.bf ... ok
10:30:51 test lang_tests::interp.bf ... ok
10:30:54 test lang_tests::hanoi-opt.bf ... ok
10:31:15 test lang_tests::bench.bf ... ok
10:31:15 test lang_tests::hanoi.bf ... ok
10:31:15
10:31:15 test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
10:31:15
10:31:15 Running tests with interpreter binary: bf_simple2_yk
10:31:15
10:31:15 running 7 tests
10:31:15 test lang_tests::bench.bf ... ignored
10:31:15 test lang_tests::hello.bf ... ignored
10:31:15 test lang_tests::hanoi.bf ... ignored
10:31:15 test lang_tests::numwarp.bf ... ignored
10:31:15 test lang_tests::interp.bf ... ignored
10:31:40 test lang_tests::hashes.bf ... ok
10:31:40 test lang_tests::hanoi-opt.bf ... ok
10:31:40
10:31:40 test result: ok. 7 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out
10:31:40
10:31:40 Running tests with interpreter binary: bf_simple_yk
10:31:40
10:31:40 running 7 tests
10:31:40 test lang_tests::bench.bf ... ignored
10:31:40 test lang_tests::interp.bf ... ignored
10:31:40 test lang_tests::hello.bf ... ignored
10:31:40 test lang_tests::hanoi.bf ... ignored
10:31:40 test lang_tests::numwarp.bf ... ignored
10:32:08 test lang_tests::hashes.bf ... ok
10:32:08 test lang_tests::hanoi-opt.bf ... ok
10:32:08
10:32:08 test result: ok. 7 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out
10:32:08 
```

PR #528, opened 2025-03-17
Fix cargo deny check license

ratmice:
*Note* this adds missing licenses to the allow list, to make the cargo deny check succeed.
Without regard to whether we actually want to allow those licenses
I've never actually read or heard of the Unicode-3.0 license.

In trying to figure out a testing strategy for wasm and the PR in #527
Previously in other things I had been running `.buildbot.sh` on a failing branch, when adding that cache check.
Running it here on a working branch  I noticed the `cargo-deny check license` script was failing, 
for a few reasons.

So now I'm wondering if the `.buildbot.sh` and the ci script are somehow out-of-sync,
This would seem to explain why I could never get #513 to fail.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -47,6 +47,6 @@ cd $root
 RUSTDOCFLAGS="-Dwarnings" cargo doc --no-deps
 
 which cargo-deny | cargo install cargo-deny || true
-if [ "X`which cargo-deny`" != "X"]; then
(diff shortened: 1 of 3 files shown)


ratmice:
Setting as draft until people have chimed in at least on the additional licenses.

ltratt:
These are all good: we've used this exact set of licenses elsewhere. Please squash.

ratmice:
Squashed.

PR #530, opened 2025-03-17
Initial experiment with direct bincode/optional serde.

ratmice:
Currently the derive macro for the bincode crate has a bug that fails to automatically derive the trait for types default type parameters.

This patch implements the `Decode` trait manually for YaccGrammar. It contains a modified copy of the derive macro output.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ codegen-units = 1
 panic = 'unwind'
 
 [workspace.dependencies]
-bincode = { version = "2.0", features = ["serde"] }
+bincode = "2.0"
 cactus = "1.0"
 filetime = "0.2"
 fnv = "1.0"
@@ -40,10 +40,10 @@ quote = "1.0"
 regex = "1.3"
 regex-syntax = "0.8"
 serde = "1.0"
-sparsevec = "0.2"
+sparsevec = "0.2.2"
 static_assertions = "1.1"
 unicode-width = "0.1.11"
-vob = ">=3.0.2"
+vob = "3.0.4"
 proc-macro2 = "1.0"
 prettyplease = "0.2.31"
 syn = "2.0"
--- a/cfgrammar/src/lib/yacc/grammar.rs
+++ b/cfgrammar/src/lib/yacc/grammar.rs
@@ -1,6 +1,8 @@
 #![allow(clippy::derive_partial_eq_without_eq)]
 use std::{cell::RefCell, collections::HashMap, fmt::Write};
 
+#[cfg(feature = "bincode")]
+use bincode::{Decode, Encode};
 use num_traits::{AsPrimitive, PrimInt, Unsigned};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
@@ -19,6 +21,7 @@ const IMPLICIT_START_RULE: &str = "^~";
 pub type PrecedenceLevel = u64;
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "bincode", derive(Encode, Decode))]
 pub struct Precedence {
     pub level: PrecedenceLevel,
     pub kind: AssocKind,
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -40,10 +40,10 @@ quote = "1.0"
 regex = "1.3"
 regex-syntax = "0.8"
 serde = "1.0"
-sparsevec = "0.2"
+sparsevec = { git = "https://github.com/softdevteam/sparsevec.git", rev = "dcf739917357c4a5bd3c358e48de9613360b40fe" }
```
ratmice, line 43: We'll want to fix this before committing.

ratmice:
I'm guessing above the hard coded commit in Cargo.toml we'll probably also want automatically derive before committing?

ratmice:
One thing I thought worth bringing up because this branch isn't really ready, was
the relative comparison between crates like grmtools, and crates like sparsevec.

```
$ time cargo build --features bincode
real    0m1.412s
user    0m2.107s
sys     0m0.676s
```

```
$ time cargo build --features serde
real    0m5.282s
user    0m6.799s
sys     0m1.487s
```

With the difference there being for sparsevec using bincode allows us to avoid
`serde`, `syn`, `quote`, `serde_derive`, `proc_macro2`, `unicode_ident`.

Where with grmtools even when maing `serde`/`serde_derive` fully optional, we still end up depending on all of those except `serde`,`serde_derive` I think.

Part of that is because all the recent codegen refactoring uses `syn`, `quote`, 
The `prettyplease` crate does too, but we could easily make that an optional feature.
(In fact when I started doing that codegen refactor one of the thoughts I had was that if we're requiring serde/syn/quote we might as well use it, never expecting there would be an opportunity that we could actually be able to remove it).
Lastly the `cttests_macro` used by the `ctfails`  requires `proc_macro2`.

Anyhow in summary, the benefits for crates like `sparsevec` are pretty clear, but for grmtools both the relative size of the crate is much larger compared to the savings and we only partially able to remove all the things serde depends on, so only partial savings.

ltratt:
Anything is better than nothing! Remind me: have I forgotten to make new releases of any of the crates we updated yesterday?

ratmice:
Just `sparsevec` has yet to be released.

ltratt:
sparsevec should be updates on crates.io now -- thanks for the reminder!

ltratt:
@ratmice Where are we on this PR?

ratmice:
Everything works, I'm just a bit hesitant to commit the manual implementation of  `impl Decode`,
But I haven't seen any PR's for fixing the derive macro.  It is related to https://github.com/bincode-org/bincode/issues/763

ratmice:
Oh, sorry I can/will fix the git dependency, that can be bumped to stablevec 0.2.2.

Cargo.toml
ratmice: Should be fixed in [51e7f08](https://github.com/softdevteam/grmtools/pull/530/commits/51e7f08cb0b055b580a597388e273ab1af2ce1a5)

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -88,6 +93,74 @@ pub struct YaccGrammar<StorageT = u32> {
     expectrr: Option<usize>,
 }
 
+#[cfg(feature = "bincode")]
+impl<StorageT, __Context> Decode<__Context> for YaccGrammar<StorageT>
+where
```
ratmice, line 98: Ideally this `impl Decode` and the impl of `BorrowDecode` below would just be part of `#[cfg_attr(feature = "bincode", derive(Encode, Decode))]`.

This impl works, it's just not very pretty, so the question is do we want to wait for that or not.

ratmice: I haven't seen any activity/progress towards fixing it, but the bug has been acknowledged.
So I really can't say.  I could look into fixing it myself, but the bincode code base is new to me.

I had for some reason had thought that because we only export serde derives with features enabled that
this could be changed in a non-breaking fashion.

But because of trait bounds, and `_reconstitute` it really does seem to be a breaking change.

ltratt: Then I think this hack (of sorts) is OK. It's probably worth documenting it with something like `// This can be simplified when github.com/xyz/issue/abc is fixed by ...` or similar?

ratmice: Added a comment in [c0ab286](https://github.com/softdevteam/grmtools/pull/530/commits/c0ab286694b76c4972a6fad845ef8b9b29bba3ed)

ltratt:
Please squash.

ratmice:
Squashed.

PR #531, opened 2025-03-18
Make prettyplease an optional dependency

ratmice:
I have mixed feelings on this patch, 

Currently by default we run the prettyplease code formatter over the generated sources.
Which was very useful during development because the unpretty sources are now very unpretty, they get emitted all on a single line.  So without pretty printing they are unreadable.

But I doubt most people do much reading of the generated sources, so I suppose if they do they could enable an option?

This also contains a fix, bringing lrlex in line with lrpar, on pretty printer failure.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -574,14 +574,9 @@ where
         };
         let outs = out_tokens.to_string();
         #[cfg(feature = "prettyplease")]
-        let outs = if let Ok(syntax_tree) = syn::parse_str(&outs) {
-            prettyplease::unparse(&syntax_tree)
-        } else {
-            // We failed to parse the source string before pretty printing.
-            // This is likely due to a syntax error in the source text.
-            // We should still emit the unformatted source text.
-            outs
-        };
+        let outs = syn::parse_str(&outs)
+            .map(|syntax_tree| prettyplease::unparse(&syntax_tree))
+            .unwrap_or(outs);
         // If the file we're about to write out already exists with the same contents, then we
         // don't overwrite it (since that will force a recompile of the file, and relinking of the
         // binary etc).
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -572,11 +572,16 @@ where
                 }
             }
         };
-
-        // Pretty print it.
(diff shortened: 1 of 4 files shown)


lrlex/src/lib/ctbuilder.rs
```diff
@@ -572,11 +572,16 @@ where
...
-        // Pretty print it.
-        let out_file = syn::parse_file(&out_tokens.to_string()).unwrap();
-        let outs = prettyplease::unparse(&out_file);
-
+        let outs = out_tokens.to_string();
+        #[cfg(feature = "prettyplease")]
+        let outs = if let Ok(syntax_tree) = syn::parse_str(&outs) {
+            prettyplease::unparse(&syntax_tree)
+        } else {
+            // We failed to parse the source string before pretty printing.
```
ratmice, line 580: I'm not really enamored with this

```
let outs = ...;
#[cfg(...)]
let outs = if ... else { outs };
```

Pattern, but like trying to make it `let mut outs` and then going `#[cfg(...)] if Ok(_) { outs = ...  }` lead to clippy warnings because `outs` doesn't need to be mut unless the feature is enabled.

I couldn't think of any nicer way, but would love a better one.

ratmice: I guess using the combinators is preferred here, so I did that in [96549f1](https://github.com/softdevteam/grmtools/pull/531/commits/96549f1e912f80a55c230cb5c471311b39959c00)

ltratt:
Personally I'd be inclined to make pretty-printing the default, at least in debug mode? For example, when I make a mistake in action code, I sometimes have to load the generated file up and take a peek.

ratmice:
I'm not aware of any mechanism to do debug-only default features, even from build.rs.
I'll just make it default for all profiles.

ratmice:
It's actually quite funny that `--no-default-features` still builds it because of the examples.
So really it's only going to end up saving time for things depend on grmtools and don't build the whole workspace.

I'm a bit ambivalent on whether we want to just remove the feature/pretty print all the time.
I guess if you really want to minimize dependencies, you'll check in the generated sources?

ltratt:
I think this is a situation where the dependency is worth it.

ratmice:
Yeah, I think I'll close this one and just make a separate PR for syncing the lex/yacc code.
I suppose that we can always revisit this in the future though, if someone wants to avoid building it.

PR #536, opened 2025-03-21
Install wasmtime manually rather than using the debian package

ratmice:
https://ci.soft-dev.org/#/builders/3/builds/3180/steps/3/logs/stdio

notes that the rust-wasmtime package doesn't exist for the version of debian we're using yet,
only on the next version.

This tries to run the install script manually,
I've had difficulty with this rerunning the install over and over regardless of whether it is currently installed.
thus it is gated in the ``if [ "X`which wasmtime`" == "X" ]`` check


ratmice:
This seems to have failed because of memory corruption in `nodejs`,
It is possible that a manual install of that is necessary too, perhaps the version on debian is too old.

ltratt:
Ouch! Well, we can give it a go.

ratmice:
Pushed that, also forgot to mention it didn't get to running wasmtime before it ran into the nodejs failure last time.
But hopefully both work, at least they seem to work when run locally.

ltratt:
Fingers crossed!

ratmice:
Progress at least, it got past `nodejs` this looks like perhaps I bungled the `which wasmtime` check, *looks*.

ratmice:
I *think* it was failing because of a bashism, but I'm not totally confident.
But the PATH setup being to late was also a problem for that check to ever succeed.

ratmice:
Squashed.

ratmice:
Ouch, #535 Forgot to git add files, but that failed to get caught by CI while the `apt-get install rust-wasmtime` failure was occurring, I would guess for the same reason the wasm commits succeeded ci, but I haven't tracked down the logs yet.

ratmice:
Read the logs, and everything looks good. Thank you for your patience and sorry about getting that bad commit inbetween breaking and fixing it.

PR #537, opened 2025-03-21
Issue 473 using a span on Production.

ratmice:
So here is a similar thing to PR #476 except instead of adding an empty symbol, it adds an 
`Option<Span>` to `Production`.

the assertion in `nimbleparse` should now be okay, but only because nimbleparse will always produce a grammar from a parser which ensures that *either* symbols has a symbol, containing a span or `empty_span` is `Some`.

That is, this isn't true in general from an `AST` built using the public constructors.

```
4| Expr: %empty | Factor ;
         ------- Reduced productions

4| Expr: Factor | %empty ;
                  ------- Reduced productions

4| Expr: Factor | ;
                  - Reduced productions

4| Expr: | Factor ;
         - Reduced productions
```

This entails adding a `span` argument to `add_prod` and the `Productions` struct.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -107,6 +107,7 @@ pub struct Production {
     pub symbols: Vec<Symbol>,
     pub precedence: Option<String>,
     pub action: Option<String>,
+    pub prod_span: Span,
 }
 
 #[derive(Clone, Debug)]
@@ -185,12 +186,14 @@ impl GrammarAST {
         symbols: Vec<Symbol>,
         precedence: Option<String>,
         action: Option<String>,
+        prod_span: Span,
     ) {
         self.rules[&rule_name].pidxs.push(self.prods.len());
         self.prods.push(Production {
             symbols,
             precedence,
             action,
+            prod_span,
         });
     }
 
@@ -436,7 +439,7 @@ mod test {
         let empty_span = Span::new(0, 0);
         grm.start = Some(("A".to_string(), empty_span));
         grm.add_rule(("B".to_string(), empty_span), None);
-        grm.add_prod("B".to_string(), vec![], None, None);
+        grm.add_prod("B".to_string(), vec![], None, None, empty_span);
         match grm.complete_and_validate() {
             Err(YaccGrammarError {
                 kind: YaccGrammarErrorKind::InvalidStartRule(_),
@@ -452,7 +455,7 @@ mod test {
         let empty_span = Span::new(0, 0);
         grm.start = Some(("A".to_string(), empty_span));
         grm.add_rule(("A".to_string(), empty_span), None);
-        grm.add_prod("A".to_string(), vec![], None, None);
+        grm.add_prod("A".to_string(), vec![], None, None, empty_span);
         assert!(grm.complete_and_validate().is_ok());
     }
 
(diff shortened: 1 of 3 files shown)


ratmice:
The error messages, and underlining is not great.  It just just kind of the position we are at when we finally can figure out that we have no symbol.  E.g. underlining the `;` or the `|`.

ratmice:
There, that is better:

```
4| Expr: %empty | Factor ;
         ------ Reduced productions

4| Expr: Factor | %empty ;
                  ------ Reduced productions
```

ltratt:
This looks like a good start!

Can we simplify this so we always give `Production` a `Span`? There might be other reasons later for people to want to know the `Span` of non-empty productions, so the simpler API might be the more powerful one!

ratmice:
I actually don't know, part of the difficulty is that the production parsing code is interspersed with the
type of the rule *and* the action parsing code.

So while it is pretty easy to substitute the current position at the point where we *know* a span doesn't previously exist.
But it's going to be a bit more difficult to actually separate out all the non-production spans from the production specific ones, including the non-symbols.

So I don't have an answer, I imagine it's possible but it sounds like it is going to be hairy.

ltratt:
> I think the ideal value for the Production's span is something like:
Production.span == Span::new(symbol.spans.minimum_start() , symbol.spans.maximum_end())

Agreed, that would be great if possible! Looking at `yacc/parser.rs` I *think* the crucial bit might be:

```rust
while i < self.src.len() {
    if let Some(j) = self.lookahead_is("|", i) {
	self.ast.add_prod(rn.clone(), syms, prec, action);
	syms = Vec::new();
	prec = None;
	action = None;
	i = self.parse_ws(j, true)?;
	continue;
    } else if let Some(j) = self.lookahead_is(";", i) {
	self.ast.add_prod(rn, syms, prec, action);
	return Ok(j);
    }
   ...
```

That's mostly at the start of the loop to deal with `R: | ...` i.e. an empty first production AFAICS. We can fairly easily expand that code a bit if we need to.

ratmice:
> > I think the ideal value for the Production's span is something like:
> > Production.span == Span::new(symbol.spans.minimum_start() , symbol.spans.maximum_end())
> 
> Agreed, that would be great if possible! Looking at `yacc/parser.rs` I _think_ the crucial bit might be:
> 

One of the things that is making this specific thing difficult is that a `Symbol::Token`'s span points *inside* the quotes.
It seems as though that might be the issue I'm running into at least, in trying to hit that ideal for cases where symbols isn't empty.

ratmice:
So [94622a2](https://github.com/softdevteam/grmtools/pull/537/commits/94622a28555166ce528fd45306f2f8d78e564264) is as close as I've gotten.

It is quite often +1 or -1 when the production starts or ends with a `Symbol::Token` because 
```
Rule: "a";
       ^ symbol span between quotes
      ^^^ production span overlapping quotes.
```

ltratt:
This feels like a significant improvement! Do we have tests for all the cases you've listed above?

ratmice:
Kind of, except for the last two, which are just whitespace variations,
the first four are from the following test.

```
#[test]
fn test_empty_production_spans_issue_473

```

Except that output is when run through nimbleparse, but we don't really have any systematic way to test nimbleparse output yet.  So that test just tests that the spans which drive nimbleparse return the expected results.

I suppose I should/need to add the last two variants.

ltratt:
If we get those two tests in, I think we'll be ready to squash.

ltratt:
Please squash.

ratmice:
Oops, typos.

ratmice:
Squashed.

Issue #473, opened 2024-11-21
Can't finish or takes forever with postgres16 grammar


mingodad:
While testing this project with a fresh clone and build with a postgres16 grammar the `nimbleparse` takes forever (I've killed it) see attached grammar/lexer/input.
```
/usr/bin/time ../target/release/nimbleparse postgresql-16-naked.l postgresql-16-naked.y input.txt
```
This is the output of bison :
```
/usr/bin/time bison-nb postgresql-16-naked.y
>/usr/bin/time bison-nb postgresql-16-naked.y
postgresql-16-naked.y: warning: 8 shift/reduce conflicts [-Wconflicts-sr]
postgresql-16-naked.y: note: rerun with option '-Wcounterexamples' to generate conflict counterexamples
1.37user 0.02system 0:01.37elapsed 101%CPU (0avgtext+0avgdata 17944maxresident)k
0inputs+4016outputs (0major+10957minor)pagefaults 0swaps
```
Notice that the 8 shift/reduce conflicts are due to the changes in the grammar to be accepted by `nimbleparse` that doesn't accept empty rules with `%prec` that are used in the postgres16 grammar:
```
...
opt_existing_window_name :
	ColId
	| %empty //%prec Op /*16L*/
	;
...
json_key_uniqueness_constraint_opt :
	WITH /*13N*/ UNIQUE /*11N*/ KEYS /*12N*/
	| WITH /*13N*/ UNIQUE /*11N*/
	| WITHOUT /*13N*/ UNIQUE /*11N*/ KEYS /*12N*/
	| WITHOUT /*13N*/ UNIQUE /*11N*/
	| %empty //%prec KEYS /*12N*/
	;
...
```
You can also see how https://github.com/BenHanson/parsertl14 handle it and several others with wasm in  the browser here https://mingodad.github.io/parsertl-playground/playground/ (select `Postgresql parser (be patient)` from `Examples` then click `Parse` to see a parse tree for the content in `Input source`).
```
build master parser -> Time taken : 70ms
parse user grammar -> Time taken : 6ms
start build user grammar -> Time taken : 2ms
build user parser -> Time taken : 1077ms
dump input parser tree -> Time taken : 10ms
```

[postgres16.zip](https://github.com/user-attachments/files/17850573/postgres16.zip)

ratmice:
I haven't had a chance to actually look through the grammar yet, but this could perhaps be another occurrence of #290 
[...]

PR #539, opened 2025-03-22
Experiment with bincode variable width encoding config

ratmice:
This is mostly an RFC, it should in theory just modify the encoding but not affect the runtime size of tables (one would hope).  I have only lightly tested it.

When migrating to bincode 2.0, I just used the `bincode::config::legacy()`, since it was compatible with old bincode.
The new default for bincode is `bincode::config::standard()`.  Which defines a variable width encoding.

This is mostly documented here https://docs.rs/bincode/latest/bincode/config/

- Legacy fixed width encoding. (little endian)
- Standard variable width encoding. (little endian)
- Other encoding/endian combinations are available by manually building a `Config`.

I had been playing around with seeing the affected sizes of files by switching to a variable width encoding.
Most of the files i've tested have been fairly small, e.g. the `%grmtools` section parser saved about 8k on a 50k `legacy()` file.

I figured I would see what the saving were using the large grammar in #473 `legacy()` produces a `postgres.y.rs` 37M while `standard()` produces a 23M source file.  I mostly wanted to just point out that this was available, I haven't tried to measure things like how it affects parser load time.  Whether it is possible to let users decide on a configuration (It may not be easy due to the way that `Configuration` has generic parameters, legacy produces a different type than standard)

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -981,8 +981,8 @@ where
             _ => unreachable!(),
         };
 
-        let grm_data = encode_to_vec(grm, bincode::config::legacy())?;
-        let stable_data = encode_to_vec(stable, bincode::config::legacy())?;
+        let grm_data = encode_to_vec(grm, bincode::config::standard())?;
+        let stable_data = encode_to_vec(stable, bincode::config::standard())?;
         Ok(quote! {
             const __GRM_DATA: &[u8] = &[#(#grm_data,)*];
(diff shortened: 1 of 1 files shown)


ltratt:
Variable width here is an excellent idea here because saving bytes in the binary is nearly always worth it: it was on my radar in the past, but I couldn't work up the enthusiasm to implement this myself. Please undraft and we can merge this.

ratmice:
Alright I went ahead and reworded the commit message, and unmarked as draft,

Issue #490, opened 2025-02-21
Add LexKind


ratmice:
This is a follow up to PR #483 we now have a number of ways in which [`RegexOptions`](https://docs.rs/lrlex/latest/lrlex/index.html#reexport.RegexOptions) can be used to alter the lrlex format primarily by modifying the interpretation of regular expressions.

In that PR it had been mentioned doing follow up work adding a LexKind similar to [`YaccKind`](https://docs.rs/cfgrammar/0.13.9/cfgrammar/yacc/enum.YaccKind.html)

One thing I have been thinking about which is kind of tangential but related to this is issue #349 which delays the time that `YaccKind`  is known until parser reading time. What seems like it should be a trivial refactor, but also a breaking change. I think it ends up a little bit less trivial than I think it first appears since it introduces some new kinds of propositions for directives that current directives doesn't require and the format parsing code has no current way to enforce.

1. YaccKind affects other directives in the file format so presumably it needs to be first before other directives
2. It should also probably be unique in that we don't want multiple `*Kind`s.

Thus my thoughts have been that it would be good to try and come to some kind of conclusion on #349 as we introduce a LexKind, just to either avoid introducing any new code that could require this same delicate refactoring, or close #349 abandoning the idea of having self describing formats.

Anyhow I'll post some more new thoughts related to that other issue where they belong.

ltratt:
I think a couple of things are in our favour here:

1. `%yacckind` would only make sense for Yacc inputs. So we already know that we're dealing with a Yacc-ish format. If we ever support non-Yacc formats, they'll go down a different codepath.
2. We could easily say "`%yacckind`, if it exists, must be the first declaration in the file. If it exists and is not the first declaration, an error is raised".

Would those be enough to enable progress?

ratmice:
I put a bunch more comments in #349 some of it overlaps the big issue is backwards compatibility, we currently *need* to specify the `YaccKind` not through the builder, but for [YaccGrammar::new](https://docs.rs/cfgrammar/0.13.9/cfgrammar/yacc/grammar/struct.YaccGrammar.html#method.new) so there is kind of a chicken-egg problem in my comment over there I tried to get around that by introducing a new variant `YaccKind::SelfDescribing`.  I think *that* is the biggest impediment to progress.

ratmice:
[...]

PR #542, opened 2025-03-24
Add comma to grmtools section syntax

ratmice:
This uses the modified grmtools section grammar from #490 which adds a comma to make it unambiguous.
This isn't what is currently parsed by the section parsing code in `parser.rs`.

using a self hosted grammar isn't something we currently do,
This patch doesn't intend to add it.  However it occurred to me that we could still run it in
`cttests/` over the `lrpar/example/*.{l,y}` and those generated by cttests.

This currently works for all the files in the repository,
you need to have at least 3 entries in the grmtools section before it realizes something is awry.
because it reads the first and second entries as key/value.

Marking this as draft for the following purposes
1. until there is agreement on whether this change of syntax is acceptable and good...
2. until we change the parser.rs files to accept it.

Edit:
  The other thing this doesn't try to do is add a specialized parser for lex or one for yacc files.
  Accepting just the keys for the respective files, I guess we could do a separate parser each?

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -304,6 +304,15 @@ where
                             });
                         }
                         i = self.parse_spaces(i)?;
+                        if let Some(j) = self.lookahead_is(":", i) {
+                            i = j
+                        } else {
+                            return Err(LexBuildError {
+                                kind: LexErrorKind::InvalidGrmtoolsSectionValue,
+                                spans: vec![Span::new(i, i)],
+                            });
+                        }
+                        i = self.parse_spaces(i)?;
                         let j = self.parse_digits(i)?;
                         // This checks that the digits are valid numbers, but currently just returns `None`
                         // when the values are actually out of range for that type. This could be improved.
@@ -351,11 +360,18 @@ where
                     &mut grmtools_section_span_map,
                     &mut grmtools_section_lex_flags,
                 )?;
-                if i == self.src.len() {
-                    return Err(self.mk_error(LexErrorKind::PrematureEnd, i));
+                if let Some(j) = self.lookahead_is(",", i) {
+                    i = self.parse_ws(j)?;
+                    continue;
+                } else {
+                    break;
                 }
             }
-            i = self.lookahead_is("}", i).unwrap();
+            if let Some(j) = self.lookahead_is("}", i) {
+                i = j
+            } else {
+                return Err(self.mk_error(LexErrorKind::PrematureEnd, i));
+            }
             i = self.parse_ws(i)?;
         }
         grmtools_section_lex_flags.merge_from(&self.force_lex_flags);
@@ -864,7 +880,9 @@ mod test {
                 .expect_err("Parsed ok while expecting error");
             assert_eq!(errs.len(), 1);
             let e = &errs[0];
-            assert!(e.kind.is_same_kind(&kind));
+            if !e.kind.is_same_kind(&kind) {
+                panic!("expected {:?}.is_same_kind({:?})", &e.kind, &kind);
+            }
             assert_eq!(
                 e.spans()
                     .iter()
@@ -1864,7 +1882,7 @@ b "A"
     fn test_grmtools_section_fails() {
         let src = r#"
 %grmtools {
-  !unicode
+  !unicode,
   unicode
 }
 %%
@@ -1879,8 +1897,8 @@ b "A"
 
         let src = r#"
 %grmtools {
-  size_limit 5
-  size_limit 6
+  size_limit: 5,
+  size_limit: 6,
 }
 %%
 . "dot";
@@ -1894,7 +1912,7 @@ b "A"
 
         let src = r#"
 %grmtools {
-  !size_limit 5
+  !size_limit: 5,
 }
 %%
 . "dot"
@@ -1905,6 +1923,22 @@ b "A"
             LexErrorKind::InvalidGrmtoolsSectionValue,
             3,
             3,
+        );
+        // The following is missing comma separators for more than 2 elements
+        // This is to avoid parsing it as "key value" number of elements.
+        // However we actually error after the first element since the parser
+        // knows "case_insensitive" is a flag with no arguments.
+        let src = r#"
+%grmtools {unicode, case_insensitive posix_escapes allow_comments}
+%%
+. "dot"
+\n+ ;
+"#;
+        LRNonStreamingLexerDef::<DefaultLexerTypes<u8>>::from_str(src).expect_error_at_line_col(
+            src,
+            LexErrorKind::PrematureEnd,
+            2,
+            38,
         );
     }
 
--- a/lrpar/cttests/src/cgen_helper.rs
+++ b/lrpar/cttests/src/cgen_helper.rs
@@ -79,10 +79,10 @@ pub(crate) fn run_test_path<P: AsRef<Path>>(path: P) -> Result<(), Box<dyn std::
         // Create grammar files
         let base = path.file_stem().unwrap().to_str().unwrap();
         let mut pg = PathBuf::from(&out_dir);
-        pg.push(format!("{}.y.rs", base));
+        pg.push(format!("{}.test.y", base));
         fs::write(&pg, grm).unwrap();
         let mut pl = PathBuf::from(&out_dir);
-        pl.push(format!("{}.l.rs", base));
+        pl.push(format!("{}.test.l", base));
         fs::write(&pl, lex).unwrap();
 
         // Build parser and lexer
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -15,21 +15,20 @@ export PATH=`pwd`/.cargo_install/bin/:$WASMTIME_HOME/bin:$PATH
 
 # Install wasmtime once debian trixie is stablized
 # we can likely just use rust-wasmtime.
+#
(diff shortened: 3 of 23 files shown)


lrlex/src/lib/parser.rs
```diff
@@ -1906,6 +1916,22 @@ b "A"
             3,
             3,
         );
+// The following is missing comma separators for more than 2 elements
+// This is to avoid parsing it as "key value" number of elements.
+// However we actually error after the first element since the parser
+// knows "case_insensitive" is a flag with no arguments.
+        let src = r#"
+%grmtools {unicode, case_insensitive posix_escapes allow_comments}
+%%
```
ratmice, line 1925: This comment indicates a difference between the `grmtools_section.y` parser, and the
lrlex/src/lib/parser.rs.

the `grmtools_section.y` parser returns a `case_insensitive posix_escapes` key/value pair, then fails on `allow_comments`.

I suppose we should consider using `key: value` here

ratmice: I guess another option would be `%yacckind`...

lrlex/src/lib/parser.rs
```diff
@@ -351,11 +351,19 @@ where
                     &mut grmtools_section_span_map,
                     &mut grmtools_section_lex_flags,
                 )?;
-                if i == self.src.len() {
-                    return Err(self.mk_error(LexErrorKind::PrematureEnd, i));
+                eprintln!("foo '{:?}'", &self.src[i..]);
```
ratmice, line 354: Should be removed.

ratmice: fixed in [57b94aa](https://github.com/softdevteam/grmtools/pull/542/commits/57b94aac5173e0220476172e445b3c844586fc4f)

lrlex/src/lib/parser.rs
ratmice: I tried `key: value` in [10e6d74](https://github.com/softdevteam/grmtools/pull/542/commits/10e6d74b2f435bc203ccbb0f86e0e36ce65854ac)

ratmice:
Well, it adds both comma and colon, so not just s/comma/colon/ but I'll reword it appropriately.

lrlex/src/lib/parser.rs
```diff
@@ -340,6 +340,7 @@ where
         let mut grmtools_section_span_map = HashMap::new();
         let mut grmtools_section_lex_flags = UNSPECIFIED_LEX_FLAGS;
         if let Some(j) = self.lookahead_is("%grmtools", i) {
+            // lrlex currently doesn't have any `key: value' settings.
```
ratmice, line 343: This is actually wrong, I totally forgot that `regex` defined some limit related values.

ratmice:
Will need another squash, I fixed `field-like` values on lrlex, and updated the docs.

ratmice:
Squashed, and cleaned up the commit message regarding fields in lex.

ratmice:
Curious curious, I have a theory about what is wrong, the build logs aren't informative about it,
The new test I added runs on native and wasm32.

I believe it is probably wasmtime sandboxing filesystem access to the current directory, and that test tries to read lex/yacc files in `../examples/` and `OUT_DIR`.

I'll just run it locally to confirm

ratmice:
So, yes that seems to be the issue. Unfortunately I'm not certain how to fix it (properly)...

My attempt was to add a `.cargo/scripts/wasi_test_runner.sh`
.cargo/config.toml

```
runner = ".cargo/scripts/wasi_test_runner.sh"
```

Unfortunately it looks like the test runner may get executed *without* the environment variable `OUT_DIR` set.
(assuming I'm not doing something silly)

```
#!/bin/sh -x
export top_dir="$CARGO_MANIFEST_DIR"
while [ -f "${top_dir}/Cargo.toml" ]; do
        export top_dir="$top_dir/.."
done
wasmtime run --dir . \
        --dir $CARGO_MANIFEST_DIR \
        --dir $top_dir \
        --dir $OUT_DIR \
        --env CARGO_MANIFEST_DIR="$CARGO_MANIFEST_DIR" \
        --env OUT_DIR=$OUT_DIR \
        $1
```

So if that is the case, this would seem to indicate a cargo problem with setting that environment variable on that target,
it is possible we may have to conditionally disable the test, perhaps: `#[cfg(all(not(wasm32_wasip2), test)]`

ltratt:
This is outside my comfort zone, but are we suggesting this is a cargo bug? or a wasmtime limitation? or ... ?

ratmice:
It's actually just me misunderstanding cargo's weirdness.

Cargo sets the environment variable, but only when `build.rs` exists in the project,
and the various crates exist in the workspace, some of them have a `build.rs` and some don't.
If the project doesn't have a build.rs a projects tests shouldn't rely on it since it can't have any effect.

So I think it is just normal behavior that my test runner needs to account for the fact that it may or may not be set.

ltratt:
Can we fix things here?

ratmice:
I *think* so but still working on it, at worst we just disable the test on that particular target.
I'm also not certain whether or not the hacks required to fix it are be so terrible that we'd rather just disable it or not.

E.g. there is no way I know of to just get the `workspace` dir, and `CARGO_MANIFEST_DIR/../examples` isn't valid for *all* crates in the workspace. So what i'm trying to do is search in `manifest_dir/../`, `manifest_dir/../..`, up one directory at a time looking for a `Cargo.toml` with the `[Workspace]` key, and passing that path.

I *think* it is possible we end up with something that fixes all issues of this kind in the workspace once and for all with a unfortunate but perhaps tolerable hack.

But I'm definitely not there yet.

ltratt:
If you have a bright idea after a good night's sleep, go for it, but if not: please squash.

ratmice:
There is something I can try, all the paths I was setting manually were about setting up which directory was allowed inside the sandbox...
I was still using the cargo/rust generated paths to `OUT_DIR`, and `MANIFEST_DIR`, those get set through the environment as absolute paths. It is possible that in order to get them to work in the sandbox those must be changed into relative paths.

That doesn't explain exactly why it didn't work when I tried to add `/` to the sandbox, so it could also be some interaction between wasmtime and how the `glob` crate works.  But there is no crate boundary involved with seeing if I can get relative paths in place of what cargo gives.

ratmice:
I did mange to get something working (it is kind of a terrible hack),
So for the time being i've put it up here rather than add it to the workspace,
I wasn't sure if you'd want it in the workspace or separate tool...

Note that because .cargo/config.toml is committed to the repo people will need it to run cargo.test
on wasm targets, so having it in repo allows us to say `cargo install --path grmtools_workspace_runner` or something
Where otherwise it'll be a separate tool they'll have to install.

I don't have much of a preference, but i very rarely ever `cargo publish` anything...

https://github.com/ratmice/workspace_time_test_runner

In .buildbot.sh add:
```
cargo install --git 'https://github.com/ratmice/workspace_time_test_runner.git'
```

then in `.cargo/config` set 
```
[target.wasm32-wasip2]
runner = "workspace_time_test_runner"
```

then removing the 
```
- #[cfg(not(all(target_arch = "wasm32", target_os = "wasi")))]
 fn test_grmtools_section_files() {
```

ltratt:
Wow, I must admit, this feels like a lot of complexity for one (apparently weird!) platform. Do I understand correctly that this would require normal grmtool developers to install stuff before `cargo test` works?

ratmice:
Anyhow, I added some minor comments + assertions + fixed whitespace, then squashed in case you wanted to just go ahead without the runner.

lrpar/cttests/src/cgen_helper.rs
```diff
@@ -79,10 +79,10 @@ pub(crate) fn run_test_path<P: AsRef<Path>>(path: P) -> Result<(), Box<dyn std::
         // Create grammar files
         let base = path.file_stem().unwrap().to_str().unwrap();
         let mut pg = PathBuf::from(&out_dir);
-        pg.push(format!("{}.y.rs", base));
+        pg.push(format!("{}.test.y", base));
```
ratmice, line 82: I had believed that this `.y.rs` thing was a mistake of sorts, because this is a `.y` file not a `.rs` file.
But it occurred to me that emitting of this file with a .rs extension may have been a way to get the same behavior as `cargo:rerun-if-changed=path` behavior.

In which case I don't think it's necessary we emit `rerun-if-changed` for the .test file these derive from.
But I do think that means we should also emit rerun-if-changed for `glob(src/*.rs)` which I *don't* think we're doing.

Perhaps best to do that and an audit in a follow up.

ltratt: > Perhaps best to do that and an audit in a follow up.

Good idea!

ltratt:
@ratmice OK, phew, since it's limited to that target, I'm fine with this. Please squash.

ltratt:
>  Just so I'm clear we're okay with this additional cargo install requirement when it is isolated to --target wasm32-wasip2, so I'll add that and squash.

Yep!

> I figure I'll just keep it under my personal account and upload to crates.io

This seems like a good idea to me. It might be useful to others too!

ratmice:
Updated to use the test runner, 

I moved the `cargo install` of the runners for wasm targets to be *after*
the `cargo test` line, so it was more apparent that they aren't required for default targets.

Anyhow I think this should be ready now :crossed_fingers:

Issue #541, opened 2025-03-23
Repairs "channel"


ratmice:
This is not really well thought out,

Just a random thought I had, in #461 and #527 We rely upon timers to try and figure out when enough time has passed to abandon searching for repairs and return.

When using grmtools interactively via an LSP parsing stuff on every keypress, I tend to turn off repairs because it increases the baseline time to report failures to whatever is specified in the `RECOVERY_TIME_BUDGET`.

It occurred to me that we could consider using something like a *spsc*, channel so that when the consumer closes the channel
the producer stops searching for repairs.

That would seem to have two effects:

1. Moves the RECOVERY_TIME_BUDGET out into the caller (given that they choose when to close the channel)
2. Allow errors to be reported, with recovery information to follow later.

Now, I don't exactly know how we would return errors without also doing such a thing with an "error channel",
and/or doing the recovery algorithm on a separate thread (which things like wasm tends to make threads painful from what I can tell)...

I'm not sure if this is reasonably implementable, so this might be a "wont fix"/closable in that it could be experimented with somewhere outside grmtools?

ratmice:
I think I'll probably close this one, as It doesn't seem like I'll have the focus to take this on in any reasonable time.
I did look into it a little bit, but the introduction of threads so that repairs can be sent after `parse` has failed/returned *is* probably not a small change for a code base that wasn't designed to be used that way initially.

Perhaps there is a reasonable implementation, but my exploratory attempts have all run aground pretty quickly.

PR #543, opened 2025-03-28
print cargo::rerun-if-changed directives from cttests/build.rs

ratmice:
Emit cargo::rerun-if-changed from src/*.rs, src/*.test, storaget.l, storaget.y, storaget.l.rs storaget.y.rs.

Files that are not emitted include *.test.l, *.test.y, and their respective .l.rs and .y.rs files. This is because the .test cgen_helper don't check timestamps and would always result in spurious rebuilds.

### DIFF
--- a/lrpar/cttests/build.rs
+++ b/lrpar/cttests/build.rs
@@ -11,6 +11,9 @@ use lrlex::{CTLexerBuilder, DefaultLexerTypes};
 // two modules `<filename>_y` and `<filename>_l`, which we can then import into src/lib.rs and
 // write tests for.
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    for src in glob("src/*.rs")? {
+        println!("cargo::rerun-if-changed={}", src?.display());
+    }
     for entry in glob("src/*.test")? {
         run_test_path(entry.unwrap())?;
     }
@@ -24,8 +27,16 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Since it modifies the type of the builder.
     CTLexerBuilder::<DefaultLexerTypes<u8>>::new_with_lexemet()
         .rust_edition(lrlex::RustEdition::Rust2021)
+        .output_path(format!(
+            "{}/storaget.l.rs",
+            std::env::var("OUT_DIR").unwrap()
+        ))
         .lrpar_config(|ctp| {
             ctp.rust_edition(lrpar::RustEdition::Rust2021)
+                .output_path(format!(
+                    "{}/storaget.y.rs",
+                    std::env::var("OUT_DIR").unwrap()
+                ))
                 .grammar_in_src_dir("storaget.y")
                 .unwrap()
         })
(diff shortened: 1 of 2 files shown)


ltratt:
I'm not sure if this is the right tool here, but would https://crates.io/crates/rerun_except help?

ratmice:
I'm not certain, like I'm adding `storaget.l.rs`, and `storage.y.rs`, but not any of the `.y.rs` and `.l.rs` derived from `.test` files.
This is slightly more precise than the normal build.rs mechanism with cargo where we don't normally also add `storaget.l`, and `.y` (by default we only detect .rs extensions).  But for that file we aren't generating from a test, so we should be avoiding spurious regeneration through the cache mechanism.

I *think* we could emit better filenames, like: `foo.test.l`, `foo.test.y`, `foo.test.l.rs`, `foo.test.y.rs`, and then
`rerun_except(&["*.test.l", "*.test.y", "*.test.l.rs", "*.test.y.rs"]);`

That is to say that *currently* without the `.test` being added `storaget.{l,y,l.rs,y.rs}` would match our inverted globs we want.

ltratt:
So probably best to merge as-is?

ratmice:
Let me try changing the extensions, and see if it helps/hurts, I think it'll likely help

ratmice:
I think I would have said it was an improvement, 

We have to change all the `lrlex_mod!("calc_wasm.test.l")` and `lrpar_mod!("calc_wasm.test.y")`.
to account for the new paths, however `rerun_except` doesn't actually work in this case.

We end up with something such as the following, however `rerun_except` appears to only set matches from `src/**`,
where our globs are all prefixed in `OUT_DIR`.

```
    let out_dir = std::env::var("OUT_DIR").unwrap();
    rerun_except(&[
        &format!("{out_dir}/*.test.y"),
        &format!("{out_dir}/*.test.l"),
        &format!("{out_dir}/*.test.l.rs"),
        &format!("{out_dir}/*.test.y.rs"),
    ])
    .unwrap();
```

That is to say, that the `rerun-if-changed=OUT_DIR/storaget.l`, and other OUT_DIR files isn't getting emitted by this line,
it's only emitting files from the `cttests/` source directory.

ltratt:
Ah, yes, rerun_except will almost certainly ignore `target` (because of the git traversal crate it uses). I'm inclined to merge as-is, as I think rerun_except is a mild square-ish peg for this particular round hole. OK to do so?

ratmice:
Yeah, I think so, the excessive number `format!` calls are kind of burning a hole in my eyes.
But that is borrow checker/ownership stuff because of closures to the ctbuilder methods,
So i'm not sure what I could do to fix that.  

So I *think* it is okay to merge as-is, in that I'm certain it could be improved but I'm not sure what we could do about it today.

ratmice:
Alright so, the next 2 things on my list are

1. recoverer or recoverykind for yacc `%grmtools` section
2. lexerkind for lex `%grmtools` section,

Then after that perhaps experiment with #541 at least enough to do a proof of concept and see if it is something that requires any breaking changes or might be feasibly done without (now that we use `#[non_exhaustive]`)

PR #544, opened 2025-03-29
Add a manual implementation of the grmtools_section_parser from cttests.

ratmice:
This adds a `pub` parser in cfgrammar for parsing the `%grmtools` section.

It is intended to replace both the manually written parser in GrammarAST, as well as the one in LexParser. This is also in preparation of adding values for `RecoveryKind` and `LexerKind` to their respective sections. Since those values live outside of `GrammarAST` and `LexParser` entirely.

The testsuite checks that the return values from this parser, and the yacc generated one in the testsuite are equal. This also fixes a bug in the action code of the `grammar_section.test` Which was return a "::" token in place of a value.

Note to self: need to reword the commit message to match the above.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -1,0 +1,325 @@
+use crate::Span;
+use lazy_static::lazy_static;
+use regex::{Regex, RegexBuilder};
+use std::collections::{hash_map::Entry, HashMap};
+
+#[derive(Debug)]
+pub struct HeaderError {
+    pub kind: HeaderErrorKind,
+    pub spans: Vec<Span>,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+#[non_exhaustive]
+#[doc(hidden)]
+pub enum HeaderErrorKind {
+    MissingGrmtoolsSection,
+    IllegalName,
+    ExpectedToken(char),
+    DuplicateEntry,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+#[doc(hidden)]
+pub enum Path<'a> {
+    Ident(&'a str, Span),
+    Scoped((&'a str, Span), (&'a str, Span)),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+#[doc(hidden)]
+pub enum Setting<'a> {
+    PathLike(Path<'a>),
+    ArgLike(Path<'a>, Path<'a>),
+    Num(u64, Span),
+}
+
+/// Parser for the `%grmtools` section
+#[doc(hidden)]
+pub struct GrmtoolsSectionParser<'input> {
+    src: &'input str,
+    required: bool,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum Value<'a> {
+    Flag(bool),
+    Setting(Setting<'a>),
+}
+
+lazy_static! {
+    static ref RE_LEADING_WS: Regex = Regex::new(r"^[\p{Pattern_White_Space}]*").unwrap();
+    static ref RE_NAME: Regex = RegexBuilder::new(r"^[A-Z][A-Z_]*")
+        .case_insensitive(true)
+        .build()
+        .unwrap();
+    static ref RE_DIGITS: Regex = Regex::new(r"^[0-9]+").unwrap();
+}
+
+const MAGIC: &str = "%grmtools";
+
+fn add_duplicate_occurrence(
+    errs: &mut Vec<HeaderError>,
+    kind: HeaderErrorKind,
+    orig_span: Span,
+    dup_span: Span,
+) {
+    if !errs.iter_mut().any(|e| {
+        if e.kind == kind && e.spans[0] == orig_span {
+            e.spans.push(dup_span);
+            true
+        } else {
+            false
+        }
+    }) {
+        errs.push(HeaderError {
+            kind,
+            spans: vec![orig_span, dup_span],
+        });
+    }
+}
+
+impl<'input> GrmtoolsSectionParser<'input> {
+    pub fn parse_value(
+        &'_ self,
+        mut i: usize,
+    ) -> Result<(&'_ str, Span, Value<'_>, usize), HeaderError> {
+        if let Some(i) = self.lookahead_is("!", i) {
+            let (flag_name, j) = self.parse_name(i)?;
+            Ok((
+                flag_name,
+                Span::new(i, j),
+                Value::Flag(false),
+                self.parse_ws(j),
+            ))
+        } else {
+            let (key_name, j) = self.parse_name(i)?;
+            let key_span = Span::new(i, j);
+            i = self.parse_ws(j);
+            if let Some(j) = self.lookahead_is(":", i) {
+                i = self.parse_ws(j);
+                match RE_DIGITS.find(&self.src[i..]) {
+                    Some(m) => {
+                        let num_span = Span::new(i + m.start(), i + m.end());
+                        let num_str = &self.src[num_span.start()..num_span.end()];
+                        // If the above regex matches we expect this to succeed.
+                        let num = str::parse::<u64>(num_str).unwrap();
+                        let val = Setting::Num(num, num_span);
+                        i = self.parse_ws(num_span.end());
+                        Ok((key_name, key_span, Value::Setting(val), i))
+                    }
+                    None => {
+                        let (path_val, j) = self.parse_path(i)?;
+                        i = self.parse_ws(j);
+                        if let Some(j) = self.lookahead_is("(", i) {
+                            let (arg, j) = self.parse_path(j)?;
+                            i = self.parse_ws(j);
+                            if let Some(j) = self.lookahead_is(")", i) {
+                                i = self.parse_ws(j);
+                                Ok((
+                                    key_name,
+                                    key_span,
+                                    Value::Setting(Setting::ArgLike(path_val, arg)),
+                                    i,
+                                ))
+                            } else {
+                                Err(HeaderError {
+                                    kind: HeaderErrorKind::ExpectedToken(')'),
+                                    spans: vec![Span::new(i, i)],
+                                })
+                            }
+                        } else {
+                            Ok((
+                                key_name,
+                                key_span,
+                                Value::Setting(Setting::PathLike(path_val)),
+                                i,
+                            ))
+                        }
+                    }
+                }
+            } else {
+                Ok((key_name, key_span, Value::Flag(true), i))
+            }
+        }
+    }
+
+    fn parse_path(&'_ self, mut i: usize) -> Result<(Path<'_>, usize), HeaderError> {
+        let (name, j) = self.parse_name(i)?;
+        let name_span = Span::new(i, j);
+        i = self.parse_ws(j);
+        if let Some(j) = self.lookahead_is("::", i) {
+            i = self.parse_ws(j);
+            let (scoped_val, j) = self.parse_name(i)?;
+            let scoped_span = Span::new(i, j);
+            i = self.parse_ws(j);
+            Ok((
+                Path::Scoped((name, name_span), (scoped_val, scoped_span)),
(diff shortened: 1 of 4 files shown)


lrpar/cttests/src/grmtools_section.test
```diff
@@ -24,36 +24,44 @@ grammar: |
...
+        Path::Ident(ident, ident_span)
     }
     | IDENT '::' IDENT {
-        let scope = $lexer.span_str($1.as_ref().unwrap().span());
-        let ident = $lexer.span_str($2.as_ref().unwrap().span());
-        Path::Scoped(scope, ident)
+        let scope_span = $1.as_ref().unwrap().span();
+        let scope = $lexer.span_str(scope_span);
+
+        let ident_span = $3.as_ref().unwrap().span();
```
ratmice, line 35: This is fixing a bug `$2` is changed to `$3` as well as adding a span.

ratmice:
This is preparation for the following:

1. Remove the manually written parsers in `LexParser` and `GrammarAST`
2. Add `lexerkind` and `recoverykind`.

I didn't want to go and do those removals, unless asked to first.

cfgrammar/src/lib/header.rs
```diff
@@ -0,0 +1,186 @@
...
+    MissingGrmtoolsSection(Span),
+    IllegalName(Span),
+    InvalidValue(Span),
+    MalformedPath(Span),
+    ExpectedToken(char, Span),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum ValBind<'a> {
+    // ! [A-Za-z]+
```
ratmice, line 17: Probably don't need these comments.

cfgrammar/src/lib/header.rs
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            Ok((vec![], 0))
+        }
+    }
+
+    fn parse_name(&self, i: usize) -> Result<(&str, usize), HeaderError> {
+        match RE_NAME.find(&self.src[i..]) {
+            Some(m) => {
+                assert_eq!(m.start(), 0);
+                Ok((&self.src[i..i + m.end()], i + m.end()))
```
ratmice, line 166: Should consider doing
```suggestion
                Ok((&self.src[i..i + m.end()].to_lowercase(), i + m.end()))
```

ltratt:
I agree that there's no perfect solution here: the crates are somewhat tied together. What we *could* do is make this function (and associated parts) hidden from a documentation point of view: that way we're "hiding" them from users, so they're less likely to rely on them, even though they're *technically* part of the public API.

Another possibility -- and I haven't thought this through! -- is that we could stick the format parsing stuff in its own crate. I'm not sure that would buy us much, but perhaps it would solve these problems?

ratmice:
> I agree that there's no perfect solution here: the crates are somewhat tied together. What we _could_ do is make this function (and associated parts) hidden from a documentation point of view: that way we're "hiding" them from users, so they're less likely to rely on them, even though they're _technically_ part of the public API.
 
That reminds me, I think we do this in a few places, and the structs show up in the main module docs.
This can be seen in the `Re-exports` of lrlex https://docs.rs/lrlex/latest/lrlex/
I believe that I recall hearing that glob imports makes these not show up in the main docs page.

*note to self* I think LexFlags can be made unhidden, now that it is more visible in the API.

> Another possibility -- and I haven't thought this through! -- is that we could stick the format parsing stuff in its own crate. I'm not sure that would buy us much, but perhaps it would solve these problems?

I was a little bit hesitant at first but the thought is growing on me. I wonder if this wouldn't solve a bunch more duplication.
Like we have duplicate copies of *a lot* of stuff between lrlex and lrpar some pub, like `Visibility`, `RustEdition`, and others private like `Quote*`.  It would be really nice if these just had a single source of truth, and some of them are not private for any reason other than to avoid cluttering the API.

I think if we want to try to gravitate more of that other stuff into the crate too, it might be worth doing.
I have leaned towards calling these kinds libaries `support` libraries, like `grmtools-support` or something along those lines...

ratmice:
I'm thinking just go with hidden in this patch for now, and then try out a support library in a separate branch/series.

cfgrammar/src/lib/header.rs
ratmice: Don't think that is possible, I had forgot that this is borrowed, so to_lowercase() would return a temporary.
I think callers will have to do it, unless we want to change these to `Strings` so we can lowercase them.

cfgrammar/src/lib/header.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Parser for the `%grmtools` section
+pub struct GrmtoolsSectionParser<'input> {
+    src: &'input str,
+    required: bool,
+}
+
+lazy_static! {
+    static ref RE_LEADING_WS: Regex = Regex::new(r"^[\p{Pattern_White_Space}]*").unwrap();
+    static ref RE_NAME: Regex = Regex::new(r"^[a-zA-Z][a-zA-Z_]*").unwrap();
```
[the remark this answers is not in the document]
ratmice: should be done in [34ac2cf](https://github.com/softdevteam/grmtools/pull/544/commits/34ac2cf7e9a4abcce77c97f71c764417f3aac46f)

cfgrammar/src/lib/header.rs
ratmice: Should be fixed in [a56b217](https://github.com/softdevteam/grmtools/pull/544/commits/a56b21779e29b99d3778c4b5afaf0e6f6287cbd5)

ratmice:
So, I think the main issue that I am a little uncertain about (now that I think about it)
is when it comes time to integrate this into the lex/yacc parsers, how to integrate the `HeaderError` into the respective lex/yacc error kinds.

There is kind of two ways we can go about it:

Make sure that we can map the variants of `HeaderError` onto variants of the respective `ErrorKind`.
Have an error that wraps the other `GrmtoolsSectionErrorKind(HeaderError)`.

cfgrammar/src/lib/header.rs
```diff
@@ -7,8 +7,6 @@ use regex::Regex;
 pub enum HeaderError {
     MissingGrmtoolsSection(Span),
     IllegalName(Span),
```
ratmice, line 10: I hadn't really thought about it when writing this type,
But perhaps these should be made more like the other `YaccGrammarErrorKind` etc.

Though these errors are a lot simpler in the sense that we really cannot continue parsing after any of them.

ratmice: I did some messing around with this,  There isn't really a problem translating `HeaderError` as is into a `YaccGrammarError` directly, we'll need to make sure the `Kinds` align properly though....

```
impl From<HeaderError> for YaccGrammarError {
      ....
      match header_err {
           HeaderError::MissingGrmtoolsSection(span) => {
                   kind: YaccGrammarErrorKind::MissingGrmtoolsSection,
                   spans: vec![spans],
           }
      }
}
```

The other thing we could do is make `HeaderError` more like
the others, then expose them as part of the public API.

```
struct HeaderError{ kind: HeaderErrorKind, span: Span}

enum HeaderErrorKind {
    MissingGrmtoolsSection,
    ...
}

enum YaccGrammarErrorKind{
    ...
    GrmtoolsSectionError(HeaderErrorKind),
}
```

I guess that the reason it currently has `Span` rather than `Vec<Span>` is it doesn't do any duplication checking.
which I had kind of considered more semantic checking than parsing, however I suppose if we want to enforce that property across all the sections, it probably makes more sense to do it here all in one place.

ratmice: Anyhow I guess I'm well on my way to convincing myself it should just follow the current error patten other errors have, and add duplication checking, so it too uses a `*Error{kind, spans}`.

lrpar/cttests/src/lib.rs
```diff
@@ -388,20 +394,32 @@ fn test_grmtools_section() {
...
+            eprintln!("{}", s);
+            parse_error = true;
         }
-        println!("{:?}", val);
+        let parser = cfgrammar::header::GrmtoolsSectionParser::new(src, true);
+        let res = parser.parse();
+        let (header_parsed, _) = res.unwrap();
+        assert_eq!(yacc_parsed, Some(header_parsed));
+        assert!(!parse_error);
+        println!("{:?} {:?}", src, yacc_parsed);
```
ltratt, line 422: I think we should kill the debugging `print`s/`eprint`s in tests.

cfgrammar/src/lib/header.rs
```diff
@@ -139,16 +154,17 @@ impl<'input> GrmtoolsSectionParser<'input> {
         Self { src, required }
     }
 
-    pub fn parse(&'_ self) -> Result<(Vec<ValBind<'_>>, usize), HeaderError> {
+    #[allow(clippy::type_complexity)]
+    pub fn parse(&'_ self) -> Result<(HashMap<(&'_ str, Span), Value<'_>>, usize), HeaderError> {
```
ratmice, line 158: Oops, we really *don't* want this `Span` in the key part of the hash table.

lrpar/cttests/src/lib.rs
ratmice: Yeah, It is mostly for the moment, current I *often* find myself running `cargo test -- --nocapture | grep '%grmtools'`,


But I'll remove them once this is up to form, I've gone ahead and marked as draft since it seems I still have a bit of work to go before I would be comfortable considering replacing the other parsers with it.

cfgrammar/src/lib/header.rs
ratmice: Should be fixed in [79fb45b](https://github.com/softdevteam/grmtools/pull/544/commits/79fb45b811682e274553cad36b6993069944e50f)

lrpar/cttests/src/lib.rs
ratmice: I *think* i got all of them removed in [79fb45b](https://github.com/softdevteam/grmtools/pull/544/commits/79fb45b811682e274553cad36b6993069944e50f)

cfgrammar/src/lib/header.rs
ratmice: Should be finished in [79fb45b](https://github.com/softdevteam/grmtools/pull/544/commits/79fb45b811682e274553cad36b6993069944e50f)

This is a *little* weird, because it means we need to do semantic checking in the yacc parser.
So we return a `HeaderError`, but only produce semantic checking errors `DuplicateEntry`, while `HeaderError` also contains parse error things.

ratmice:
It seems to me like it should be ready for review now, and now it should more closely match the error handling of `YaccParser`, `LexParser`, in the *other* cases we should be using a `Box<dyn Error>` already so that should work fine there too.

I also think it improved the *AST* we returned by moving to a `HashMap`

ltratt:
Please squash.

ratmice:
Sadly, I forgot to run the duplicates test, and it actually failed due to the way the test was written not coalescing errors.

Squashed with that fixed.

cfgrammar/src/lib/header.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn test_header_duplicates() {
+        let src = "%grmtools {dupe, !dupe, dupe: test}";
+        for flag in [true, false] {
+            let parser = GrmtoolsSectionParser::new(src, flag);
+            let res = parser.parse();
+            let errs = res.unwrap_err();
+            assert_eq!(errs.len(), 1);
+            assert_eq!(errs[0].kind, HeaderErrorKind::DuplicateEntry);
+            assert_eq!(errs[0].spans.len(), 3);
+
```
ratmice, line 323: newline..

ratmice: Squashed a rustfmt in [104b442](https://github.com/softdevteam/grmtools/pull/544/commits/104b44248421631504da66263eea7ffb04bbbcbe)

PR #545, opened 2025-03-30
Clean up unused dependencies, and reduce duplicates, add Zlib license

ratmice:
Clean up unused dependencies, and reduce duplicates, add Zlib license
    
    Remove duplicates
    * Bumping yaml-rust2 reduces the number of versions of hashbrown to 1.
    * Turning default features off for wasm-bindgen allows us to share a
     dev-dependency with the same features via wasm-bindgen-test.
    
    yaml-rust2 enables a feature of hashbrown that causes a Zlib licensed library
    `foldhash` to be used.
    
    Unused dependencies:
    * maplit
    * static_assertions
    
    Other crates are moved to the workspace, to avoid specifying version numbers
    in workspace members so they can't get out of sync and pull in multiple versions.

AFACT the only breaking changes in yaml-rust2 are the bumping of the version of hashlink it depends upon,
which is exported from it's public API. It seems like it cold only affect this crate if we also depended upon hashlink for anything and used it in a way that required the version to be in sync with yaml-rust2's usage.


ratmice:
This still leaves a couple of duplicate versions, both windows-sys and syn have duplicate instances of the same crate version because different features are enabled between dev-dependency and dependency.  But it wasn't obvious to me how to avoid it.

ltratt:
@ratmice Something seems to have got stuck on this PR. Can you force push a rebase against master and I can try merging that to see if it unsticks things.

ratmice:
Sure, I Rebased it against master.

PR #546, opened 2025-03-31
Integrating the new grmtools section parser into YaccParser

ratmice:
Marking as draft for now, this is kind of just a first attempt that passes the testsuite.
I'll want to read it myself tomorrow, but overall I feel like the rough edges are starting to wear down.

While integrating this it became apparent that 

1. Borrowing `&str` wasn't going to be fun, so this converts it to use `String` (Allowing lowercase() to be added)
2. The Path type could be simplified to an `(Option<...>, (String, Span))` instead of an enum with multiple variants.
This made the codepaths dealing with optional namespaces somewhat tolerable.
3. Some things could be renamed to better reflect their purpose.

This also fixes a few bugs like skipping of initial whitespace before the section, and conclude by using a position including the final "}".

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -18,19 +18,40 @@ pub enum HeaderErrorKind {
     ExpectedToken(char),
     DuplicateEntry,
 }
-
+/// Indicates a value prefixed by an optional namespace.
+/// `Foo::Bar` with optional `Foo` specified being
+/// ```rust,ignore
+/// Namespaced{
+///     namespace: Some(("Foo", ...)),
+///     member: ("Bar", ...)
+/// }
+/// ```
+///
+/// Alternately just `Bar` alone without a namespace is represented by :
+/// ```rust,ignore
+/// Namespaced{
+///     namespace: None,
+///     member: ("Bar", ...)
+/// }
+/// ```
 #[derive(Debug, Eq, PartialEq)]
-#[doc(hidden)]
-pub enum Path<'a> {
-    Ident(&'a str, Span),
-    Scoped((&'a str, Span), (&'a str, Span)),
+pub struct Namespaced {
+    pub namespace: Option<(String, Span)>,
+    pub member: (String, Span),
 }
 
 #[derive(Debug, Eq, PartialEq)]
 #[doc(hidden)]
-pub enum Setting<'a> {
-    PathLike(Path<'a>),
-    ArgLike(Path<'a>, Path<'a>),
+pub enum Setting {
+    /// A value like `YaccKind::Grmtools`
+    Unitary(Namespaced),
+    /// A value like `YaccKind::Original(UserActions)`.
+    /// In that example the field ctor would be: `Namespaced { namespace: "YaccKind", member: "Original" }`.
+    /// The field would be `Namespaced{ None, UserActions }`.
+    Constructor {
+        ctor: Namespaced,
+        arg: Namespaced,
+    },
     Num(u64, Span),
 }
 
@@ -42,9 +63,9 @@ pub struct GrmtoolsSectionParser<'input> {
 }
 
 #[derive(Debug, Eq, PartialEq)]
-pub enum Value<'a> {
+pub enum Value {
     Flag(bool),
-    Setting(Setting<'a>),
+    Setting(Setting),
 }
 
 lazy_static! {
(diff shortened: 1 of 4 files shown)


cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -382,119 +412,60 @@ impl YaccParser {
...
+                            && *value == yk_str
+                        {
+                            self.yacc_kind = Some(yk);
                         }
-                    } else {
-                        return Err(YaccGrammarError {
-                            kind: YaccGrammarErrorKind::InvalidGrmtoolsHeaderKey,
-                            spans: vec![Span::new(key_start_pos, key_end_pos)],
-                        });
                     }
```
ratmice, line 448: I hadn't settled here on how best to return an error on invalid `yacckind` value in this case or the `Constructor` case. At the very least this loop should stop after the first match... 

I guess i'll try `let x = for ... {  if ... {break Some(yk) } else { None } };` then throw an error on None.

ratmice: I did a pass over this in [23a1391](https://github.com/softdevteam/grmtools/pull/546/commits/23a1391cd40eed077eca29612d824ed17d11fec7)

Having rewritten it, I feel like the error handling here is probably more robust, in the sense
that previously we were using `starts_with` in many cases, which the `invalid_yacckind` tests I added should cover.

That said I wouldn't be suprised if this could be drastically improved somehow by avoiding `loop`.

ratmice:
Would happily split out renames or whatever to their own patch if that helps?
Other than that sort of stuff I think I'm running out of ideas for further cleanups to this PR so marking as ready for review.

cfgrammar/src/lib/header.rs
```diff
@@ -18,19 +18,18 @@ pub enum HeaderErrorKind {
     ExpectedToken(char),
     DuplicateEntry,
 }
+// Indicates a value prefixed by an optional namespace.
```
ltratt, line 21: Should be `///` I think.

cfgrammar/src/lib/header.rs
```diff
@@ -18,19 +18,18 @@ pub enum HeaderErrorKind {
...
+// `Foo::Bar` with optional `Foo` specified, becomes (Some(("Foo", ...)), ("Bar", ...))
+// Alternately just `Bar` alone without a namespace is represented by (None, ("Bar", ...)).
+pub type Namespaced = (Option<(String, Span)>, (String, Span));
```
ltratt, line 24: I wonder if this has got fiddly enough that it should be a `struct`? I must admit that I tend to confuse pairs when the LHS and RHS types are more-or-less the same.

ltratt, line 24: I think this should probably be a `struct`: at least I tend to get a bit confused with types where the LHS and RHS are more-or-less the same.

cfgrammar/src/lib/header.rs
```diff
@@ -18,19 +18,18 @@ pub enum HeaderErrorKind {
...
-}
-
-#[derive(Debug, Eq, PartialEq)]
-#[doc(hidden)]
-pub enum Setting<'a> {
-    PathLike(Path<'a>),
-    ArgLike(Path<'a>, Path<'a>),
+pub enum Setting {
+    // A value like `YaccKind::Grmtools`
+    Unitary(Namespaced),
```
ltratt, line 30: These names felt maybe a bit different in comparison to `Num`? I was wondering whether `Constructor`, for example, is "key:value" in disguise?

cfgrammar/src/lib/header.rs
ratmice: I'll give it a shot

cfgrammar/src/lib/header.rs
```diff
@@ -144,21 +143,26 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
-                Path::Scoped((name, name_span), (scoped_val, scoped_span)),
+                (
+                    Some((name, name_span)),
+                    (namespaced_val, namespaced_val_span),
+                ),
                 i,
             ))
         } else {
-            Ok((Path::Ident(name, name_span), i))
+            // Without a namespace.
```
ratmice, line 189: Comment seems superfluous now that we've made this a struct.

ratmice: Should be fixed in [1fefed0](https://github.com/softdevteam/grmtools/pull/546/commits/1fefed0770fc22f6511a8848a52330b921feec89)

Also the locals here could be more clear, being more similar to the struct field names.

ltratt:
Please squash.

ratmice:
Squashed.

PR #547, opened 2025-04-01
Initial attempt at adding RecoveryKind to yacc %grmtools section

ratmice:
Here is an initial attempt at adding `recoverer` to the grmtools section for yacc.

Things I don't like:

1. It feels to me like quite a bit of boiler plate for adding a single flag (checking all the kinds of value it could be but shouldn't be to throw errors...).
2. Not checking for unused flags `%grmtools{unknown: Foo}`
3. Parsing the section multiple times

I have some thoughts on these last two, for grmtools but lex is perhaps more complicated.

In the same way that ctparserbuilder calls into `self.yacckind = ast_validation.yacc_kind();` we could use a similar path to communicate the key/values that went unused by `YaccParser`.

On the first one, I'll try and think of helper functions, but we have things like YaccKind which can be *either* a `Unitary` or a `Constructor`, so we need to be able to check that it is `oneOf` those.

I'll try and fiddle with it more, but this is at least a naive draft implementation.

Edit: I have some general thoughts on things to attempt,
largely replacing the `HashMap` with something that tracks key consumption,
and a small mechanism for querying values based on a `repr` enum.

```
#[repr(u8)]
enum ValueRepr {
   Flag = 1 << 0,
   Setting = 1 << 1
}
```

```
#[repr(u8)]
enum SettingRepr {
    Num = 1 << 2,
    Unitary = 1 << 3,
    Constructor = 1 << 4,
}
```

Anyhow the feeling I get is that a key query API that checks the value in the hashmap against one of those in bitmask form. That along with, marking keys as "used", doesn't seem to fall into any complexity traps.

I'll elaborate tomorrow in code form...

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -4,7 +4,6 @@ use crate::{
     Span,
 };
 use lazy_static::lazy_static;
-use quote::{quote, ToTokens};
 use regex::{Regex, RegexBuilder};
 use std::{error::Error, fmt};
 
@@ -93,20 +92,6 @@ pub struct Namespaced {
     pub member: (String, Span),
 }
 
-impl ToTokens for Namespaced {
-    fn to_tokens(&self, toks: &mut proc_macro2::TokenStream) {
-        let namespace = self.namespace.as_ref().map(|(str, _)| str);
-        let member = &self.member.0;
-        toks.extend(if namespace.is_some() {
-            quote! {
-                #namespace::#member
-            }
-        } else {
-            quote!(#member)
-        })
-    }
-}
-
 #[derive(Debug, Eq, PartialEq)]
 #[doc(hidden)]
 pub enum Setting {
@@ -122,16 +107,6 @@ pub enum Setting {
     Num(u64, Span),
 }
 
-impl ToTokens for Setting {
-    fn to_tokens(&self, toks: &mut proc_macro2::TokenStream) {
-        toks.extend(match self {
-            Self::Unitary(namespaced) => quote!(#namespaced),
-            Self::Constructor { ctor, arg } => quote!(#ctor(#arg)),
-            Self::Num(num, _) => quote!(#num),
-        })
-    }
-}
-
 /// Parser for the `%grmtools` section
 #[doc(hidden)]
 pub struct GrmtoolsSectionParser<'input> {
@@ -143,15 +118,6 @@ pub struct GrmtoolsSectionParser<'input> {
 pub enum Value {
     Flag(bool),
     Setting(Setting),
-}
-
-impl ToTokens for Value {
-    fn to_tokens(&self, toks: &mut proc_macro2::TokenStream) {
-        match self {
-            Value::Flag(flag) => toks.extend(quote!(#flag)),
-            Value::Setting(setting) => toks.extend(quote!(#setting)),
-        }
-    }
 }
 
 impl Value {
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -122,7 +122,7 @@ pub enum Value {
 
 impl Value {
     pub fn matches_query_mask(&self, mask: u16) -> bool {
-        let q = self.query_bits() as u16;
+        let q = self.query_bits();
         (q & mask) == q
     }
     pub fn query_bits(&self) -> u16 {
@@ -160,9 +160,9 @@ pub enum ValueQuery {
 /// The last 8 bits is reserved for `SettingQuery`
 #[repr(u16)]
 pub enum SettingQuery {
-    Unitary = (ValueQuery::Setting as u16) | 1 << 8,
-    Constructor = (ValueQuery::Setting as u16) | 1 << 9,
-    Num = (ValueQuery::Setting as u16) | 1 << 10,
+    Unitary = (ValueQuery::Setting as u16) | (1 << 8),
+    Constructor = (ValueQuery::Setting as u16) | (1 << 9),
+    Num = (ValueQuery::Setting as u16) | (1 << 10),
 }
 
 lazy_static! {
@@ -431,7 +431,7 @@ impl Header {
         &mut self.contents
     }
 
-    pub fn mark_key_used(self: &mut Self, key: &str) {
+    pub fn mark_key_used(&mut self, key: &str) {
         let key = key.to_owned();
         let pos = self.used.binary_search(&key);
         match pos {
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -88,8 +88,8 @@ impl fmt::Display for HeaderContentsError {
 /// ```
 #[derive(Debug, Eq, PartialEq)]
 pub struct Namespaced {
-    pub namespace: Option<(String, Span)>,
-    pub member: (String, Span),
+    pub namespace: Option<(String, Option<Span>)>,
+    pub member: (String, Option<Span>),
 }
 
 #[derive(Debug, Eq, PartialEq)]
@@ -104,7 +104,7 @@ pub enum Setting {
         ctor: Namespaced,
         arg: Namespaced,
     },
-    Num(u64, Span),
+    Num(u64, Option<Span>),
 }
 
 /// Parser for the `%grmtools` section
@@ -152,30 +152,30 @@ impl From<YaccKind> for Value {
     fn from(kind: YaccKind) -> Value {
         match kind {
             YaccKind::Grmtools => Value::Setting(Setting::Unitary(Namespaced {
-                namespace: Some(("yacckind".to_string(), Span::new(0, 0))),
-                member: ("grmtools".to_string(), Span::new(0, 0)),
+                namespace: Some(("yacckind".to_string(), None)),
+                member: ("grmtools".to_string(), None),
             })),
             YaccKind::Eco => Value::Setting(Setting::Unitary(Namespaced {
-                namespace: Some(("yacckind".to_string(), Span::new(0, 0))),
-                member: ("eco".to_string(), Span::new(0, 0)),
+                namespace: Some(("yacckind".to_string(), None)),
+                member: ("eco".to_string(), None),
             })),
             YaccKind::Original(action_kind) => Value::Setting(Setting::Constructor {
                 ctor: Namespaced {
-                    namespace: Some(("yacckind".to_string(), Span::new(0, 0))),
-                    member: ("original".to_string(), Span::new(0, 0)),
+                    namespace: Some(("yacckind".to_string(), None)),
+                    member: ("original".to_string(), None),
                 },
                 arg: match action_kind {
                     YaccOriginalActionKind::NoAction => Namespaced {
-                        namespace: Some(("yaccoriginalactionkind".to_string(), Span::new(0, 0))),
-                        member: ("noaction".to_string(), Span::new(0, 0)),
+                        namespace: Some(("yaccoriginalactionkind".to_string(), None)),
+                        member: ("noaction".to_string(), None),
                     },
                     YaccOriginalActionKind::UserAction => Namespaced {
-                        namespace: Some(("yaccoriginalactionkind".to_string(), Span::new(0, 0))),
-                        member: ("useraction".to_string(), Span::new(0, 0)),
+                        namespace: Some(("yaccoriginalactionkind".to_string(), None)),
+                        member: ("useraction".to_string(), None),
                     },
                     YaccOriginalActionKind::GenericParseTree => Namespaced {
-                        namespace: Some(("yaccoriginalactionkind".to_string(), Span::new(0, 0))),
-                        member: ("genericparsetree".to_string(), Span::new(0, 0)),
+                        namespace: Some(("yaccoriginalactionkind".to_string(), None)),
+                        member: ("genericparsetree".to_string(), None),
                     },
                 },
             }),
@@ -257,7 +257,7 @@ impl<'input> GrmtoolsSectionParser<'input> {
                         let num_str = &self.src[num_span.start()..num_span.end()];
                         // If the above regex matches we expect this to succeed.
                         let num = str::parse::<u64>(num_str).unwrap();
-                        let val = Setting::Num(num, num_span);
+                        let val = Setting::Num(num, Some(num_span));
                         i = self.parse_ws(num_span.end());
                         Ok((key_name, key_span, Value::Setting(val), i))
                     }
@@ -312,8 +312,8 @@ impl<'input> GrmtoolsSectionParser<'input> {
             i = self.parse_ws(j);
             Ok((
                 Namespaced {
-                    namespace: Some((name, name_span)),
-                    member: (member_val, member_val_span),
+                    namespace: Some((name, Some(name_span))),
+                    member: (member_val, Some(member_val_span)),
                 },
                 i,
             ))
@@ -321,7 +321,7 @@ impl<'input> GrmtoolsSectionParser<'input> {
             Ok((
                 Namespaced {
                     namespace: None,
-                    member: (name, name_span),
+                    member: (name, Some(name_span)),
                 },
                 i,
             ))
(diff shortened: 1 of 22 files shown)


ratmice:
I added basic implementation of what I was thinking in [96d721d](https://github.com/softdevteam/grmtools/pull/547/commits/96d721d5c0b523c8d4873f06eaadb6c8c5584648) It doesn't get used yet anywhere, so is largely untested.

Anyhow that is the basic idea of what I had in mind.

cfgrammar/src/lib/header.rs
```diff
@@ -335,6 +395,110 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
+        // simple to implement.
+        self.contents
+            .iter()
+            .filter(|(key, _)| self.used.binary_search(key).is_ok())
+    }
+
+    pub fn query_key(
+        &self,
+        key: &str,
+        query_mask: u16,
```
ratmice, line 444: Of course there is some borrow checker issues with the combination of `query_key`/`mark_key_unused`,
So this interface seems to require some changes.  I haven't gotten the entry API to work (since we need to mutate the `header` not the `HashMap`. So might have to resort to `contains_key` -> `mutate` -> `query`.

ratmice: Should be fixed in [06458b1](https://github.com/softdevteam/grmtools/pull/547/commits/06458b1fcd1ac37d837edaba89d2ef8e5db919ff)

I'm not sure it's the best fix, the "mark as used" even on error behavior
is intended to avoid the following situation:

1. specify `%grmtools{yacckind: Foo}` 
2.  Get an "invalid yacc kind" for `Foo` error.
3.  *also* get `unused yacckind`

ratmice:
That at least seems like a first try at fixing all 3 of those issues, and I haven't thought of any more either.

nimbleparse/src/main.rs
```diff
@@ -154,7 +155,57 @@ fn main() {
...
+            Some(Err(err)) => {
+                eprintln!("{}", err);
+                process::exit(1);
+            }
+            Some(_) => {}
+            None => {}
+        }
+    }
+    let recoverykind = recoverykind.unwrap_or(RecoveryKind::CPCTPlus);
+    ast_validation.header_mut().mark_key_used("recoverer");
```
ratmice, line 199: It's worth pointing out a difference here between nimbleparse and `CTParserBuilder`.

In nimbleparse we `mark_key_used("recoverer")` even when the value is ignored because the value was set on the command line.

But in `CTParserBuilder` it will complain if you set `recoverykind(CPCTPlus)` but also have the key in the `%grmtools` section.  The thinking for allowing it to error in `CTParserBuilder` was it might be confusing if one of the build.rs or `%grmtools` section went unused.

So how it is, if you specify it in both places you should get an `unused in %grmtools section`  error.

ltratt: > There is kind of an interaction here with the forcing behavior of lrlex, we should probably consider how to handle it generally? Anyhow I'm not entirely certain what to think about it yet.

In the sense that it's not clear which setting should be forcing which?

ratmice: > > There is kind of an interaction here with the forcing behavior of lrlex, we should probably consider how to handle it generally? Anyhow I'm not entirely certain what to think about it yet.
> 
> In the sense that it's not clear which setting should be forcing which?

I think a good way to explain it is that currently, we have `Force`, `Default`, `NoDefault`, but we lack `XOR`.

* Force overrides input text.
* Default is overridden by input text.
* NoDefault forces input text to specify.

So we don't have a way to specify "one of the builder, or the input text should specify, but not both"
That is actually the behavior which `CTParserBuilder` kind of implements for `RecoveryKind` because `RecoveryKind` only gets marked used if it is `None` the `CTParserBuilder`, so if it is specified in both,
we complain that `RecoveryKind` is unused in the input file.

This wasn't really an intended thing, I only noticed it when I added my comment, but the behavior as implemented does have a correlation to an `XOR` forcing mechanism

ratmice: I guess the other thing is we originally added `YaccKindResolver`, there was only one value `yacckind` for `cfgrammar::YaccParser`, but when we got to `lrlex` there were multiple flags.  Now that we integrate `CTParserBuilder` the forcing mechanism is too specific, and it isn't clear how to expand it across these multiple crates.  Neither of the forcing mechanisms the way we did it in `cfgrammar` nor the way we did it in `lrlex` works when crossing crate boundaries.

I guess that is where my head is most fogged up.

ratmice: Part of me wonders if I shouldn't be looking towards the `mark_key_used` mechanism, and adding a flag to it
to make it less specific to `used/unused`, and then reusing the structure as an input parameter, and doing the whole `merge` thing we did with lexflags but with all these string arguments.

so: 
```
enum Mark {
  Used: 1 << 0,
  Default: 1 << 1,
  NoDefault: 1 << 2,
  Force: 1 << 3
}
fn mark_key(key, mark: Mark)
```

cfgrammar/src/lib/header.rs
```diff
@@ -335,6 +395,110 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
+pub struct Header {
+    contents: HashMap<String, (Span, Value)>,
+    used: Vec<String>,
+}
+
+impl Header {
+    pub fn new() -> Self {
+        Header {
+            contents: HashMap::new(),
+            used: vec![],
```
ratmice, line 422: Given the thoughts on extending the `mark` behavior to include more than `used`, but also forcing flags.
I've been thinking about ways to get rid of the `used` value entirely,

For instance whether we can use [`hashbrown::HashTable`](https://docs.rs/hashbrown/latest/hashbrown/hash_table/struct.HashTable.html) to store the mark flags along side the key in a way that allows them to be mutated separately without affecting the `Hash`.

I don't yet know whether this HashTable API though allows us to add to keys in a way that doesn't also require us to add an element.  At the very least relatively popular crates like `SlotMap` unfortunately don't fall under such a use case as slots are created during insertion.

The other option is just using a sorted vec instead `contents: Vec<(String, Mark, Option<(Span, Value)>>` 
which honestly seems like the simplest option to me.

nimbleparse/src/main.rs
ltratt: Just so I'm clear: XOR would raise an error if both defined incompatible settings?

ratmice: That would work, but I wasn't considering any notion of compatibility, just that it must be `None` in one place, and `Some(value)` in the other, but not both `Some(value)`.  Just enforcing that there must be one source of truth either the build.rs or the input source file itself.

But yeah I could see the argument for allowing the same value to be set in both places if e.g. they are equal or compatible.

ratmice: FWIW, here is a better tested version that implements the merge, unused checking and adds a required mark.

That is to say, it is approaching the merging behavior or `LexFlags` except in a stringly typed fashion, and on a per-key basis unlike lex_flags where we had different values for defaults and forcing.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=747e715aa04b227ad8d9d031db35cc45

Anyhow I still can't tell if it is a terrible idea or not...

There is probably quite a ways to go before it actually feels like the `HashMap` API (if we even want that).
But essentially it tries to mimic `HashMap` api which includes these `merge`, `missing`, and `unused` functions.
Which specifies on a per-key basis how to resolve conflicts, or whether they are required but missing, or present but unused.

ratmice: I went ahead and integrated the thing in [cdccd47](https://github.com/softdevteam/grmtools/pull/547/commits/cdccd4713f4b3ae9b3cc4a5c1a038ee97a592d8e) just to see how far away from getting it working in place of the `HashMap` we were.

I had to add the `Borrow` thing to a few places, and add `contains_key`, and change some method signatures for the `used` function.

But overall it did drop into place pretty well...

ratmice: So the next step making `CTParserBuilder` use these marks and merge behaviors instead of things `YaccKindResolver`, is sort of being a painful task, and showing it's warts.

I added things like `impl Into<header::Value> for YaccKind` and then insert them into a separate header than the one read from the grammar source.

One of the issues is that `Header` assumes it's going to be read from a file, so It has things like `Span`.
while the one constructed from the `CTParserBuilder` doesn't have span data.

Ideally I guess we'd have headers with/without spans and be able to merge them into one with optional spans
Anyhow, I'll press on and maybe this just my distaste for stringly typed nature of it, but I certainly don't have high hopes for this particular experiment.

ratmice: So [c16876c](https://github.com/softdevteam/grmtools/pull/547/commits/c16876c57d4b52f14ae649264f5dbc13968a48e9) is the first functioning attempt which replaces `YaccKindResolver` with this whole `merge_from` behavior, there is quite a few warts in there.

But it does appear to function.

lrtable/src/lib/itemset.rs
```diff
@@ -199,8 +205,21 @@ mod test {
     }
 
     fn eco_grammar() -> YaccGrammar {
+        let mut header = Header::new();
+        header
+            .contents_mut()
+            .set_merge_behavior(&"yacckind".to_string(), MergeBehavior::Ours);
+        header.contents_mut().mark_required(&"yacckind".to_string());
+        // FIXME seems like this this should be Eco!
```
ratmice, line 213: Searching through this patch for `eco`, it appears that there are
a couple more instances where the yacc kind seems like it should be `Eco`
but at some point most likely was accidentally changed to `GenericParseTree`.

cfgrammar/src/lib/yacc/firsts.rs
```diff
@@ -201,8 +216,20 @@ mod test {
 
     #[test]
     fn test_first_no_subsequent_rules() {
+        let mut header = Header::new();
```
ratmice, line 219: It would be really nice to simplify a lot of this testsuite stuff with a macro.
I think I'll try to make a `test_support.rs` module, with a macro `header_for_yacckind!` or something to that effect.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -25,10 +25,10 @@ impl ASTWithValidityInfo {
     /// encountered during the construction of it.  The `ASTWithValidityInfo` can be
     /// then unused to construct a `YaccGrammar`, which will either produce an
     /// `Ok(YaccGrammar)` or an `Err` which includes these errors.
-    pub fn new(yacc_kind_resolver: YaccKindResolver, s: &str) -> Self {
+    pub fn new(header: Header, s: &str) -> Self {
```
ratmice, line 28: So, I wonder here if we can take a `header: &mut Header` here,
and then avoid returning it in `YaccParser::build()`, it seems possible that would simplify things.

But I'm not sure whether the compiler will be able to tell that it isn't capture by the `GrammarAST` return value.

ratmice: It appears that we can, this was done in [4c59bb9](https://github.com/softdevteam/grmtools/pull/547/commits/4c59bb93225d6dda2b0a93bdca2fcdf9734e32c1).

It required adding a lifetime to (now) `YaccParser<'a>`, but I think that should be a private type.
Seemed like a small win to me.

cfgrammar/src/lib/yacc/firsts.rs
ratmice: I did that in [a7aee25](https://github.com/softdevteam/grmtools/pull/547/commits/a7aee25a8140bb7d36e8376474c49267fd32bfc9).

lrtable/src/lib/itemset.rs
ratmice: It appears this one at least has been this way since the beginning of lrtable.
I'll look into the others I see.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -369,15 +367,14 @@ impl YaccParser {
...
-        let section_required = matches!(self.yacc_kind_resolver, YaccKindResolver::NoDefault);
-        let header_parser = GrmtoolsSectionParser::new(&self.src, section_required);
+        let header_parser = GrmtoolsSectionParser::new(&self.src, false);
         let result = match header_parser.parse() {
-            Ok((mut header, i)) => {
-                let result = self.update_yacckind(&mut header, i);
-                self.header = header;
-                result
+            Ok((parsed_header, i)) => {
+                self.header.merge_from(parsed_header).unwrap();
```
ratmice, line 373: This is `unwrap()` is just a temporary hack that was done just to get things working, 
These errors should somehow be converted into a grammar error.

cfgrammar/src/lib/header.rs
ratmice: Removing used was done in [cdccd47](https://github.com/softdevteam/grmtools/pull/547/commits/cdccd4713f4b3ae9b3cc4a5c1a038ee97a592d8e)

ratmice:
So, I *think* this is at a good place to decide if we should put any more effort into it, or cut our losses here.
There is quite a bit of small nuisances, and code organization that can be improved, which I have been ignoring, because I didn't want to fine tune it all and then decide I didn't even like it the basic idea...

But from a big picture perspective of having this header kind of tree, it's basic purpose of merging things CTParserBuilder or nimbleparse and the input sources.   How it gets passed around and reused across crates because it just holds settings as a bunch of `header::Values` aka `Strings`...

I at least find the tree structure moderately interesting, and I can't remember seeing anything generalized of it's sort (Although it isn't really generalized, and not having seen anything like it might be a reason enough to avoid it!)...

cfgrammar/src/lib/yacc/parser.rs
ratmice: I fixed that by trying to map this error to one of the `Duplicate` errors `DuplicateGrmtoolsSectionEntry`.

[06390fa](https://github.com/softdevteam/grmtools/pull/547/commits/06390fa0030dda28769140947d431ad97df480be)
This subsequently required me to add a new spans kind for none-or-more spans in [6a7f5c5](https://github.com/softdevteam/grmtools/pull/547/commits/6a7f5c52d73ff978f4479ef0a4d0af31fd8e3ab6)

I'm not certain that reusing `DuplicateGrmtoolsSectionEntry` is the right thing, e.g. usually we see duplicate because of something silly like:

```
%grmtools{yacckind: Grmtools, yacckind: Grmtools}
```

however this is like

```
cp_builder.yacckind(YaccKind::Grmtools)
```

Parsing a source file with:

```
%grmtools{yacckind: Grmtools}
```

nimbleparse/src/diagnostics.rs
```diff
@@ -153,11 +153,14 @@ impl<'a> SpannedDiagnosticFormatter<'a> {
             } else {
                 out.push('\n');
                 let s = match e.spanskind() {
-                    SpansKind::DuplicationError => {
+                    SpansKind::DuplicationError | SpansKind::OptionalDuplicationSpan => {
```
ratmice, line 156: This is a bit awkward, we can only count the duplicate occurances by their spans, so for `OptionalDuplicationSpan`, we can't accurately know how many duplicates there are, just how many we see.

This is kind of akin to the old C compiler problem where you can say get C preprocessor variables
`cc -DFOO` where their source location is on `<command-line>`.

Except here because we're dealing with `Option<Span>` ↝ `Vec<Span>`, these `None` values end up dropped on the floor.

ratmice: Anyhow it seems likely to me that these should be something else besides `Option<Span>`, perhaps something to the effect of:

```
enum Location {
   Span(Span),
   CommandLine,
   Other(String),
}
```

Allowing `Other("From CTParserBuilder".to_string())` or some such.

ratmice: Attempted a fix in [714d9da](https://github.com/softdevteam/grmtools/pull/547/commits/714d9dab9ec5e64320499de5e2d98ac409b21297) anyways

ratmice:
So, For now I feel like I should probably just close this, while it is finally in a state where it is both pretty much complete and I can't think of any glaring holes, It isn't exactly in a state where I feel it could reasonably be reviewed.

So I'll spend some time reordering the commits, so that free standing things can be committed separately
and squashing out any interim changes... hopefully I'll manage to get something more review friendly...

ltratt:
Thanks!

PR #549, opened 2025-04-07
Lrpar section parser ctbuilder take2

ratmice:
Here is a sequence of patches, which comprise my attempts to clean up #547 
This is still a bit of a doozy, but I'm not sure how to split it up in any nicer way.
I hope the commit messages, and inline docs help a little in providing the motivation for each patch.

I have left out the `RecoveryKind` additions for a subsequent patch.

I'm going to keep this as a draft for now, I at least want to go through and do some manual testing,
and trying to figure out how terribly the errors this patch produces have regressed.
I also think I'm probably leaving in some `YaccGrammarErrorKind` variants that clippy is not discovering are `not used`
that should probably be removed.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -1,9 +1,19 @@
-use crate::{yacc::ParserError, Location, Span};
+use crate::{
+     markmap::MarkMap,
+    yacc::{ParserError, YaccKind, YaccOriginalActionKind},
+    Location, Span,
+};
 use lazy_static::lazy_static;
 use regex::{Regex, RegexBuilder};
 use std::collections::{hash_map::Entry, HashMap};
 use std::{error::Error, fmt};
 
+/// An error regarding the `%grmtools` header section.
+///
+/// It could be any of:
+///
+/// * An error during parsing the section.
+/// * An error resulting from a value in the section having an invalid value.
 #[derive(Debug, Clone)]
 pub struct HeaderError {
     pub kind: HeaderErrorKind,
@@ -32,6 +42,7 @@ pub enum HeaderErrorKind {
     ExpectedToken(char),
     DuplicateEntry,
     InvalidEntry(&'static str),
+    ConversionError(&'static str, &'static str),
 }
 
 impl fmt::Display for HeaderErrorKind {
@@ -42,6 +53,9 @@ impl fmt::Display for HeaderErrorKind {
             HeaderErrorKind::ExpectedToken(c) => &format!("Expected token: '{}", c),
             HeaderErrorKind::InvalidEntry(s) => &format!("Invalid entry: '{}'", s),
             HeaderErrorKind::DuplicateEntry => "Duplicate Entry",
+            HeaderErrorKind::ConversionError(t, err_str) => {
+                &format!("Converting header value to type '{}': {}", t, err_str)
+            }
         };
         write!(f, "{}", s)
     }
@@ -64,6 +78,7 @@ impl fmt::Display for HeaderErrorKind {
 /// }
 /// ```
 #[derive(Debug, Eq, PartialEq)]
+#[doc(hidden)]
 pub struct Namespaced {
     pub namespace: Option<(String, Location)>,
     pub member: (String, Location),
@@ -91,7 +106,12 @@ pub struct GrmtoolsSectionParser<'input> {
     required: bool,
 }
 
+/// The value contained within a `Header`
+///
+/// To be useful across diverse crates this types fields are limited to types derived from `core::` types.
+/// like booleans, numeric types, and string values.
 #[derive(Debug, Eq, PartialEq)]
+#[doc(hidden)]
 pub enum Value {
     Flag(bool, Location),
     Setting(Setting),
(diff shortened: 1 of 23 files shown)


ratmice:
I also wanted to ask you to please feel free to have adverse or allergic reactions to this code,

I will admit that it is starting to grow on me... primarily because *reasonable* people will only ever need to pass in `Header::new()` if all their things are in `%grmtools{...}` or never see it if they only use `nimbleparse` or a `ctbuilder`.

cfgrammar/src/lib/markmap.rs
```diff
@@ -0,0 +1,780 @@
...
+    Required,
+    MergeBehavior(MergeBehavior),
+}
+
+impl Mark {
+    fn repr(self) -> u16 {
+        match self {
+            Self::Used => 1 << 0,
+            Self::Required => 1 << 1,
+            Self::MergeBehavior(mb) => (1 << 2) | ((mb as u16) << 8),
```
ratmice, line 66: It is probably worth noting that `MergeBehavior(_)`  and it's inner element are both using more bits than necessary, I felt it kept the logic uniform to just give each behavior a 1 bit, and since these are opaque, and not pressed for space, it seemed like we could change it at any time...

I don't know if it is worth a comment, if we were pressed for space we could:
* Drop the `1 << 2` entirely for `MergeBehavior(_)`.
* Just make `MutuallyExclusive` always a zero bit.

cfgrammar/src/lib/header.rs
```diff
@@ -345,6 +347,132 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
+}
+
+impl Header {
+    pub fn new() -> Self {
+        Header {
+            contents: MarkMap::new(),
+        }
+    }
+
+    pub fn merge_from(
```
ltratt, line 363: We'll need to document these functions since they're part of the public interface.

cfgrammar/src/lib/header.rs
```diff
@@ -345,6 +347,132 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
+    ///
+    /// Regardless of whether this query mask check succeeds or fails,
+    /// if the `contents` contains the key, will mark the key as being used.
+    pub fn query(
+        &mut self,
+        key: &str,
+        query_mask: u16,
+    ) -> Option<Result<(&Location, &Value), HeaderError>> {
+        self.contents_mut().mark_used(&key.to_owned());
+        if let Some((key_loc, value)) = self.contents.get(key) {
```
ltratt, line 389: I wonder if clippy will complain about this not being `self.contents.get(key).map(...)`?

cfgrammar/src/lib/header.rs
```diff
@@ -345,6 +347,132 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
+        } else {
+            None
+        }
+    }
+}
+
+/// Repr vaues bit representation should not overlap `SettingQuery`,
+/// The first 8 bits is reserved for `ValueQuery`.
+#[repr(u16)]
+pub enum ValueQuery {
```
ltratt, line 435: I wonder if we need to expose the maskiness of this to users? The `enum` seems fine (users never need to know what the integer discriminant values are) but when we get `matches_query_mask(u16)` in the public API I do worry a bit about the cognitive load.

A dumb question is: do users ever need to query more than one thing at a time?

cfgrammar/src/lib/header.rs
```diff
@@ -474,6 +474,41 @@ impl Setting {
     }
 }
 
+impl From<YaccKind> for Value {
```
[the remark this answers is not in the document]
ltratt: The challenge with `From` is that it's infallible. At the moment it looks OK but it is a strong future commitment. In that sense `TryFrom` gives more wriggle room in the future. So it partly depends how confident we are about the API as-is.

ltratt:
I will admit that I'm struggling to pick up the high-level gist of where this is going. I don't think I've fully understood `ValueQuery`, for example: perhaps you can give me a layman's explanation at a high level?

cfgrammar/src/lib/header.rs
ratmice: More than one thing like more than one `key`, or more than one `mask`?

It isn't set up to query more than one key at a time (I never needed it, I don't think we ever do?).
It is used to query more than one value mask value for `yacckind` which can be a unitary `YaccKind::Grmtools` or a constructor `YaccKind::Original(_)`, this is used in the last patch [here](https://github.com/softdevteam/grmtools/pull/549/commits/34272ccaed6e9753612410223db3be8dfa51e7f0#diff-3f17b8629b7e418df92ecbab5e58f5a28a5adcf69d532f5fdb004526d0529ba4R460-R464) if that link works.

```
        if let Some(result) = self.header.query(
            "yacckind",
            SettingQuery::Unitary as u16 | SettingQuery::Constructor as u16,
        )
```

ltratt: Ah, then I really have misunderstood this!

If we do go down this route (and I can be convinced -- I just haven't understood it yet!) I think we would want to find a means that avoids `as`. I have become allergic to `as` because it has surprising effects on some, but not other, types.

cfgrammar/src/lib/header.rs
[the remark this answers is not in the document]
ratmice: Hmm, I had kind off assumed that because e.g. `nimbleparse` converts strings to `YaccKind` we'd always be able to convert or there would be some applicable string version of a `YaccKind`.

The most complex situation I can think of arises where we have a non-string parameter #541 comes to mind,
where I proposed to add a `RecoveryKind::CPCTPlusChannel(mut channel)`, and one cannot come up with a textual version for passing a channel, however map this value to the normal `RecoveryKind::CPCTPlus` via text.

It didn't occur to me we could have one with *no* mapping to text whatsoever, so I'll have to think about that.
My thoughts for the situation with `"CPCTPlus"` is that it's string intrepretation couold be mapped via the program to be interpreted *either* as `RecoveryKind::CPCTPlus` or as a `CPCTPlusChannel(...)` with a channel constructed by the program.

That is to say I can at least imagine having a surjective mapping, but I hadn't thought of cases where we don't have a mapping.

cfgrammar/src/lib/header.rs
ratmice: I don't know of any way to avoid `as` (I will look, maybe i've missed a way being added, but last time I looked it was the only way to convert from an `enum` to it's `#[repr(...)]`) 

At the very least we can wrap it in a function like `fn repr(self) -> u16`  though.

ratmice:
There is kind of two purposes to the `query` function (I already mentioned one but I'll repeat it here).

* In the case of `YaccKind` we want a value that is either a unitary like `YaccKind::Grmtools` or a constructor like `YaccKind::Original`, but we don't care about `bool` or `num` values like `case_insensitive` or `size_limit`.

The query mechanism thus checks if the value *mask* contains the `query_bits()` of the actual repr that was requested,
and *if not* it produces an error.

In that call to query for `YaccKind`, we only check the variants which were queried, and the rest are handled by this
[ _ => unreachable!("Handled all settings queried for"),](https://github.com/softdevteam/grmtools/pull/549/commits/34272ccaed6e9753612410223db3be8dfa51e7f0#diff-3f17b8629b7e418df92ecbab5e58f5a28a5adcf69d532f5fdb004526d0529ba4L541-L542) fallthrough branch. So this whole query mechanism is there to allow this error conversion in just one place so the main code can stick to the happy path.

cfgrammar/src/lib/header.rs
ratmice: It didn't for me, I don't know why it wouldn't though like maybe option is a special case?

I do kind of dislike `Option::map` and `Result::map`, because I can't tell you how many times i've gotten mixed up by it when I really intended to map over `Some(iterator)`.

ltratt:
Dumb question: will users tend to think of these things as query bits? Is there a higher level query API that might make more sense to them?

ratmice:
Maybe I can just take the `u16` usage out of the public API, make a way to pass in a `Iterator<Item=ValueQuery>`, that then internally does all the bit masking and abstracts it away

cfgrammar/src/lib/header.rs
ratmice: I partially did this in [c9756c8](https://github.com/softdevteam/grmtools/pull/549/commits/c9756c8fcadd82f6cb6784c3e83218c02368bf3f)

Then a did a pass over [f671bd4](https://github.com/softdevteam/grmtools/pull/549/commits/f671bd45aa03462cee260b05591bc26f9921624d) since the first commit references `MarkMap`, but `merge_from` was already documented there.

ratmice:
> I will admit that I'm struggling to pick up the high-level gist of where this is going. I don't think I've fully understood `ValueQuery`, for example: perhaps you can give me a layman's explanation at a high level?

I think there are 2 major things this is trying to do and both of those have a kind of common theme.

Problems:
* We have values across both crates and within crates via visibility restrictions that are isolated from one another.
   Examples: (cfgrammar, 'YaccKind'), (lrpar, 'RecoveryKind'), and in lrlex there is LexFlags from LRNonStreamingLexer and LexKind from CTLexeBuilder.
* We have multiple forcing/default behaviors across these, with `YaccResolver` having it's own for `YaccKind` and `LexFlags` having it's own.   But none for `RecoveryKind`, or `LexerKind`.

The experiment herein kind of takes the approach of using `Header` which is limited to core types, to pass these values around in a data type which contains only `core::` types, to their respective crates where they can be converted into their respective concrete types like `RecoveryKind`, `YaccKind`, etc.

It does this by passing the `core::` type restricted `Header` structure *to* places like `YaccParser` where it can then produce the `YaccKind`.  Instead of resolving these beforehand in `CTParserBuilder` and `nimbleparse` separately then passing `YaccKind` in after the fact.  So that command line options etc work as though they were built from a synthesized `%grmtools` section.

The forcing/defaulting mechanism then gets built by a notion of `merge` on this synthesized header and the one that was actually parsed from the `%grmtools` section.  So now that allows us to just automatically get forcing/defaulting for all the all the values that can be put in a header.

That in itself introduces some problems, primarily dealing with non-parsed `spanless` locations, so part of this is solving problems introduced by the mechanisms herein.

ratmice:
> Dumb question: will users tend to think of these things as query bits? Is there a higher level query API that might make more sense to them?

Hopefully I've covered the query bits part already by getting rid of them, but there is still the nagging question of "Will users tend to think of these things?"

So far, I've primarily been running under the assumption that users *rarely*, if ever need to see this part of the API,
it is mostly `pub` due to the cross-crate nature of the data involved, users probably rarely use it because they mostly put stuff *into* the header, but rarely take anything *out* of the header.

I guess I can think of use cases similar to metadata that people might want to put stuff in the header to consume.
There currently isn't a way to leverage this from `ctbuilder`, which actively complain if a key goes unused.
At some point if we want to allow this, we could add function similar to `lrpar_config` that takes a callback/closure,

```
.header_config(|header| {
    // do something with the header.
})
```

This *could* be useful, but likely needs some extensions, like a `Value::String` variant
```
%grmtools {
   valid_test_inputs: "../tests/*.input",
   invalid_test_inputs: "../tests/invalid/*.input",
}
```

then people could write something like
```
cpbuilder.header_config(|header| {
     for input in glob(header.query("valid_test_inputs")) {
         header.mark_key_used("valid_test_inputs");
     }
})
```

But currently lacking that kind of an API the header value is mostly constructed inside the `build()` method,
And even if the user calls `GrmtoolsSectionParser` on the source file directly, the builder will complain *unless*
the field is one which the builder specifically knows about and uses like `yacckind`.

So, there is *potential* we could expose it more to users, but it really isn't going to be of much use outside of programs like nimbleparse, nimbleparse_lsp, and the `CT*Builder`

cfgrammar/src/lib/header.rs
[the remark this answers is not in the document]
ratmice: I changed this to `TryFrom` in [553ce4d](https://github.com/softdevteam/grmtools/pull/549/commits/553ce4de73cb084f067a2bc6e8d7a2a7ead8f61b)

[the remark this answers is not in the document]
ratmice: Curiously trying to change the other direction to use `TryFrom` (`TryFrom<Value> for YaccKind` like my initial comment), runs into the thing that we try to avoid using the query API (We can't use the query API because we already have the value), so we must duplicate all the error handling for each variant of `Value`.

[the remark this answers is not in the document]
ratmice: switched to `TryFrom<Value> for YaccKind` in [57285ca](https://github.com/softdevteam/grmtools/pull/549/commits/57285ca04c70dbf428d66e2c9f8c3a45b3704305)

Seems like an improvement to me even if each `TryFrom` implementation is a bit longer than with query,
I think the simplicty of getting rid of the `Header` wrapper and making it an alias alone is worth it.

ltratt:
Right, I get the high-level motivation now: sorry I was a bit dense! So to some extent we can mark this as an unstable, internal API that just so happen to be `pub`. I can live with that: we can e.g. mark all the relevant parts as `#[doc(hidden)]` to make this clear to anyone perusing the code.

> I guess I can think of use cases similar to metadata that people might want to put stuff in the header to consume.
> There currently isn't a way to leverage this from ctbuilder, which actively complain if a key goes unused.
At some point if we want to allow this, we could add function similar to lrpar_config that takes a callback/closure,

This does sound intriguing, in the sense that it would reduce the quantity of "internal but pub" stuff I think?

ratmice:
> Right, I get the high-level motivation now: sorry I was a bit dense! So to some extent we can mark this as an unstable, internal API that just so happen to be `pub`. I can live with that: we can e.g. mark all the relevant parts as `#[doc(hidden)]` to make this clear to anyone perusing the code.
> 

I really should have spent time making a table of contents to give the overall big picture too, sorry about that.
I think the only difficulty to making this `#[doc(hidden)]` is that `Header` is going to show up as a parameter to documented functions.

I wonder if we can `#[doc(hidden)]` every function but `fn new()`?
I'll give that a try in the morning.

> > I guess I can think of use cases similar to metadata that people might want to put stuff in the header to consume.
> > There currently isn't a way to leverage this from ctbuilder, which actively complain if a key goes unused.
> > At some point if we want to allow this, we could add function similar to lrpar_config that takes a callback/closure,
> 
> This does sound intriguing, in the sense that it would reduce the quantity of "internal but pub" stuff I think?

I suppose it would, in the sense of perhaps giving people a reason to use it, and building library integrations for things like testing?

I personally am intrigued by it for `nimbleparse_lsp` (which I have been neglecting :cry:) for the purposes of getting rid of the `.toml` file, it only contains a few necessary pieces of info (optional location of tests as I mentioned above) and the file extension to associate with the parser.

In that sense it is intriguing to me because it could allow me to get rid of the entire nimbleparse.toml file.

ratmice:
There wasn't a super simple way to document *just* new, since there is no attribute to negate a hidden structure,
thus I had to remove the top-level docs on `struct Header` for now, and scatter `#[doc(hidden)]` about.
It would be nice if we could do something like the following, but afaict there isn't any way currently.

```
#[doc(hidden)] 
struct Header {...}

impl Header {
  /// ...
  #[doc(hidden = false)]
  #[doc(visible)]
  pub fn new()
```

ratmice:
I think in another PR, I'd probably want to try and do it along side an example that uses it,
and this PR seems plenty large enough already.  I could develop it on-top of this PR but in
another branch though, in case we stumble upon issues on the way?

There is also migration of `LexFlags`, `LexerKind`, and `RecoveryKind` to this API to do either on top of this PR or as separate ones, I had been planning on doing those as separate PRs too but maybe it makes sense to do that work on top of this one, both to make certain we're happy with it, and that it comes out of those without any major changes?

ltratt:
I think we can get this one in and iterate further from there. I think this is then ready to squash?

ltratt:
Thanks!

PR #551, opened 2025-04-12
ctbuilder: Parse lexerkind from grmtools section

ratmice:
This adds `lexerkind` parsing to `CTLexerBuilder`.

This patch contains an interim change which modifies the manual grmtools section parser in `LexParser` so that it recognizes the `lexerkind` field, and doesn't produce an error.

In a follow up, the plan is to remove that parser.  We are most likely going to have to parse it twice, just so that the second parse can find the position into the source text where the grmtools section ends (Or alternately strip the header off of the source text passed to the `LexParser`, but I think the former because we don't want users having to do that).  But replacing the manual parser + the custom forcing mechanism, and passing the header through seemed like a change that was going to dwarf adding the `lexerkind` field to the header.

Let me know however if you would prefer those changes to be done all at once instead?


ratmice:
FWIW, I have been having some trouble with the patch after this one (applyting to `LexParser` the same style of patch as #549, not so much any new or insurmountable problems, more just discontent with it's execution.  But I don't think it is anything that affects this patch.

ltratt:
This looks fine to me and I think we can merge it as-is?

ratmice:
I think so, will have to revisit the parsing code but it's good to get the feature and it's tests in while that is being redone.

PR #553, opened 2025-04-14
Migrate `LexFlags` parsing to GrmtoolsSectionParser

ratmice:
So here is what I've managed to come up with for migrating the `LexFlags` parsing code in the `%grmtools` section to use the `GrmtoolsSectionParser`, with this patch we parse the `%grmtools` section twice.

1. The first in `CTParserBuilder`, to pull out `LexerKind`,
2. Then again in one of the `LRNonStreamingLexerDef` constructors to get and merge the `LexFlags`.
3. As a consequence by the time `LexParser` has run `Header` has been lowered to a `LexFlags`, 
    and the whole `%grmtools` section has been stripped off, so we don't need to parse it a third time to find the beginning 
    of the lex file's contents.
 
We do all the merging in step 2, partially because `LexFlags` lacks a notion of `MergeBehavior` and that is the point where we still have enough information to merge without introducing multiple parameters like `forcing` and `default`.
This also kindly isolates the `LexBuildHeaderError` (for lack of any good name) out of `LexParser`.
This allows us to keep around the `new_with_lex_flags` constructor but simplified a little bit with fewer arguments.

One very important thing to note is that this changes the return type of the trait method `from_str`.
A second thing to note is that I noticed that `from_rules` doesn't have any method to change the `LexFlags`, it is still relying on the defaults.
A third thing to note is that this may not produce the nicest of errors, but I think we should be able to improve them without breaking semver now.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -16,8 +16,12 @@ use std::{
 
 use bincode::Encode;
 use cfgrammar::{
-    header::{GrmtoolsSectionParser, HeaderError, HeaderErrorKind, Namespaced, Setting, Value},
+    header::{
+        GrmtoolsSectionParser, Header, HeaderError, HeaderErrorKind, Namespaced, Setting, Value,
+    },
+    markmap::MergeBehavior,
     newlinecache::NewlineCache,
+    span::Location,
     Spanned,
 };
 use lazy_static::lazy_static;
@@ -27,7 +31,7 @@ use proc_macro2::TokenStream;
 use quote::{format_ident, quote, ToTokens, TokenStreamExt};
 use regex::Regex;
 
-use crate::{DefaultLexerTypes, LRNonStreamingLexerDef, LexFlags, LexerDef, UNSPECIFIED_LEX_FLAGS};
+use crate::{DefaultLexerTypes, LRNonStreamingLexerDef, LexBuildHeaderError, LexFlags, LexerDef};
 
 const RUST_FILE_EXT: &str = "rs";
 
@@ -203,8 +207,7 @@ where
     rule_ids_map: Option<HashMap<String, LexerTypesT::StorageT>>,
     allow_missing_terms_in_lexer: bool,
     allow_missing_tokens_in_parser: bool,
-    force_lex_flags: LexFlags,
-    default_lex_flags: LexFlags,
+    header: Header,
     #[cfg(test)]
     inspect_lexerkind_cb: Option<Box<dyn Fn(LexerKind) -> Result<(), Box<dyn Error>>>>,
 }
@@ -250,8 +253,7 @@ where
             rule_ids_map: None,
             allow_missing_terms_in_lexer: false,
             allow_missing_tokens_in_parser: true,
-            force_lex_flags: UNSPECIFIED_LEX_FLAGS,
-            default_lex_flags: UNSPECIFIED_LEX_FLAGS,
+            header: Header::new(),
             #[cfg(test)]
             inspect_lexerkind_cb: None,
         }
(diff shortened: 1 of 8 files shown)


lrlex/src/lib/ctbuilder.rs
```diff
@@ -463,29 +465,34 @@ where
...
-                                format!("{} at line {line} column {column}", e)
-                            } else {
-                                format!("{}", e)
-                            }
-                        })
-                        .collect::<Vec<_>>()
-                        .join("\n")
-                })?;
+                let lexerdef =
+                    LRNonStreamingLexerDef::<LexerTypesT>::new_with_header(&lex_src, header)
```
ratmice, line 473: I didn't get to the trailhead before the defect in this patch came to me,
We aren't using `mark_used` on the `header`, but I'm not even sure we can, passing it by value.
`merge_from` isn't preserving `mark_used`, it's kind of assuming that all of that will happen *after* merging.

It isn't exactly clear to me whether we can pass this to `LRNonStreamingLexerDef` not as an owned value, but as a `&mut` ref.

ltratt: I haven't totally understood the details, but on the face of it (2) does sound like the better route to me. Is this PR the right place to do it or ...?

lrlex/src/lib/ctbuilder.rs
```diff
@@ -463,29 +465,34 @@ where
...
+                                                e.spans().first().unwrap().start(),
+                                            )
+                                        {
+                                            format!("{} at line {line} column {column}", e)
+                                        } else {
+                                            format!("{}", e)
+                                        }
+                                    }
+                                    LexBuildHeaderError::Header(e) => {
+                                        format!("{}", e)
```
ltratt, line 490: Dumb question: can we not give line number information for the header?

lrlex/src/lib/ctbuilder.rs
ratmice: I don't think this PR is the right place for it, it actually ends up undoing a lot of what is in this PR.

I do think (1) can go in this PR, it is basically the same approach taken here, I think maybe a good idea is to try the first approach, finishing it off in this PR if only to give us a fallback option that we can hold off on merging.  Just in case something goes awry with (2) which I also think seems preferable but also has some complexity trade-offs.

ltratt: Works for me.

lrlex/src/lib/ctbuilder.rs
ratmice: I think even that would be awkward, e.g. `YaccGrammarError` assumes spans exist (the least `SpansKind` assumes there is one span), we could add a `SpansKind::None`, and `YaccGrammarErrorKind::Header(HeaderErrorKind)`, It feels to me somewhat like throwing the baby out with the bath water though, and e.g. we still have to make changes of similar scope to adding `Location`.  In that sense `Location::CommandLine` is somewhat type equivalent to not having a span.

lrlex/src/lib/ctbuilder.rs
ratmice: the first approach is implemented in this PR in [7468581](https://github.com/softdevteam/grmtools/pull/553/commits/74685813d647cc992ce488a339c84ad4a8227966) this didn't require trait modifications at all, and was actually a relatively minor change to the patch.

I'll try and start work on a new branch implementing second approach tomorrow.

ltratt:
Might we be ready for squashing with that latest commit?

ratmice:
Sure, but I still want to hold off/try that other approach, because it would seem like if we can keep `Location` out of these interior errors I can only imagine it'll be much cleaner.

ratmice:
Think we can close this one, and will start over migrating `LexFlags` on top of #554

PR #554, opened 2025-04-17
First take at removing `HeaderError` from `YaccParser` internals

ratmice:
This is a first take at removing the `Location`/Header error from the internals of `Parser`,
I haven't gone over the diff extremely carefully.

My original intent was to make `Header` generic on either `Span` or `Location`, and then use `Header<Span>` inside, and `Header<Location>` in `CTParserBuilder` and `nimbleparse`.

What I noticed though, that it was possible to do it without that change, by returning a `Result<..., HeaderError>` from what is essentially `YaccParser::new` rather than from `parse` where we currently return `YaccGrammarErrors`.

We absolutely could still make `Header` generic, the parsed header results all return `Location::Span`.
I thought it was worth noting this before I went ahead and did it, since the generic `Span` type permeates through both the errors, and the `Value` type.


cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -44,6 +46,33 @@ impl ASTWithValidityInfo {
         }
     }
 
+    /// Parses the `%grmtools section` expecting it to contain a `yacckind` entry.
+    pub fn from_src(src: &str) -> Result<Self, Vec<HeaderError>> {
```
ratmice, line 50: I named this `from_src` rather than `from_str` like LrLex is because for some reason clippy complains about this one, but not about LrLex, I assume because that one is a trait method?

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -362,117 +360,78 @@ fn add_duplicate_occurrence(
...
+            yacc_kind,
             src,
             num_newlines: 0,
             ast: GrammarAST::new(),
             global_actiontype: None,
         }
     }
 
+    #[cfg(test)]
+    pub(crate) fn from_src(src: &str) -> Result<YaccParser, Vec<HeaderError>> {
```
ratmice, line 374: This one we mostly needed for the testsuite, because we had lost any internal methods that lack a `YaccKind`.
Perhaps these tests should be moved elsewhere.

cfgrammar/src/lib/yacc/ast.rs
ltratt: What's the Clippy error? It might be that we want to turn that one off (or maybe not!).

cfgrammar/src/lib/yacc/parser.rs
ltratt: It sounds like it's worth trying. If that move simplifies the code a bit, we probably have a winner!

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -362,117 +360,78 @@ fn add_duplicate_occurrence(
...
+            Err(vec![HeaderError {
+                kind: HeaderErrorKind::InvalidEntry("yacckind"),
+                locations: vec![Location::Span(Span::new(pos, pos))],
+            }])
+        }
+    }
+
     pub(crate) fn parse(&mut self) -> YaccGrammarResult<usize> {
         let mut errs = Vec::new();
+        let (_, pos) = GrmtoolsSectionParser::new(self.src, false).parse().unwrap();
```
ltratt, line 401: I wasn't expecting `unwrap` here -- should it be `?`?

nimbleparse/src/main.rs
```diff
@@ -176,7 +176,35 @@ fn main() {
...
+                        .collect::<Vec<_>>()
+                        .join("\n")
+                );
+                std::process::exit(1);
+            }
+        }
+    }
+    let yk_val = header.get("yacckind");
+    if yk_val.is_none() {
+        eprintln!("Cannot determine yacckind, add it to the `%grmtools` section in the yacc file, or pass the `-y` argument");
```
ltratt, line 202: I might phrase this one as "yacckind not specified in the `%grmtools` section of the grammar or via the '-y' parameter", just to try and fit a typical terminal width a bit better.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Ahh, good catch.  That was supposed to be a temporary fix until I migrated to `Header<Span>` so we could actually return ` YaccGrammarError` at this point.  I.e. this is unwrap because the error still has a `Location`, even though it is always a `Location::Span(...)`.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Another option of course is implementing the `FromStr` trait.
https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait

Here is  playground link in case you want to see it
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e9f0dd766fa27e37b9249db7150fc29e

ltratt: Good point, `FromStr` makes sense.

nimbleparse/src/main.rs
```diff
@@ -176,7 +176,35 @@ fn main() {
...
+        let parsed_header = GrmtoolsSectionParser::new(&yacc_src, true).parse();
+        match parsed_header {
+            Ok((parsed_header, _)) => {
+                header
+                    .merge_from(parsed_header)
+                    .expect("Specified merge behavior cannot fail");
+            }
+            Err(errs) => {
+                eprintln!(
+                    "Errors parsing `%grmtools` section:\n {}\n",
```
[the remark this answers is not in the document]
ratmice: Should be fixed in [c2e8838](https://github.com/softdevteam/grmtools/pull/554/commits/c2e8838bf37d5ef6b5e9fe55702b8a648906577f)

cfgrammar/src/lib/yacc/ast.rs
ratmice: This is partially done in [34906f9](https://github.com/softdevteam/grmtools/pull/554/commits/34906f9956caa755a76c183e741af5138ed90b37)

There is another `from_src` function that needs to move into `YaccGrammar<_>`, but is blocked on
adding generics to `HeaderError`.

cfgrammar/src/lib/yacc/parser.rs
ratmice: So fixing this was a little more involved than I like (Which is why I was so happy to have avoided doing so, until you found I had forgotten this unwrap!)

But should be fixed in [bb6225f](https://github.com/softdevteam/grmtools/pull/554/commits/bb6225f5e6dc0af2e200230f4bf4000e76864f44)

ltratt: Wow, yeah, that's a big change!

ltratt:
I think this one is ready now?

cfgrammar/src/lib/yacc/grammar.rs
```diff
@@ -181,6 +187,18 @@ impl YaccGrammar<u32> {
...
 }
 
+impl<StorageT: 'static + PrimInt + Unsigned> FromStr for YaccGrammar<StorageT>
+where
+    usize: AsPrimitive<StorageT>,
+{
+    type Err = Vec<YaccGrammarError>;
+    fn from_str(s: &str) -> YaccGrammarResult<Self> {
+        let ast_validation = ast::ASTWithValidityInfo::from_str(s)
+            .map_err(|errs| errs.iter().cloned().map(|e| e.into()).collect::<Vec<_>>())?;
```
ratmice, line 197: On my walk it occurred to me that it'd be better to use `drain` here over `cloned`, for some reason I thought there was another instance of this, which would allow us to remove `T: Clone` bounds, but I'm not seeing it right this moment.

ratmice: Should be fixed in [edf677f](https://github.com/softdevteam/grmtools/pull/554/commits/edf677fcba9a09dc60d4ef64257b7e370b5b49a8)

Must have been imagining the other `cloned()` call, because

ratmice:
I think it is ready too, but I'd like to read through the patch more closely which I won't have time to do until tomorrow most likely.

cfgrammar/src/lib/header.rs
```diff
@@ -14,21 +16,35 @@ use std::{error::Error, fmt};
...
+            kind: YaccGrammarErrorKind::Header(e.kind, e.spanskind()),
+            spans: e.locations,
+        }
+    }
+}
+
+// This is essentially a tuple that needs a newtype so we can implement `From` for it.
+// Thus we aren't worried about it being `pub`.
+#[derive(Debug, PartialEq)]
+pub struct HeaderValue<T>(pub T, pub Value<T>);
```
ratmice, line 43: Likely needs `#[doc(hidden)]`, and I should just go through the docs again, to ensure there isn't anything else I added that also needs it.

cfgrammar/src/lib/yacc/parser.rs
ratmice: Woops, I actually forgot the specific `unwrap()` here, since that unwrap was hidding the type conversion error.

Should really be fixed in [3df2592](https://github.com/softdevteam/grmtools/pull/554/commits/3df259282deb755e4e66fe785eb2ff9e6eefb1c8)

ratmice:
I did manage to get my self-review done before heading out which was mostly just documentation stuff,
but also hiding `HeaderError` from the public facing API, that allowed basically hiding the entire module from top-level.

I suppose I'm happy with this now, the public facing API seems good at least.

The callers/header building and what not seems like it could probably use some refinement,
but that all seems well within the scope of stuff that can be done as we come across things that can be improved (primarily error messages).

cfgrammar/src/lib/header.rs
ratmice: I believe all the doc changes should be fixed across various commits,
this one in [7895675](https://github.com/softdevteam/grmtools/pull/554/commits/7895675b1309b354179855f1c1979d70c52d7518)

ltratt:
Please squash.

ratmice:
Is that alright? (It leaves the `unwrap()` in the first patch to avoid squashing those two fairly large patches together)

ratmice:
I went ahead and switched moving the tests and added the `FromStr` implementation to the first patch
(modified to temporarily use HeaderError and unwrap) 

That removes the `#[cfg(test)]` function entirely from the series, but that means adding a second call to unwrap in the first patch which then gets fixed in the second patch.

ltratt:
> Is that alright? (It leaves the unwrap() in the first patch to avoid squashing those two fairly large patches together)

Yep!

PR #556, opened 2025-04-20
Migrate `LexFlags` parsing to `GrmtoolsSectionParser` and `Header`.

ratmice:
This rebuilds my previous PR #553 on top of the current master branch.
I didn't manage to avoid *all* problems associated with Span/Location conversion.

In particular, because we need to synthesize a `LexFlags` for default values, and merge those in with the `%grmtools` section in the `.l` file, we have to synthesize a made up span for said default values.

It isn't to difficult to see though that when the `LexFlags::try_from` in `from_str` encounters a default value with a synthesized span, the hard coded values in the default shouldn't cause an error, and therefore the span should go unused. After which the header parameter to try_from gets dropped. So as much as I dislike just making up spans, in this case it doesn't really seem like it can materialize into a problem where the user could see the synthesized span.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -17,10 +17,12 @@ use std::{
 use bincode::Encode;
 use cfgrammar::{
     header::{
-        GrmtoolsSectionParser, HeaderError, HeaderErrorKind, HeaderValue, Namespaced, Setting,
-        Value,
+        GrmtoolsSectionParser, Header, HeaderError, HeaderErrorKind, HeaderValue, Namespaced,
+        Setting, Value,
     },
+    markmap::MergeBehavior,
     newlinecache::NewlineCache,
+    span::Location,
     Spanned,
 };
 use lazy_static::lazy_static;
@@ -30,7 +32,7 @@ use proc_macro2::TokenStream;
 use quote::{format_ident, quote, ToTokens, TokenStreamExt};
 use regex::Regex;
 
-use crate::{DefaultLexerTypes, LRNonStreamingLexerDef, LexFlags, LexerDef, UNSPECIFIED_LEX_FLAGS};
+use crate::{DefaultLexerTypes, LRNonStreamingLexerDef, LexFlags, LexerDef};
 
 const RUST_FILE_EXT: &str = "rs";
 
@@ -206,8 +208,7 @@ where
     rule_ids_map: Option<HashMap<String, LexerTypesT::StorageT>>,
     allow_missing_terms_in_lexer: bool,
     allow_missing_tokens_in_parser: bool,
-    force_lex_flags: LexFlags,
-    default_lex_flags: LexFlags,
+    header: Header<Location>,
     #[cfg(test)]
     inspect_lexerkind_cb: Option<Box<dyn Fn(LexerKind) -> Result<(), Box<dyn Error>>>>,
 }
@@ -242,6 +243,8 @@ where
     ///     .build()?;
     /// ```
     pub fn new_with_lexemet() -> Self {
+        let mut header = Header::new();
+        header.set_default_merge_behavior(MergeBehavior::Ours);
         CTLexerBuilder {
             lrpar_config: None,
             lexer_path: None,
@@ -253,8 +256,7 @@ where
             rule_ids_map: None,
             allow_missing_terms_in_lexer: false,
             allow_missing_tokens_in_parser: true,
-            force_lex_flags: UNSPECIFIED_LEX_FLAGS,
-            default_lex_flags: UNSPECIFIED_LEX_FLAGS,
+            header,
             #[cfg(test)]
             inspect_lexerkind_cb: None,
         }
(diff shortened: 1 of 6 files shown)


lrlex/src/lib/ctbuilder.rs
```diff
@@ -794,7 +815,15 @@ where
     ///
     /// Setting this flag will override the same flag within a `%grmtools` section.
     pub fn dot_matches_new_line(mut self, flag: bool) -> Self {
-        self.force_lex_flags.dot_matches_new_line = Some(flag);
+        let key = "dot_matches_new_line".to_string();
+        self.header.set_merge_behavior(&key, MergeBehavior::Ours);
```
ratmice, line 819: We could add a function here `set_default_merge_behavior`, and avoid adding this line in all of the setters here.

lrlex/src/lib/lexer.rs
```diff
@@ -252,7 +365,7 @@ where
         Self: Sized;
 
     /// Instantiate a lexer from a string (e.g. representing a `.l` file).
-    fn from_str(s: &str) -> LexBuildResult<Self>
+    fn from_str(s: &str) -> Result<Self, Vec<LexBuildError>>
```
ratmice, line 368: Unnecessary change.

ratmice: Should be fixed in [223d257](https://github.com/softdevteam/grmtools/pull/556/commits/223d257af1349f8a3f49fb7ebcde2a3b0462c5a1)

lrlex/src/lib/lexer.rs
```diff
@@ -40,47 +45,155 @@ pub struct LexFlags {
...
+        cvt_num!(dfa_size_limit);
+        cvt_num!(nest_limit);
+
+        header
+    }
+}
+
+pub(crate) fn default_lex_flags_header() -> Header<Span> {
+    // Because this is called in `from_str`, which needs a span to convert to `LexBuildError`
+    // there doesn't seem to be any way to avoid making up this span out of thin air.
```
ratmice, line 176: So, One thing we could do here is just remove this function and implement the defaults manually on `LexFlags` so we don't have to synthesize `Span` like this.

It unfortunately gets called from a few places, so perhaps `pub(crate) fn setup_default_lex_flags(&mut LexFlags)`.

ratmice: The best I could think of is in [82069cc](https://github.com/softdevteam/grmtools/pull/556/commits/82069ccbbd9787c5d96209024dbfeb47bcae00a1)

We couldn't do this so late as when constructing rules, because they affect parsing of the `.l` file.
Also I didn't document these shouldn't be marked required, since the constructors already document their
default values it seemed redundant.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -446,8 +444,6 @@ where
         let lex_src = read_to_string(lexerp)
             .map_err(|e| format!("When reading '{}': {e}", lexerp.display()))?;
         let mut header = self.header;
-        let defaults = default_lex_flags_header();
-        defaults.merge_into(&mut header)?;
```
ratmice: This was the only call to `merge_into` so we could remove that function if we want to?

lrlex/src/lib/ctbuilder.rs
ratmice: I implemented this in [b399a72](https://github.com/softdevteam/grmtools/pull/556/commits/b399a721e8b78619a9fff81f7fae34172a38aed3) but happy to back it out if you would prefer the minimalist API or would implementing it as a separate PR.

ltratt: Works for me!

lrlex/src/lib/ctbuilder.rs
ltratt: Might we need it again later or is this likely to be the only user of that function ever?

ratmice: I cannot see us needing to call it later, in the sense that I think this series has all the `merge_from` calls that the repo currently needs.  In order to need to call it we would need to add another tool, otherwise it would be that a user needs to call it if we decide to make this `pub` + unhidden.

I needed it for approximately two attempts of various patch series, there are two reasons `merge_into` is usable in more cases than `merge_from`:

1. It is more flexible when merging the `DEFAULT_LEX_FLAGS` since it requires no knowledge about the destination header to know that the `from` header will be overridden by the destinations values via `Theirs`.
2. It allows merging the defaults into another structure *without* moving that struct since it takes `other` as an owned value.

Since we now know the contents of the source header, the first is not necessary,
and since we're not passing headers by mutable reference, the second is no longer necessary.
Thus we currently fall into the more narrow band where `merge_from` can be used to the same effect.

It is a fairly simple derivative of copy/paste `merge_from` and reverse the src/destination values so it isn't the end of the world to just add it at that time, if we find we need it either.

ltratt: OK, so should we remove it or keep it?

ltratt:
Please squash.

ratmice:
Squashed, I tried to pull out the few free-standing commits I could find.

ratmice:
Okay, well I think unless anything comes up that should conclude that patch series.
It turned out much more involved than I had expected, I feel like I made
some painful choices along the way and paid the price, but that has largely been rectified.

I don't really have a whole lot in the way of breaking changes left on my todo list,
with most of the stuff I'm looking towards implementing I think should be non-breaking.

PR #558, opened 2025-04-24
Differentiate between quoted tokens and %token directives fixes #557

### LINKED ISSUE #557: Having production name equal to one of the tokens results in an invalid state graph
With these dependencies in a new binary project:
```toml
[dependencies]
cfgrammar = { git = "https://github.com/softdevteam/grmtools", rev = "3997be1" }
lrtable = { git = "https://github.com/softdevteam/grmtools", rev = "3997be1" }
```
(latest main)

And this `main.rs`:
```rust
use cfgrammar::yacc::{YaccGrammar, YaccKind, YaccOriginalActionKind};
use lrtable::Minimiser;

const YACC: &str = r###"

%start start

%%

start
  : "a" a
  | "b" b
  ;
  
a
  : "c"
  ;
  
b
  : "d"
  ;

"###;

fn main() {
    let grammar = YaccGrammar::new(
        YACC,
    )
    .expect("failed to load parser definition");
    let (automaton, _) = lrtable::from_yacc::<u32>(&grammar, Minimiser::Pager)
        .expect("failed to create parsing tables");

    println!("{}", automaton.pp(&grammar, false));
}
```

I get:
```
0: [start -> . 'b' 'b', {'$'}]
   [^ -> . start, {'$'}]
   [start -> . 'a' 'a', {'$'}]
   'b' -> 1
   start -> 2
   'a' -> 3
1: [start -> 'b' . 'b', {'$'}]
   'b' -> 4
2: [^ -> start ., {'$'}]
3: [start -> 'a' . 'a', {'$'}]
   'a' -> 5
4: [start -> 'b' 'b' ., {'$'}]
5: [start -> 'a' 'a' ., {'$'}]
```

There are no `'c'` or `'d'` anywhere and production `b` is treated like token `"b"`.

I believe this is a bug since when I change the grammar to:
```
%start start

%%

start
  : "a" x
  | "b" y
  ;
  
x
  : "c"
  ;
  
y
  : "d"
  ;
```
I get:
```
0: [start -> . 'b' y, {'$'}]
   [^ -> . start, {'$'}]
   [start -> . 'a' x, {'$'}]
   'b' -> 1
   start -> 2
   'a' -> 3
1: [start -> 'b' . y, {'$'}]
   [y -> . 'd', {'$'}]
   y -> 4
   'd' -> 5
2: [^ -> start ., {'$'}]
3: [start -> 'a' . x, {'$'}]
   [x -> . 'c', {'$'}]
   x -> 6
   'c' -> 7
4: [start -> 'b' y ., {'$'}]
5: [y -> 'd' ., {'$'}]
6: [start -> 'a' x ., {'$'}]
7: [x -> 'c' ., {'$'}]
```
which is correct.

ratmice:
Here is an attempt to fix issue #557, below is the new output for the first (failing) example given by @umut-sahin with this patch, which seems right to me.

```
0: [start -> . 'b' b, {'$'}]
   [^ -> . start, {'$'}]
   [start -> . 'a' a, {'$'}]
   'b' -> 1
   start -> 2
   'a' -> 3
1: [start -> 'b' . b, {'$'}]
   [b -> . 'd', {'$'}]
   b -> 4
   'd' -> 5
2: [^ -> start ., {'$'}]
3: [start -> 'a' . a, {'$'}]
   [a -> . 'c', {'$'}]
   a -> 6
   'c' -> 7
4: [start -> 'b' b ., {'$'}]
5: [b -> 'd' ., {'$'}]
6: [start -> 'a' a ., {'$'}]
7: [a -> 'c' ., {'$'}]
```

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -111,9 +111,14 @@ impl FromStr for ASTWithValidityInfo {
 #[derive(Debug)]
 pub struct GrammarAST {
     pub start: Option<(String, Span)>,
-    // map from a rule name to indexes into prods
+    // map from a rule name to indexes into `prods`
     pub rules: IndexMap<String, Rule>,
     pub prods: Vec<Production>,
+    // A set of indexes into `tokens` for all tokens in `%token` directives.
+    // e.g. given a `%token a` and a token "b" not specified in any `%token` directive
+    // we have `self.tokens.get_index_of("a") ∈ self.token_directives`. However for
+    // token "b" we have `self.tokens.get_index_of("b") ∉ self.token_directives`.
+    pub token_directives: HashSet<usize>,
     pub tokens: IndexSet<String>,
     pub spans: Vec<Span>,
     pub precs: HashMap<String, (Precedence, Span)>,
@@ -195,6 +200,7 @@ impl GrammarAST {
             prods: Vec::new(),
             spans: Vec::new(),
             tokens: IndexSet::new(),
+            token_directives: HashSet::new(),
             precs: HashMap::new(),
             avoid_insert: None,
             implicit_tokens: None,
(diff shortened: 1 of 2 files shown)


cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -114,6 +114,7 @@ pub struct GrammarAST {
     // map from a rule name to indexes into prods
     pub rules: IndexMap<String, Rule>,
     pub prods: Vec<Production>,
+    pub token_directives: IndexSet<String>,
```
ratmice, line 117: I guess an alternate implementation here makes this a `Vec<Tidx<StorageT>>`, so it would take an index into the `tokens` field instead of cloning the string?

ratmice: I went ahead and did that in [6b0ea90](https://github.com/softdevteam/grmtools/pull/558/commits/6b0ea909c7a79a7d191c34e99fb8f315dc41fb4b)

The if statement isn't quite as nice, but I felt since the field was added `pub` like the rest, it was a better choice
than cloning the string. Perhaps it will look nicer with a `let chain` once those are stabilized?

ltratt:
Thanks for doing this so quickly! If we can tweak that docstring a bit, I think we're good to go.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -111,9 +111,11 @@ impl FromStr for ASTWithValidityInfo {
 #[derive(Debug)]
 pub struct GrammarAST {
     pub start: Option<(String, Span)>,
-    // map from a rule name to indexes into prods
+    // map from a rule name to indexes into `prods`
     pub rules: IndexMap<String, Rule>,
     pub prods: Vec<Production>,
+    // set of indexex into `tokens` for all tokens in `%token` directives.
```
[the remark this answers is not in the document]
ratmice: does [2ae0715](https://github.com/softdevteam/grmtools/pull/558/commits/2ae07158a7cab92b783521f3092c50daa3bb7e89) help at all? (also typo fixed)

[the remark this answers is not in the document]
ratmice: 2nd try in [f9eb018](https://github.com/softdevteam/grmtools/pull/558/commits/f9eb0186c3dfb7bd5509a39a4fee959e9f5e36be)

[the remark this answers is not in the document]
ratmice: I also added a test case in [cda4da5](https://github.com/softdevteam/grmtools/pull/558/commits/cda4da53ed0cf66c14a45bc190ca361c65378e1d) which mirrors the `e.g.` text, and tests a corner case.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -380,8 +380,8 @@ impl YaccParser<'_> {
                 i = self.parse_ws(j, false)?;
                 while i < self.src.len() && self.lookahead_is("%", i).is_none() {
                     let (j, n, span, _) = self.parse_token(i)?;
-                    self.ast.token_directives.insert(n.clone());
-                    if self.ast.tokens.insert(n) {
+                    if let (idx, true) = self.ast.tokens.insert_full(n) {
+                        self.ast.token_directives.insert(idx);
```
ratmice, line 384: I think this should actually go outside the if statement, we don't actually appear to insert "a" into tokens yet in the case of something like the following where "a" appears in `%left "a"` before `%token a`, however this code is kind of written in a way that the token could already be in `self.tokens`.  So I think it is better to put this call outside the `if`.

```
%left "a"
%token a
```

ratmice: should be fixed in [af08b7b](https://github.com/softdevteam/grmtools/pull/558/commits/af08b7bbba3ac3ffce9531d3f27020a81c022f63)

ltratt:
Please squash.

ratmice:
Squashed.

PR #560, opened 2025-05-03
Add grmtools book documentation for new header fields

ratmice:
It occurred to me I should probably do a pass over the docs, this documents the new lex and yacc header fields that were added: `test_files` `recoverykind` and `lexerkind`.


ltratt:
Good idea!

PR #561, opened 2025-05-07
Try to improve the error when we encounter an unquoted glob

ratmice:
This tries to improve the error messages in the test cases,
```yacc
"%grmtools {test_files: *.test,}",
"%grmtools {test_files: foo*.test,}",
```

Before this change these headers would produce:

```console
Error(s) parsing `%grmtools` section:
 Illegal name

Error(s) parsing `%grmtools` section:
 Expected token: '}
```

Afterwords they both produce the following:
```console
Error(s) parsing `%grmtools` section:
 Unxpected token: '*', perhaps this is a glob, in which case it requires string quoting.
```

This also fixes the quoting and provides more useful span information for that `ExpectedToken: '}` error.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -57,6 +57,7 @@ pub enum HeaderErrorKind {
     MissingGrmtoolsSection,
     IllegalName,
     ExpectedToken(char),
+    UnexpectedToken(char, &'static str),
     DuplicateEntry,
     InvalidEntry(&'static str),
     ConversionError(&'static str, &'static str),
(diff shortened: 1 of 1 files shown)


ratmice:
Fwiw, I wasn't quite sure how to approach this error, and we still apparently aren't writing source context in `HeaderError` in nimbleparse which could perhaps would help a lot in displaying context.

That seems like perhaps a good thing to follow up with?

ltratt:
> That seems like perhaps a good thing to follow up with?

Agreed! But this PR is already an improvement so I'm happy with it!

PR #564, opened 2025-05-11
Add line/column info to nimbleparses printing of `HeaderError`.

ratmice:
In the process replace some usage of `Option` with `OnceCell`.

It also restores some spanned error printing that was lost in commit 277ae03622cb956455d6e98ee5e7efa4f71b3855 when `Location` was in `YaccGrammarError`.

This is a follow up to PR #561 in that PR the errors when run through nimbleparse looked like:

```console
Error(s) parsing `%grmtools` section:
 Unxpected token: '*', perhaps this is a glob, in which case it requires string quoting.
```

After this patch they include source context information,

```
[Error] parsing the `%grmtools` section: in foo.y
    3|     test_files: *.test,
                        Unxpected token: '*', perhaps this is a glob, in which case it requires string quoting.
[Error] parsing the `%grmtools` section: in bar.y
    3|     test_files: input*.test,
                            ^ Unxpected token: '*', perhaps this is a glob, in which case it requires string quoting.
```

### DIFF
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -141,12 +141,12 @@ fn main() {
     };
     let entry = match header.entry("yacckind".to_string()) {
         Entry::Occupied(_) => unreachable!("Header should be empty"),
-        Entry::Vacant(v) => v.occupied_entry(),
+        Entry::Vacant(v) => v,
     };
     match matches.opt_str("y") {
         None => {}
         Some(s) => {
-            entry.insert(HeaderValue(
+            entry.insert_entry(HeaderValue(
                 Location::CommandLine,
                 Value::try_from(match &*s.to_lowercase() {
                     "eco" => YaccKind::Eco,
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -1,7 +1,7 @@
 mod diagnostics;
 use crate::diagnostics::*;
 use cfgrammar::{
-    header::{GrmtoolsSectionParser, Header, HeaderValue, Value},
+    header::{GrmtoolsSectionParser, Header, HeaderError, HeaderValue, Value},
     markmap::Entry,
     yacc::{ast::ASTWithValidityInfo, YaccGrammar, YaccKind, YaccOriginalActionKind},
     Location, Span,
@@ -165,13 +165,13 @@ fn main() {
 
     let lex_l_path = PathBuf::from(&matches.free[0]);
     let lex_src = read_file(&lex_l_path);
+    let lex_diag = SpannedDiagnosticFormatter::new(&lex_src, &lex_l_path);
     let mut lexerdef = match LRNonStreamingLexerDef::<DefaultLexerTypes<u32>>::from_str(&lex_src) {
         Ok(ast) => ast,
         Err(errs) => {
-            let formatter = SpannedDiagnosticFormatter::new(&lex_src, &lex_l_path).unwrap();
-            eprintln!("{ERROR}{}", formatter.file_location_msg("", None));
+            eprintln!("{ERROR}{}", lex_diag.file_location_msg("", None));
             for e in errs {
-                eprintln!("{}", indent(&formatter.format_error(e).to_string(), "    "));
+                eprintln!("{}", indent(&lex_diag.format_error(e).to_string(), "    "));
             }
             process::exit(1);
         }
@@ -179,6 +179,7 @@ fn main() {
 
     let yacc_y_path = PathBuf::from(&matches.free[1]);
     let yacc_src = read_file(&yacc_y_path);
+    let yacc_diag = SpannedDiagnosticFormatter::new(&yacc_src, &yacc_y_path);
     let yk_val = header.get("yacckind");
     if yk_val.is_none() {
         let parsed_header = GrmtoolsSectionParser::new(&yacc_src, true).parse();
@@ -190,12 +191,12 @@ fn main() {
             }
             Err(errs) => {
                 eprintln!(
-                    "Error(s) parsing `%grmtools` section:\n {}\n",
-                    errs.iter()
-                        .map(|e| e.to_string())
-                        .collect::<Vec<_>>()
-                        .join("\n")
+                    "{ERROR}{}",
+                    yacc_diag.file_location_msg(" parsing the `%grmtools` section:", None)
                 );
+                for e in errs {
+                    eprintln!("{}", indent(&yacc_diag.format_error(e).to_string(), "    "));
+                }
                 std::process::exit(1);
             }
         }
@@ -211,7 +212,25 @@ fn main() {
     let recoverykind = if let Some(HeaderValue(_, rk_val)) = header.get("recoverer") {
         match RecoveryKind::try_from(rk_val) {
             Err(e) => {
-                eprintln!("{e}");
+                eprintln!(
+                    "{ERROR}{}",
+                    yacc_diag.file_location_msg(" parsing the `%grmtools` section:", None)
+                );
+                let spanned_e: HeaderError<Span> = HeaderError {
+                    kind: e.kind,
+                    locations: e
+                        .locations
+                        .iter()
+                        .map(|l| match l {
+                            Location::Span(span) => *span,
+                            _ => unreachable!("All reachable errors should contain spans"),
+                        })
+                        .collect::<Vec<_>>(),
+                };
+                eprintln!(
+                    "{}",
+                    indent(&yacc_diag.format_error(spanned_e).to_string(), "    ")
+                );
                 process::exit(1)
             }
             Ok(rk) => rk,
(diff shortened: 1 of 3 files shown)


ratmice:
One thing I notice in the output is that the first instance doesn't have a `^` caret and I'm not certain why, but it will be a few days before I can investigate it.

I am also wondering whether you think it is worth trying larger follow up where we try moving this diagnostics formatting code to a `support` library, and use them from within `CTParserBuilder`, `CTLexerBuilder` and `nimbleparse`, I believe all my other failed attempts have been to just make API so people can use whatever error printing library they want within their `build.rs`.

This would be a different approach where since it doesn't try to add any API where people can customize error printing, but just shares the nimbleparse error printing code between the aformentioned things?

nimbleparse/src/main.rs
```diff
@@ -211,7 +220,33 @@ fn main() {
...
         match RecoveryKind::try_from(rk_val) {
             Err(e) => {
-                eprintln!("{e}");
+                let formatter =
+                    yacc_diag.get_or_init(|| spanned_formatter(&yacc_src, &yacc_y_path));
+                eprintln!(
+                    "{ERROR}{}",
+                    formatter.file_location_msg(" parsing the `%grmtools` section:", None)
+                );
+                // I believe that this can only ever be true, but I haven't tried to prove it nor do the types ensure it.
```
ltratt, line 229: If we're fairly confident of this, I would make this an `unwrap` (or similar). If we are wrong, we'll find out, and can then readd the complexity back in.

nimbleparse/src/main.rs
```diff
@@ -86,7 +87,13 @@ where
     (missing_from_lexer, missing_from_parser)
 }
 
+fn spanned_formatter<'a>(input: &'a str, path: &'a Path) -> SpannedDiagnosticFormatter<'a> {
+    SpannedDiagnosticFormatter::new(input, path).unwrap()
+}
+
 fn main() {
+    let yacc_diag: OnceCell<SpannedDiagnosticFormatter> = OnceCell::new();
```
ltratt, line 95: I'm not totally sure I've understood why this needs to be a `OnceCell`. I may have misunderstood the control flow, but I *think* the `SpannedDiagnosticFormatter` is initialised on two separate codepaths: couldn't we just use a normal `let yacc_diag = SpannedDiagnosticFormater::...` on each of those paths? Feel free to tell me I've missed out on something!

ltratt:
> I am also wondering whether you think it is worth trying larger follow up where we try moving this diagnostics formatting code to a support library (or just making it a hidden module in cfgrammar)

My gut reaction is to first see what we can break into a separate module *within* nimbleparse; and, depending on how that goes, maybe we can then put that in another crate. Warning: this might not be a good idea on my part!

nimbleparse/src/main.rs
```diff
@@ -325,7 +353,8 @@ fn main() {
                 eprintln!();
             }
             if let Some(token_spans) = missing_from_parser {
-                let formatter = SpannedDiagnosticFormatter::new(&lex_src, &lex_l_path).unwrap();
+                let formatter =
+                    yacc_diag.get_or_init(|| spanned_formatter(&yacc_src, &yacc_y_path));
```
ratmice, line 357: This one is using `yacc_src` where it should be using `lex_src`, definitely needs to change.

nimbleparse/src/main.rs
ratmice: The main reason to build it lazily and once is that it involves building a `NewLineCache` and a pass over the entire source building the data structure containing the offsets of each line.  So the current code (and the usage of once cell in this patch) is trying to do two things: 

1. Avoid building a `NewlineCache` entirely unless an error actually occurs.
2. Avoid rebuilding a `NewlineCache` when one was already built for the same source text.

I believe the only case where second thing happens *across branches* is when a *warning* happens and then, 
in a later branch a conflict occurs and gets printed.  We could also perhaps relax the error handling to also print conflict errors.

There is also the `missing_from_lexer` and `missing_from_parser`, we could non-fatal errors in header parsing,
like we might want to complain about `%grmtools{ some_random_unknown_header_value }`, (I think we do in both `CTParserBuilder` and `CTLexerBuilder` but I may have forgotten that in nimbleparse.

ratmice:
> My gut reaction is to first see what we can break into a separate module _within_ nimbleparse; and, depending on how that goes, maybe we can then put that in another crate. Warning: this might not be a good idea on my part!

The `diagnostics.rs` stuff is basically already is it's own module, but the problem/need to move it outside nimbleparse is more so that it can be used within `CTParserBuilder` and `pub` so that it can be used within `CTLexerBuilder` that was the motiviation behind proposing that it be `#[doc(hidden)]`.

It seemed like something I haven't tried yet at least

nimbleparse/src/main.rs
ratmice: We can't exactly use `unwrap`, but we can just pull this out of the `if` and remove the `else`, I guess that is a worthwhile cleanup

```
                  let spanned_e: HeaderError<Span> = HeaderError {
                      kind: e.kind,
                      locations: e
                          .locations
                          .iter()
                          .map(|l| match l {
                              Location::Span(span) => *span,
                              _ => unreachable!(),
                          })
                          .collect::<Vec<_>>(),
                  };
                  eprintln!(
                      "{}",
                      indent(&formatter.format_error(spanned_e).to_string(), "    ")
                  );
```

The reason I believe that this case cannot occur, because: 

1. Values passed during argument parsing are pre-checked for validity and would have already produced errors.
2. Other values lacking spans that derive from a e.g `From<RecoveryKind> for Value<Location>` impl, cannot produce errors during the corresponding inverse `TryFrom<Value<Location>> for RecoveryKind`
3. All other values have corresponding spans.
4. The merge process being used isn't mutually exclusive so can't introduce errors between otherwise valid values.

Making thats `unreachable!("args have already been checked, values derived from From impls are infallible, and all other values should have spans");`

The first 3 of those are straight-forward it's cases like that last one though where this line of thinking becomes somewhat tenuous, and give me pause mainly due to questioning the fragility and whether that will always be the case.

ltratt: I tend to agree that these are probably `unreachable`.

nimbleparse/src/main.rs
ltratt: Right, got it.

ratmice: One thing we could try is moving to `OnceCell<NewlineCache>` inside `SpannedDiagnosticFormatter` then initializing it in an accessor method `self.nlc()`, that also allows us to remove the `unwrap` from the lazy initialization code I guess.

nimbleparse/src/main.rs
ratmice: It'll be a few days before I'm able to make/test this change, will try the above change out then.

nimbleparse/src/main.rs
ltratt: If that makes it clearer, go for it, but if it makes it worse, let's keep what we've got.

ratmice: I did that in [fb27a50](https://github.com/softdevteam/grmtools/pull/564/commits/fb27a5008b306c02f4f3859e364d4229857f020d), and I felt it improved things, curious what you think?

nimbleparse/src/main.rs
ratmice: I ran across an unrelated issue while testing this, fixed in [8c626cc](https://github.com/softdevteam/grmtools/pull/564/commits/8c626ccb0d74ab504106bfe57655468dfff90fab)
But otherwise did this change in [11648c6](https://github.com/softdevteam/grmtools/pull/564/commits/11648c67cadb29e7b14147be30fcc55663accc4a)

nimbleparse/src/main.rs
ltratt: Works for me!

ltratt:
Please squash.

ratmice:
Squashed, but I left a long commit message for [be35407](https://github.com/softdevteam/grmtools/pull/564/commits/be354073054128046e8bea7d80a2420f0282aa67) describing that the API is somewhat prone to misuse and perhaps some future changes that could be made to avoid this kind of error in the future.

I'm not sure it is imperative that we fix that immediately, considering it's hidden API.

nimbleparse/src/main.rs
```diff
@@ -141,12 +141,12 @@ fn main() {
     };
     let entry = match header.entry("yacckind".to_string()) {
         Entry::Occupied(_) => unreachable!("Header should be empty"),
-        Entry::Vacant(v) => v.occupied_entry(),
+        Entry::Vacant(v) => v,
```
ratmice, line 144: I think rather than what I discussed in the commit message, of adding a different `key_entry`, it's best to just implement the `mark_required` and `set_merge_behavior` functions for `VacantEntry` directly, perhaps `mark_used` too but I can't think of a reason to mark a vacant entry used.

I still think it's best to do as a separate patch, but let me know if you feel otherwise.

ltratt: Doing it in another PR is fine!

PR #567, opened 2025-05-19
Lrpar diagnostics

ratmice:
This patch moves the `diagnostics.rs` from nimbleparse, to `lrpar`, so it can be used in `CTParserBuilder`.
It modifies the `YaccGrammarError` and `CTConflictsError` and some portions of the `HeaderError` error printing to inline the source snippets in the error messages.

There is a couple of things to note:
* I haven't done a great job of manually going through and looking at the output of each error yet.
* I wasn't terribly fond of `CTConflictsError` changes here, I'll explain the hurdles I ran into below.

`CTConflictsError{stable}` is used via `downcast_ref` in the testsuite, and the `SpannedDiagnosticsFormatter` only borrows the string, owned by `build()` which causes lifetime problems with trying to put the `SDF` into the `CTConflictsError` with the source string and the formatter, thus it gets pre-formatted into a `String` and embed that in the `CTConflictsError`.

Further, since the source string contains temp filenames when run during the testsuite, it is difficult to do some form of string comparison with what we expect the error to output.

Thus the somewhat weird implementation where we also include the otherwise unused `stable`, and jump through clippy hoops to keep the existing tests working, without a string comparison replacement.  Otherwise I probably would likely have just used `ErrorString`, and removed `CTConflictsError`.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -14,14 +14,16 @@ use std::{
     sync::Mutex,
 };
 
-use crate::{LexerTypes, RTParserBuilder, RecoveryKind};
+use crate::{
+    diagnostics::{DiagnosticFormatter, SpannedDiagnosticFormatter},
+    LexerTypes, RTParserBuilder, RecoveryKind,
+};
 use bincode::{decode_from_slice, encode_to_vec, Decode, Encode};
 use cfgrammar::{
     header::{GrmtoolsSectionParser, Header, HeaderValue, Value},
     markmap::{Entry, MergeBehavior},
-    newlinecache::NewlineCache,
     yacc::{ast::ASTWithValidityInfo, YaccGrammar, YaccKind, YaccOriginalActionKind},
-    Location, RIdx, Spanned, Symbol,
+    Location, RIdx, Symbol,
 };
 use filetime::FileTime;
 use lazy_static::lazy_static;
@@ -38,6 +40,9 @@ const ACTIONS_KIND_PREFIX: &str = "Ak";
 const ACTIONS_KIND_HIDDEN: &str = "__GtActionsKindHidden";
 
 const RUST_FILE_EXT: &str = "rs";
+
+const WARNING: &str = "[Warning]";
+const ERROR: &str = "[Error]";
 
 lazy_static! {
     static ref RE_DOL_NUM: Regex = Regex::new(r"\$([0-9]+)").unwrap();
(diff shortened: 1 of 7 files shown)


ratmice:
I was thinking about it and, while I anticipate not having any issues with `CTLexerBuilder`, it still isn't exactly trivial,
so probably worth knowing what we're going to run into end up with there before making any decisions on this.

ratmice:
I did the easy part of `CTLexerBuilder`, what is missing is the `missing from lexer` and `missing from parser`, 
corresponding to the following nimbleparse formatting code https://github.com/softdevteam/grmtools/blob/master/nimbleparse/src/main.rs#L312-L351

This doesn't appear to be something that we can even achieve via `CTParserBuilder` callbacks, because this happens after `CTParserBuilder::build()` which takes `self` by value.

I'm uncertain, but wondering if it will be enough to add the yacc sources to `CTParser` in the `build()` return value.
I think that code in nimbleparse also needs the `&YaccGrammar` which is only present with conflicts unfortunately.

My hope is that perhaps that can be moved to it's own field outside of conflicts, and returned unconditionally.
I *think* those two changes will allow parity of errors between nimbleparse and the ctbuilder implementations.
If that is the case it seems like the only semver breakage is regarding [`CTParser::conflicts()`](https://docs.rs/lrpar/latest/lrpar/struct.CTParser.html#method.conflicts) which would need changing but is documented as an unstable interface that may change at any time.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -1547,6 +1528,10 @@ where
     }
 }
 
+fn indent(s: &str, indent: &str) -> String {
```
ltratt, line 1531: Let's document this one. I suspect we should also reorder the parameters (I misread the calls above!).

lrpar/src/lib/diagnostics.rs
```diff
@@ -19,6 +19,17 @@ pub struct SpannedDiagnosticFormatter<'a> {
     nlc: OnceCell<NewlineCache>,
 }
 
+trait PushLine {
+    fn push_line<S: AsRef<str>>(&mut self, s: S);
+}
+
+impl PushLine for String {
+    fn push_line<S: AsRef<str>>(&mut self, s: S) {
```
ltratt, line 27: How about we call this `pushln` to mirror `println`? Also, a quick docstring on the trait to make that clear would be great.

ltratt:
A couple of small comments but I think we can easily get this in: I appreciate that there are more rough edges we could smooth off, but this already feels like an improvement.

lrpar/src/lib/ctbuilder.rs
ratmice: Is it okay if I reorder the parameters in a separate PR?

Since this function is copy/paste from nimbleparse I'd want to update that as well, and I fear updating all the callers over seems like it might make these patches a bit more messy.

ratmice:
Well, I guess looking at the output now, it looks like cargo adds it's own text at the beginning and that is
causing misalignment of the underlined portions of text.  We probably need to start them with an initial newline,
also I missed some newlines in places it switched from `eprintln` to `push_str`.

I think it is probably best to at least fix these misalignments and errors in transition here now.
But leave further refinements to things like indentation (as cargo is adding it's own too) for later tweaking.

ltratt:
>I think it is probably best to at least fix these misalignments and errors in transition here now.
>But leave further refinements to things like indentation (as cargo is adding it's own too) for later tweaking.

Works for me!

ratmice:
Alright, I believe I fixed all the changes you requested, and the formatting issues I've found.

ltratt:
Please squash.

ratmice:
Squashed, I wasn't really sure If I wanted to squash [eb2f213](https://github.com/softdevteam/grmtools/pull/567/commits/eb2f213f83a487d052b1e1987ebbfea672d0d1fd) and [92f97e9](https://github.com/softdevteam/grmtools/pull/567/commits/92f97e928ff046055f8aacccf04cc2c66e59a000) together, didn't have a strong opinion so I just left them separate for now?

PR #574, opened 2025-05-22
treat allow_missing_tokens_in_parser as a warning

ratmice:
I don't think we have any way to test this, so I manually tested by modifying one of the examples.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -39,6 +39,7 @@ use crate::{DefaultLexerTypes, LRNonStreamingLexer, LRNonStreamingLexerDef, LexF
 const RUST_FILE_EXT: &str = "rs";
 
 const ERROR: &str = "[Error]";
+const WARNING: &str = "[Warning]";
 
 lazy_static! {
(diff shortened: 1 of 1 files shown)


ratmice:
Oh I should probably say more:

This one follows from the conversation at the end of #571

It flips the `allow_missing_tokens_from_parser` default value to `false` but changes the meaning to produce a warning.

The values here are implemented slightly differently than the defaults for `CTParserBuilder`.
In `CTLexerBuilder` the `warnings_are_errors` defaults to `false` while it is defaulting to `true` in `CTParserBuilder`.

As such it will now print a warning, but not produce an error.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -672,7 +672,7 @@ where
                     }
                 }
 
-                has_unallowed_missing = self.warnings_are_errors;
+                has_unallowed_missing |= self.warnings_are_errors;
```
ratmice, line 675: In the case where `has_unallowed_missing = true` and `self.warnings_are_errors = false` this would,
unset the `has_unallowed_missing` value back to false.

This was caught by lrpar/cttests/src/ctfails/missing.test

ltratt:
Please squash.

ratmice:
Squashed.

PR #577, opened 2025-05-23
Export an array of all tokens from `ct_token_map`

taminomara:
This helps with writing structured input adapters for fuzzing. When fuzzing a parser specifically (as opposed to fuzzing lexer and parser at the same time), we'd like to supply it with an array of valid lexemes. This export helps us build such an array as we don't have to manually list all tokens in a fuzzing entry point.

Note that I didn't implement this functionality for generated lexers because there's already a way to get all tokens via `mod_l::lexerdef().iter_rules()`.

Example of a fuzzing implementation after this PR:

```rust
#[derive(Debug)]
struct Token(u32, String);

impl<'a> Arbitrary<'a> for Token {
    fn arbitrary(u: &mut Unstructured<'a>) -> libfuzzer_sys::arbitrary::Result<Self> {
        Ok(Token(*u.choose(token_map::TOKENS)?, u.arbitrary()?))
    }
}

fuzz_target!(|data: Vec<Token>| {
    let mut text = String::new();
    let lexemes = data.into_iter().map(|tok| {
        let lexeme = DefaultLexeme::new(
            tok.0,
            text.len(),
            tok.1.len(),
        );
        text.push_str(&tok.1);
        lexeme
    }).collect();

    // Run parser...
}
```

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -1181,14 +1181,17 @@ impl CTLexer {
 
 /// Create a Rust module named `mod_name` that can be imported with
 /// [`lrlex_mod!(mod_name)`](crate::lrlex_mod). The module contains one `const` `StorageT` per
-/// token in `token_map`, with the token prefixed by `T_`. For example with `StorageT` `u8`,
-/// `mod_name` `x`, and `token_map` `HashMap{"ID": 0, "INT": 1}` the generated module will look
-/// roughly as follows:
+/// token in `token_map`, with the token prefixed by `T_`. In addition, it will
+/// contain an array of all token IDs `TOK_IDS`.
+///
+/// For example with `StorageT` `u8`, `mod_name` `x`, and `token_map`
+/// `HashMap{"ID": 0, "INT": 1}` the generated module will look roughly as follows:
 ///
 /// ```rust,ignore
 /// mod x {
 ///   pub const T_ID: u8 = 0;
 ///   pub const T_INT: u8 = 1;
+///   pub const TOK_IDS: &[u8] = &[T_ID, T_INT];
 /// }
 /// ```
 ///
@@ -1201,6 +1204,7 @@ impl CTLexer {
 /// mod x {
 ///   pub const T_PLUS: u8 = 0;
 ///   pub const T_ID: u8 = 1;
+///   pub const TOK_IDS: &[u8] = &[T_PLUS, T_ID];
 /// }
 /// ```
 pub fn ct_token_map<StorageT: Display + ToTokens>(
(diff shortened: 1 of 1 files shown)


ltratt:
This is a part of the system I haven't thought about for a while. Is it possible to do the same thing with `mod_l::lexerdef().iter_rules().map(|x| x.name()).collect()` or similar? [Warning: untried!]

lrlex/src/lib/ctbuilder.rs
```diff
@@ -1215,31 +1215,47 @@ pub fn ct_token_map<StorageT: Display + ToTokens>(
...
+        .map(|(k, _)| {
+            let name = match rename_map {
+                Some(rmap) => *rmap.get(k.as_str()).unwrap_or(&k.as_str()),
+                _ => &k,
+            };
+            let tok_ident = format_ident!("T_{}", name.to_ascii_uppercase());
+            quote! {
+                #tok_ident,
+            }
+        })
```
ratmice, line 1249: Since these are such similar cases, it might make sense to return a tuple

```
(quote!{pub const #tok_ident: #storaget = #id;}, quote!{#tok_ident})
```

then use [unzip](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.unzip) to make separate arrays

ltratt:
> If there's a custom lexer with ct_token_map, then there's no way to get a full array of tokens.

I take your point.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -1215,31 +1215,47 @@ pub fn ct_token_map<StorageT: Display + ToTokens>(
...
             };
             let tok_ident = format_ident!("T_{}", name.to_ascii_uppercase());
-            let storaget = str::parse::<TokenStream>(type_name::<StorageT>()).unwrap();
             // Code gen for the constant token values.
             quote! {
                 pub const #tok_ident: #storaget = #id;
             }
         })
         .collect::<Vec<_>>();
+    // Export an array of tokens to facilitate easier fuzzing.s
```
ltratt, line 1237: I think that fuzzing is just one possible use case. So `// Export an array of tokens.` is adequate.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -1215,31 +1215,47 @@ pub fn ct_token_map<StorageT: Display + ToTokens>(
...
+        })
+        .collect::<TokenStream>();
     // Since the formatter doesn't preserve comments and we don't want to lose build time,
     // just format the module contents.
     let unformatted = quote! {
         mod #mod_ident {
             #![allow(dead_code)]
             #(#tokens)*
+
+            pub const TOKENS: &[#storaget] = &[#token_array];
```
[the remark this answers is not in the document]
taminomara: > Dumb question: do we need to specify the length of the array?

No, because we export a reference to an array, not the array itself.

ratmice:
Took a bit of head scratching until I grokked it (building a token stream directly rather than an intermediate vector!),
But once it clicked, it all seemed fine to me.

Seems fine to me now, unless Laurence has any further comments.

PR #506, opened 2025-03-12
Add lookahead to reduce/reduce conflicts

quernd:
This adds more diagnostic information to reduce/reduce conflicts based on the observation that different lookahead symbols don't necessarily cause the same conflicts.

For instance, the LR automaton for this grammar:

```
%start S
%%
S: A X | B X | A Y;
A: '0' ; B: '0' ; C: '0' ;
X: '1' ; Y: '2' ;
```

has the following item set for state 1:

```
1:  [B -> '0' ., {'1'}]
    [A -> '0' ., {'1', '2'}]
```

which causes a reduce/reduce conflict if the lookahead is '1', but not if the lookahead '2'.

PR #578, opened 2025-05-23
Print source snippet for lookahead value of rr-conflicts

ratmice:
This adds printing using spans for the lookahead information that was added in #506
using the example from that PR

```
%start S
%%
S: A X | B X | A Y;
A: '0' ; B: '0' ; C: '0' ;
X: '1' ; Y: '2' ;
```

Adds the last line to this conflict printing output:
```console
Reduce/Reduce conflict, can reduce 'A' or 'B' with lookahead '1' at rr.y:5:1
5| A: '0' ; B: '0' ; C: '0' ;
   ^ First reduce
5| A: '0' ; B: '0' ; C: '0' ;
            - Second reduce
6| X: '1' ; Y: '2' ;
       + Lookahead
```

Not certain we need to be using different underline characters for each thing here, but it was currently doing that so I kept the tradition?

### DIFF
--- a/lrpar/src/lib/diagnostics.rs
+++ b/lrpar/src/lib/diagnostics.rs
@@ -267,7 +267,14 @@ impl<'a> SpannedDiagnosticFormatter<'a> {
             let (_r1_prod_names, _r1_prod_spans) = pidx_prods_data(ast, *r1_prod_idx);
             let (_r2_prod_names, _r2_prod_spans) = pidx_prods_data(ast, *r2_prod_idx);
 
-            let t_name = grm.token_name(*t_idx).unwrap_or("$");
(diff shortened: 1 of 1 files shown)


ltratt:
> Not certain we need to be using different underline characters for each thing here, but it was currently doing that so I kept the tradition?

Hmm, *probably* it would be better to use `^` for all of them (which seems more consistent with other tools)? But we can do that in a follow-up PR if you agree.

ratmice:
Yeah, I think my original attempts probably had more of a footnote style attempting to reduce duplicate line output.
By having multiple underlines on a single line, and then footnote text on their own line.

But as we can see from the duplicate line output that isn't actually being done, so I agree that keeping to `^` is probably better.

ratmice:
Force pushed a clippy/rustfmt update

PR #580, opened 2025-05-23
Update lrlex to parse grmtools section, diagnostics.

ratmice:
Trying to think about loose ends and what not, I remembered I hadn't updated lrlex for using the %grmtools section, nor updated diagnostics.

So this adds all of that, I'm not certain the grmtools section parsing code is really necessary.
It keeps things in sync with ctbuilder and prepares for the addition of new `lexerkinds`, without these changes the `LexFlags` would still be applied just through a different code path, but it wouldn't do things like error out on an unrecognized `LexerKind` field.

### DIFF
--- a/lrlex/src/main.rs
+++ b/lrlex/src/main.rs
@@ -1,16 +1,38 @@
 use getopts::Options;
 use std::{
     env,
+    error::Error,
+    fmt,
     fs::File,
     io::{stderr, stdin, Read, Write},
     path::Path,
     process,
-    str::FromStr,
 };
 
-use cfgrammar::{newlinecache::NewlineCache, Spanned};
-use lrlex::{DefaultLexerTypes, LRNonStreamingLexerDef, LexerDef};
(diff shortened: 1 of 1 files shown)


lrlex/src/main.rs
```diff
@@ -42,43 +47,71 @@ fn read_file(path: &str) -> String {
...
+                eprintln!(
+                    "{}",
+                    &indent("     ", &lex_diag.format_error(e).to_string())
+                );
             }
             process::exit(1);
-        });
+        }
+    };
+    header.mark_used(&"lexerkind".to_string());
```
ratmice, line 84: Marking as used here, but never doing the check for unused values.

ratmice: Should be fixed in [51b4b97](https://github.com/softdevteam/grmtools/pull/580/commits/51b4b9727dbc8c60a81f5a4cd38977d37046349a)

ratmice:
Squashed/slightly reworded the commit message.

PR #583, opened 2025-05-25
Update nimbleparse help text regarding default values

ratmice:
I had missed/failed to notice that in nimbleparse there is some help text regarding default values.
This missed getting updated when the default values were changed with the `%grmtools` section changes.
I'm not actually certain how to get this text to print from nimbleparse, as things like`-y list`, etc didn't seem to work.

But I also didn't dig at all into the getopt crate, and the changes seemed straightforward enough.

### DIFF
--- a/nimbleparse/src/main.rs
+++ b/nimbleparse/src/main.rs
@@ -108,13 +108,13 @@ fn main() {
         .optopt(
             "r",
             "recoverer",
-            "Recoverer to be used (default: cpctplus)",
+            "Recoverer to be used (default: cpctplus unless specified in grammar)",
             "cpctplus|none",
         )
         .optopt(
             "y",
             "yaccvariant",
-            "Yacc variant to be parsed (default: original)",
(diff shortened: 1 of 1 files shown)


nimbleparse/src/main.rs
```diff
@@ -108,13 +108,13 @@ fn main() {
         .optopt(
             "r",
             "recoverer",
-            "Recoverer to be used (default: cpctplus)",
+            "Recoverer to be used (default: cpctplus unless specified in grammar)",
```
ratmice, line 111: This text could perhaps be more leading towards the user finding how to specify it.
E.g. "unless specified in the %grmtools section of the grammar", but I feared making it too long
and I don't see any obvious way it gets printed, so I stuck with "unless specified in the grammar"
which is already way longer than the current text.

ltratt:
I don't *think* you can get these printed out, but having them be less misleading in the source is a worthwhile goal in and of itself.

Issue #484, opened 2025-02-03
Feature Request: All Rules are "Top-Level" Functions


rneswold:
This is a great project and I've used it in a handful of Rust applications. Thanks!

There is one feature from the OCaml tools (`ocamllex` and `ocamlyacc`) have that I found convenient. In the code that `ocamlyacc` generates, all the grammar rules create functions you can call (if you need to parse a subset of the grammar.)

For instance, I'm working on a project that uses a string to describe data acquisition. There's a "name" field, optional range specification, option field names, and a field for the event on which to sample. All of that is straightforward to implement. However, we also have an API which would like just the event portion and another API that just wants the device name.

We could have a struct with a bunch of `Option` fields, but I'd rather be able to have `Event` as a parameter to a function.

If this feature would be too disruptive, I wonder how others would solve this with the current tools.

ltratt:
Right now, grmtools doesn't generate functions in this way. I guess it's possible to do so and to call "into the middle of" the LR statetable. I haven't thought about that before and it might need a bit of thought about exactly what it means.

In the interim, I think there is a (horrible) hack one can do: you can duplicate the grammar (including in a `build.rs` file), change the `%start` line and output to a different file(s).

ratmice:
Comes to mind that with the horrible hack, is seems undoubtedly likely to produce the 'unused rule' and 'unused token' warnings/errors.  You'll likely need to set at least [`warnings_are_errors`](https://docs.rs/lrpar/latest/lrpar/struct.CTParserBuilder.html#method.warnings_are_errors) and more than likely want to set [`show_warnings`](https://docs.rs/lrpar/latest/lrpar/struct.CTParserBuilder.html#method.show_warnings) to false entirely.

ltratt:
@ratmice Definitely! It would be good to do something nicer here, though an interesting question is what "unused" means if you have multiple start rules.

rneswold:
> Comes to mind that with the horrible hack, is seems undoubtedly likely to produce the 'unused rule' and 'unused token' warnings/errors.

I would delete the unused rules in the "horrible hack". However, I was hoping to use the same lex file, so the "unused token" warnings would be a problem.

ratmice:
> [@ratmice](https://github.com/ratmice) Definitely! It would be good to do something nicer here, though an interesting question is what "unused" means if you have multiple start rules.

Interesting question, my inclination would be to define `unused` as unreachable from *any* start rule.

[...]

PR #585, opened 2025-05-26
Allow building parsers from `GrammarAST` with modified start rules.

ratmice:
I'm going to mark this as a draft, it isn't something I personally am likely to use,
While thinking through #584 and what uses `GrammarAST::new` has or could have, 
issue #484 came to mind, the current workaround described in that issue is to create multiple copies of yacc grammars
and modify the start rule.

As such cc'ing @rneswold just to see whether or not this approach is an improvement on that hack,
or if this is something which if we added it might just go unused?

The current hacks have some drawbacks like needing to disable warnings as errors because if you change the start rule, unused warnings can come up from the `top-level` start rule.

The idea of this patch is to provide an alternative hack mechanism, where you build a `GrammarAST`, clone it
and then modify the `start` field of each clone, in an unsafe manner.  Since you only do the actual checking for the first rule, you don't run into running the checks multiple times and subsequent warnings/errors caused by checking a partial grammar with a start rule somewhere in the middle.
 
It doesn't really *fix* #484 or share parsing tables or anything *nice*.
This may well be just a solution looking for a problem.

Also, because of the `lrlex` dependency I wasn't able to fully test this within cfgrammar (building parsers/lexers and passing input), I should probably be concerned though that the whole `&mut self -> &mut GrammarAST` seems like it might cause borrow checker issues once we try to actually finish the parser/lexer build process with the modified grammar?


ratmice:
Somewhat silly of me not to think of it, but there is no way to get `CTParserBuilder` to actually use this modified AST since it always works from paths.  So this is fairly unlikely to help.

rneswold:
> The idea of this patch is to provide an alternative hack mechanism, where you build a `GrammarAST`, clone it and then modify the start field of each clone, in an unsafe manner.

This sounds like it could solve my case: 6 production rules which 3 could be used as a top-level rule.

Making more than one top-level rule doesn't seem like an `unsafe` operation to me. But maybe it means all rules would be more expensive(?) Or it's the only way to do it with the current `GrammarAST`? I just hate having to add `unsafe` code to a project.

I know, I'm asking for a lot! 🤣 Thanks for thinking about this.

ltratt:
Dumb idea: could we make something like `fn clone_and_change_start_rule(rule: ...) -> Self` to satisfy this use case?

ratmice:
> Dumb idea: could we make something like `fn clone_and_change_start_rule(rule: ...) -> Self` to satisfy this use case?

While perhaps providing a safer abstraction (I think we'd have to return a `Result` in case `rule` isn't a rule in the grammar), this I think runs into much the same problem, of how are we going to call this function from `CTParserBuilder`.

With either mechanism we could perhaps use a function `CTParserBuilder::set_ast(self, ASTWithValidationInfo)`,
this function would then bypass the filesystem read/parse/validation phase, so it'd only use `grammar_path` if `self.ast` is `None`.

One of the reasons here we need to bypass validation is because we no longer have source which matches the AST, and it is also likely that spans are messed up so if we *did* do checking and what not and an error did occur it isn't clear how we'd do error printing (which now even relies on source availability within `CTParserBuilder`!

Both these mechanisms can bypass this problem it seems, (one by being `unsafe`, the other by potentially returning a `Result` at `clone_and_change_start_rule` time).

ratmice:
> Would that supersede this PR?

Yup

ltratt:
It sounds like it's worth giving it a go. Perhaps it will be an "unofficial" API of sorts, at least until we gain more confidence in it, but it is a feature that some people really want, so even unofficial/unstable is better than nothing (and double points if it's not `unsafe`).

ltratt:
Thanks -- I think this is a good intermediate step towards future perfection! Please squash.

cfgrammar/src/lib/mod.rs
```diff
@@ -3,7 +3,7 @@
 #![allow(clippy::new_without_default)]
 #![allow(clippy::unnecessary_wraps)]
 #![allow(clippy::upper_case_acronyms)]
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
```
ratmice, line 6: Change no longer needed.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -195,7 +209,7 @@ impl fmt::Display for Symbol {
 }
 
 impl GrammarAST {
-    pub fn new() -> GrammarAST {
+    pub(crate) fn new() -> GrammarAST {
```
ratmice, line 235: This change of visibility was only included in this patch accidentally, I feel like it is good to consider whether we should do this, but probably doesn't belong in this series.

So I'd rather drop it from this PR.

ratmice:
Squashed with the visibility modification, and `unsafe_code` changes omitted.
I squashed this into two commits, hoping that [5edbfc5](https://github.com/softdevteam/grmtools/commit/5edbfc530f41b4447744442b4ef6b9fdb884a7e4) will also suffice as an example of usage

ltratt:
I think we can mark this as undraft and merge.

ratmice:
Sorry about that, it's no longer in draft now.

PR #588, opened 2025-06-05
Replace `ct_token_map` with `CTTokenMapBuilder`

taminomara:
This commit adds `CTTokenMapBuilder` that replaces `ct_token_map`. It provides a few new features:

- generated module is no longer `#[allow(dead_code)]` unless manually enabled. This helps to spot errors in custom lexers where some of the tokens are never emitted;                                

- if generated module contains invalid identifiers, the generation will fail and suggest using `CTTokenMapBuilder::rename_map`;

- `CTTokenMapBuilder::rename_map` accepts any iterable that yields pairs of strings, eliminating the need to convert static arrays of pairs of strings before passing them to builder.

The `ct_token_map` function is deprecated since the next minor release (`0.14`).

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -1,19 +1,5 @@
 //! Build grammars at run-time.
 
-use std::{
-    any::type_name,
-    borrow::Borrow,
-    collections::{HashMap, HashSet},
-    env::{current_dir, var},
-    error::Error,
-    fmt::{self, Debug, Display, Write as _},
-    fs::{self, File, create_dir_all, read_to_string},
-    hash::Hash,
-    io::Write,
-    path::{Path, PathBuf},
-    sync::{LazyLock, Mutex},
-};
-
 use bincode::Encode;
 use cfgrammar::{
     header::{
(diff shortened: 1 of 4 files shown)


taminomara:
> I wonder then if it isn't worth doing something similar here,
instead of setting a boolean flag which specifies whether a specific lint should be enabled,
allow a String parameter to specify inner attributes within the generated code directly,
and then if you don't want the default lint set you could specify the empty string?

I don't have a strong opinion here, either approach is fine. I'll try to make a builder interface, though. This will be cleaner than passing booleans around 😄

ratmice:
> I don't have a strong opinion here

Just for the record, I don't either it was just a thought.

ltratt:
`ct_token_map` was always a mildly strange function to me (and I'm the person responsible for it!). I can imagine that other people might want other settings when generating the module, so I do think a simple builder is the right thing to go for here, rather than set ourselves up for a proliferation of similarly named functions. We can still keep (and mark as deprecated) `ct_token_map` so people have a simple upgrade path.

I would think that something as simple as `CTTokenMapBuilder::new(mod_name, token_map, rename_map).silence_dead_code(false).build()` or similar might do the trick, if everyone's amenable.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -1183,99 +1183,228 @@ impl CTLexer {
...
+    /// is `HashMap{"+": 0, "ID": 1}` and `rename_map` is `HashMap{"+": "PLUS"}`
+    /// then the generated module will look roughly as follows:
+    ///
+    /// ```rust,ignore
+    /// mod x {
+    ///   pub const T_PLUS: u8 = 0;
+    ///   pub const T_ID: u8 = 1;
+    /// }
+    /// ```
+    pub fn set_rename_map<M, I, K, V>(&mut self, rename_map: Option<M>)
```
ltratt, line 1253: Do we need `set_rename_map` and `rename_map`? It seems more idiomatic to me to have just the latter.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -1183,99 +1183,201 @@ impl CTLexer {
...
-/// For example with `StorageT` `u8`, `mod_name` `x`, and `token_map`
+/// For example, if `StorageT` is `u8`, `mod_name` is `x`, and `token_map` is
 /// `HashMap{"ID": 0, "INT": 1}` the generated module will look roughly as follows:
 ///
 /// ```rust,ignore
 /// mod x {
 ///   pub const T_ID: u8 = 0;
 ///   pub const T_INT: u8 = 1;
-///   pub const TOK_IDS: &[u8] = &[T_ID, T_INT];
+///   pub const TOK_IDS: &[u8] = &[0, 1];
```
ltratt, line 1201: Rereading this, I realise that the constants `0` and `1` are more accurate, but less helpful: I briefly had to remind myself what they refer to! I think it would be better to keep the variable names in here. *Technically* (I think...) they both lead to the same observable values because of `const`.

ltratt:
One minor comment, and then I think we're good to go.

ltratt:
Thanks!

Issue #589, opened 2025-06-05
Serde support for `ParseGenericTree`


ratmice:
Currently the [`parse_generictree`](https://docs.rs/lrpar/latest/lrpar/struct.RTParserBuilder.html#method.parse_generictree) function can be used to return a `Node` from there [`Node::pp`](https://docs.rs/lrpar/latest/lrpar/enum.Node.html#method.pp) can convert a parse tree, into a textual representation as used by `nimbleparse`.

One thing I'm slightly interested in is extending the output format to be more machine readable.
As such one thought is that we could include a function similar to `Node::pp` that also takes a `serde::Serializer`,
but instead of returning a `String` returns something from the serde data model.

Here are a couple questions:

1. This can likely be done outside of the repo, in the same way that we can copy/paste the `pp` function implementation and build it outside the crate as part of a binary to customize the output.
2. It isn't clear to me what exactly the best translation into the serde data model would be.
3. type signature/type of the return value?
4. Reusing `Node` or not?
5. We likely would need to choose some specific `Serializer` impl for e.g. nimbleparse/`test_files`
6. Then does it make sense to 

##### In repo or out?

One reason to include this in the repo is that it could then be used by the [`test_files`](https://softdevteam.github.io/grmtools/master/book/yaccextensions.html), which would make it possible to introduce a known file extension which contains both parser input, as well as a serialized form of parser output.

That seems to me like the most direct argument for it's inclusion, but usage in tools like `nimbleparse` would also likely be deciding on a concrete `Serializer` type too.  So perhaps it is worth experimenting with this outside the repo before making these kinds of decisions.

##### AST translation into serde data model

One way would be to just use [`serialize_map`](https://docs.rs/serde/latest/serde/trait.Serializer.html#associatedtype.SerializeMap) with `String` keys, recursively use `serialize_map` for the values. This seems possible, because `serialize_map` allows duplicate keys. However I somewhat fear that it seems likely that only a subset of serde serializers implementations will allow this duplicate keys.

So another option would be to use something more [`S-Expression`](https://en.wikipedia.org/wiki/S-expression) like, using [`serialize_seq`](https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_seq) combined with [`serialize_tuple`](https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_tuple)

[...]

PR #591, opened 2025-06-07
Initial prototype for using serde generic parse trees.

ratmice:
This is an initial prototype for #589, I think the code is pretty gross, however I also believe it might work,
just from minimal testing with 

```
cargo build --all-features
./target/debug/nimbleparse -t ./lrpar/examples/calc_parsetree/src/calc.{l,y} lrpar/examples/calc_parsetree/src/input.txt
```

For instance, if I compare the following output with and without `-t` passed to nimbleparse, it seems like I got the correct nesting (I think, unless my sleep deprived brain is mistaken).

```ron
{
    "src": "5 + 4 * 3",
    "ast": [
        Nonterm("Expr", [
            Nonterm("Expr", [
                Nonterm("Term", [
                    Nonterm("Factor", [
                        Term("INT", "5"),
                    ]),
                ]),
            ]),
            Term("+", "+"),
            Nonterm("Term", [
                Nonterm("Term", [
                    Nonterm("Factor", [
                        Term("INT", "4"),
                    ]),
                ]),
                Term("*", "*"),
                Nonterm("Factor", [
                    Term("INT", "3"),
                ]),
            ]),
        ]),
    ],
}
```

```console
Expr
 Expr
  Term
   Factor
    INT 5
 + +
 Term
  Term
   Factor
    INT 4
  * *
  Factor
   INT 3
```

It currently only supports `Serialize`, doesn't do anything when parsing multiple inputs (`parse_many`) or when inputs are specified via  `test_files` in the `%grmtools` section.

But it seemed like perhaps a good place to stop and evaluate.


lrpar/src/lib/parser.rs
```diff
@@ -43,72 +43,85 @@ pub enum Node<LexemeT: Lexeme<StorageT>, StorageT> {
...
+        }
 
-            fn insert_inner<'a>(top: &mut VecDeque<NamedNode<'a>>, node: NamedNode<'a>, limit: usize, acc:usize) {
-                match top.back_mut() {
-                    Some(NamedNode::Nonterm(_, entries)) => {
-                        if acc < limit {
-                            insert_inner(entries, node, limit, acc + 1)
-                        } else {
-                            top.push_back(node)
-                        }
```
ratmice: I think this whole `insert_inner` function I think is pretty terrible, and is still using the first algorithm
that came to mind.  In particular the recursion seems problematic for large a AST.

Perhaps there is a better way to implement it using something like a `BinaryHeap<NamedNode>`, and a `Vec<NamedNode>`, so that the `inner` elements are being stored/pushed to in the binary heap.
So that rather than starting from the root and recursively traversing the `back_mut()` to find where we `push_back`.  The element to push onto would be at some lower priority of the `BinaryHeap`.
Then the `Vec<NamedNode>` would be the rest of the nodes that in the order they get popped from that heap.

Anyhow I haven't totally thought it through, instead more focused on what the actual AST representation should be, but figured I should drop a note that this here probably needs a rethink.

PR #593, opened 2025-06-13
Initial experiments with serialized testing

ratmice:
This is mostly just for collecting feedback, as this patch is fairly incomplete,
It implements the serialization stuff, and it adds a `-s` switch to nimbleparse to *output* the serialized format,

It also adds some design docs which is a basic overview of what I *intend* to do, but this is necessarily incomplete,
and at times speculative based upon how I hope or imagine it might be made to work.

I don't think it is too far from actually deserializing tests and doing the pass/fail checking, but before going through the trouble of implementing that, it seemed worth doing a fairly complete proof of concept of the serialization format stuff.

The basic goals here have been to keep the syntax fairly noise free while being well typed,
I.e. instead of dynamically doing tests and vectors of tests, using a separate extension for the vector of tests.
and work well with both raw strings/and escaped strings, for grammars containing quotation marks, newlines, raw strings, etc.

While doing this, I noted that `test_files` currently only allows a single `glob`, but when we start dealing with multiple file extensions, we probably are going to need a way to specify multiple globs.


ratmice:
I'm not certain I understand the second part specifically w.r.t `subset`.
But I believe the answer is yes, the overall testing process is this:

1. Build an `RTParserBuilder` from the parser
2. Read the `.grmtest`, or `.grmtests` file,
3. For each `GrammarTest` pass the `input` field to the `RTParserBuilder` in step 1.
    Compare the output of `parse_map` building an `ASTRepr` to the fields of the
    `GrammarTest` deserialized in step 2 using the following:
    * If `GrammarTest.pass` is `true` and `GrammarTest.ast.is_none()`, check that the return value `errors.is_empty()`
    * If `GrammarTest.pass` is `false` and `GrammarTest.errors.is_none()` check that the returned errors match the errors field.
    * If `GrammarTest.pass` is `true` and `GrammarTest.ast.is_some()` check that the `ASTRepr` built by `parse_map` matches the `Some(GrammarTest.ast)`.
    * If `GrammarTest.pass` is `false` and `GrammarTest.errors.is_some()` check that the errors returned by `parse_map` matches `GrammarTest.errors`.

So essentially there are 4 kinds of tests supported from `it parses successfully`, `it should fail to parse`, and more specifially `it should parse successfully and produce this exact AST`, and `it should fail to parse and produce this exact error`. One of these tests gets run for each `input`.

grammar_testing/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+pub struct GrammarTest {
+    input: String,
+    #[serde(default)]
+    ast: Option<ASTRepr>,
+    // Ideally we would derive this value from `errors`` being empty/none but doing that
+    // within serialization is a painful exercise, so use the accessor `should_pass`.
+    #[serde(default)]
+    pass: Option<bool>,
+    #[serde(default)]
+    errors: Option<Vec<String>>,
```
ratmice, line 38: FWIW, I'm not really a big fan of this structure layout,
The fact that you can specify both `ast`, `errors`, and `pass` leads to non-sensical values like:

```
{
   pass: true,
   errors: ["some error"],
}
```

The layout was largely driven by the serialized input, I think ideally this would be some kind of enum
```
   struct GrammarTest {
        input: String,
        expected_output: Option<GrammarTestOutput>
   }

   enum GrammarTestOutput {
       PassFail{pass: bool},
       AST{ast: ASTRepr},
       Errors{errors: Vec<String>},
   }
```

I will see if I cannot get the same or effectively similar serialized input to map to that changed structure,
perhaps using the [`unwrap_variant_newtypes`](https://github.com/ron-rs/ron/blob/master/docs/extensions.md#unwrap_variant_newtypes), it doesn't sound like it but it seems worth playing around with.

ratmice: I messed around an alternative structure, using the `unwrap_variant_newtypes` extension,

```
#[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
#[serde(deny_unknown_fields)]
pub struct GrammarTest {
    input: String,
    /// A `None` value indicates expecting input to successfully parse.
    expect: Option<TestOutput>,
}

#[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
pub enum TestOutput {
    Fail,
    AST(ASTRepr),
    Errors(Vec<String>),
}
```

```ron
/// Not real parsing output, just testing round tripping of the struct serialization.
[
            (input: "b"),
            (input: "a", expect: AST("start", [("character", "a")])),
            (input: "abc", expect: Fail),
            (input: "abc", expect: Errors(["some error"])),
]
```

Even though it is a bit wordier, it feels like it might be an improvement, and it isn't *that* much wordier?

ltratt: AFAICS you only need an `enum` here? SO I think one would have:

```rust
enum Test {
  TestError { input: String, errors: Vec<String> },
  TestFail {input: String},
  TestSuccess { input: String, ast: Option<...> },
}
```

One could split `TestSuccess` into two. I also wonder if `Fail` and `Error` even need splitting. They could probably be `TestError { input: String, errors: Option<Vec<String>> }`?

ratmice: Perhaps, I'll give that a try.

ltratt: That's a good point: IIUC, it's perfectly valid for a test to produce an AST *and* errors?

ratmice: I think only with error correction, but in that case both are non-deterministic, and as such we can't really encode expected output, so I think we'll have to run tests with that off?

ratmice: So I guess what I'm thinking is that situations that return an AST *and* errors, are kind of exceptional, and I'm a bit hesitant to try and enable it without without having a use case fairly well pinned down, my thinking was if the `parse()` call *does* return an AST for the failure case, we can just ignore the returned AST for now.

If we *do* end up wanting that, we can add a new variant?

The main thing I was thinking regarding `TestFail` vs `TestError` is that `TestFail` seems more inverse to `TestSuccess`, and the `errors:` field is both plural, and optional, so I just have a slight preference for `TestFail` from the naming perspective, but no real behavioral change intended.

ratmice: Ahh, I see.  I hadn't intended to do more than the simplest things, either `this exact ast/error` occurs, or `an unspecified ast from a successful parse, or an unspecified error from a failed parse".

I do wonder if for that kind of thing the `expect: TestOutput` style isn't more flexible
In the sense that you could have `TestOutput::ExactError(errors)`,  `TestOutput::ErrorContains(errors)`,
and some other variant for AST matching.

Anyhow my preference here would be to build it in a way that adding new output matching mechanisms in the future seems natural?  Rather than try and build in all the flexibility on a first take?

ltratt: Right, so matching against serde probably isn't a good idea. Can we define our own, stable, output perhaps? Could it be as simple as the indented "sideways tree!" output that nimbleparse gives?

ratmice: I guess some things worth pointing out, is the paper I had linked to *pattern matching in trees*, asks a slightly different problem than what we want, it's algorithm produces *all* the subtrees that match a pattern tree.

While what I believe we're after is more akin to the *tree inclusion problem*, which asks, given a pattern tree P and a tree T, whether we can produce `P` from `T` by deleting nodes in `T`.

That is to say that the *tree inclusion problem* omits wildcards, and holes.  While the *pattern matching in trees* admits them.  If we slightly modify the rules of the tree inclusion problem, to restrict the number of deletions allowed... e.g. a hole might allow only one deletion, while a wildcard allows any number of deletions.

I suppose it's interesting for two reasons, because of the lack of wildcards in the original formulation, but also just the slightly different formulation of the problem.

ltratt:
I quickly started looking at wildcards, got distracted, and dropped the ball. Sorry!

Let me take a step back and say what I think we've got here:

1. The ability to take a grammar and input in and produce a stable, textual, version of the AST.
2. The ability to compare two stable, textual, versions of the AST. [Do we compare the text or the ASTs?

Have I got that right?

Next question: if things don't match in (2), do we just get "fail" or is the aim to give a detailed diff of some sorts?

ratmice:
> Quick dumb question, which might just be terminological: when you say "AST" I _think_ I would say "parse tree"? In other words, we mean a tree that still has "all the stuff in it from the grammar", much of which a compiler might later choose to discard because it has no (subsequent) semantic effect.

Yeah, when I say "AST" I mean "parse tree", or "serializable generic parse tree", (I have a genetic issue with my hands that makes most of my fingers not actually bend, so I tend to fall back on names that are as few characters as possible, and shorten them by frequency of use, sorry)

> I think the fundamental question in terms of effort then becomes:
> 
>     1. Do we just aim to facilitate "pure text matching" i.e. that grammar G and input I always give output T. We could even do that, I think, by just using `nimbleparse`'s current output and simple textual equality.
> 
>     2. _But_ pure text matching means you can't change G without effecting T, which makes it of limited use.
> 
> 
> So because (2) is so limited, it makes us think "how do we introduce some sort of fuzziness" to matching so that not every test is invalidated as soon as G changes? There are various tree-based algorithms which could help with the fuzziness. I have seen someone implement those in the past, and it's doable, but not a 60 minute job by any means!

Yeah indeed, let me also take a few steps back, and lay out some more basic thoughts,
first let me just show the original nimbleparse-lsp  testing mechanism this evolved from, and it's limitations.
The main toml file is here, which specifies globs akin to `%grmtools{test_files: "*.test"}
and it only supports the boolean options 

https://github.com/ratmice/crimson/blob/main/nimbleparse.toml
https://github.com/ratmice/crimson/tree/main/test

One of the big limitations I was hoping to lift with this was the ability to encode multiple tests into a single file, using a vector of inputs. Simply because I found myself having a lot of really small input files, with silly names like "struct1.test", "struct2.test", which would be nice to have fewer files and combine into one file containing a vector of test inputs named "struct.test".  

From there the next logical step from that are also encoding the expected parse tree, and the expected error if any.
But I would expect while the grammar G is still changing rapidly the basic pass/fail, to kind of be the primitive but primary use case when developing a grammar because it's still useful but isn't so subject to change.
In that mode it basically specifies the shape of the input, without specifying the shape of the output except as a very coarse boolean.

For embedding inputs/outputs my concern here is more that we aren't making any format choices that preclude us from encoding the expected outputs in nice way. But it was not really my focus so much as being the primary mode of use.

> Another question is "how do we want users to encode these tests?" Do we want them to express them directly in Rust? Or some sort of DSL? Or both?

For something like nimbleparse_lsp it was testing occurred at every character press (within reason) of the change to the test files, or the change to the grammar/lexer, (Essentially there was a timeout which was reached when you stopped editing for a few seconds it would generate a rtparser builder, and start feeding that rtparserbuilder the test input.

I had also experimented with using `catch_unwind` and using an `AtomicBool` to communicate with `ActionFn` and a custom `parse_actions`, so that when the grammar was modified, the bool would be flipped, and after that the test run would panic/abort parsing the currently running tests.

So I would definitely prefer it *not* be in Rust, because that requires a whole compilation cycle which isn't really feasible at interactive speeds.  Generating an RTParserBuilder is already pushing it for large grammars.  What I would do was wrap the `ActionFn` using `parse_actions` using an `AtomicBool` to specify if we should abort testing due to an out of date grammar.

So I definitely was going for DSL rather than anything directly testing rust user actions, I think the latter is likely to need a significantly different design. (At the very least for nimbleparse_lsp which intends to be run in the browser, so we're trying to avoid porting the rust compiler to run in the browser and generate web-assembly which will subsequently get run in the browser)

So my focus has *definitely* been on DSL over rust for expressing tests.

> I'm actually starting to wonder if what we're slowly groping our way towards is a "grammar testing DSL". I don't know if any thing like this exists, but we can fairly easily imagine something. Let's imagine I have a calculator grammar G; I might then write multiple tests in our imaginary DSL (which, because I'm unimaginative, I'm thinking of in [lang_tester](https://crates.io/crates/lang_tester)-esque syntax) like this:
> 
> ```
> input: 2+3
> tree:
>   Start
>     Plus
>       Int 2
>       Int 3
> ```

So I guess my difficulty with this syntax is that it isn't really explicit about newlines is that `"2+3"`  or `"2+3\n"`,
I assume that for multi-line inputs, it's whitespace sensitive and relies on the indentation level of the input string
being indented.  Presumably people may need to specify either. It is nice that it doesn't seem to induce any quoting requirements around quotation marks?

This explicit input is why I went with [Ron](https://github.com/ron-rs/ron) for the first try, because
`"2+3\n"`, `"2+3"`, and 

```
r#"2 + 3
#"
```

So it has both the power of raw string for longer multi-line examples, but also explicit quoting requiring escapes
which can be used depending on whichever is most convenient.

>
> What about fuzzy matching? Well, I've deliberately used nimbleparse-ish parse-tree-indented syntax above. I _think_ that lends itself well to very basic fm-textual wildcarding:
> 
> ```
> input: 2 + 3 * 4
> tree:
>   Start
>     ...
>       Mul
>         Int 3
>         Int 4
> ```

Yeah here is where I start worrying about the case where the parse tree has embedded `...`
like `[0...5]` and this accidentally triggering the wildcard match this feels like a similar issue to me as the one for explicitness and escaping of input, but because the name of tokens can contain pretty much any quoted thing.
It seems like we'll also have to deal with an output tree like:

```
input: [0...5]
tree:
   Start
      ...
         "..."
             Int 0
             Int 5
```

It just seems to me like it'll end up with special cases where we need to escape to avoid considering output to be a wildcard which we _can_ avoid with the tree matching in the case of something like:

```
input: "[0...5]",
tree:
```

Because wildcards are part of the tree format, rather than being a parameter to a terminal or non-terminal node.
We no longer have to consider any potential matches occurring within the tree data that could conflict.

I'm still using ron with explicit types here, using the `implicit` types this would reduce to something like:

```
input: "[0...5]",
tree: 
   ("Start", 
         Wildcard(
              ("...", [
                   ("Int", 0),
                   ("Int", 5)
              ])))
```

Using a DSL we can replace `Wildcard` with the ellipses while retaining the same typing discipline which keeps
wildcards out of the data stream.

```
   ("Start", 
         ...
              ("...", [
                   ("Int", 0),
                   ("Int", 5)
              ]))

```

I guess my concern is that if we treat `tree:` as a serialized string, and do fuzzy matching on that string.
We now have to deal with all the co-mingling of data and match signal, since there is now a bunch of corner cases if we treat it like so, since it isn't well specified whether wildcards are allowed or disallowed inside/outside the quotes anymore.

```
input: "[0...5]",
tree: r#"
   Start
      ...
         "..."
             Int 0
             Int 5
#"
```

Hopefully that goes into more detail at least about my whole hang-up on this whole approach.

ratmice:
Makes sense, Initially I guess my hesitation to doing a custom format was just that the rust raw string format isn't something we can easily replicate with `lrlex`, because it's very context sensitive (even the limited context sensitivity given by start states is not enough), so if we want raw strings that will require a manual lexer.

Above and beyond that I'm not certain we can rely on lrpar without introducing any bootstrapping issues, I *think* since it is in lrlex (`CTLexerBuilder`) and `nimbleparse`, and both *already* depend on lrpar we can probably depend on an lrpar defined grammar for the test DSL without any actual problems, that is my hope at least!

Anyhow I think I have a good enough idea now that I should probably be able to start playing with implementations.

ltratt:
Bootstrapping introduces all sorts of fun and games. I'm not sure I'd advocate for it TBH if there's even the slightest whiff of problems! For such a simple DSL it isn't too difficult to write a recursive descent parser (a la cfgrammar) if neccessary.

ratmice:
Yeah, indeed. I also realized that even if we did avoid raw string context sensitivity we probably still couldn't use lrlex, just because of bootstrapping it would introduce a circular dependency.

ltratt:
So coming back to this: I think I am still strongly in favour of the DSL route. The exact syntax we could endlessly bikeshed if we wanted to :) I think the crucial thing is (1) to allow multiple tests in a file (fortunately this is easy!) (2) some concept of wildcarding so that users can have tests that don't require full rewriting for every minor tweak of a grammar.

ratmice:
Anyhow, I guess my question of how we want to handle the multi-test case files remains,
because I don't think the normal comma separated sequence of elements within `[]` works very well for this 
which is why I was asking about the yaml style vectors e.g.
```
- input: ()
  tree:
    Start
- input: a
  tree:
    Start
      A
```

otherwise perhaps we could throw curly braces around it and use the normal vector style?
```
[{
  input: ()
  tree:
    Start
}, {
  input: a
  tree:
    Start
     A
}]
```

ltratt:
Curly brackets are fine with me!

ltratt:
Totally understood, and I for one am grateful for whatever time you're able to find, whenever that is!

PR #600, opened 2025-10-07
Fix compile error in generated code.

rbartlensky:
I was testing out the `master` branch of `grmtools` (because it has an API I need) and noticed that the generated Rust code has a compile error:

```
error[E0412]: cannot find type `Node` in the crate root
     --> /home/<redacted>/target/debug/build/rust-parser-992f753d99ae5fa2/out/rust.y.rs:24154:24
      |
24154 |                 crate::Node<
      |                        ^^^^ not found in the crate root
      |
help: consider importing one of these items
      |
6     +         use crate::rust_y::Node;
      |
6     +         use lrpar::Node;
      |
help: if you import `Node`, refer to it directly
      |
24154 -                 crate::Node<
24154 +                 Node<
      |
```

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -1179,7 +1179,7 @@ where
                 }
             }
             YaccKind::Original(YaccOriginalActionKind::GenericParseTree) => quote! {
-                (::std::option::Option<crate::Node<<#lexertypest as ::lrpar::LexerTypes>::LexemeT, #storaget>>,
+                (::std::option::Option<Node<<#lexertypest as ::lrpar::LexerTypes>::LexemeT, #storaget>>,
                     ::std::vec::Vec<::lrpar::LexParseError<#storaget, #lexertypest>>)
             },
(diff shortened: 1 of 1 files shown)


rbartlensky:
> I haven't looked at the details or experienced this yet, but it seems reasonable, if as you say the code here is generated within the module where the `Node` type is also generated. I'm not certain though why/how you experienced this in a way that wasn't caught/covered by the testsuite or the `lrpar/examples`

I can upload some code tomorrow and link it here.

> The only question I have is if this isn't a case where we should use the `self` keyword, by removing the `crate` is is possible that this could be confused by something to the effect of
> 
> ```
> mod child {
>     use bar::Node;
>     lrpar_mod!("rust.y");
> }
> ```
> 
> Or does that just not a problem because e.g. we'd get a compile error at the time at which the type `Node` is declared due to a naming conflict?

This won't be an issue because the generated parser has a top-level module. It looks like:

```rust
```

ratmice:
It occurred to me the `self` keyword in module context is only rarely used, and since it is more familiar in i.e. self variable context, I should probably link to the docs for it.

https://doc.rust-lang.org/rust-by-example/mod/super.html

PR #601, opened 2025-10-16
Update cttest names

ratmice:
Just a minor thing I noticed when I went to work on cttests, the `name` field was copy/pasted in a few places
where the tests no longer reflected the actual purpose.  It isn't clear to me though when or whether this gets printed.


ltratt:
I must admit that I'm not sure if this is used or not, but either way, if it's there, it might as well be accurate!

PR #603, opened 2025-10-17
Add initial support for extra_files to cttests

ratmice:
To say this isn't very pretty is a bit of an understatement, but this is my first successful attempt at implementing
`extra_files` in the .yaml test output.

We want to use the [`CTLexerBuilder::lrpar_config`](https://docs.rs/lrlex/latest/lrlex/struct.CTLexerBuilder.html#method.lrpar_config) callback which implements `%grmtools{test_files}` behavior.
In order to do that though, we must fix/hack up the lifetimes of the `cgen_helper::run_test_files`

We hack up those lifetimes in two ways:
1. leak the `Path` sent to `run_test_files`
2. reread and reparse the `.test` file yaml from within the closure.

The main issue being the borrow checker can't tell that the `Path` variable isn't used *after* the `run_test_files` function exits.
I spent quite a lot of time fighting this, and this is the only thing I could manage to get working.

I manually tested that when you change one of the `.calc_input` files to include any kind of malformed input that it is caught correctly.

### DIFF
--- a/lrpar/cttests/src/calc_input.test
+++ b/lrpar/cttests/src/calc_input.test
@@ -1,0 +1,49 @@
+name: Test with calculator input from %grmtools{test_files}
+grammar: |
+    %grmtools {
+        yacckind: Original(YaccOriginalActionKind::UserAction),
+        recoverer: RecoveryKind::None,
+        test_files: "*.calc_input"
+    }
+    %start Expr
+    %actiontype Result<u64, ()>
+    %avoid_insert 'INT'
+    %%
+    Expr: Expr '+' Term { Ok($1? + $3?) }
+        | Term { $1 }
+        ;
+
+    Term: Term '*' Factor { Ok($1? * $3?) }
+        | Factor { $1 }
+        ;
+
+    Factor: '(' Expr ')' { $2 }
+          | 'INT' {
+                let l = $1.map_err(|_| ())?;
+                match $lexer.span_str(l.span()).parse::<u64>() {
+                    Ok(v) => Ok(v),
+                    Err(_) => {
+                        let ((_, col), _) = $lexer.line_col(l.span());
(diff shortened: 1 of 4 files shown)


ratmice:
So just in the spirit of trying to brainstorm alternate ideas, if we had some sort of API for running the `test_files` manually, instead of using `lrpar_config` and relying on that function to call it, we could call it ourselves then continue to use the separate builders with the token map?

lrlex/src/lib/ctbuilder.rs
```diff
@@ -226,7 +226,7 @@ where
     LexerTypesT::StorageT: Debug + Eq + Hash + ToTokens,
     usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
 {
-    lrpar_config: Option<Box<dyn Fn(CTParserBuilder<LexerTypesT>) -> CTParserBuilder<LexerTypesT>>>,
+    lrpar_config: Option<Box<dyn Fn(CTParserBuilder<LexerTypesT>) -> CTParserBuilder<LexerTypesT> + 'a>>,
```
ratmice: If we're willing to change the lifetime of this closure from being `'static` to another lifetime like `'a` here,
we can entirely get rid of the leak variables hacks.

This is here because Box<dyn Trait> is defaulting to 'static unless otherwise specified.
With it specified to something like 'a, the closure no longer needs to outlive the builder and main.

With that we can get rid of the leak hacks in this branch (at least, haven't looked into further cleanups yet)

ltratt: Going from `'static` -> `'a` seems fully backwards compatible (i.e. it's more general) to me, so I don't see a problem with this!

ltratt:
Thanks!

ratmice:
Had to rustfmt the first commit, should be fixed.

PR #604, opened 2025-10-20
Grmtools section array values

ratmice:
Here is a first working attempt at adding array values to the grmtools section, and migrating the `test_files` to using that.

I also did some manual testing, e.g. that adding `"*.bad_input"` to the `calc_input.test` causes a testsuite failure.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -210,56 +210,6 @@ impl From<Value<Span>> for Value<Location> {
     }
 }
 
-impl<T> Value<T> {
-    pub fn expect_string_with_context(&self, ctxt: &str) -> Result<&str, Box<dyn Error>> {
-        let found = match self {
-            Value::Flag(_, _) => "bool".to_string(),
-            Value::Setting(Setting::String(s, _)) => {
-                return Ok(s);
-            }
-            Value::Setting(Setting::Num(_, _)) => "numeric".to_string(),
-            Value::Setting(Setting::Unitary(Namespaced {
-                namespace,
-                member: (member, _),
-            })) => {
-                if let Some((ns, _)) = namespace {
-                    format!("'{ns}::{member}'")
-                } else {
-                    format!("'{member}'")
-                }
-            }
-            Value::Setting(Setting::Array(_, _, _)) => "array".to_string(),
-            Value::Setting(Setting::Constructor {
-                ctor:
-                    Namespaced {
-                        namespace: ctor_ns,
-                        member: (ctor_memb, _),
-                    },
-                arg:
-                    Namespaced {
-                        namespace: arg_ns,
-                        member: (arg_memb, _),
-                    },
-            }) => {
-                format!(
-                    "'{}({})'",
-                    if let Some((ns, _)) = ctor_ns {
-                        format!("{ns}::{ctor_memb}")
-                    } else {
-                        arg_memb.to_string()
-                    },
-                    if let Some((ns, _)) = arg_ns {
-                        format!("{ns}::{arg_memb}")
-                    } else {
-                        arg_memb.to_string()
-                    }
-                )
-            }
-        };
-        Err(format!("Expected 'String' value, found {}, at {ctxt}", found).into())
-    }
-}
-
 static RE_LEADING_WS: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[\p{Pattern_White_Space}]*").unwrap());
 static RE_NAME: LazyLock<Regex> = LazyLock::new(|| {
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -56,23 +56,35 @@ echo "2 + 3 * 4" | cargo run | grep "Result: 14"
 touch src/main.rs && CACHE_EXPECTED=y cargo build
 cd $root/lrpar/examples/calc_actions
 echo "2 + 3 * 4" | cargo run --package nimbleparse -- src/calc.l src/calc.y -
+# Invoke `%grmtools{test_files}`
+cargo run --package nimbleparse -- src/calc.l src/calc.y
 echo "2 + 3 * 4" | cargo run | grep "Result: 14"
 touch src/main.rs && CACHE_EXPECTED=y cargo build
 cd $root/lrpar/examples/calc_ast
 echo "2 + 3 * 4" | cargo run --package nimbleparse -- src/calc.l src/calc.y -
+# Invoke `%grmtools{test_files}`
+cargo run --package nimbleparse -- src/calc.l src/calc.y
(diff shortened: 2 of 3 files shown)


lrpar/cttests/src/lib.rs
```diff
@@ -355,16 +355,15 @@ fn test_expect() {
...
     use std::io::BufReader;
     use std::io::{BufRead as _, Read as _};
 
-    let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
     let examples_glob = format!("{manifest_dir}/../examples/**");
     let examples_l_glob = format!("{examples_glob}/*.l");
     let examples_y_glob = format!("{examples_glob}/*.y");
-    let out_dir = env::var("OUT_DIR").unwrap();
+    let out_dir = env!("OUT_DIR");
```
ratmice: Maybe this should go into a separate patch, it was helpful for debugging this, because `env!` caches the environment variables in the static section of the binary, so they can be run outside of cargo, like when running under a debugger.

ratmice:
The two in https://github.com/softdevteam/grmtools/pull/604/commits/60df6f94635a9b7c1ea3ccb331255bce3318f5e2 calc_input and ctfails/calc_bad_input both use multiple values. In the first one, i added the same glob twice.
In the second one, it parses both the good inputs from `calc_input` and the `.bad_input` files.

But actually now that I think about it this latter case is probably unreliable because of multi-threaded test runners and the globbing may be spuriously matching an empty set of files.

These should really only be matching files within the .test

ltratt:
Good idea!

lrpar/cttests/src/lib.rs
ltratt: Let's leave it as-is in your PR: if it breaks, we'll find out at some point!

lrlex/src/lib/ctbuilder.rs
```diff
@@ -540,24 +547,31 @@ where
...
+                                                    if ext.starts_with("grm") {
+                                                        Err(ErrorString("test_files extensions beginning with `grm` are reserved.".into()))?
+                                                    }
+                                                }
+                                            }
+                                            let input = fs::read_to_string(&path)?;
+                                            let l: LRNonStreamingLexer<LexerTypesT> =
+                                                closure_lexerdef.lexer(&input);
+                                            for e in rtpb.parse_map(&l, &|_| (), &|_, _| ()).1 {
+                                                Err(format!("parsing {}: {}", path.display(), e))?
```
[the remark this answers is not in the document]
ratmice: Essentially `l` is the lexer for the contents of one of the files we've globbed, 
`rtpb.parse_map(&l, &|_| (), &|_, _| ())` is essentially running something similar to `parse_noaction` (deprecated)
That is returning a `(Option<()>, Vec<LexParseError<...>>)`, because of the unit return of `&|_| ()`, etc.  We ignore the first value of the tuple.

Thus the `format` is essentially adding the fully evaluated globbed path onto the error message, since the rtpb never encounters the path it's just dealing with a string contents/lexer that has already been read from the filesystem.

I *think* that is what you're referring to here, if not I might need a bit more help understanding.
The other cases in this block are just ensuring the `test_files` key is an array of string value mostly.

cfgrammar/src/lib/header.rs
```diff
@@ -295,7 +295,88 @@ fn add_duplicate_occurrence<T: Eq + PartialEq + Clone>(
...
+                            Ok((Setting::Unitary(path_val), i))
+                        }
+                    }
+                }
+            },
+        }
+    }
+    fn parse_value(&'_ self, i: usize) -> Result<(Value<Span>, usize), HeaderError<Span>> {
+        let (setting, pos) = self.parse_setting(i)?;
+        Ok((Value::Setting(setting), pos))
```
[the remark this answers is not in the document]
ratmice: Should be fixed in [970f734](https://github.com/softdevteam/grmtools/pull/604/commits/970f7346348c5d63bdb0e55610dc546c08448ef8)

lrlex/src/lib/ctbuilder.rs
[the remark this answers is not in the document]
ltratt: I think my question is the following: is:

```rust
for x in y {
  Err(...)?
}
```

not going to return `Err(...)` on the first iteration of the loop?

lrlex/src/lib/ctbuilder.rs
```diff
@@ -547,24 +547,31 @@ where
...
+                                            }
                                         }
                                     }
-                                }
-                                let input = fs::read_to_string(&path)?;
-                                let l: LRNonStreamingLexer<LexerTypesT> =
-                                    closure_lexerdef.lexer(&input);
-                                for e in rtpb.parse_map(&l, &|_| (), &|_, _| ()).1 {
-                                    Err(format!("parsing {}: {}", path.display(), e))?
+                                _ => return Err("Invalid value for setting 'test_files'".into()),
```
ratmice: This `_ =>` branch looks like it is at the wrong indentation level, but for whatever reason rustfmt gives up on it here.

lrlex/src/lib/ctbuilder.rs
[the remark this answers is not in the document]
ratmice: Looking at the output still leaves quite a bit to be desired, still working on it.

1. Not actually utilizing the pretty spanned error formatting code.
2. When the error itself spans multiple lines, we should add the appropriate indent for each line...
3. There is some sync that should happen between nimbleparse to give similar error text.

ltratt:
I think this is ready to squash; and then we're ready to make a release?

lrlex/src/lib/ctbuilder.rs
```diff
@@ -547,24 +547,31 @@ where
...
-                            let path_joined = grm_path.parent().unwrap().join(test_files);
-                            for path in
-                                glob(&path_joined.to_string_lossy()).map_err(|e| e.to_string())?
-                            {
-                                let path = path?;
-                                if let Some(ext) = path.extension() {
-                                    if let Some(ext) = ext.to_str() {
-                                        if ext.starts_with("grm") {
-                                            Err(ErrorString("test_files extensions beginning with `grm` are reserved.".into()))?
+                        Some(HeaderValue(_, Value::Setting(Setting::Array(test_globs, _, _)))) => {
```
ratmice: There should/needs to be the same change to this one in nimbleparse.
We should also probably add nimbleparse invocations without inputs to `.buildbot.sh`
so the testsuite could catch it.

ratmice: Should be fixed in [c22415f](https://github.com/softdevteam/grmtools/pull/604/commits/c22415f6a805e067bc985a76cf95ef6ecf48c41c)

PR #615, opened 2025-12-02
More mdbook

ltratt:
Two not-entirely-related commits relating to the grmtools book, including testing it in CI (oops).

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -35,6 +35,15 @@ cargo fmt --all -- --check
 rustup toolchain install stable
 rustup default stable
 
+# Later on we are going to need to install cargo-deny and mdbook. We kick the
+# install jobs off now so that at least some work (e.g. downloading crates) can
+# happen in parallel, speeding up the overall process.
+
+cargo_deny_mdbook_tmp=$(mktemp)
+( cargo install --locked cargo-deny ; cargo install --locked mdbook ) \
+  >"${cargo_deny_mdbook_tmp}" 2>&1 &
+cargo_deny_mdbook_pid=$!
+
 cargo test
 cargo test --release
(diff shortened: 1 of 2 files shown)


.buildbot.sh
```diff
@@ -89,7 +98,12 @@ cd $root
...
-fi
+# Check licenses.
+wait "${cargo_deny_mdbook_pid}" || ( cat "${cargo_deny_mdbook_tmp}" && exit 1 )
+cargo-deny check license
+
+# Build the docs
+cd doc
+mdbook build
+test -d book
+cd ..
```
ratmice: Only a fairly minor nit, feel free to ignore if you prefer how it is.

I might suggest since the build root is in the `$root` variable already
`cd $root/doc` and `cd $root` instead of `cd doc` and `cd ..`

.buildbot.sh
```diff
@@ -35,6 +35,15 @@ cargo fmt --all -- --check
 rustup toolchain install stable
 rustup default stable
 
+# Later on we are going to need to install cargo-deny and mdbook. We kick the
+# install jobs off now so that at least some work (e.g. downloading crates) can
+# happen in parallel, speeding up the overall process.
+
+cargo_deny_mdbook_tmp=$(mktemp)
+( cargo install --locked cargo-deny ; cargo install --locked mdbook ) \
```
ratmice: I see there are also some more `cargo install` calls happening (for I believe decently sized wasm packages).
I'm not sure if they benefit from this parallelism too, perhaps another patch?.

.buildbot.sh
ltratt: Good idea! Fixed in the force push.

PR #621, opened 2026-01-20
Pass `LexerKind` by reference.

ltratt:
At the moment the code happens to work only because of an oversight in rustc: that oversight is likely to be fixed soon (https://github.com/rust-lang/rust/pull/150681).

@meithecatte If you're able to do so, I'd be grateful if you can confirm this addresses the rustc issue properly. If so, I will then make a new release shortly after so that we're not blocking the rustc PR.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -248,7 +248,7 @@ where
     show_warnings: bool,
     header: Header<Location>,
     #[cfg(test)]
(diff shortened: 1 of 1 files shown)


ratmice:
I'm happy to approve.

I was able to reproduce the error from the [crater log](https://crater-reports.s3.amazonaws.com/pr-150681-2/try%2336577a2b3cb42b8f23dec84e9203116bad6051e8/reg/lrlex-0.14.0/log.txt) in #620 on current rustc by adding a second variant to `LexerKind` and compiling with `cargo build --all-targets` since the error only occurs with `cfg(test)`.

Edit: The test worked as expected (failed on the master branch, and worked with this one)

PR #631, opened 2026-04-28
Avoid repeated parser table decode and cut parse setup overhead (issue 630)

### LINKED ISSUE #630: Generated parse() re-decodes grammar tables on every call
Hi! First, thanks for grmtools — I'm looking at using it in a project
that parses a lot of small inputs in a tight loop, and while profiling I
noticed something I wanted to flag.

The generated `parse()` from `lrpar_mod!` calls
`::lrpar::ctbuilder::_reconstitute(__GRM_DATA, __STABLE_DATA)` on every
invocation, which does two `bincode::decode_from_slice` calls to rebuild
the `YaccGrammar` and `StateTable`. For a single whole-buffer parse this
is invisible, but if you call `parse()` many times — e.g. parsing many
small chunks, or parsing inside a hot loop — that decode ends up
dominating. `RTParserBuilder::new` only borrows the tables, so the work
is pure repeat overhead.

In addition, the per-call setup in `lrpar::parser` has a couple of small
redundancies:

 * `Parser::token_cost` is stored as `Box<&dyn Fn>` — a heap allocation
   around an already-borrowed reference, on every parse.
 * `parse_map` / `parse_actions` collect the lexer iterator into a
   `Vec<Result<_, _>>` and then iterate it again to build the lexeme
   vec, instead of a single `collect::<Result<Vec<_>, _>>()`.
 * The `iter_tidxs` cost-positive `assert!` loop runs in release builds,
   even though it's really a builder-side contract.

On the `calc_actions` example, with 200k iterations of a few short
expressions in release mode, I'm seeing roughly **~2.06 µs/parse → ~0.80 µs/parse**
(about a 2.5× speedup) once these are addressed. Most of that is the
table decode; the rest is the small setup items.

I have a working patch and would be happy to send a PR. Two things I'd
appreciate guidance on before I do:

1. **Public API.** To cache the tables across calls, generated code
   needs a concrete type to put in a `OnceLock`. Naming
   `(YaccGrammar<S>, StateTable<S>)` directly would force every
   downstream crate to depend on `lrtable`, which seems like a
   regression. My current approach is to introduce an opaque
   `lrpar::ParserTables<S>` newtype with `grm()` / `stable()`
   accessors, re-exported at the crate root, so generated code only
   names `::lrpar::ParserTables`. Does that shape sound reasonable, or
   would you prefer it `#[doc(hidden)]`, or somewhere else entirely?

2. **`_reconstitute` signature.** It's already `#[doc(hidden)]`, but its
   return type would change from a tuple to `ParserTables<S>`. Since
   generated code is regenerated against the matching `lrpar` version,
   I think this is fine, but happy to keep the old function and add a
   new one if you'd rather not touch it.

Happy to open the PR as-is, or to iterate on the design first — whatever
works best for you. Thanks again!

avityuk:
While profiling a workload that parses many small inputs in a tight loop, I found two sources of avoidable per-parse overhead in `lrpar`.

First, generated `parse()` functions from `lrpar_mod` were calling  `_reconstitute(__GRM_DATA, __STABLE_DATA)` on every invocation, which meant re-decoding the grammar and state table every time even though RTParserBuilder::new only borrows them. This change caches the reconstituted tables in generated code and reuses them across calls.

Second, there were a couple of smaller setup costs in lrpar::parser:
- parser::token_cost was stored as `Box<&dyn Fn(...)>`, introducing a heap allocation around an already-borrowed callback on every parse.
- parse_map and parse_actions collected the lexer iterator into `Vec<Result<...>>` and then walked it again to build the lexeme vector.

This PR removes those extra costs by:
- caching generated parser tables behind a `OnceLock`.
- introducing an opaque `lrpar::ParserTables` wrapper so generated code does not need to name lrtable types directly
- storing token_cost as a borrowed callback rather than boxing it
- collecting lexemes in one pass

Since the time spent in _reconstitute is proportional to grammar size, this change is particularly impactful there.
 
However, even on a very small grammar, such as calc_actions example, these changes brought a tight-loop parse benchmark down from roughly ~2.06 µs/parse to ~0.80 µs/parse.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -1093,7 +1093,7 @@ where
         let run_parser = match self.yacckind.unwrap() {
             YaccKind::Original(YaccOriginalActionKind::GenericParseTree) => {
                 quote! {
-                    ::lrpar::RTParserBuilder::new(&grm, &stable)
+                    ::lrpar::RTParserBuilder::new(grm, stable)
                         .recoverer(#recoverer)
                         .parse_map(
                             lexer,
@@ -1104,7 +1104,7 @@ where
             }
             YaccKind::Original(YaccOriginalActionKind::NoAction) => {
                 quote! {
-                    ::lrpar::RTParserBuilder::new(&grm, &stable)
+                    ::lrpar::RTParserBuilder::new(grm, stable)
                         .recoverer(#recoverer)
                         .parse_map(lexer, &|_| (), &|_, _| ()).1
                 }
@@ -1145,7 +1145,7 @@ where
                                     #action_fn_parse_param_ty
                             ) -> #actionskind #type_generics
                         > = ::std::vec![#(&#wrappers,)*];
-                    match ::lrpar::RTParserBuilder::new(&grm, &stable)
+                    match ::lrpar::RTParserBuilder::new(grm, stable)
                         .recoverer(#recoverer)
                         .parse_actions(lexer, &actions, #action_fn_parse_param) {
                             (Some(#actionskind::#action_ident(x)), y) => (Some(x), y),
@@ -1205,6 +1205,14 @@ where
             const __GRM_DATA: &[u8] = &[#(#grm_data,)*];
             const __STABLE_DATA: &[u8] = &[#(#stable_data,)*];
 
+            fn __lrpar_parser_data() -> &'static ::lrpar::ParserData<#storaget> {
+                static DATA: ::std::sync::OnceLock<::lrpar::ParserData<#storaget>>
+                    = ::std::sync::OnceLock::new();
+                DATA.get_or_init(
+                    || ::lrpar::ctbuilder::_reconstitute(__GRM_DATA, __STABLE_DATA)
+                )
+            }
+
             #[allow(dead_code)]
             pub fn parse #generics (
                  lexer: &'lexer dyn ::lrpar::NonStreamingLexer<'input, #lexertypest>,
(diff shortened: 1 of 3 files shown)


lrpar/src/lib/ctbuilder.rs
```diff
@@ -1601,16 +1611,33 @@ where
     }
 }
 
+/// Bundles the parser tables (`YaccGrammar` + `StateTable`) so that generated parsers can hold
+/// them in a `OnceLock` without naming `lrtable` directly.
+pub struct ParserTables<StorageT: Eq + Hash> {
```
ltratt: I've been chewing on the right name for this. It's not "tables", quite. I think I might go for `ParserBundle`. I also think we should "nodoc" this `struct`, because we don't want users relying on this part of the API (ideally they won't even know it exists).

ltratt:
I have one easy comment: @ratmice does this look OK to you?

ratmice:
I haven't yet had a chance to look, but I will try and have a gander this evening, in a couple of hours.

lrpar/src/lib/ctbuilder.rs
ratmice: One thing I kind of like is something along the lines of `ParserData`, or `ParserDataBundle`, but I'm not too picky, and `ParserBundle` seems okay too.

ratmice:
Looks like what it says on the tin, so this looks OK to me, couldn't help but join the bikeshed a little though.

lrpar/src/lib/ctbuilder.rs
avityuk: I went with `ParserBundle` and added `[doc(hidden)]` (I assume that's what "nodoc" is?).

Let me know if you change your mind with naming. `ParserData` doesn't sound bad either.

ltratt: Agreed: `ParserData` would be better. If you force push that, I think we're ready to merge!

avityuk: Done, changed it to `ParserData`. PTAL. Thanks!

PR #632, opened 2026-04-30
Prepare a 0.14.2 release.

ltratt:
@ratmice I _think_ the CHANGES here are a sensible way of describing things, but please let me know if you think they can be improved!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+# grmtools 0.14.2 (2026-04-30)
+
+* Only initialise internal parser state once at run-time. For programs that
+  create many new parser instances, this can be a substantial win.
+
+* Improve a number of error messages: several of these were previously
+  inscrutable.
+
+
 # grmtools 0.14.1 (2026-01-20)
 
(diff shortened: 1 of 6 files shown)


ratmice:
When going to review the changes between the prior release and the current one,
I noticed that the grmtools-0.14.1 tag appears as though it didn't get set.

I'm not actually familiar with the release process, whether it is automated or manual but wanted to note it in case it was automated, perhaps we may need to look into it lest it might happen again?

ltratt:
Mea culpa! I sometimes forget to do `git push --tags _upstream_`. Should be fixed now.

ratmice:
This seems reasonable to me.

I definitely didn't get to all the errors in #623 but there is nothing that was committed that is in any kind of limbo, but a few PRs left before that can be closed that haven't been completed yet. None of that seems worth holding up a release.

ltratt:
Separately, I noticed that https://softdevteam.github.io/grmtools/ had broken since the mdbook `multilingual` thing -- I've now unbroken that, so we'll keep getting docs generated.

PR #635, opened 2026-05-03
Stop trimming escaped spaces off the end regex

### LINKED ISSUE #634: regex/token name separator for regexes that end with trailing spaces
This is an issue that affects current grmtools which was brought to light when considering the feature request in #597 it seemed like it would be good to split this out into it's own issue than discuss it there.

If we take the following lex file (foo.l) which ends in a trailing space, current grmtools accepts it,
but it seems like it should probably reject it.

```
%%
\ \ "weirdness"
\n ;
```

The thing is the trailing space in `\ \ ` is part of a regex, but it gets treated as the space separator that separates regex from the token name. Then it seems like perhaps it gets trimmed off of the regular expression before being passed to regex.

```
$ echo " " | cargo run --quiet -p lrlex ./foo.l -
weirdness
$ echo "  " | cargo run --quiet -p lrlex ./foo.l -
weirdness
weirdness
```

I would have expected the echo with two spaces to match a single weirdness token, and the one with a single token to not match.

The following regex *is* rejected, but by the regex crate, I assume it is just sending the '\' token alone to regex.

```
%%
\ "rejectedByRegex"
\n ;
```

ratmice:
This is a first attempt to fix #634 basically it continues to interpret escaped spaces
as a separator between regex and token ident, however it stops trimming the escaped space.

This has pretty thorough testing of the trim function, but only minimal testing of the stream of tokens produce by lrlex in cttests. Understand if you want me to expand on that, but this seemed like a good start.

### DIFF
--- a/lrlex/src/lib/parser.rs
+++ b/lrlex/src/lib/parser.rs
@@ -490,7 +490,7 @@ where
 
         if !dupe {
             let (start_states, re_str) =
-                self.parse_start_states(i, line[..rspace].trim_end_matches(matches_whitespace))?;
+                self.parse_start_states(i, trim_end_unescaped(&line[..rspace]))?;
             let rules_len = self.rules.len();
             let tok_id = LexerTypesT::StorageT::try_from(rules_len)
                            .unwrap_or_else(|_| panic!("StorageT::try_from \
@@ -682,6 +682,20 @@ where
         } else {
             None
         }
+    }
+}
+
+fn trim_end_unescaped(s: &str) -> &str {
+    let trimmed = s.trim_end_matches(matches_whitespace);
+    if trimmed.len() == s.len() {
+        return s;
+    }
+    // If the number of backslashes is odd then the first space in the trimmed portion is escaped so re-add it.
+    if trimmed.chars().rev().take_while(|&c| c == '\\').count() % 2 == 1 {
+        // Panic safety: the trimmed portion is at least one char long.
+        &s[..trimmed.len() + s[trimmed.len()..].chars().next().unwrap().len_utf8()]
+    } else {
+        trimmed
     }
 }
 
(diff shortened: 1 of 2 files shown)


lrlex/src/lib/parser.rs
```diff
@@ -685,6 +685,43 @@ where
...
 
+fn trim_end_unescaped(s: &str) -> &str {
+    use unicode_width::UnicodeWidthChar;
+    let mut cbuf = [0; 4];
+    let mut initial_space_bytes = 0;
+    let mut last_char_width = 0;
+    // First loop over spaces
+    for ch in s.chars().rev().into_iter() {
+        if RE_SPACE_SEP.is_match(ch.encode_utf8(&mut cbuf)) {
+            last_char_width = ch.width().unwrap_or(0);
```
ratmice: I think this is probably wrong, and we want the width of the character in bytes, and what this `width` function gives is the width of the character in columns.

So I think this should be `ch.len_utf8()`

ratmice: Should be fixed in [26e8886](https://github.com/softdevteam/grmtools/pull/635/commits/26e88866471c33c008345289ec7c49a3d174d995)

lrlex/src/lib/parser.rs
```diff
@@ -685,6 +685,42 @@ where
     }
 }
 
+fn trim_end_unescaped(s: &str) -> &str {
+    let mut cbuf = [0; 4];
+    let mut total_ws_bytes = 0;
+    let mut last_ws_char_len = 0;
+    // First loop over spaces
+    for ch in s.chars().rev().into_iter() {
```
ltratt: Could this be something like `in s.chars().rev().into_iter().take_while(|x| RE_SPACE_SEP.is_match(...)`?

ltratt:
I have one shallow comment: otherwise, this looks good, particularly with the test suite!

lrlex/src/lib/parser.rs
ratmice: I don't think so for a couple of reasons, it seems like `take_while` takes ownership of the self iterator, and returns an iterator over the matching elements.

What we'd really want is `skip_while`, so the iterator resumes from where the skipping returned false. Unfortunately `skip_while` doesn't lend itself to having a `fold` like accumulator, this kind of needs something that does both.

Beyond that I looked at a couple of things, 
* `try_reduce` could work but is nightly only.
* `scan` seemed like it could work, but I couldn't find a way that improved clarity.

So part of the problem is that we want to accumulate into a value,
and then continue iteration from that point exactly where accumulation stopped.

Here is a `try_fold` version after which we do still need to make a second iterator on a sub-slice of the string.
```
    // First loop over space characters
    let (last_ws_char_len, mut total_ws_bytes) = match s.chars().rev().into_iter().try_fold((0, 0), |(last_ch_len, total_ws_bytes), ch| {
        if RE_SPACE_SEP.is_match(ch.encode_utf8(&mut cbuf)) {
            let this_ch_len = ch.len_utf8();
            ControlFlow::Continue((this_ch_len, total_ws_bytes + this_ch_len))
        } else {
            ControlFlow::Break((last_ch_len, total_ws_bytes))
        }
    }) {
        // The nightly function into_value would help here
       // https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html#method.into_value
        ControlFlow::Continue((x, y)) => (x, y),
        ControlFlow::Break((x, y)) => (x, y),
    };
```

I kind of think this fold might be an improvement, opinions?

ratmice: ~~Oops, I forgot to `cargo test` that `try_fold` implementation, after getting it to compile, so of course it doesn't seem to work yet.~~

Nevermind, that was me messing up some *other* part of the function.

ratmice: I was curious if AI had any ideas to improve the code, and it did have a novel idea.
Using the `trim_end_matches` function to trim all spaces, then re-add the space if needed.

I ended up rewriting basically all the AI code, but this seems like it might be more idiomatic, or at least less loopy.

```
fn trim_end_unescaped(s: &str) -> &str {
    let trimmed = s.trim_end_matches(matches_whitespace);
    if trimmed.len() == s.len() {
        return s;
    }
    // If the number of backslashes is odd then the first space in the trimmed portion is escaped so re-add it.
    if trimmed.chars().rev().take_while(|&c| c == '\\').count() % 2 == 1 {
        // Panic safety: the trimmed portion is at least one char long.
        &s[..trimmed.len() + s[trimmed.len()..].chars().next().unwrap().len_utf8()]
    } else {
        trimmed
    }
}
```

ratmice: I went ahead and pushed that in [dbbc4af](https://github.com/softdevteam/grmtools/pull/635/commits/dbbc4af77e6a4d18e437d4cd0db3791228642cb6)

Let me know if you'd rather something else though.

ltratt: Less code is generally a code sign!

ltratt:
Please squash.

ratmice:
Squashed.

PR #637, opened 2026-05-04
Add some docs on lex syntax

### LINKED ISSUE #597: Production names can't contain whitespace
I wanted to have a `\/[A-Za-z0-9-]+\/ "unknown directive"` lex rule, but this turns out to produce an error. I haven't checked if Lex or Flex accept those inputs, but at least my syntax highlighter does.

IMO, regardless of whether spaces become allowed, it might be a good idea to document the accepted charset for token names?

ratmice:
I was looking to fix #597 by documenting the limitation of not having spaces in token names
there was one place in the [major differences](https://softdevteam.github.io/grmtools/master/book/lexcompatibility.html#major-differences) section where it mentions token names, but doesn't define them explicitly.

I figured it might be good to just add a node on lex syntax. Here was a very quickly thrown together overview of the lex syntax to test out whether we like that idea, and if it's worth expanding upon.

I didn't go through and document quirks though like regexes with trailing spaces or anything, not sure if we need to.

### DIFF
--- a/doc/src/lexsyntax.md
+++ b/doc/src/lexsyntax.md
@@ -1,0 +1,33 @@
+The syntax of the lrlex `.l` format, aims to be familiar to the format used by posix lex.
+It uses the same basic structure as lex.
+
+```
+Definitions
+%%
+Rules
+%%
+User Subroutines
+```
+
+## Definitions
+
+Within the definitions section, you can add an option `%grmtools` section
+documented in [extensions](./lexextensions.md)
+
+## Rules 
+
+Each rule is given by the following elements in sequence:
+
+1. Optional Start State, a name given between angle brackets for example `<Start_State>` documented in [Start States](./start_states.md)
+2. Regex, syntax defined by the rust [regex](https://docs.rs/regex) crate with optional escaping for any character.
+3. Separator space, any horizontal space character in the unicode `Pattern_White_Space` character set
+4. Optional State operator, given between angle brackets documented in [Start States](./start_states.md)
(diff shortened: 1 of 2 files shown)


doc/src/lexsyntax.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Definitions
+
+### Optional %grmtools section
+
+Defined in [extensions](./lexextensions.md)
+
+## Rules 
+
+Each rule is given by
```
ltratt: I think this should end in `:` and the following bits be an itemised list? Though "State operator" may well be best left in a subsubsection "State operator\n\nOptionally, a start state can be ..." etc.

doc/src/lexsyntax.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Token Name
+
+Token name may contain any non-space character between double quotes, including double quotes
+
+### End of line
+
+The end of line finishes a rule.
+
+## UserSubroutines
```
ltratt: Missing space?

ltratt (commented):
With a couple of quick edits, I think getting this in one would be good!

doc/src/lexsyntax.md
ratmice: I updated some text too, but I tried leaving state operator as an itemized list.
It seemed okay with the list being in a paragraph? Let me know what you think.

ltratt:
Thanks! Please squash.

ratmice:
Squashed.

PR #639, opened 2026-05-13
Stop emitting double `to_string()`

hseg:
d70b2942 wrapped `regex` in a `QuoteToString` whose `ToTokens` implementation nominally handles the necessary `to_string()` invocation prior to splicing. However, it didn't make the parallel change to drop the call to `to_string` from the splice. This results in a redundant `regex.to_string().to_string()` in the generated code.
A minor tweak, with potentially negligible but positive impact if the compiler isn't clever enough. Mostly motivated on my part because the import model for the generated parsers (`include!`ing them directly in the code) makes it impossible to get Clippy to shut up about this.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -823,7 +823,7 @@ where
                     // We cannot `impl ToToken for Rule` because `Rule` never stores `lex_flags`,
                     // Thus we reference the local lex_flags variable bound earlier.
                     quote! {
-                        Rule::new(::lrlex::unstable_api::InternalPublicApi, #tok_id, #n, #n_span, #regex.to_string(),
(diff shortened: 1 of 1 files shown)


ratmice:
Thanks, good catch!

> Mostly motivated on my part because the import model for the generated parsers (`include!`ing them directly in the code) makes it impossible to get Clippy to shut up about this.

I just wanted to note that I've also run into this sort of problem in the past, at that time it was possible to work around with the following kind of mod/use workaround code, I haven't tested this in a long time though.

```
mod __foo__ {
  #![allow(clippy::...)]
  include!(...)
}
use __foo__::*;
```

hseg:
I tried doing that, but I kept getting complaints that the inner module is private, and so inaccessible outside the wrapper.

ratmice:
Oh, indeed I think it would be `pub use __foo__::*;` but as I said I haven't tested this in a long long time I could see some visibility lint getting in the way perhaps.

PR #642, opened 2026-07-12
Clippy fixes

ratmice:
The first patch is hand written, so that the subsequent `cargo clippy --fix` won't introduce compilation errors with the borrow checker.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -594,10 +594,10 @@ impl<T: Clone> TryFrom<&Value<T>> for YaccKind {
                 namespace,
                 member: (yk_value, yk_value_loc),
             })) => {
-                if let Some((ns, ns_loc)) = namespace {
-                    if ns != "yacckind" {
-                        err_locs.push(ns_loc.clone());
-                    }
+                if let Some((ns, ns_loc)) = namespace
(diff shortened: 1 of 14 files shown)


ratmice:
Perhaps rather than adding the trait impl, instead I should put more effort into rewriting the code block?

Maybe a `match parsed_header` block would fix it because it currently errors later when we `parsed_header.unwrap()`, and when using match the compiler knows both arms are disjoint.

Edit: That will work, but it *is* a 200+ line `Ok(...)`  block for the rest of the function, which the `unwrap` was preventing indentation of. However the clippy `if let ... && ... ` has reduced the indentation on the inner block significantly, so it really doesn't too terribly.

I went ahead and pushed that, happy to back it out though...

ltratt:
I think this is good. Do you want to squash?

ratmice:
I went ahead and squashed, I'd rather keep the clippy/manual rewrites separate, but it seemed like a pain to do.
(Easier to just rewrite the whole patch if we want to preserve that)

ltratt:
Hmm, we've hit a now-classic GitHub problem and the only way I know how to resolve it is: can you amend the last commit (e.g. just update the time) and force push? [I wish I knew a better way to fix this...]

ratmice:
Done.

PR #643, opened 2026-07-12
Fix #641

### LINKED ISSUE #641: MarkMap::merge_from not always preserving marks on merge
Was taking a breather this weekend, from some other projects. Was looking into various `#[allow()]` lints.

I noticed two, (three?) issues with the `MarkMap::merge_from` implementation that I wrote.
These unlikely to cause any effect in grmtools, as grmtools always sets marks like required values on the "ours" map then merges a theirs map where the theirs map never has any marks to preserve. 

https://github.com/softdevteam/grmtools/blob/master/cfgrammar/src/lib/markmap.rs#L437-L445

1. in the first if statement in the highlighted code doesn't call, so isn't preserving marks for exclusive merge.

```
*my_mark = their_mark;
```

2. The second issue is that marks can be set on keys, without their presence in the markmap. e.g.

```
let mut m1 = MarkMap::new();
m1.mark_required(&"foo")
let mut m2 = MarkMap::new();
m1.mark_required(&"bar")
m2.merge_from(m1);
assert!(m2.is_required(&"foo") && m2.is_required(&"bar"));
```

3. We currently override marks entirely with `=` assignment.

It isn't clear that there is a single correct merge behavior for all keys. These various marks get merged together via bitwise operators, so we should be using bitwise operators to preserve the marks on the original map probably.

MarkMap when I originally wrote it was intended as a data structure for rho polymorphic record types, and in that situation it makes sense some marks need to be preserved.  But there are probably other marks that should not be preserved.
In the rho polymorphic record case you probably want a mark for the opposite of required `x.prohibited(&"baz")`, and there it may not make sense that `x.merge_from(y)` brings in the prohibitions from `y`, since `x` itself may contain a `baz`, and the prohibition is to ensure that they can be merged without conflict. So it doesn't actually make sense to prohibit it, part of the problem here is that this is working based on mutation rather than creating a new merged record.

ratmice:
This should everything in #641 which is in the grmtools impl, the 3rd issue isn't present because I don't think there are any marks in grmtools that shouldn't carry over the only marks are (required, and used), it doesn't have one like prohibited which shouldn't carry over to the merge destination.

It isn't toally clear that `used` should carry over, but at the very least, the behavior is now consistent across merge behaviors, if we merge a value we now also merge it's mark. While keeping the merge target's marks intact, and it's merge behavior intact.

This also gets rid of the `#[allow(clippy::unnecessary_unwrap)]` I was looking at.

### DIFF
--- a/cfgrammar/src/lib/markmap.rs
+++ b/cfgrammar/src/lib/markmap.rs
@@ -420,28 +420,42 @@ impl<K: Ord + Clone, V> MarkMap<K, V> {
                     let ours_mark = Mark::MergeBehavior(MergeBehavior::Ours).repr();
                     let exclusive_mark =
                         Mark::MergeBehavior(MergeBehavior::MutuallyExclusive).repr();
+                    // Their marks with their merge behavior stripped.
+                    let mut just_their_marks = their_mark;
+                    let merge_reprs = Mark::MergeBehavior(MergeBehavior::MutuallyExclusive).repr()
+                        | Mark::MergeBehavior(MergeBehavior::Ours).repr()
+                        | Mark::MergeBehavior(MergeBehavior::Theirs).repr();
+                    // Zap just the MergeBehavior bits.
+                    just_their_marks ^= just_their_marks & merge_reprs;
                     let mut merge_behavior = (*my_mark & exclusive_mark)
                         | (*my_mark & ours_mark)
                         | (*my_mark & theirs_mark);
                     if merge_behavior == 0 {
                         merge_behavior = Mark::MergeBehavior(self.default_merge_behavior).repr();
                     }
-                    if merge_behavior == exclusive_mark {
-                        // If only clippy could convince me and the borrow checker this is actually unnecessary.
-                        #[allow(clippy::unnecessary_unwrap)]
-                        if my_val.is_some() && their_val.is_some() {
-                            return Err(MergeError::Exclusivity(
-                                their_key,
-                                Box::new(their_val.unwrap().into()),
-                            ));
-                        } else if my_val.is_none() {
+                    match their_val {
+                        Some(their_val) if merge_behavior == exclusive_mark => {
+                            if my_val.is_some() {
+                                return Err(MergeError::Exclusivity(
+                                    their_key,
+                                    Box::new(their_val.into()),
+                                ));
+                            } else {
+                                *my_mark |= just_their_marks;
+                                *my_val = Some(their_val.into());
+                            }
+                        }
+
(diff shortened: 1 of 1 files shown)


ratmice:
I don't think this one should be conflict with the clippy PR, as I cherry-picked it from a branch based off of that.

cfgrammar/src/lib/markmap.rs
```diff
@@ -420,28 +420,42 @@ impl<K: Ord + Clone, V> MarkMap<K, V> {
                     let ours_mark = Mark::MergeBehavior(MergeBehavior::Ours).repr();
                     let exclusive_mark =
                         Mark::MergeBehavior(MergeBehavior::MutuallyExclusive).repr();
+                    // Their marks with their merge behavior stripped.
+                    let mut just_their_marks = their_mark;
+                    let merge_reprs = Mark::MergeBehavior(MergeBehavior::MutuallyExclusive).repr()
+                        | Mark::MergeBehavior(MergeBehavior::Ours).repr()
+                        | Mark::MergeBehavior(MergeBehavior::Theirs).repr();
+                    // Zap just the MergeBehavior bits.
```
ratmice, line 428: It would probably be better to move the merge behavior out of the mark bits, 
A better design would be to have two `u8` fields to replace the current `u16`.

Mark's which will get merged into the merge target would in one of these, and marks which
would not could go in the other.

(So, merge behaviors would go into the ephemeral pile, then we wouldn't have to strip them out)

We'd need to mess with the internal representation a little, but it allow us to remove a lot of bit fiddling.
Should at least make things a bit easier to reason about?

ratmice: Woke up with a small question of whether stripping merge behavior is always correct.

It's certainly incorrect to always set it via an assignment operator (overwriting any merge bits or marks in the merge target).
It's also incorrect to always bitwise or it, this possibly sets multiple merge behaviors for the merge target.

In this patch we never propagate merge bits. However there could be an argument for propagating them if the target has no merge behavior set.

This patch is still an improvement in that at least we're not knowingly doing incorrect things. even if we really haven't settled on all aspects of what it should be doing.

ltratt: I'm not quite sure what perfection looks like here, but this is at least an improvement, so I think we should merge now, and then consider perfecting in a subsequent PR?

ratmice: Yeah I agree, also there isn't really a place where this code path gets used by anything currently.
So this is just one step closer towards that (perfecting might be a bit strong though)

PR #644, opened 2026-07-12
Remove an #[allow(unnecessary_unwrap)]

ratmice:
This just attempts to remove some allow lints, but slightly modifying the code to not need it anymore.
I'm not sure if this one will conflict with the clippy patch. If it does I will rebase one or the other not a big deal.


ratmice:
rustfmt & squashed that.

ratmice:
This ran fine with my old version of `cargo-deny`, until I upgraded,
`cargo-deny check license` appears to no longer work, and the correct command appears to be pluralized
`cargo-deny check licenses`, I'll make a PR for CI

PR #645, opened 2026-07-12
Previously cargo-deny accepted 'check license' now it only seems to accept 'check licenses'

ratmice:
This should fix the CI license check with newer versions of cargo-deny.
I couldn't find a ChangeLog entry/bug report/PR in cargo-deny that indicated this change though.


ltratt:
Good spot: I've seen this one in another repo too (didn't think of this one!).

PR #646, opened 2026-07-16
Since bincode is unmaintained use wincode instead.

### LINKED ISSUE #616: Bincode dependency now unmaintained
In the last day or so, the bincode crate has been marked as unmaintained.
The situation is described in the crates.io readme https://crates.io/crates/bincode

Since the code generated by grmtools uses bincode, I'm struggling to see any
way to resolve this without affecting semver (short of new maintainers for the bincode crate).
That seems very unfortunate, since it wasn't that long ago we did a semver breaking release (and bincode reached 2.0).

I don't know of any issues, bugs or otherwise with the current bincode library
that affect grmtools or would otherwise make this an urgent issue other than the maintenance status.
I am not well informed but I believe the current maintenance status is the result of a doxxing that was preceded by the mistaken suspicion of a supply chain attack.

ratmice:
This is the final patch in the series, this should fix #616. Serialization with our complex bounds is quite a bit less straight forward with `wincode` than it was with `bincode`. Because the `Configuration` is parameterized in the serialization bounds.
Thus we end up having to put bounds on multiple serialization configurations in order to serialize data with that configuration.
At least that was the only way I could figure out how to implement it.

This doesn't provide any `CTParserBuilder` mechanisms for configuring the encoding used, so it currently always defaults to the `VarInt` encoding, but the plumbing is all there to do so.

This is part of a series of patches to various crates moving from the bincode crate to the wincode crate.
I believe the reverse order of dependencies is:

1. packedvec
2. sparsevec
3. vob
4. grmtools

I'm currently leaving how we want to deal with semver up in the air. Since these do change trait impls around,
It may be that as we release these we need to fiddle with the version numbers in `Cargo.toml` too.


lrpar/src/lib/ctbuilder.rs
```diff
@@ -1198,19 +1241,48 @@ where
...
 
             fn __lrpar_parser_data() -> &'static ::lrpar::ParserData<#storaget> {
                 static DATA: ::std::sync::OnceLock<::lrpar::ParserData<#storaget>>
                     = ::std::sync::OnceLock::new();
-                DATA.get_or_init(
-                    || ::lrpar::ctbuilder::_reconstitute(__GRM_DATA, __STABLE_DATA)
-                )
+                // We have to call reconstitute like this because the config parameter takes a trait
+                // which uses const generics. Thus the two config parameters here are not actually of the same type.
+                match __ENCODING_CONFIG {
```
ratmice: Since in https://github.com/softdevteam/grmtools/pull/631 this was calling this function in hot loop,
I should try and put this match statement in the closure body I think.

ratmice: Should be fixed in [d74ca12](https://github.com/softdevteam/grmtools/pull/646/commits/d74ca121fdf85cc4e212a84cde0c5c29ee986fe7)

ratmice:
This was tested with local versions of all the crates using the following Cargo.toml

```
diff --git a/Cargo.toml b/Cargo.toml
index c1f3ce15..662a5ae8 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,8 @@ proc-macro2 = "1.0"
 prettyplease = "0.2.31"
 syn = "2.0"
 yaml-rust2 = "0.10.1"
+
+[patch.crates-io]
+vob = {path = "../vob", version = "3.0.4"}
+sparsevec = {path = "../sparsevec", version = "0.2.2"}
+packedvec = {path = "../packedvec", version = "1.2"}
```

lrlex/src/lib/ctbuilder.rs
```diff
@@ -257,12 +257,28 @@ impl CTLexerBuilder<'_, DefaultLexerTypes<u32>> {
...
 where
-    LexerTypesT::StorageT:
-        'static + Debug + Eq + Hash + PrimInt + Encode + TryFrom<usize> + Unsigned + ToTokens,
+    LexerTypesT::StorageT: 'static
+        + Debug
+        + Eq
+        + Hash
+        + PrimInt
+        + SchemaWrite<FixIntConfig, Src = LexerTypesT::StorageT>
+        + SchemaWrite<VarIntConfig, Src = LexerTypesT::StorageT>
```
ratmice, line 278: It is a bit unfortunate, but as far as I can tell it doesn't appear that we can just do something like:

```
LexerTypesT::StorageT: SchemaWrite<Config, Src = LexerTypesT::StorageT> + ...;
Config: for<c: wincode::config::Config>, c,
```

Because higher ranked trait bounds are limited to lifetimes.
`Config` here is also not dyn compatible because of the associated types and consts.
So if there are other wincode configurations we do want to try encoding with afaik we have to add a bounds
here for each of those.

ratmice:
Working on the follow up commit that configures encoding between variable, and fixed...

lrpar/src/lib/ctbuilder.rs
```diff
@@ -255,15 +276,32 @@ where
             ) -> Result<(), Box<dyn Error>>,
         >,
     >,
+    encoding_config: EncodingConfiguration,
```
ratmice: `Option<EncodingConfiguration>`.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -305,6 +343,7 @@ where
             visibility: Visibility::Private,
             rust_edition: RustEdition::Rust2021,
             inspect_rt: None,
+            encoding_config: EncodingConfiguration::VariableSizedInteger,
```
ratmice: Should be `None` at this point like `yacckind` and `recoverer`, above.
Otherwise we can't differentiate between the default value, and user having set it.

ratmice (commented):
Still needs some work to prepare it for follow up patches.

ltratt:
@ratmice Let me handle vob and sparsevec to save you the effort,

ratmice:
Woops, think I might have already pushed to my branch, but feel free.

lrpar/src/lib/ctbuilder.rs
ratmice: I went ahead and fixed this in [f0da9ba](https://github.com/softdevteam/grmtools/pull/646/commits/f0da9ba996cc42a00d4dcaa21c79377ed4de4a8e)

Which contains the whole configuration scheme, I did some looking at the generated sources,
and the variable in encoding files are both about half the size of the fixed int encodings,
but the fixed int encodings should should load faster. 

If we don't want that, parts of the first patch can be simplified (Like the multiple Schema bounds on different configurations)

ratmice:
:crossed_fingers: other than version bumping, I think this one is ready for being looked at whenever you've got time.

ltratt:
I'm working my way through. I must admit that I didn't anticipate wincode having quite such an impact! Still, this is definitely the right thing to do.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -257,12 +257,28 @@ impl CTLexerBuilder<'_, DefaultLexerTypes<u32>> {
         CTLexerBuilder::<DefaultLexerTypes<u32>>::new_with_lexemet()
     }
 }
-
+type FixIntConfig = wincode::config::Configuration;
+
+type VarIntConfig = wincode::config::Configuration<
```
ltratt, line 262: I wonder if it's worth documenting that the only difference between `FixIntConfig` and `VarIntConfig` is the final type parameter? I had to look that up to be sure, and I bet I'll soon forget!

lrpar/src/lib/ctbuilder.rs
```diff
@@ -220,6 +223,151 @@ impl Visibility {
         }
     }
 }
+#[non_exhaustive]
+#[derive(SchemaRead, SchemaWrite, Debug, Clone, Copy)]
+pub enum EncodingConfiguration {
+    FixedSizeInteger,
```
ltratt, line 234: These names are fairly intuitive, but a brief doc hint might be worthwhile. Perhaps we should just literally point people at the wincode documentation for each (e.g. https://docs.rs/wincode/latest/wincode/int_encoding/struct.VarInt.html)? Actually, I naively didn't expect quite so many caveats with `VarInt` (though they're all reasonable upon reflection)!

lrpar/src/lib/ctbuilder.rs
```diff
@@ -220,6 +223,151 @@ impl Visibility {
...
+    }
+}
+
+impl<T: Clone + Debug> TryFrom<&Value<T>> for EncodingConfiguration {
+    type Error = HeaderError<T>;
+    fn try_from(value: &Value<T>) -> Result<EncodingConfiguration, HeaderError<T>> {
+        let mut err_locs = Vec::new();
+        match value {
+            // Someone has tried to `%grmtools{ encoding_config: false }`
+            Value::Flag(_, loc) => Err(HeaderError {
```
ltratt: I wondered if it might be easier to put all the error cases into one and say something like "EncodingConfiguration takes a string". Expanding out the cases like this is a bit of a code reading burden.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -220,6 +223,151 @@ impl Visibility {
...
+                    })
+                }
+            }
+        }
+    }
+}
+
+// We export this for generated code to refer to.
+#[doc(hidden)]
+pub use wincode;
```
ltratt, line 316: Dumb question: could the generated code refer to `::wincode` directly? I must admit that I've forgotten what our convention here is...

ltratt:
I have a few minor comments and one bikeshed: do we think it's obvious what "encoding_config" will be to users? I'm somewhat neutral on this: I did wonder if "serialisation_format" (ok, ok, I could live with a "z" in there :)) might be clearer, but I'm not sure.

lrpar/src/lib/ctbuilder.rs
ratmice: Not a dumb question all, this tripped me up when I was writing it.
We actually can't without requiring that users depend upon grmtools,
This export is for the two lines here.
```
1435:                                ::lrpar::ctbuilder::_reconstitute(__GRM_DATA, __STABLE_DATA, ::lrpar::ctbuilder::wincode::config::Configuration::default().with_fixint_encoding())
1438:                                ::lrpar::ctbuilder::_reconstitute(__GRM_DATA, __STABLE_DATA, ::lrpar::ctbuilder::wincode::config::Configuration::default().with_varint_encoding())
```

Here we know the project building this generated code depends upon lrpar, but not wincode.

ltratt: Right, that makes sense. Sorry!

lrpar/src/lib/ctbuilder.rs
ratmice: Totally, this code was basically just copy/paste from `cfgrammar::header` module, I *think* what it's trying to do is provide more precise errors with line numbers but it reads incredibly badly.

I for one didn't find the errors helpful when I was debugging. But I was also making internal errors that should not be possible.

Maybe we can add a function like `Value::primary_span()`, function
then have `val => make_some_error_based_on(val.primary_span())`

ltratt: That sounds like a good idea to me!

ltratt:
I think all of the necessary dependencies are now on crates.io.

lrpar/src/lib/ctbuilder.rs
ratmice: I tried to clean this up in [af75298](https://github.com/softdevteam/grmtools/pull/646/commits/af75298ce8509f60a9d7b26f8f5bb001946e5bb7)

ratmice:
I went with SerialisationFormat in [2822302](https://github.com/softdevteam/grmtools/pull/646/commits/28223026af42f65af5b4a59fdc20ae25359ab08d) so at least we're consistently british spelling!

lrpar/src/lib/ctbuilder.rs
ratmice: Should also be fixed in [af75298](https://github.com/softdevteam/grmtools/pull/646/commits/af75298ce8509f60a9d7b26f8f5bb001946e5bb7)

lrlex/src/lib/ctbuilder.rs
ratmice: Should hopefully also be fixed in [af75298](https://github.com/softdevteam/grmtools/pull/646/commits/af75298ce8509f60a9d7b26f8f5bb001946e5bb7)

cfgrammar/src/lib/header.rs
```diff
@@ -210,6 +210,32 @@ impl From<Value<Span>> for Value<Location> {
     }
 }
 
+impl<T> Value<T> {
+    pub fn primary_location(&self) -> &T {
```
ratmice, line 214: This is a pretty awkward API, the thing going on is that we've parameterized between
logical locations (command line arguments in nimbleparse, builder parameters in CTParserBuilder, and source locations read from %grmtools sections).

In the first two cases we have `Value<Span>` while in the second we have `Value<Location>`.

We *could* return a `Vec<T> where T: Clone` here, but in all the `Value<Location>` areas, this is going to be extremely unhelpful, and in the `Span` case, `PrimaryLocation` will still return at least one relevant location.

So we may drop some information regarding source locations when dealing with `Span`,
but we also avoid emitting redundant locations, when all those locations are like `CommandLine`

ratmice: So, it's kind of like we're either going to have this information loss, where we drop "secondary" locations,
or we're going to have a bunch of redundancy. So I think picking one of the locations is ok.

cfgrammar/src/lib/header.rs
```diff
@@ -676,6 +673,10 @@ impl<T: Clone> TryFrom<&Value<T>> for YaccKind {
                     })
                 }
             }
+            val => Err(HeaderError {
+                kind: HeaderErrorKind::ConversionError("From<YaccKind>", "Cannot to YaccKind"),
```
ratmice: Hmm, the existing code says `From<YaccKind>` however it's actually `TryFrom<Value<T>> for YaccKind`.
So it looks like this needs a bit of work at least on error presentation.

Ditto for the `SerializationFormat` code below.

ratmice: In [b8d30c1](https://github.com/softdevteam/grmtools/pull/646/commits/b8d30c11790207fbf103521e8df1622c0b7eb16c) it should at least have the benefit of being accurate while not being super informative about why exactly it is invalid.

ratmice:
Also let me know if you want me to split this one back into two separate PRs.
I just didn't want to get into a situation where some choice made in the first patch meant we had to make some incompatible change to move forward on making the serialization_fromat configurable.

I'm pretty sure though now that isn't the case.

cfgrammar/src/lib/header.rs
ltratt: Since we've got a major version bump coming up, we could consider an API change if it would help?

ltratt:
> I went with SerialisationFormat in [2822302](https://github.com/softdevteam/grmtools/pull/646/commits/28223026af42f65af5b4a59fdc20ae25359ab08d) so at least we're consistently british spelling!

Sorry :)

cfgrammar/src/lib/header.rs
ltratt: Works for me: I think it's better than having the whole type name in (the user will work that out soon enough from rustc's type messages).

ltratt:
I'm fine with this as one PR.

cfgrammar/src/lib/header.rs
ratmice: I hadn't given it a whole lot of thought, In that the whole mixing of spans and logical error locations is kind of new territory for me, so I've kind of been winging it. One thing we could consider is the way that the AST `Value<T>` is currently an enum which has the pair of e.g. `String`, and `Span`, but also `String`, and `Location`.

We could consider making it a 

```
enum Setting {
   Num(u64, usize)
   Flag(bool, usize),
   Array(Vec<Setting>, usize, usize)
}
struct ValueRepr<T> {
    value: Value,
    locations: Vec<T>   
}
```

With that the `Location::CommandLine` and `Location::Other("From<...>")` could just push one location on the `locations` stack, and we'd use the `usize` indexes to point to that one location.

With spans, each of these could refer to different locations on the `locations` stack.

I *think* one of the questions is whether we could get the lrpar/cttests/grmtools_section.test to emit that structure.
Because we test that the AST structure is exactly the same for the handwritten parser, and the generated parser in the testsuite. It's hard for me to imagine how this structure representation could be emitted from the generated parser's rules. I'll try and give it a think overnight.

ltratt: Hmm, perhaps the current muddle is the best we can do right now.

ratmice: Well, I did think of one more place that could be cleaned up in [e8a5dab](https://github.com/softdevteam/grmtools/pull/646/commits/e8a5dab4a2a861666c96a1ac19029053464d8233), Edit: the other thing to note is that this *is* all `#[doc(hidden)]` too.

ltratt: Good point!

ltratt:

ratmice:
Squashed, but unfortunately while I was almost done rebasing, I noticed a pesky line that shouldn't be there (but was actually inconsequential because of the way `quote` encodes `Option` types as their contents, and `None` as empty strings)

Plus I apparently accidentally rebased on top of current master, rather than what I originally based the patch on.
So there are a few changes.

Edit: Actually I may have introduced one issue in the squash accidentally.

ratmice:
Alright, squashed correctly this time.

ltratt:
Before we consider making a release: since we've got a major release bump coming, is there anything else breaking we want to include?

ratmice:
I'm pretty much heads down in another project, and don't really have any API breaking changes that would be helpful paged in.
One thing that comes to mind is a way to run the codegen/builders and say capture the output to a string, or `Write` trait.
So you could then run the code through something like rust-analyzer, or static analysis, from a .y file passed as a string
rather than as a filename.

With the intent of eventually having rust-analyzer `cargo check` the embedded actions as you're editing yacc files.
This is kind of where I was heading when I started cleaning up the codegen process using the `quote` crate.
But never got started on that aspect of it. That seems like something to consider.

I don't know how the rust-analyzer aspects will work, but we probably don't need that figured out to get grmtools in shape for it.

ltratt:
> One thing that comes to mind is a way to run the codegen/builders and say capture the output to a string, or Write trait.

I would love this! Do you think it would require breaking changes? Do you have a rough sense fo how one might go about it?

ratmice:
My hope was to try and do it without breaking changes, by adding a new module, `codegen.rs`,
It would largely be taking the existing `ctbuilder.rs`, from the `build()` on down, but removing the filesystem interactions.
Then replacing the builder implementation with calls into the codegen module, so the existing builder would go:

```
let input_yacc_str = input_file.read(..)
let mut output_yacc_rs_str = String::new();
codegen::generate_yacc_rust(input_yacc_str, &mut output_yacc_rs);
output_file.write(output_yacc_rs_str)
```

If only to avoid introducing more API surface to CTParserBuilder, which it seems good to keep as the primary API for cargo/`build.rs`. I *think* it's mostly just a refactor, but not totally positive. 

Edit:
There are probably a lot of things I'm not taking into account, like all the configuration methods the builder provides,
but those can largely be configured in the yacc file now, I'm not sure how to do it without making the `Header` datatype more public.

One thing we could do is consider a separate builder-like API *just* for configuring, which would largely mirror the configuration aspects of `CTParserBuilder`. So we could split it into two modules (builder settings, and codegen) then CTParserBuilder would just be a wrapper around these that adds cargo + filesystem interaction. (So essentially builder settings struct would just be for hiding the `Header` struct, and providing the current builder configuration to the codegen phase)

ratmice:
Let me try a more concrete thought, and a less wordy one...

Currently `CTParserBuilder`

```
pub struct CTParserBuilder<'a, LexerTypesT: LexerTypes>
where
    LexerTypesT::StorageT: Eq + Hash,
    usize: AsPrimitive<LexerTypesT::StorageT>,
{
    // Anything stored in here (except `output_path`, `conflicts`, and `error_on_conflict`) almost
    // certainly needs to be included as part of the rebuild_cache function below so that, if it's
    // changed, the grammar is rebuilt.
    grammar_path: Option<PathBuf>,
    // If specified rather than reading source from `grammar_path`, use this string directly
    grammar_src: Option<String>,
    // If specified along with `grammar_src`, use this rather than building an ast from `grammar_src`.
    from_ast: Option<ASTWithValidityInfo>,
    output_path: Option<PathBuf>,
    mod_name: Option<&'a str>,
    recoverer: Option<RecoveryKind>,
    yacckind: Option<YaccKind>,
    error_on_conflicts: bool,
    warnings_are_errors: bool,
    show_warnings: bool,
    visibility: Visibility,
    rust_edition: RustEdition,
    inspect_rt: Option<
        Box<
            dyn for<'b> FnMut(
                &'b mut Header<Location>,
                RTParserBuilder<LexerTypesT::StorageT, LexerTypesT>,
                &'b HashMap<String, LexerTypesT::StorageT>,
                &PathBuf,
            ) -> Result<(), Box<dyn Error>>,
        >,
    >,
    serialisation_format: Option<SerialisationFormat>,
    // test function for inspecting private state
    #[cfg(test)]
    inspect_callback: Option<Box<dyn Fn(RecoveryKind) -> Result<(), Box<dyn Error>>>>,
    phantom: PhantomData<LexerTypesT>,
}
```

The thought would be move many of data these fields to their own struct
```
struct CTBuilderSettings {
    mod_name: Option<&'a str>,
    recoverer: Option<RecoveryKind>,
    yacckind: Option<YaccKind>,
    error_on_conflicts: bool,
    warnings_are_errors: bool,
    show_warnings: bool,
    visibility: Visibility,
    rust_edition: RustEdition,
    serialisation_format: Option<SerialisationFormat>,
};
```

Then CTParserBuilder could own one of those
```
struct CTParserBuilder {
  settings: CTBuilderSettings,
  ...
}

fn build() {
   let input_str = read(..)
   let output_file = File::create(...);
   codegen::generate(self.settings, &input_str, &mut output_str);
   output_file.write(output_str)
}
```

I should probably move this line of discussion to it's own issue?

ltratt:
>I should probably move this line of discussion to it's own issue?

Probably a good idea, as I can see this one's going to require a bit of thought!

In terms of breaking out CTBuilderSettings, is the idea that a different struct/function can make use of that too?

ratmice:
Cool, I'll try and open up a new issue, but to respond to your quetsion

> In terms of breaking out CTBuilderSettings, is the idea that a different struct/function can make use of that too?

Yes the key thing to note is that some of the current CTParserBuilder options affect codegen,
Primarily the ones derived from `Regex`, but also the `serialisation_format` we just added.

So if we want to break codegen out into it's own module, the codegen function/method would take the `CTBuilderSettings` as a parameter. So generally the idea is that `CTParserBuilder` would pass a reference to it's collected settings to codegen.
But programs that want to interface with codegen directly for rust-analyser integration can also just construct their own `CTBuilderSettings`.

ltratt:
> So if we want to break codegen out into it's own module, the codegen function/method would take the CTBuilderSettings as a parameter.

Right, got it!

PR #647, opened 2026-07-16
Fix typo, and test in cttests

ratmice:
This test was expecting to run with a recoverer. So fixing the typo, broke the test.
Because of the typo it was falling back to the default (CPCTPlus).

### DIFF
--- a/lrpar/cttests/src/cgen_helper.rs
+++ b/lrpar/cttests/src/cgen_helper.rs
@@ -31,9 +31,10 @@ pub(crate) fn run_test_path<P: AsRef<Path>>(path: P) -> Result<(), Box<dyn std::
             Some(s) => panic!("YaccKind '{}' not supported", s),
             None => None,
         };
-        let recoverer = match docs[0]["revoverer"].as_str() {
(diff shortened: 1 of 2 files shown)


lrpar/cttests/src/cgen_helper.rs
```diff
@@ -31,7 +31,7 @@ pub(crate) fn run_test_path<P: AsRef<Path>>(path: P) -> Result<(), Box<dyn std::
             Some(s) => panic!("YaccKind '{}' not supported", s),
             None => None,
         };
-        let recoverer = match docs[0]["revoverer"].as_str() {
+        let recoverer = match docs[0]["recoverer"].as_str() {
```
ratmice, line 34: I guess we should learn our lesson here and add a `Some(_) => panic!("Unknown recovery kind")`

ratmice: Should be fixed in 72afa11

I didn't see anything else that wasn't checking for unrecognized `Some(_)` values.

ltratt:
Good find!

Issue #191, opened 2020-05-30
Expose more than one rule?


utkarshkukreti:
Question / Feature Request: Is there any way to parse a specific rule as the starting parser? For example, if I have:

```
%start Expr
%%
Expr -> ...;

Int -> ...;
%%
```

I also want to be able to parse a string as `Int`, not just `Expr`.

(I'm trying to port my parser from LALRPOP to lrpar (mainly because of the operator precedence feature) which exposes a parser for any rule prefixed with the keyword `pub`.)

ltratt:
I must admit, this isn't a feature that I'd thought about. At first glance, it seems an awkward fit with LR parsing, in the sense that, at least conceptually, you need 1 statetable per start rule. However, in practise I think you might be able to use a single statetable and "emulate" the *accept* state for an arbitrary start rule such that you can get away with a single statetable. However, I might be wrong about that. LALRPOP, uses the [lane table algorithm](http://smallcultfollowing.com/babysteps/blog/2017/03/17/the-lane-table-algorithm/), which might or might not make this stuff easier -- I haven't familiarised myself with the algorithm. @nikomatsakis might have a thought or two on this.

So, at the moment, unfortunately, your only option is to duplicate the Yacc grammar for each start rule. However, what I can fairly easily do is remove some of the assumptions the grmtools libraries have about start states. That won't get us all of the way to the feature you're asking for, but it will make it easier for someone else to implement it -- and I hope they do, because this is the sort of feature that I think grmtools should be flexible enough to accommodate!

nikomatsakis:
I don't think Lane Table helps in particular but I've not given it a lot of thought. I think LALRPOP permits you to tag multiple rules as `pub` but IIRC it just generates separate parsers for each one, there is no shared code or state. (I could be misremembering.)

ltratt:
Sorry @nikomatsakis for not saying thanks for your comment (I'm only nearly a year late)!

Coming back to this one with the benefit of hindsight, I think that I'd be fine with generating 1 parser per start rule at first. It's probably suboptimal, but I'm fine with getting the functionality in and working out how to make it more efficient in the future.

AFAICS neither Yacc nor Bison supports this, so that gives us both the freedom to do what we want, but also the difficulty of sifting amongst the design choices. My first thought is that I think it would be reasonable to allow the `%start` rule to specify more than 1 start rules (e.g. `%start R1 R2`).

[...]

PR #649, opened 2026-07-19
Add docs about multiple start rules.

ratmice:
Here is a first attempt at documenting the current ability to perform the hack for faking multiple start rules by jump starting the lexer with a synthesized token.

When doing this I noticed some issues in the Hand-written lexer section, but will handle those in a separate patch.
I don't know if the progression I used here is good, e.g. it goes from the hard implementation explaining the underlying details, to the easier one.

But it was just a first attempt I suppose.

### DIFF
--- a/doc/src/parsing_startrules.md
+++ b/doc/src/parsing_startrules.md
@@ -1,0 +1,148 @@
+# Multiple start rules
+
+lrpar parsers start with a single start rule, this chapter discusses
+the case where you want _multiple_ start rules.
+
+## Simulating multiple start rules
+
+An approach to simulating multiple start rules from a single start rule
+is to synthesize tokens not produced by the lexer as initial start tokens.
+Then produce those as the initial token passed to the parser to select between
+parser entry points.
+
+### Initial project
+First create a new project, add the dependencies and build dependencies.
+
+```console
+cargo new multistart
+cd multistart
+cargo add lrlex lrpar --build
+cargo add lrlex lrpar cfgrammar
+```
+
+### Parser
+Here is our grammar `src/multistart.y` the tokens `START_A` and `START_B` are
+our synthetic tokens. While `ARule` and `BRule` are our different start rules
+we want to select between.
+
+```
+%grmtools{yacckind: Grmtools}
+%token START_A START_B
+%start meta_start
+%%
+meta_start -> Result<String, DefaultLexeme> :
+  START_A ARule { $2 }
+| START_B BRule { $2 }
+;
+
+ARule -> Result<String, DefaultLexeme>
+: 'A' { Ok($lexer.span_str($1?.span()).to_string()) }
+| BRule { $1 }
+;
+
+BRule -> Result<String, DefaultLexeme>
+: 'B' { Ok($lexer.span_str($1?.span()).to_string()) }
+;
+%%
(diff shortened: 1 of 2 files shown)


ratmice:
Should link in #484 and #191

doc/src/parsing_startrules.md
```diff
@@ -0,0 +1,248 @@
...
+First we need a lexer. We'll include the `START_A` and `START_B` tokens in the grammar, using rules that
+cannot match any input characters.
+
+`src/multistart.l`
+```
+%%
+A "A"
+B "B"
+[^\p{any}] "START_A"
+[^\p{any}] "START_B"
```
ratmice: These are mostly just to silence the `missing from lexer` warning, it seemed like the easiest way at the time.
There is also `allow_missing_terms_in_lexer`, though that doesn't appear to allow you to say set a list of terms.

Although these regex should be very fast, I suppose their existence in the list of lex rules, still means you're
going to check the empty regex for each character.

I don't know in practice how much of a difference that makes, for however many synthetic start tokens people have, over each unmatched character.  I suppose that if they're at the very end, they'll only match unmatched characters in practice.

As such, I should probably move the next line above these rules.

ratmice: This is at least improved in [c9b377c](https://github.com/softdevteam/grmtools/pull/649/commits/c9b377cf75f0a55bb5fa9bf5673b3381116101c1)

ltratt: > allow_missing_terms_in_lexer

Good point. I wonder if we should add an equivalent of lrpar's `%expect-unused` to lrlex? Arguably we could then remove the Rust API `allow_missing_terms_in_lexer` entirely (or, perhaps, make it a no-op and deprecated for 0.15.0?).

ltratt:
Thanks for this! I think you've got the right technical solution in here, which is the most important thing to get right. What I would probably suggest is that perhaps slightly cut down some of the lead-up and getting to the solution quicker. We can briefly allude to lesser alternatives if we want, but I think for this sort of thing most readers most want to know what we recommend (and less what we don't recommend).

[Part of me can't quite believe that I had never thought of this way of doing things as a possibility! It's a really neat trick.]

doc/src/parsing_startrules.md
ratmice: That sounds like a good idea, I'll try work on this some more mid week.

ratmice:
Yeah, I'm also a bit surprised I've never thought of it either (nor run across it before). I'll work on the text some more as soon as I can.

doc/src/parsing_startrules.md
ratmice: I filed #651 for the new lrlex directive.

ratmice:
Oops, I rebased this on top of that `%expect-missing` branch, but didn't realize how far behind the master branch this was :(

Anyhow I tried to trim down the chapter to the bare minimum.

ltratt:
I think the docs are good but we need to rebase the `%expect-missing` bits out?

ratmice:
Yeah, I don't know what I was thinking the docs don't even link to the library in any way.
I'll just rebase it back onto the current master branch.

ratmice:
Rebased, and squashed it into a single commit.

Issue #611, opened 2025-10-26
User defined keys in grmtools section


ratmice:
Something I avoided putting much thought into during the 0.14.0 development cycle. Currently the `%grmtools` directive is *very* conservative in the sense that the `CTBuilder`s and `nimbleparse` when they read a `%grmtools` section will complain if any of the keys in the section are unrecognized.

There are a number of keys that `nimbleparse_lsp` uses which don't have an equivalent in the `%grmtools` section.
These are currently (and there is no problems keeping it this way) stored in an external `nimbleparse.toml` file.

The keys are `extension` which gives a file extension for the parser input.
`l_file` an `y_file`, one could imagine the lexer having a `parsers` array,  and the parser having a `lexer` string
(where there may be multiple parsers for a lexer, and a single lexer for a parser)

I kind of see 3 options or ways that this could go:

1. Do nothing (keep these in an external file like they currently are in nimbleparse.toml).
2. Add some mechanism for user defined keys.
3. Allow the ad-hoc set of keys to the allowed list of keys even if nimbleparse and the builders don't need them.

This isn't something I'm in any rush to decide, I'm currently trying to rewrite the lsp so that it is more portable to web assembly,
and can be deployed on the web. For at least the first iteration of this development I plan on keeping it to an external .toml file.

Overall i'm somewhat inclined to explore and see if we can get something reasonable for option two.
Because it is impossible to say _what_ keys might be useful, like perhaps entry points for the `clone_and_change_start_rule`
At the same time, we probably don't want user defined keys standing in the way of adding new keys to the grmtools section
in the future.

One common way people do user defined keys (e.g. Cargo.toml/rustdoc) is to have a metadata key (the contents of which becomes a free for all).
Another way, is to restrict the key namespace such that user defined keys can be recognized.

```xyz.y
%grmtools {
  yacckind: Grmtools,
  @nimbleparse_lsp.extension: ".xyz",
  @nimbleparse_lsp.l_file: "xyz.l",
}
```

```xyz.l
%grmtools {
   @nimbleparse_lsp.parsers: ["xyz.y"]
}
```

We could even consider the unadorned keys to be something like: `@grmtools.yacckind`, or `@lrpar.yacckind`, and with that we could continue to complain about unknown keys within the `@grmtools` or `lrpar` namespace, and ignore external keys.
This is completely setting aside though question of how user defined keys are retrieved from the files.

[...]

PR #650, opened 2026-07-20
Prepare a 0.15.0 release.

ltratt:


CHANGES.md
```diff
@@ -1,3 +1,19 @@
...
+## Breaking changes
+
+* Move from `bincode` to `wincode` for serialisation. This also changes the
+  feature name that grmtools exposes, hence the major version bump.
+
+## Other changes
+
+* Use the correct default for `DEFAULT_LEX_FLAGS.case_insensitive`.
+
+* Add documents for the syntax lrlex supports.
```
ratmice: I think this should be `documentation`

CHANGES.md
```diff
@@ -1,3 +1,19 @@
...
+
+* Fix mark merging sometimes not preserving marks.
```
ratmice: I think we can drop this one, grmtools doesn't yet use mark merging,
it was kind of there for https://github.com/softdevteam/grmtools/issues/611

where you have multiple consumers of the %grmtools section, and each consumer
want's to mark their own keys as `used`, so the header checker doesn't complain about unused keys.

But we currently only have one thing doing marking (lrlex, or lrpar, or nimbleparse)
So the dropped merges didn't really have any effect.

ratmice: Anyhow if you do want to keep it, i'd suggest including something can will relate it to:

Marks on keys in the %grmtools section are preserved on merge.

ltratt:
Fixed in the force push (hopefully).

PR #651, opened 2026-07-23
Add lrlex `%expect-missing-tokens` directive.

ratmice:
This adds a lrlex directive `%expect-missing-tokens` for the docs PR #649 

The syntax is like: `%expect-missing-tokens "TOK"` including quotes.

This is so that it behaves exactly like `%expect-unused` which accepts quoted tokens, but unquoted *rules*. Unlike `%expect-unused`, the new directive doesn't take unquoted rules as those are a parser thing.

I tested that the new test fails if we remove the new directive.

### DIFF
--- a/lrlex/src/lib/ctbuilder.rs
+++ b/lrlex/src/lib/ctbuilder.rs
@@ -640,7 +640,7 @@ where
             );
         }
 
-        let (missing_from_lexer, missing_from_parser) = match self.rule_ids_map {
+        let (mut missing_from_lexer, missing_from_parser) = match self.rule_ids_map {
             Some(ref rim) => {
                 // Convert from HashMap<String, _> to HashMap<&str, _>
                 let owned_map = rim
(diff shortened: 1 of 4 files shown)


ltratt:
Quick thought: would it be ambiguous if we shortened this to `%expect-missing` to match `%expect-unused`?

ratmice:
I don't think it would be ambiguous, at worst we might add `%expect-missing` to lrpar someday, and grep might match both. But that said I'm not sure what justification there is for having the parser not handle a token the lexer is capable of producing. But that grepability is kind of spurred me to throw `tokens` on the end. I'm happy to remove it though.

ratmice:
I went ahead and used the shorter name in 80a6b98

ltratt:
This looks good! Please squash.

ratmice:
Squashed, I went ahead and wrote a better commit message, please have a look.

ltratt:
Thanks!

Issue #648, opened 2026-07-18
Separate codegen from filesystem interactions


ratmice:
For doing things like running the codegen process on in-editor yacc and lex files, It would be helpful
if we could run the codegen process passing it e.g. an input string, and perhaps a mutable output string to where codegen writes.

This could hopefully allow us someday to integrate into rust-analyzer and/or run rusts static analysis tools on the embedded actions within
`.y` files. Currently the codegen process assumes it can read the grammar from the filesystem, and then write the generated code to the filesystem. Requiring you to save the files before static analysis can occur.

Below are my ideas on this, it *seems* like it should be possible to do this without breaking `CTParserBuilder`, and `CTLexerBuilder`.
I'm open minded about other ideas, these are just the thoughts that occur to me.

1. Add a module like `lrpar::codegen::CodeGenerator`, and similar for `lrlex`
2. Some of the ctbuilder options in both lrpar (serialisation_format) and lrlex (regex), both (mod_name) affect codegen.
3. Some of the ctbuilder.rs in both affect parsing (yacckind, etc).
4. Other options seem entirely builder specific like `grammar_path`, and `output_path.

To handle 2. and 3. we need some way to pass settings from the Builders to the `CodeGen`. 
The thought is we could stick these into a struct like `CTBuilderSettings`, or `CodeGenSettings`.
`CTParserBuilder` and `CTLexerBuilder` could own one of these, and passing a reference to the code generator.

Things that want to bypass CTParserBuilder and avoid filesystem interaction can construct these directly.
But mostly CTParserBuilder could become a wrapper around these two structures, plus it's own settings and cargo integration behaviors (printing to stderr, environment variables).

ratmice:
This kind of came up since we are making a breaking release, but like I said, I'm not sure it requires any breaking changes.
One thing we could do is an experiment to see how it looks and if it appears not to require breaking changes, it wouldn't have to hold up the release? That way we wouldn't have to be rushed in to settling on an API for it.

ratmice:
Saw the following project which is an interface to the rust compiler which has been compiled to wasm (bjorn3's patches), which is doing both code gen can checking. Once we have a solution for this issue, that will be an interesting thing to explore for both running and checking actions from the web.

https://github.com/AngelOnFira/weblings
https://weblings.forest-anderson.ca/

ltratt:
[...]

PR #652, opened 2026-07-27
Very rough first draft at codegen module

ratmice:
Here is a (not very pretty) first draft which is mostly just pulling code as-is out of `CTLexerBuilder::build()` into smaller, more manageable functions. I don't think anything here is really properly named, `codegen` at times seems like it's doing more parsing/validation than actual code generation (where it never uses it's self type!)

I thought that the splitting of the `CTTokenMapBuilder` turned out okay, however it is quite a bit simpler with less moving parts.

My thoughts are to try and rename the `LexCodegen` structure to something like `LexCodegenBuilder`,
and then try see if we can manage to come up with a good `Self` type for the codegen parts.

Edit:
I haven't tried to make anything `pub` here yet just `pub(crate)`, aiming for close to verbatim code movement.
I do have an idea of what a second draft might look like, we'll see if it works soon...

But so far trying to make e.g. a two pass API for checking then codegen has proven elusive,
Primarily because of data dependencies between the checking phases, and callbacks like `lrpar_config`, I've managed to get some nicer API to compile, but not pass the testsuite.

### DIFF
--- a/lrlex/src/lib/codegen.rs
+++ b/lrlex/src/lib/codegen.rs
@@ -1,0 +1,541 @@
+use crate::{LRNonStreamingLexerDef, LexFlags, LexerDef, Visibility, ctbuilder::LexerKind};
+use cfgrammar::{
+    header::{GrmtoolsSectionParser, HeaderValue},
+    markmap::MarkMap,
+    span::{Location, Span},
+};
+use lrpar::{
+    LexerTypes,
+    diagnostics::{DiagnosticFormatter, SpannedDiagnosticFormatter},
+};
+use proc_macro2::{Ident, TokenStream};
+use quote::{ToTokens, TokenStreamExt, format_ident, quote};
+use regex::Regex;
+use std::{
+    any::type_name,
+    borrow::Borrow,
+    collections::HashMap,
+    error::Error,
+    fmt::{self, Display, Write as _},
+    marker::PhantomData,
+    path::Path,
+    sync::LazyLock,
+};
+
+static RE_TOKEN_ID: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z_][a-zA-Z_0-9]*$").unwrap());
+
+use crate::ctbuilder::ERROR;
+/// The quote impl of `ToTokens` for `Option` prints an empty string for `None`
+/// and the inner value for `Some(inner_value)`.
+///
+/// This wrapper instead emits both `Some` and `None` variants.
+/// See: [quote #20](https://github.com/dtolnay/quote/issues/20)
+struct QuoteOption<T>(Option<T>);
+
+impl<T: ToTokens> ToTokens for QuoteOption<T> {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        tokens.append_all(match self.0 {
+            Some(ref t) => quote! { ::std::option::Option::Some(#t) },
+            None => quote! { ::std::option::Option::None },
+        });
+    }
+}
+
+/// The wrapped `&str` value will be emitted with a call to `to_string()`
+struct QuoteToString<'a>(&'a str);
+
+impl ToTokens for QuoteToString<'_> {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let x = &self.0;
+        tokens.append_all(quote! { #x.to_string() });
+    }
+}
+
+/// This wrapper adds a missing impl of `ToTokens` for tuples.
+/// For a tuple `(a, b)` emits `(a.to_tokens(), b.to_tokens())`
+struct QuoteTuple<T>(T);
+
+impl<A: ToTokens, B: ToTokens> ToTokens for QuoteTuple<(A, B)> {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let (a, b) = &self.0;
+        tokens.append_all(quote!((#a, #b)));
+    }
+}
+
+/// Currently this is kind of a hodge podge of everything between parsing/validation
+/// and code generation, from lex source input strings, to rust source output strings.
+///
+/// This probably needs a better name, as note that the self variable only gets used
+/// for the former parsering/validation stages, and the latter codegen phases are all
+/// implemented through associated functions.
+pub(crate) struct LexCodegen<'a> {
+    src: &'a str,
+    // We store the path here so we can generate a module name from it if needed.
+    // But should never use it for filesystem interaction within this module.
+    path: &'a Path,
+    diagnostics: SpannedDiagnosticFormatter<'a>,
+    header: MarkMap<String, HeaderValue<Location>>,
+}
+
+impl<'a> LexCodegen<'a> {
+    pub(crate) fn new(
+        src: &'a str,
+        path: &'a Path,
+        header: MarkMap<String, HeaderValue<Location>>,
+    ) -> LexCodegen<'a> {
+        let diagnostics = SpannedDiagnosticFormatter::new(src, path);
+        LexCodegen {
+            src,
+            path,
+            header,
+            diagnostics,
+        }
+    }
+
+    pub(crate) fn lex_diag(&self) -> &SpannedDiagnosticFormatter<'a> {
+        &self.diagnostics
+    }
+
+    pub(crate) fn merge_headers(&mut self) -> Result<(), Box<dyn Error>> {
+        let (parsed_header, _) = self.parse_header()?;
+        Ok(self.header.merge_from(parsed_header)?)
+    }
+
+    fn parse_header(&self) -> Result<(MarkMap<String, HeaderValue<Span>>, usize), Box<dyn Error>> {
+        GrmtoolsSectionParser::new(self.src, false)
+            .parse()
+            .map_err(|es| {
+                let mut out = String::new();
+                out.push_str(&format!(
+                    "\n{ERROR}{}\n",
+                    self.lex_diag()
+                        .file_location_msg(" parsing the `%grmtools` section", None)
+                ));
+                for e in es {
+                    out.push_str(&indent(
+                        "     ",
+                        &self.lex_diag().format_error(e).to_string(),
+                    ));
+                    out.push('\n');
+                }
+                ErrorString(out).into()
+            })
+    }
+
+    pub(crate) fn extract_lexerkind(
+        &mut self,
+        lexerkind: Option<LexerKind>,
+    ) -> Result<LexerKind, Box<dyn Error>> {
+        self.header.mark_used(&"lexerkind".to_string());
+        match lexerkind {
+            Some(lexerkind) => Ok(lexerkind),
+            None => {
+                if let Some(HeaderValue(_, lk_val)) = self.header.get("lexerkind") {
+                    Ok(LexerKind::try_from(lk_val)?)
+                } else {
+                    Ok(LexerKind::LRNonStreamingLexer)
+                }
+            }
+        }
+    }
+
+    pub(crate) fn extract_lexerdef<LexerTypesT>(
+        &mut self,
+        lexerkind: &LexerKind,
+    ) -> Result<(LRNonStreamingLexerDef<LexerTypesT>, LexFlags), Box<dyn Error>>
+    where
+        LexerTypesT: LexerTypes,
+        LexerTypesT::StorageT: TryFrom<usize>,
+        usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+    {
+        let (lexerdef, lex_flags): (LRNonStreamingLexerDef<LexerTypesT>, LexFlags) = match lexerkind
+        {
+            LexerKind::LRNonStreamingLexer => {
+                let lex_flags = LexFlags::try_from(&mut self.header)?;
+                let lexerdef =
+                    LRNonStreamingLexerDef::<LexerTypesT>::new_with_options(self.src, lex_flags)
+                        .map_err(|errs| {
+                            let mut out = String::new();
+                            out.push_str(&format!(
+                                "\n{ERROR}{}\n",
+                                self.lex_diag().file_location_msg("", None)
+                            ));
+                            for e in errs {
+                                out.push_str(&indent(
+                                    "     ",
+                                    &self.lex_diag().format_error(e).to_string(),
+                                ));
+                                out.push('\n');
+                            }
+                            ErrorString(out)
+                        })?;
+                let lex_flags = lexerdef.lex_flags().cloned();
+                (lexerdef, lex_flags.unwrap())
+            }
+        };
+        Ok((lexerdef, lex_flags))
+    }
+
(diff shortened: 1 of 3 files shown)


ratmice:
Well, I did manage to get a second draft up and working, and it feels like it's starting to resemble an API...
The data dependencies ended up being the `lexerdef.set_rule_ids_spanned` call, and the `rule_ids_map`.

So the way I tried to work around this was
1. Construct the `LexCodegen` pretty early, before all the elements it's composed of have been checked for validity.
2. Do the checking, and grab mutable references to any elements that need to be updated from the `codegen`.
3. Finally once all the checking and updating has been done, generate code.

So it isn't *exactly* a 2 pass system of validation/generation, but it kind of behaves like one unless you're taking mutable references.

I still haven't made anything `pub` though.
I need to do another pass over the changes, I had added some interim cloning (which ended up causing issues, before switching to mut refs) and need to make sure I got rid of *all* that.

lrlex/src/lib/codegen.rs
```diff
@@ -217,7 +230,59 @@ impl<'a> LexCodegen<'a> {
         Ok(mod_name)
     }
 
-    pub(crate) fn codegen_lex_flags(lex_flags: LexFlags) -> TokenStream {
+    pub(crate) fn build<LexerTypesT>(
+        &mut self,
```
ratmice, line 259: It was only towards the end of getting this working that I realized we need to keep the builder around
after `build()` to call `cgb.lex_diag()` some more.

Thus we should probably come up with a better name than `LexCodegenBuilder` since it isn't exactly the builder pattern anymore. I was thinking maybe `LexCodegenSourceEnv` or something

ratmice: One more issue with this `build` method, is the list of arguments isn't easily extensible in the way a real builder would be. Maybe we should put all these args actually into the struct.

Edit: For this we'll probably need to add the `Clone` trait to `LexerKind`. I'm a bit uncertain about that
In particular when it comes to future expansion with `LRStreamingLexerer` I'd always imagine we would need to pass a data channel in through `LexerKind` using some kind of spsc which wouldn't necessarily need to impl clone?

I'll explore a different option.

ratmice: [ba0891b](https://github.com/softdevteam/grmtools/pull/652/commits/ba0891bbcbad5b4f46cb4428bc082da89e5ed6aa) was my attempt at both these issues,
I renamed `LexCodegenBuilder` to `LexSrcEnv`, and introduced a struct `LexCodegenArgs` for stuff which `LexSrcEnv` takes ownership of but can't easily be owned by it's `Self`.

ratmice:
This is starting to feel like we could start to consider pulling out of draft mode, and discuss making these `pub(crate)` functions `pub`. One thing I'm uncertain of though is whether these structures are going to translate directly into `lrpar`.

Edit: Regarding lrpar, at least one thing I'm noticing is that lrpar needs mechanisms for dealing with warnings.
While lrlex doesn't currently have warnings, this will probably end up changing the abstraction a little (not certain about the specifics).
It might be worth adding an empty `warnings` and vector and type into lrlex, just to keep the abstractions the same.
Perhaps that might be done separately though?

Edit 2x: The caveat to this is that I'm pretty uncertain in my ability to actually anticipate *where* warnings would arise from though in the api

lrlex/src/lib/codegen.rs
```diff
@@ -0,0 +1,635 @@
...
+        self.visibility = visibility;
+        self
+    }
+}
+
+impl<'a> LexSrcEnv<'a> {
+    pub(crate) fn new(
+        src: &'a str,
+        path: &'a Path,
+        header: MarkMap<String, HeaderValue<Location>>,
```
ratmice: Perhaps this `header` value seems like the biggest impediment to making any of this API public, as it is currently a #[doc(hidden)] part of the API. It stores all the things like all the `LexFlags` coming from the `CTLexerBuilder`.
It the `parsed_header` is subsequently built by the `%grmtools` section. Those two header values then get merged below. 

I say this partly because there was a moment at some point where I was confused why we weren't dropping a bunch of CTLexerBuilder options on the floor.

One thing we can perhaps consider is making the constructor public,
so downstream you can construct an empty one, outside of `CTLexerBuilder` codegen options could only be set when embedded in the `%grmtools` section.

I haven't had much in the way of other good ideas here. It isn't something I actually paid any mind to when initially making the patch.

ratmice: Probably a better way is that we could just make a pub version of new without the `header` argument at all like...

```
  pub fn new(src: &'a str, path: &'a path) -> Self {
     Self::new_with_header(src, path, Header::new())
  }
```

Not sure why I didn't think of it earlier.

ratmice: One slight step further to allow configurability of default values *without* making header public might be to make the type public but keep all the methods hidden, then we can have a way to obtain a `Header` from a `CTLexerBuilder`,
So you could call `LexSrcEnv::new(src, path, ct_lexer_builder.header())`. With that people could still configure it through the builders, but the type stays mostly opaque.

Edit:
Alternately we could consider just adding method like `code_generator` to `CTLexerBuilder` so the `LexerSrcEnv` could stay entirely private. I suppose the difficulty with that is that then there are two routes to generating code from the builder, the one people should use `build()`, and `code_generator()` that people will trip over.

ratmice:
Should mention this implements the lrlex parts of #648

ltratt:
I'm a bit behind on reviewing, but will hopefully get to this one soon!

ratmice:
no worries, I understand, and also am also still trying to figure out an impl for lrpar.

ratmice:
One thing that has occurred to me is that maybe it would have been nicer if I did one commit for each moved code block, instead of just moving a bunch of code blocks.

Let me know if you think that kind of approach would help, I could probably split it up.

Edit: On second thought it may not be possible without adding `Clone` bounds to some variables.

ltratt:
Apologies again for being slow on this one: I wanted to do my best to digest it and think about the consequences! I broadly get where this is going: I agree with you that trying to think of the right names is always challenging when doing this sort of refactoring.

I'm intrigued by this:

>But so far trying to make e.g. a two pass API for checking then codegen has proven elusive. Primarily because of data dependencies between the checking phases, and callbacks like lrpar_config, I've managed to get some nicer API to compile, but not pass the testsuite.

Let's assume we could get such an API past the test suite: what would the impact on users be? Would we be able to provide them a painless upgrade path or would they have to hack things around?

ratmice:
I did actually manage to get a two-pass API to work (I guess it would be two-ish, i'll explain), there were two primary things that allowed it to work,
moving ownership into the codegen, and then allowing a `lexerdef_mut` accessor to borrow a mutable reference.
The other was a code organization regarding logical error handling.

When I say logical error handling, I mean cases where the control flow is roughly like
```
let foo = ...;
if something {
   return Err(...)
} else {
   let bar = foo.do_stuff();
}
```

The thing to note is that if `foo.do_stuff()` has no observable side effects (like the codegen passes) this can be reordered equivalently to
```
let foo = ...;
let bar = foo.do_stuff();
if something {
   return Err(...);
}
```

These two things, grabbing mutable borrows, and the reordering went a bit hand-in-hand.
But basically ended up with two-ish passes:

```
let src_env = LexSrcEnv::new(...);
let codegen_args = LexCodeGenArgs{...};
let codegen = src_env.code_generator(args)?;
// borrow code_gen mutably and e.g. modify lexerdef
let output_rs_src = codegen.generate();
```

At least if you consider "code_generator(args)?" and `generate()` to be the two passes.
Most of the other separation between those structs is just ownership.
With that I was able to get it at least pretty close to two pass and pass the test suite.

I did forget to update my initial message.

ltratt:
Interesting! I wonder if we should take a brief step back and ask ourselves: knowing what we know now, about how people use these tools, and how we would like them to be able to use them in the future, what would the perfect API look like? We could sketch that out, and then work out (a) how big a change it would be for us to make a reality (b) if there's a plausible upgrade path for users. The reason for this is that it might free us from some of the accidental nastiness that, inevitably, has been baked in over time as we've evolved things (and, of course, many bad decisions I took back in the early days!).

ratmice:
I'm actually pretty happy with they way (this series) turned out at the moment.

It's funny actually a lot of the things that were proving difficult are some of the things I've added (fairly recently even).
Things like the running an `RTParserBuilder` with test inputs (These are exactly the kinds of things that we needed to reorder the error handling/codegen building).

The other thing that requires some finesse is things like the communication with cargo being `println("cargo:{}", ...)`,
which doesn't belong in the codegen. That sort of stuff is something we're always going to have to work around with all APIs (this has more impact in lrpar due to `cargo:warning=`.

So I *think* because of cargo we're always going to have this kind of control flow where we're borrowing references
between building the code generator and generating code?

ratmice:
I'll definitely have a think on these things though, but definitely nothing instantly jumps out at me!

ltratt:
Dumb idea: perhaps one of the problems is that we've hard-coded the output of codegen? I wonder if we had some sort of `trait Codegen` (perhaps even more abstract an idea than "codegen"?!) it would help us isolate more of the bits that are currently intertwined?

ratmice:
I had kind of thought about it when I started implementing `lrpar`, there are a couple of things I noticed:

There is a bit of overlap in `LexSrcEnv` in that we could almost make it `SrcEnv` if we rename `lex_diag()` to just `diagnostics()` or something. The whole `merge_headers()` and `parse_headers()` functions are reusable across lex/yacc.

We could somehow make `SrcEnv::code_generator(...)` and `Codegen::generate()`, into traits the thing to note about `SrcEnv::code_generator(CodegenArgs) -> CodeGenerator` is it would have two associated types, one for args, and one for the generator itself.

I had kind of discounted this as a bunch of extra complexity, but now I'm perhaps wondering where you are going with using this trait, e.g. if you have an array of `Vec<Box<dyn CodeGen>>` you could do both `lrlex` and `lrpar` in a single builder? 

Edit: Off-hand I'm not certain whether the trait would be `dyn compatible` or `object safe`, but if so I could imagine a world where that allows us to have a single unified builder which builds both lrlex and lrpar?

Edit 2x: My thoughts right now are just I need to get an lrpar version working before we can start trying to make them both work under the same trait, I *think* making `Codegen::generate() -> Result<String, Box<dyn Error>>` might help, where it currently is `generate(self) -> String`

ltratt:
Please don't overestimate how much I have thought about this :) In a sense what I'm thinking -- very vaguely! -- is: there's no huge reason we just have to dump out a `.rs` file with a serialised table in a specific directory. We could, for example, output a LALR-pop style "generate normal Rust code" parser; or something very different. Right now we (well... I!) hard-coded in one specific style, but that was probably a mistake. Admittedly, undoing that might be more disruption than it's worth (at least for now). So in this view of the world we might get something like:

```rust
trait CodeGen { ... }

impl CodeGen for SerialisedInRust { ... }

impl CodeGen for GenerateRecursiveDescentParser { ... }
```

and so on. It may well be that this is entirely orthogonal to the current PR anyway, and I'm just creating noise!

ratmice:
Ahh, I see, I hadn't thought of that usage of traits.  I thought you meant a trait that did codegen for e.g. all of `lrpar`, `lrlex`, and the perhaps the `token_map`. For which the only thing I could think of was some kind of unified builder, but that makes sense.

Trait enabled code generators might also make the streaming lexers more feasible to write, since they'd probably need different codegen too.

Anyhow, I kind of need to get lrpar even working before I can think about how to get it working nicely :)

ltratt:
Understood! Like I said, it might be trying to do too much in one go. [Or, perhaps, it suggests a different ordering of refactorings to us? Dunno!]

ratmice:
I think my hope has been just as a first take/prototype trying to stick with just a private API, hopefully without any semver breakage. Then once we've got it functioning having a think on public API surface and traits and that sort of thing, once we can tinker with that more easily on branches.

ltratt:
I agree that's a solid plan. [FWIW, I was -- at least for now -- imagining this trait being private, in the sense that we could use it to decouple currently-entangled parts of `codegen.rs` but I appreciate that it might not be the right time to think about it now.]

ratmice:
I should have asked if you had preferences on how I push out `lrpar` prototype?
While it isn't working yet, it feels like it's getting closer to working on top of current `HEAD`, so doing it on top of this one would require a rebase (and the two don't really overlap in any way, there is little reason they need to be in the same PR).

So I'm wondering if I should aim for a separate PR, or rebase it on top this one. If the latter should I try and squash these commits down to get rid of interim changes? Or even close this one and just make a new PR?

ratmice:
FWIW, I think I'm at the stage where I'm now convinced that the approach taken here with lrlex, just can't possibly work for lrpar. The key things to note is that lrpar has more things that need to be owned, and more stages/passes that it goes through, than there are owners in this current codegen module.

In particular CTParserBuilder goes like:

```
Source String -> ASTWithValidityInfo? -> YaccGrammar -> Rust String -> CTParser (owns a YaccGrammar);
```
Where this patch currently has: `(SrcEnv, CodeGenArgs) -> CodeGen`


```
fn generate(self) -> String
```

The closest I got was modifying it to: 

```
fn generate(&self, src_env: &ParserSrcEnv) -> Result<String, Box<dyn Error>>
```
The ownership relationship here was that `Codegen` owned an `ASTWithValidityInfo` (so that given a reference to a codegen you could get warnings out).  But leaves the `generate` function having a local variable for the `YaccGrammar` with no means for it to escape via `CTParser`. The only way I see it working introducing another owner to the mix:

```
(SrcEnv, BuildEnvArgs) -> BuildEnv -> CodeGen
```

Now BuildEnv: Owns the `ASTWithValidityInfo` and CodeGen would own the `YaccGrammar`,
and it's possible to get both the source string out (via `generate`), and return a `CTParser`.
So I think we need a Third pass for lrpar, where we could squeak by with two for lrlex.

Edit: I've gotten that sort of structure working now (but had to introduce a clone, seeing if I can avoid that before push)...

ltratt:
Dumb question: do we want to hold back on this PR while we get the lrpar PR sorted out and then can come back to this with the lessons learned?

ratmice:
That seems fine to me, like I'm not sure what to do with this one in the light of the differences between it and the additional complexity required for `lrpar`.

Like this one does work with the simpler two-phase implementation, rather than `lrpar` which is kind of `three-phase`.
`SrcEnv`, `BuildEnv`, and `Codegen`.  One question is should we stick to the simpler implementation.

Make a unitary `BuildEnv` with no fields in it, just to keep the same structure? (Would we need it if lrlex became more complex somehow?)

But I'm happy to let this one rest. While we hammer lrpar codegen into the appropriate shape, if you'd like to close it for now.

ltratt:
I'm open-minded, and since you're driving this, am happy to take your steer.

ratmice:
Alright, I'll just close thus one then. I'd much prefer if we end up in a place where the lrlex and lrpar codegen process resemble one another, and lrpar seems to be driving that due to it's additional complexity.

PR #654, opened 2026-08-04
Lrpar build cleanups

ratmice:
This is mostly just reformatting/reordering code,

In 63f32e7e02eda7b307022ec5 I had reformatted this code to use a match statement, which increased the indent
to avoid a clippy lint.

I found that by using `return Err(ErrorString(out).into)` instead of `return Err(ErrorString(out))?;` we can avoid the clippy lint without increasing indentation levels.

The second patch pulls some if guards out of a match statement, turning
```
match result {
  Ok(...) if warnings_are_errors && !warnings.is_empty() => {
   }
   Ok(...) => {
       if !warnings.is_empty() > {
       }
  }
  Err(errs) => ...,
}
```

Into a simpler sequence of statements:
```
if warnings_are_errors && !warnings.is_empty() {
   ...
} else !warnings.is_empty() {
   ...
}

match result {
   Ok(...) => ...,
   Err(errs) => ...,
}
```

These are kind of helpful for trying to split out the codegen, since in the patches for separating codegen out,
the match statements have been moved into part of codegen the bring-up, while the if statements/warning handling stay in CTParserBuilder since they deal with cargo.

### DIFF
--- a/lrpar/src/lib/ctbuilder.rs
+++ b/lrpar/src/lib/ctbuilder.rs
@@ -777,40 +777,36 @@ where
 
         self.yacckind = Some(ast_validation.yacc_kind());
         let warnings = ast_validation.ast().warnings();
-        let res = YaccGrammar::<StorageT>::new_from_ast_with_validity_info(&ast_validation);
-        let grm = match res {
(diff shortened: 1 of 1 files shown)


ltratt:
Agreed!

PR #656, opened 2026-08-06
Adding a  `lrpar::codegen` module.

ratmice:
This the beginning of attempt No. 2 at pr #655 just trying to split up the giant commit from that patch into smaller more easily reviewable ones. Below is the original pr description:

Here is an attempt at pulling a codegen module out of `lrpar`, it migrates the `CTParserBuilder` to use it, and passes
the testsuite. I haven't gone over this with a fine toothed comb, there have been some obscure timing related problems I introduced during development that didn't cause any testsuite failures.

(Like calling `check_unused_header_keys()` too early before the callback.

I wasn't able to do this patch in a way that was even remotely incremental because of ownership issues. Nor really figure out a way to do it in a way that didn't require making slight changes to things as they moved over (mostly changing references to `self`, removing calls `unwrap()`).

There are 3 passes to this and a 4th structure `ParserBuildEnvArgs`:
1. `ParserSrcEnv`
2. `ParserBuildEnv`
3. `ParserCodeGen`

The general idea is:

`ParserSrcEnv`: source, source path, diagnostics generator, `header` collection which contains the default values for the `%grmtools` section.
`ParserBuildEnvArgs`, these are essentially the builder arguments including the essential `Option<>` types. It's just a minimal builder.
`ParserBuildEnv`: This structure contains fields derived from the `BuildEnvArgs` and also the inner types from options in `BuildEnvArgs`. By derived fields I mean things like `ASTWithValidationInfo`
`ParserCodegen`: This one contains a `YaccGrammar`, `StateTable`, and `StateGraph`. In order to generate code, though it still needs the `SrcEnv` and the `BuildEnv`. 

Currently the `BuildEnv` takes ownership of the `BuildEnvArgs`, it was easiest this way because the `rebuild_cache` code expects `Options` it may be we should drop `BuildEnvArgs` from `BuildEnv`.

But generally the idea (in a pseudo functional syntax) is something to the effect of:

`(SrcEnv?, BuildEnvArgs) -> BuildEnv? -> CodeGen::generate(&build_env, &src_env)?;`

There is a pretty high probability that this contains some code duplicated from `CTParserBuilder`, which
didn't go fully unused in `CTParserBuilder`, given that the diff +/- is only a few hundred lines, I don't expect an enormous amount. But perhaps if it is really used in both places like `fn indent` appears to be we can `use ctbuilder::indent` instead.
I'll try and read through it in this regard tomorrow.

### DIFF
--- a/lrpar/src/lib/codegen.rs
+++ b/lrpar/src/lib/codegen.rs
@@ -1,10 +1,17 @@
 #![deny(unfulfilled_lint_expectations)]
 #![expect(dead_code)]
 
-use std::{any::type_name, error::Error, fmt, hash::Hash, marker::PhantomData, path::Path};
+use std::{
+    any::type_name,
+    error::Error,
+    fmt::{self, Write},
+    hash::Hash,
+    marker::PhantomData,
+    path::Path,
+};
 
 use cfgrammar::{
-    Location, Span,
+    Location, Span, Symbol,
     header::{GrmtoolsSectionParser, Header, HeaderValue},
     yacc::{YaccGrammar, YaccKind, ast::ASTWithValidityInfo},
 };
@@ -12,7 +19,9 @@ use proc_macro2::{Literal, TokenStream};
 
 use crate::{
     LexerTypes, RecoveryKind, RustEdition, SerialisationFormat, Visibility,
-    ctbuilder::{CTConflictsError, ERROR, FixIntConfig, VarIntConfig, indent},
+    ctbuilder::{
+        ACTION_PREFIX, CTConflictsError, ERROR, FixIntConfig, VarIntConfig, indent, make_generics,
+    },
     diagnostics::{DiagnosticFormatter, SpannedDiagnosticFormatter},
 };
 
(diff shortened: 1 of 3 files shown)


ratmice:
A couple of oopsies where I copied some code but didn't delete it, or still had things refer to `self` when they could/should refer to `build_env`. But now that all the codegen doesn't have access to self that shouldn't be possible.

At least I *think* that is all of it in that the `use` items for `quote`, `syn`, `proc_macro2`, used by codegen.

Edit: Anyhow I'm not sure if that helps making it easier to review or not, it's still a lot.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -887,7 +887,7 @@ where
     ) -> Result<(), Box<dyn Error>> {
         let grm = code_gen.grm();
         let mod_name = build_env.mod_name();
-        let visibility = self.visibility.clone();
+        let visibility = build_env.visibility();
```
ratmice: Probably would be better if I squashed all these little changes to `output_file` into one patch.

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1226 @@
...
+                "Required values were missing from the header: {}",
+                missing_keys.join(", ")
+            )
+            .into())
+        } else {
+            Ok(())
+        }
+    }
+
+    fn extract_ast_validation(
```
ratmice, line 196: I think rather than `extract_ast_validation` this should be something like `derive_ast_validation`.
For the other `extract` functions, I wonder if they would be better named `resolve_*`?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1226 @@
...
+        &mut self,
+        from_ast: Option<&ASTWithValidityInfo>,
+    ) -> Result<ASTWithValidityInfo, Box<dyn Error>> {
```
ratmice, line 199: With my nimbleparse_lsp hat on for that project it wants the raw errors, 
so it can send spans directly to the editor for highlighting rather than `Box<dyn Error>`.

Probably something to think about later

lrpar/src/lib/codegen.rs
```diff
@@ -451,6 +460,176 @@ where
         quote!(#cache_info_str)
     }
 
+    /// Generate the user action functions (if any).
+    pub(crate) fn gen_user_actions(
+        &self,
+        src_env: &ParserSrcEnv,
+    ) -> Result<TokenStream, Box<dyn Error>> {
+        let grm = self.grm();
+        let diag = src_env.yacc_diag();
```
ratmice, line 469: I probably should have done the minor modifications to the signature, and added the let bound variables at the top in place, rather than when moving the code 👎 

```
        let grm = self.grm();
        let diag = src_env.yacc_diag();
```

Same during the other code moves... Let me know if you want me to do some kind of interactive rebase which adds these changes before the code gets moved so it shows up more cleanly in the diff?

ratmice:
Sorry, I think i'll just close this, and do a 3rd pass via interactive rebase.  At the very least there is enough to clean up here that it feels probably worthwhile...

PR #657, opened 2026-08-07
Lrpar codegen

ratmice:
I think this is probably as reviewable as I'm going to manage to make this large of a patch.
The basic idea behind this patch is to have a kind of "functional pipeline", for doing code generation,

`(SrcEnv?, BuildEnvArgs) -> BuildEnv? -> Codegen(SrcEnv, BuildEnv) -> rust_code?`

A bit of an oversimplification, as there are some other minor details...

1. `BuildEnvArgs` is kind of a minimalist equivalent of the current builder, it's just full of `Option` values.
2. `BuildEnv` is full of `derived` values, it mostly strips off the `Option`, but it also contains values like `ASTWithValidityInfo` that are derived from all the other args.
3. `Codegen` then owns the `YaccGrammar`, `StateTable` and `StateGraphs`, which you can take ownership of after generating code.

This may not be totally perfect basis for traits and external usage, for example it currently returns `Box<dyn Error>` instead of typed errors. But it should be a pretty faithful conversion of the existing process, into a more targeted/self contained module.

### DIFF
--- a/lrpar/src/lib/codegen.rs
+++ b/lrpar/src/lib/codegen.rs
@@ -1,11 +1,11 @@
 #![deny(unfulfilled_lint_expectations)]
 #![expect(dead_code)]
 
-use std::{error::Error, fmt, marker::PhantomData, path::Path};
+use std::{error::Error, fmt, hash::Hash, marker::PhantomData, path::Path};
 
 use crate::{
     LexerTypes, RecoveryKind, RustEdition, SerialisationFormat, Visibility,
-    ctbuilder::{ERROR, indent},
+    ctbuilder::{CTConflictsError, ERROR, FixIntConfig, VarIntConfig, indent},
     diagnostics::{DiagnosticFormatter, SpannedDiagnosticFormatter},
 };
 
@@ -14,7 +14,9 @@ use cfgrammar::{
     header::{GrmtoolsSectionParser, Header, HeaderValue},
     yacc::{YaccGrammar, YaccKind, ast::ASTWithValidityInfo},
 };
-use lrtable::{StateGraph, StateTable};
+
+use lrtable::{Minimiser, StateGraph, StateTable, from_yacc};
+use wincode::SchemaWrite;
 
 pub(crate) struct ParserSrcEnv<'a> {
     src: &'a str,
@@ -328,6 +330,103 @@ where
 
     pub(crate) fn yacc_kind(&self) -> YaccKind {
         self.ast_with_validity_info.yacc_kind()
+    }
+
+    pub(crate) fn code_generator(
+        &self,
+        src_env: &ParserSrcEnv,
+        timestamp: &str,
+    ) -> Result<ParserCodegen<LexerTypesT>, Box<dyn Error>> {
+        let grm = match YaccGrammar::<LexerTypesT::StorageT>::new_from_ast_with_validity_info(
+            &self.ast_with_validity_info,
+        ) {
+            Ok(grm) => grm,
+            Err(errs) => {
+                let mut out = String::new();
+                out.push_str(&format!(
+                    "\n{ERROR}{}\n",
+                    src_env.yacc_diag().file_location_msg("", None)
+                ));
+                for e in errs {
+                    out.push_str(&indent(
+                        "     ",
+                        &src_env.yacc_diag().format_error(e).to_string(),
+                    ));
+                    out.push('\n');
+                }
+                return Err(ErrorString(out).into());
+            }
+        };
+
+        let (sgraph, stable) = from_yacc(&grm, Minimiser::Pager)?;
+        if self.cache_args.error_on_conflicts
+            && let Some(c) = stable.conflicts()
+        {
+            match (grm.expect(), grm.expectrr()) {
+                (Some(i), Some(j)) if i == c.sr_len() && j == c.rr_len() => (),
+                (Some(i), None) if i == c.sr_len() && 0 == c.rr_len() => (),
+                (None, Some(j)) if 0 == c.sr_len() && j == c.rr_len() => (),
+                (None, None) if 0 == c.rr_len() && 0 == c.sr_len() => (),
+                _ => {
+                    let conflicts_diagnostic = src_env.yacc_diag().format_conflicts::<LexerTypesT>(
+                        &grm,
+                        self.ast_with_validity_info.ast(),
+                        c,
+                        &sgraph,
+                        &stable,
+                    );
+                    return Err(Box::new(CTConflictsError {
+                        conflicts_diagnostic,
+                        phantom: PhantomData,
+                        #[cfg(test)]
+                        stable,
+                    }));
+                }
+            }
+        }
+
+        Ok(ParserCodegen {
+            grm,
+            stable,
+            sgraph,
+            timestamp: timestamp.to_string(),
+        })
+    }
+}
+
+impl<LexerTypesT> ParserCodegen<LexerTypesT>
+where
+    LexerTypesT: LexerTypes,
+    usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
+    LexerTypesT::StorageT: 'static
+        + fmt::Debug
+        + Hash
+        + num_traits::PrimInt
+        + SchemaWrite<FixIntConfig, Src = LexerTypesT::StorageT>
+        + SchemaWrite<VarIntConfig, Src = LexerTypesT::StorageT>
+        + num_traits::Unsigned,
+    LexerTypesT: LexerTypes,
+{
+    pub(crate) fn grm(&self) -> &YaccGrammar<LexerTypesT::StorageT> {
+        &self.grm
+    }
+
+    pub(crate) fn stable(&self) -> &StateTable<LexerTypesT::StorageT> {
+        &self.stable
+    }
+
+    pub(crate) fn sgraph(&self) -> &StateGraph<LexerTypesT::StorageT> {
+        &self.sgraph
+    }
+
+    pub(crate) fn finish(
+        self,
+    ) -> (
+        YaccGrammar<LexerTypesT::StorageT>,
+        StateGraph<LexerTypesT::StorageT>,
+        StateTable<LexerTypesT::StorageT>,
+    ) {
+        (self.grm, self.sgraph, self.stable)
     }
 }
 
--- a/lrpar/src/lib/codegen.rs
+++ b/lrpar/src/lib/codegen.rs
@@ -1,23 +1,32 @@
 #![deny(unfulfilled_lint_expectations)]
 #![expect(dead_code)]
 
-use std::{error::Error, fmt, hash::Hash, marker::PhantomData, path::Path};
+use std::{
+    any::type_name,
+    error::Error,
+    fmt::{self, Write},
+    hash::Hash,
+    marker::PhantomData,
+    path::Path,
+};
 
 use crate::{
     LexerTypes, RecoveryKind, RustEdition, SerialisationFormat, Visibility,
-    ctbuilder::{CTConflictsError, ERROR, FixIntConfig, VarIntConfig, indent},
+    ctbuilder::{
+        ACTION_PREFIX, CTConflictsError, ERROR, FixIntConfig, VarIntConfig, indent, make_generics,
+    },
     diagnostics::{DiagnosticFormatter, SpannedDiagnosticFormatter},
 };
 
 use cfgrammar::{
-    Location, Span,
+    Location, Span, Symbol,
     header::{GrmtoolsSectionParser, Header, HeaderValue},
     yacc::{YaccGrammar, YaccKind, ast::ASTWithValidityInfo},
 };
 
 use lrtable::{Minimiser, StateGraph, StateTable, from_yacc};
 use proc_macro2::TokenStream;
-use quote::{ToTokens, TokenStreamExt, quote};
+use quote::{ToTokens, TokenStreamExt, format_ident, quote};
 use wincode::SchemaWrite;
 
 pub(crate) struct ParserSrcEnv<'a> {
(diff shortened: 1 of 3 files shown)


lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+const GLOBAL_PREFIX: &str = "__GT_";
+const ACTIONS_KIND: &str = "__GtActionsKind";
+const ACTIONS_KIND_PREFIX: &str = "Ak";
+const ACTIONS_KIND_HIDDEN: &str = "__GtActionsKindHidden";
+
+pub(crate) struct ParserSrcEnv<'a> {
+    src: &'a str,
+    // We store the path here so we can generate a module name from it if needed.
+    // But should never use it for filesystem interaction within this module.
+    path: &'a Path,
```
ltratt: I _think_ that in some future use-cases we might not even have a `Path` for a given input (e.g. if it's from stdin or a buffer in memory)? If so, I wonder if we should just make this `Option<String>` for the path? [Maybe not for this PR, but as a simple follow-up.] That might avoid a (borderline-but-not-quite pathological) case where the `Path` disappears and we don't get a valid pathname out of it later.

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+    diagnostics: SpannedDiagnosticFormatter<'a>,
+    header: Header<Location>,
+}
+
+pub(crate) struct ParserBuildEnvArgs<'a> {
+    /// This allows the parser to originate from from a pre-parsed AST, rather than
+    /// parsing a grammar definition given as source string into an AST.
+    ast_originated: Option<&'a ASTWithValidityInfo>,
```
ltratt: I ideally wondered if there are only two routes to us getting the AST: pre-parsed AST or string. I think "there are only those two routes" but perhaps I'm being narrow-minded!

Either way, I wondered if we could maybe tweak this field name a bit. Maybe just `ast_validity_info` with a docstring like "If the input came from a preparsed AST, this will be `Some`"?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+        self
+    }
+    pub(crate) fn warnings_are_errors(mut self, warnings_are_errors: bool) -> Self {
+        self.warnings_are_errors = warnings_are_errors;
+        self
+    }
+}
+
+impl<'a> ParserSrcEnv<'a> {
+    pub(crate) fn new_with_defaults(
```
ltratt: Will there ever be another `new` method? Should we just call this `new`?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+                "Required values were missing from the header: {}",
+                missing_keys.join(", ")
+            )
+            .into())
+        } else {
+            Ok(())
+        }
+    }
+
+    fn extract_ast_validation(
```
ltratt: Any reason to prefix these methods with `extract_`? I'm not saying it's wrong, but I wondered if it has more semantic implication than is immediately obvious to me. I also suspect these should document that they mark headers as used?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+                        break;
+                    }
+                    stem = new_stem;
+                }
+                format!("{}_y", stem)
+            }
+        }
+    }
+
+    pub(crate) fn build_env<LexerTypesT>(
```
ltratt: Dumb question: could/should this ever be called more than once? If "no" should this consume `self`?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+            cache_args: args,
+            recoverer,
+            serialisation_format,
+            mod_name,
+            phantom_storaget: PhantomData,
+        })
+    }
+}
+
+impl<'a, LexerTypesT> ParserBuildEnv<'a, LexerTypesT>
```
ltratt, line 402: Given that this is struct is private to the crate, I wondered if these getter methods are serving a useful purpose?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+{
+    pub(crate) fn grm(&self) -> &YaccGrammar<LexerTypesT::StorageT> {
+        &self.grm
+    }
+
+    pub(crate) fn stable(&self) -> &StateTable<LexerTypesT::StorageT> {
+        &self.stable
+    }
+
+    pub(crate) fn take_parser(
```
ltratt: This one's interesting because it consumes `self`. Why/when do we use it in that way?

lrpar/src/lib/codegen.rs
```diff
@@ -0,0 +1,1258 @@
...
+            _ => unreachable!(),
+        };
+
+        let additional_decls = if let YaccKind::Original(YaccOriginalActionKind::GenericParseTree) =
+            build_env.yacc_kind()
+        {
+            // `lrpar::Node`` is deprecated within the lrpar crate, but not from within this module,
+            // Once it is removed from `lrpar`, we should move the declaration here entirely.
+            Some(quote! {
+                        #[allow(unused_imports)]
```
ltratt: We can probably dedent this a bit (`cargo fmt` doesn't tend to deal well with macros, so we often have to do it by hand).

lrpar/src/lib/codegen.rs
ratmice: I originally had it named `new`, but changed it in case we want  a method new like so:

```
pub fn new(src, path) -> Self {
   self.new_with_defaults(src, path, Header::new())
}
```

The thought was that this could allow us to make `Codegen` public without making `Header` public,
the public API though would only be usable with grammars that have filled out grmtools sections.
And wouldn't allow specifying default values via the header. So a little less flexible than what the builder can do, but something.

ltratt: Maybe this method should be `new_with_header`?

lrpar/src/lib/codegen.rs
ratmice: I *think* we kind of need something here stdin, e.g. we derive the default `mod_name` from it, and error formatting will need some label for error printing, I do think it could be a string though.

e.g. often in the case of `stdin` this will just be the text "stdin" or something to that effect, for instance the gnu assembler will use the text "standard input"

```
$ echo bad_instruction | as -
Assembler messages:
{standard input}:1: Error: no such instruction: `bad_instruction'
```

Part of the difficulty with turning it into string then is also that `SpannedDiagnosticFormatter::new()` accepts a path, but it seems like maybe despite it's convenience the diagnostic formatter might not actually belong in the `SrcEnv`, if we just derive a module name from it we should be able to turn it into a string.

lrpar/src/lib/codegen.rs
ratmice: That works for me.

lrpar/src/lib/codegen.rs
ratmice: I had meant to rename these `resolve_` but I forgot, is resolve any better?
Essentially these are resolving them to some final/absolute value, from all the possible ways they could have defaults set for them.

lrpar/src/lib/codegen.rs
ratmice: I think calling it more than once would be a waste and inefficient, there is seems no actual harm in it though.
(I should say, there is no reason to, but there is no reason it couldn't either)

However his can't really consume `self` though:
`Codegen::generate()` and many functions (all error handling) after `build_env` require a `&ParserSrcEnv` reference. I didn't really think about this, and whether `BuildEnv` should take ownership of the `SrcEnv` after the call to `build_env` though.

lrpar/src/lib/codegen.rs
ratmice: This is for the return values to `CTParserBuilder::build` we construct a `CTParser` from it in the `Ok` case.
And I think we need to take ownership of the `StateTable` and `StateGraph` for `CTConflictsError` in some of the error cases too.

lrpar/src/lib/codegen.rs
ratmice: So, the only real vital getter that I recall is `fn ast_validation` it is vital, because that that is how we get the list of warnings, to error from `warnings_are_errors` in `CTParserBuilder`.

A lot of the other getters are less vital, the reason they were added was that I couldn't keep into `self` while codegen was partially migrated from `CTParserBuilder` to `codegen`, where we didn't have access to the private variables from outside the module.

But what that does mean, is that if we have an external `Codegen` trait, that the `ParserCodegen` struct could be derived from an external source. So if we *did* make it public, there are getters for all the things needed by the codegen module itself.

I don't exactly know about trait boundaries and the like given that e.g. `BuildEnv::code_generator` actually returns a `CodeGen` whether that would make it impossible to use with some external codegen impl.

That was my thinking behind keeping them anyways.

lrpar/src/lib/codegen.rs
ratmice: Definitely a better name.

lrpar/src/lib/codegen.rs
ratmice: FWIW, it looks like `Codegen::generate` also uses it for error handling.
One thing to note though is what I said in the PR description:

> This may not be totally perfect basis for traits and external usage, for example it currently returns Box<dyn Error> instead of typed errors.

As an example `nimbleparse_lsp` wants to send spans directly to the editor, so it wants a totally separate error formatting code than tools like `nimbleparse` or `CTParserBuilder`.

Perhaps that changes things.

ratmice: So, it appears that not all of these errors actually have typed errors (aren't coming from a `YaccGrammarError`).

For instance the following error, I'm not exactly certain why this error is only being caught so late.
It's the only one that can happen during `generate(...)`, and so is why we need to pass in a `src_env`.
https://github.com/softdevteam/grmtools/blob/3a9d88f0c7bb22b29a891224659940ec408a3b89/lrpar/src/lib/ctbuilder.rs#L1659-L1663

I'm assuming if we can't catch it earlier for some reason, we could always add a `CodegenError` type?

lrpar/src/lib/codegen.rs
ratmice: Should be fixed in [9694268](https://github.com/softdevteam/grmtools/pull/657/commits/9694268c5b6814cfc7502ba91ff67dedfac96f44)

lrpar/src/lib/codegen.rs
ratmice: I went ahead and renamed these `resolve_*` in [e53376e](https://github.com/softdevteam/grmtools/pull/657/commits/e53376ef28a64ea580ce8819a0f0d11a27160dc3) and docs added in [c452dd4](https://github.com/softdevteam/grmtools/pull/657/commits/c452dd472ed55aed73c930d2c27ef25834a0990a).

Let me know if you prefer/can think of a better naming scheme

lrpar/src/lib/codegen.rs
ratmice: Renamed in [c0330c6](https://github.com/softdevteam/grmtools/pull/657/commits/c0330c68231ca6be8daff349da329f3dd9cfcbcb)

lrpar/src/lib/codegen.rs
ratmice: I did this also renaming the `ast_validation()` method in [b29e3a4](https://github.com/softdevteam/grmtools/pull/657/commits/b29e3a4fb89fbcd031384e7a89f641f8f65128d2). I also noticed a couple of places that we could use the shorter `yacc_kind()` method instead (It just calls `ast_with_validity_info.yacc_kind()` so there is no change.

lrpar/src/lib/codegen.rs
ratmice: I *think* that the reason this is caught so late, was that long ago `GrammarAST` could be constructed manually (without a `YaccKind`), and you could create a `YaccGrammar` from that. Thus there was nowhere to know whether or not an action type was required or not, because by that point the `yacc_kind` was lost.

There seems multiple places we could do it, but I might gravitate towards passing the `YaccKind` in as a parameter to `complete_and_validate`? Though perhaps not, I don't immediately see how to get the actiontype from the ast, ~because the `YaccParser.global_actiontype` is outside the AST~. There is the `Rule.actiont`, which global_actiontype gets cloned into for `YaccKind::Original`, and part of the rule syntax for `YaccKind::Grmtools`.

ratmice: FWIW, there was more error handling in `gen_user_actions` I hadn't noticed regarding action code like `$var` parsing during `gen_user_actions`.

I *do* think it'll probably end up needing a `enum CodegenError`, just for *other* kinds of errors though
for separating `YaccGrammarError` from those originating from `proc_macro2`, etc.
So I don't know if we absolutely *need* to catch these earlier?

lrpar/src/lib/codegen.rs
ratmice: I was wondering if `fn finish(self) -> (YaccGrammar, StateGraph, StateTable)` or `finished` might be a better
name, since technically it's taking more than just the parser?

I don't know if it's really finishing or finalizing anything, but maybe it conveys better that we're done with `self` and taking ownership of all the interesting values.

lrpar/src/lib/codegen.rs
ltratt: Sorry, I wasn't very clear! What I mean is: if we _only_ use this for a later mod name, we could create the mod name here, rather than carry around a `Path`.

lrpar/src/lib/codegen.rs
ltratt: I've confused myself as to whether this could be `self` now! :)

lrpar/src/lib/codegen.rs
ltratt: Fair enough!

lrpar/src/lib/codegen.rs
ratmice: I don't think we can quite yet, still working on it. Got to move all the `yacc_diag()` back out into `CTParserBuilder`, and make sure there isn't any more `yacc_diag()` usage inside `codegen`.
We've gotten rid of the major impediment I've looked at so far though.

The next issue is we don't really have a public error type conflicts, `CTConflictsError` is private and basically just a wrapper around a formatted string. But nothing like `YaccGrammarError` which is a typed error with spans.

Edit: It seems like rather than wrapping this in an error it might be better to just borrow `lrtable::Conflicts` from the `ParserCodeGen`, and forgo the error.

There is [lrtable::Conflicts](https://docs.rs/lrtable/latest/lrtable/statetable/struct.Conflicts.html) which isn't an error though. I think there is probably still a decent amount of work left before we can make this take `self`. The only other issue I see is that `HeaderError<Span>` may need to be a bit more public, which is currently `#[doc(hidden)]` for errors in `%grmtools` sections, that seems like it might be reasonable though. 

The stuff that is left seems like it might be easier to do on top of the codegen patch than on master (saves us from introducing code only to move it later)?

lrpar/src/lib/codegen.rs
ratmice: That makes sense, so once we've removed `SpannedDiagnosticsFormatter` I think that should work!

lrpar/src/lib/codegen.rs
ratmice: Oh, I forgot to answer your question here.

it is correct that "There are only two routes"

lrpar/src/lib/codegen.rs
ratmice: I should ask, the rest of the work to remove the `diagnostics` field from `ParserSrcEnv` is going to require a rebase on top of PR #659 

Let me know whether you prefer this to be done on top of a new PR, or rebasing this one.

Edit: FWIW I've kind of got a plan that doesn't require rebasing (since the code changed in master is all moving in this PR, we can just manually update the moved code). This branch will temporarily lack some error checking, but unless that perturbs the testsuite it seems like it should merge fine. It seems like the easiest thing but if it doesn't work out we can just rebase.

ratmice: I did manage to switch it to `self` in [9039f51](https://github.com/softdevteam/grmtools/pull/657/commits/9039f518ceebf25bd90586747e04baa3231b70e1) and the commits leading up to that.

I did run into one other borrow of `src_env` which I had forgotten about, but just moved that variable into `build_env`.

lrpar/src/lib/codegen.rs
ratmice: There was one more usage of the path, just an entry in the cache, which gets lossily converted to string (before and after), I just did that early while creating `build_env` rather than while generating the cache string now.

Should be fixed in [f158c51](https://github.com/softdevteam/grmtools/pull/657/commits/f158c51c85b1827660eaf0c71fb716bb86542e46)

ratmice:
My attempt to avoid conflicts didn't work (We modified lines that are removed too!),
 so this'll need a rebase or reopening as a new PR.

But this does run though `cargo test`, despite some errors are temporarily panics, because it doesn't have the error handling we moved into cfgrammar.

lrpar/src/lib/ctbuilder.rs
```diff
@@ -1078,717 +922,19 @@ where
...
-                })
-            } else {
-                None
-            };
-
-        let mod_name =
-            match syn::parse_str::<proc_macro2::Ident>(mod_name) {
-                Ok(s) => s,
-                Err(e) => return Err(format!(
-                    "CTParserBuilder::mod_name(\"{}\") is not a valid rust identifier due to '{}'",
```
ratmice, line 1125: So I think ideally we'd be able to retain some mention of `CTParserBuilder` in this error text, which
in my patch series we've lost, to me it indicates that `mod_name` might want to be routed through `Header`, so it can pick up a `location` of `CTParserBuilder` or so.

There seem like a number of places in this patch where the origin of some error condition might be improved, but I'm hesitant to try and do too much/everything in this specific series.

lrpar/src/lib/codegen.rs
ltratt: I agree, that's clearer.

lrpar/src/lib/codegen.rs
ltratt: Should we just flat out say "modname" (or similar here) rather than sort-of-pretending it's a `Path` (even though we've now stored it as a `String`)?

lrpar/src/lib/codegen.rs
ltratt: Works for me!

lrpar/src/lib/ctbuilder.rs
ltratt: I agree: we don't have to do everything in one go.

ltratt:
I suspect it's worth merging `master` in as a merge commit sooner rather than later, as it'll make the eventual rebase easier.

lrpar/src/lib/codegen.rs
ratmice: Oops, looks I'd changed the one in `ParserBuildEnv`. We could turn this into two fields:

```
/// A potential modname derived from a path name, only used if `mod_name` is unspecified.
default_modname: String,
/// The path name has an entry (`GRAMMAR_PATH`) in the cache, this should only be used for that.
path_name_cache_entry: String,
```

I guess my question is in the constructor it's probably easier to take a `Path`, and derive these two fields from it. Or have a separate constructor that takes both these fields as strings. I guess we could also turn the cache entry into an option and take an `Option<Path>` to the constructor?

ratmice: I did the best I could think of in [b7decac](https://github.com/softdevteam/grmtools/pull/657/commits/b7decacdf4e34ce72f5d59490694eb425b33b9f7) how do you feel about that?

ratmice:
I'm actually not really worried about the rebase, since all the conflicts should be of the form `delete both the incoming and current changes` (probably famous last words). But will try and figure out how to merge master without rebasing.

I believe that [5143009](https://github.com/softdevteam/grmtools/pull/657/commits/514300994372939469608b3bc1bea2888b6d88e5) should bring it in line with the current master branch

lrpar/src/lib/codegen.rs
ratmice: Changed to `finish` in [57cd704](https://github.com/softdevteam/grmtools/pull/657/commits/57cd70483b0a1c2fc87379c7004880ae101c99ff)

lrpar/src/lib/codegen.rs
ltratt: Aha, I hadn't thought about the cache name! I actually like the split above, even though in one sense it's mildly wasteful, it'll stop future me from forgetting about the two very different uses.

ltratt:
I think we're probably ready to squash?

ratmice:
I think so, but I wanted to think given all the simplifications we've done, whether we can move back to the more simpler model like we started with with `lrlex` (or if we'd want to).

I think it might be feasible now to have the two-phase thing, but I kind of like the `SrcEnv`, `BuildEnv` split, and the differences between errors are pretty stark. (e.g. `Span` on errors only happens in `SrcEnv` which makes sense).
I think we'd basically have to squash those errors together in the same type to go back to the two phase approach.
But if we ignore that, now that we aren't borrowing both the `SrcEnv` and the `BuildEnv` it seems like it could be possible combine them?

I'd like to mull that over a bit, hear what you think first.

ltratt:
Taking a step back is always a good idea if it might help us reach a better decision. I'm a little bit unsure about the best way ahead, as I admit I've slightly muddled myself in the detail.

ratmice:
I think *if* it's going to work, basically we'd be combining the `ParserSrcEnv` and `ParserBuildEnv` struct fields,
then we'd move the `ParserSrcEnv::build_env` function into `ParserSrcEnv::new_with_header` and make that return a `Result<Self, ParserSrcEnvError>`. 

So, `ParserSrcEnv` would now build a `ASTWithValidationInfo` on `new`. Then we could put the `code_generator` method on that. We maybe could keep the error split by having different errors returned by `new_with_header` and `code_generator` on the same type.

But I *think* it might allow us to squash those two types together. I believe the main reason I had to separate them is we needed to borrow `ParserSrcEnv::yacc_diag()`, and then also borrow `ParserBuildEnv::ast_with_validation_info()`, but we removed the former.

That said, I do kind of like how this current patch is structured in the sense that building a src env, and a build env, etc seems like a logical sequence.

Edit: There is probably a simple way we can test this by making a constructor on `ParserBuildEnv` that internally calls `ParserSrcEnv::new_with_header().build_env()` I'll play around with the idea tomorrow

ratmice:
So, I did end up quickly testing that out, and it does work but it didn't really spark joy, mostly because I had forgotten about the `args` parameter to the constructor, which it has to bring in too.

```
    pub(crate) fn new_with_header(
        src: &'a str,
        path: Option<&Path>,
        header: Header<Location>,
        args: ParserBuildEnvArgs<'a>,
    ) -> Result<ParserBuildEnv, ParserSrcEnvError> {
        ParserSrcEnv::new_with_header(src, path, header).build_env(args)
    }
```

It is perhaps one less type we have to make public, but the construction progression doesn't seem to feel nearly as nice.

ltratt:
I tend to agree.

ratmice:
Working on squashing, but it might take me a bit before I can get the history into nice shape.
Just doing minor commits that make the history look better.

Because in the back and forth some code got moved around that we can clean up...

ratmice:
@ltratt I squashed and moved some commits around to hopefully make it somewhat reasonable.
I would have kind of liked to have gotten rid of the [1208709](https://github.com/softdevteam/grmtools/pull/657/commits/120870942ac193bb179b9dfb16f17f2d37a76471) commit entirely,
but it seemed like it was going to be way too much effort to fit that into all the rest of the patches.

Does that seem okay to you, or keep working on it?

ltratt:
Perfection is for the next life: if we're confident that this moves us in the right direction overall, I'm fine with it! So, ready to merge I think?

ratmice:
I think so, it feels like a step in the right direction and isn't committing us to any API surface yet.

ltratt:
Agreed!

ratmice:
Thanks, I'll start working on a version of the lrlex patches using the same overall structure. 
Before I start working on making anything public, I should get a version of nimbleparse-lsp
up to speed for test out the code generation API in practice.

PR #659, opened 2026-08-08
Move actiont checking from codegen to ast validation.

ratmice:
Here is an attempt to validate actiont earlier,
adding it to `YaccGrammarError`.

I'm uncertain it's possible to trigger for `YaccKind::Grmtools` I would imagine you'd get a syntax error before it could `complete_and_validate`. So I checked that with a manually constructed AST.

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -15,6 +15,7 @@ use super::{
 use crate::{
     Span,
     header::{GrmtoolsSectionParser, HeaderError, HeaderErrorKind, HeaderValue},
+    yacc::YaccOriginalActionKind,
 };
 
 /// Any error from the Yacc parser returns an instance of this struct.
@@ -55,7 +56,9 @@ impl ASTWithValidityInfo {
             let mut yp = YaccParser::new(yacc_kind, s);
             yp.parse().map_err(|e| errs.extend(e)).ok();
             let mut ast = yp.build();
-            ast.complete_and_validate().map_err(|e| errs.push(e)).ok();
+            ast.complete_and_validate(Some(yacc_kind))
+                .map_err(|e| errs.push(e))
+                .ok();
             ast
         };
         ASTWithValidityInfo {
@@ -122,7 +125,9 @@ impl FromStr for ASTWithValidityInfo {
                 let mut yp = YaccParser::new(yacc_kind, src);
                 yp.parse().map_err(|e| errs.extend(e)).ok();
                 let mut ast = yp.build();
-                ast.complete_and_validate().map_err(|e| errs.push(e)).ok();
+                ast.complete_and_validate(Some(yacc_kind))
+                    .map_err(|e| errs.push(e))
+                    .ok();
                 ast
             };
             Ok(ASTWithValidityInfo {
@@ -307,9 +312,19 @@ impl GrammarAST {
     ///   3) Every token reference references a declared token
     ///   4) If a production has a precedence token, then it references a declared token
     ///   5) Every token declared with %epp matches a known token
-    ///
-    /// If the validation succeeds, None is returned.
-    pub(crate) fn complete_and_validate(&mut self) -> Result<(), YaccGrammarError> {
+    ///   6) If `yacc_kind` is specified, perform any kind specific validation.
+    ///      * If the kind requires an action type, check that each rule has one
+    ///      * That each production has action code
+    ///      * That `$` variables referred to in action code are recognised.
+    pub(crate) fn complete_and_validate(
+        &mut self,
+        yacc_kind: Option<YaccKind>,
+    ) -> Result<(), YaccGrammarError> {
+        let kind_requires_action_checks = matches!(
+            yacc_kind,
+            Some(YaccKind::Original(YaccOriginalActionKind::UserAction)) | Some(YaccKind::Grmtools)
+        );
+
         match self.start {
             None => {
                 return Err(YaccGrammarError {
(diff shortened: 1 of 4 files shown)


lrpar/src/lib/ctbuilder.rs
ratmice: There is a similar error here that should probably also be moved in this patch

ratmice: Ditto for this error.

ratmice: Should be fixed in [9e956c2](https://github.com/softdevteam/grmtools/pull/659/commits/9e956c2d6dacd60f06e60bde81555ba0d0240b1d)

ratmice: Should be fixed in [6313b96](https://github.com/softdevteam/grmtools/pull/659/commits/6313b9653f2e1e907126391f1fb7db467fd58fc4) and tested in [ca3a9b1](https://github.com/softdevteam/grmtools/pull/659/commits/ca3a9b1317ff5e3064d2dcfaf681992c6c35f5b4)

And with that the `diag` variable becomes unused and can be removed from the last stages of codegen.

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -59,6 +59,7 @@ pub enum YaccGrammarErrorKind {
     InvalidString,
     NoStartRule,
     UnknownSymbol,
+    UnrecognizedDollarVariable,
```
ratmice: Would `UnrecognizedActionVariable` be better?

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -307,9 +312,17 @@ impl GrammarAST {
...
-    pub(crate) fn complete_and_validate(&mut self) -> Result<(), YaccGrammarError> {
+    ///   6) If `yacc_kind` is specified, perform any kind specific validation.
+    ///      * If the kind requires an action type, check that each rule has one.
+    pub(crate) fn complete_and_validate(
+        &mut self,
+        yacc_kind: Option<YaccKind>,
+    ) -> Result<(), YaccGrammarError> {
+        let kind_requires_actiont = matches!(
+            yacc_kind,
+            Some(YaccKind::Original(YaccOriginalActionKind::UserAction)) | Some(YaccKind::Grmtools)
```
ratmice, line 325: We could consider making this boolean check a method on `YaccKind`?
With a different name like `YaccKind::requires_actions`?

I think it might also simplify a similar check in `CTParserBulder`.

cfgrammar/src/lib/yacc/parser.rs
ratmice: I renamed it in [c2d027c](https://github.com/softdevteam/grmtools/pull/659/commits/c2d027c1c253a835a97c387bf4d60ad2eeb97dc9)

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -349,12 +349,36 @@ impl GrammarAST {
...
+                            {
+                                return Err(YaccGrammarError {
+                                    kind: YaccGrammarErrorKind::UnrecognizedDollarVariable,
+                                    spans: vec![Span::new(
+                                        action_span.start() + last + off + "$".len(),
+                                        action_span.end(),
+                                    )],
+                                });
+                            } else {
+                                last = last + off + "$$".len();
```
ratmice: This `else` is probably a  bit wonky, but it doesn't seem totally wrong in that we know each match is going to be at least 2?

There is one corner case here though where we get a *weird* span.
If we slightly modify the test I added:

(Modified test)
```
            r#"
%token a
%%
start -> () : "a" {$};
"#,
        );
        assert_eq!(
            ast_validity.errors(),
            vec![YaccGrammarError {
                kind: YaccGrammarErrorKind::UnrecognisedActionVariable,
                spans: vec![Span::new(33, 33)],
            }]
        );
```

We see that the cursor with the removed space before the `$` points to a zero length span
beginning right after the `$` at the very end of the action. I would probably expect a span of like `(32, 33)`. Morning brain hasn't started working yet to figure out what's going on with this calculation.


```
        let ast_validity = ASTWithValidityInfo::new(
            YaccKind::Grmtools,
            r#"
%token a
%%
start -> () : "a" { $foo; };
"#,
        );
        assert_eq!(
            ast_validity.errors(),
            vec![YaccGrammarError {
                kind: YaccGrammarErrorKind::UnrecognisedActionVariable,
                spans: vec![Span::new(33, 37)],
            }]
        );
```

ratmice: Span calculation should be fixed in [d244993](https://github.com/softdevteam/grmtools/pull/659/commits/d244993fcfca5fb7cc79f0923c02a19208924d9b) this turned to be partly because the span for the action code was untrimmed, while the string itself was trimmed, but that was being obscured skipping the leading "$" from the span.

The easiest fix seemed to be not trimming the string, and not skip the leading `$` since it leads to an empty string in the trailing $ sign cases like `{$}`.

cfgrammar/src/lib/yacc/ast.rs
ltratt: We could happily do that in a follow-up PR I think?

ltratt:
Please squash.

ratmice:
There was one american english typo that I fixed in YaccGrammarErrorKind::Display, before squashing.

ratmice:
Squashed. (After going through the patch again, and noticing a few interim changes I squashed it further)

ltratt:
Thanks!

PR #661, opened 2026-08-19
Lrlex codegen

ratmice:
This is an attempt at lrlex codegen, in the style of the recently added lrpar codegen module.
One thing this doesn't do is change the `CTTokenMapBuilder`.

This tries to sequence the move over a couple of commits, with the first commit just generating module and ownership structure. With each subsequent commit moving one aspect of code generation, until it's all in the codegen module.

One slightly suprising thing was the placement of lexerdef within the `LexerBuildEnv`, because we need it to generate the `rule_ids_map`, with that we can wait to generate the `LexerCodegen` util the `rule_ids_map` is built, and pass it in to the `code_generator` function, this is a slight departure from the `lrpar` api, which merely took a `timestamp`.

Edit: One thing that the additional `rule_ids_map` parameter I immediately notice is that unlike `BuildEnvArgs` and other builder patterns, it's hard for us to add add an additional parameter should we need one.  I wonder if we should consider `LexerCodegenArgs`?

### DIFF
--- a/lrlex/src/lib/codegen.rs
+++ b/lrlex/src/lib/codegen.rs
@@ -10,7 +10,12 @@ use proc_macro2::{Ident, TokenStream};
 use quote::{ToTokens, TokenStreamExt, format_ident, quote};
 use regex::Regex;
 use std::{
-    any::type_name, collections::HashMap, fmt, marker::PhantomData, path::Path, sync::LazyLock,
+    any::type_name,
+    collections::HashMap,
+    fmt::{self, Write as _},
+    marker::PhantomData,
+    path::Path,
+    sync::LazyLock,
 };
 
 static RE_TOKEN_ID: LazyLock<Regex> =
@@ -119,7 +124,6 @@ where
     usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
 {
     rule_ids_map: Option<HashMap<String, LexerTypesT::StorageT>>,
-    #[expect(unused)]
     timestamp: String,
 }
 
@@ -458,6 +462,26 @@ where
                 #(#token_consts)*
             }
         })
+    }
+
+    pub(crate) fn generate(
+        &self,
+        build_env: &LexerBuildEnv<LexerTypesT>,
+    ) -> Result<String, LexerCodegenError> {
+        // Try and run a code formatter on the generated code.
+        let unformatted = self.generate_lex_module(build_env)?.to_string();
+        let mut outs = String::new();
+        // Record the time that this version of lrlex was built. If the source code changes and rustc
+        // forces a recompile, this will change this value, causing anything which depends on this
+        // build of lrlex to be recompiled too.
+        let timestamp = &self.timestamp;
+        write!(outs, "// lrlex build time: {}\n\n", quote!(#timestamp),).ok();
+        outs.push_str(
+            &syn::parse_str(&unformatted)
+                .map(|syntax_tree| prettyplease::unparse(&syntax_tree))
+                .unwrap_or(unformatted),
+        );
+        Ok(outs)
     }
 }
 
--- a/lrlex/src/lib/codegen.rs
+++ b/lrlex/src/lib/codegen.rs
@@ -375,6 +375,24 @@ where
             let start_states: Vec<StartState> = vec![#(#start_states),*];
         }
     }
+
+    pub(crate) fn gen_lexerdef_ty(&self, build_env: &LexerBuildEnv<LexerTypesT>) -> TokenStream {
+        match build_env.lexerkind() {
+            LexerKind::LRNonStreamingLexer => {
+                quote!(::lrlex::LRNonStreamingLexerDef)
+            }
+        }
+    }
+
+    pub(crate) fn gen_instantiate_lexerdef(
+        &self,
+        build_env: &LexerBuildEnv<LexerTypesT>,
+    ) -> TokenStream {
+        let lexerdef_ty = self.gen_lexerdef_ty(build_env);
+        quote! {
+            #lexerdef_ty::from_rules(start_states, rules)
+        }
+    }
 }
 
 /// The quote impl of `ToTokens` for `Option` prints an empty string for `None`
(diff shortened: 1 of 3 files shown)


lrlex/src/lib/codegen.rs
```diff
@@ -0,0 +1,525 @@
...
+            lexerkind,
+            visibility: args.visibility,
+            header: self.header,
+            lex_flags,
+            lexerdef,
+        })
+    }
+}
+
+pub(crate) enum LexerBuildEnvError {}
```
ratmice, line 232: I had forgotten to double check this, for some reason I thought a variant had been added to this,
but that was `LexerCodegenError`.

lrlex/src/lib/ctbuilder.rs
```diff
@@ -632,34 +576,39 @@ where
...
-                    x.map(|a| a.iter().map(|&b| b.to_string()).collect::<HashSet<_>>()),
-                    y.map(|a| {
-                        a.iter()
-                            .map(|(b, span)| (b.to_string(), *span))
-                            .collect::<HashSet<_>>()
-                    }),
-                )
+        let code_gen = build_env
+            .code_generator(self.rule_ids_map, env!("VERGEN_BUILD_TIMESTAMP"))
+            .map_err(|e| match e {})?;
```
ratmice, line 588: This wacky `map_err`/`match` was necessary at the time because `?` couldn't convert the error.
I figured by the time I was done we'd have a variant added. Although I must admit I'm a bit uncertain what the type returned is, I assume it's inferring `!` or some such.

lrlex/src/lib/codegen.rs
```diff
@@ -0,0 +1,525 @@
...
+        let lexerdef_ty = self.gen_lexerdef_ty(build_env);
+        quote! {
+            #lexerdef_ty::from_rules(start_states, rules)
+        }
+    }
+
+    pub(crate) fn gen_token_consts(&self) -> TokenStream {
+        let mut token_consts = TokenStream::new();
+        if let Some(rim) = self.rule_ids_map() {
+            let mut rim_sorted = Vec::from_iter(rim.iter());
```
ratmice, line 419: This thing just emits an empty `TokenStream` if `rule_ids_map` is `None`, I assume that is okay?
Meaning something like `lrpar` didn't run? This copies over the existing code, but I was curious whether this needs to be as-is, or if it could/should be checked like an error condition?

ratmice: For some reason I forgot we had lrlex/examples, sure enough if we add `assert!(self.rules_id_map.is_some())`, the example `lrlex/examples/calclex` fails. So this seems correct as is.

lrlex/src/lib/codegen.rs
```diff
@@ -0,0 +1,525 @@
...
+use lrpar::LexerTypes;
+
+use crate::{LRNonStreamingLexerDef, LexBuildError, LexFlags, LexerDef, LexerKind, Visibility};
+use proc_macro2::{Ident, TokenStream};
+use quote::{ToTokens, TokenStreamExt, format_ident, quote};
+use regex::Regex;
+use std::{
+    any::type_name,
+    collections::HashMap,
+    fmt::{self, Write as _},
```
ltratt, line 15: In the `as _` necessary? It's a new one on me!

lrlex/src/lib/codegen.rs
```diff
@@ -0,0 +1,525 @@
...
+pub(crate) enum LexerCodegenError {
+    InvalidRustIdentifierModName { mod_name: String, error: syn::Error },
+}
+
+impl fmt::Display for LexerCodegenError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&match self {
+            Self::InvalidRustIdentifierModName { mod_name, error } => {
+                format!(
+                    "Invalid mod_name '{mod_name}' is not a valid rust identifier due to '{error}'"
```
ltratt: I think I'd drop the "Invalid " prefix: the rest of the message makes clear it's an error.

lrlex/src/lib/ctbuilder.rs
ltratt: It's a new one on me! I imagine it's inferring the unit type `()`.

lrlex/src/lib/codegen.rs
ratmice: It is not (only there is a `Write` in naming conflict, and this import is last).
It basically imports the trait *without* binding the `Write` name.
With it this way you can call a method in the `Write` trait, but haven't bound it to a name.

You'd need to remove the `as _` before you could do `impl Write for ... {}`

ltratt: Ah!

lrlex/src/lib/ctbuilder.rs
ratmice: I'd be surprised if `()` would convert into `Box<dyn Error>` via `?` though, which is why I think it must be `!`
or `Infallible`.

ltratt: I see what you mean. I guess the ultimate point here is that the closure passed to `map_err` can't ever be called because `e` is of a type that cannot be constructed.

lrlex/src/lib/codegen.rs
ratmice: Done in [741cecf](https://github.com/softdevteam/grmtools/pull/661/commits/741cecfbf54f2442afa1e32dd8974e424a2061aa)

I checked lrpar/codegen, and the latter didn't include "Invalid" already, so the two errors looked in sync.

lrlex/src/lib/codegen.rs
ltratt: It is odd, I admit, but I can imagine we will add a variant here, so at most it doesn't hurt.

ltratt:

ratmice:
Squashed the stray commits into the first one.

PR #662, opened 2026-08-20
Initial parsing of %user section as a `Header`.

ratmice:
This still needs some work, but is an initial attempt at a `%user` section we can parse by reusing the `%grmtools` section parser, then restrict the types within it as a post processing stage.

In a subsequent patch we can iterate over the `%user` section and convert it into a nicer to use `HashMap`. Throwing invalid value errors on the more rust-like constructs available via `HeaderValue`

This is also currently missing any way to access the information from the section.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -6,8 +6,11 @@ use crate::{
     },
 };
 use regex::{Regex, RegexBuilder};
-use std::{error::Error, fmt, sync::LazyLock,
-    collections::{HashSet, HashMap}
+use std::{
+    collections::{HashMap, HashSet},
+    error::Error,
+    fmt,
+    sync::LazyLock,
 };
 
 #[derive(Debug)]
@@ -15,7 +18,7 @@ use std::{error::Error, fmt, sync::LazyLock,
 pub struct FileHeaders {
     pub grmtools: Header<Span>,
     pub grmtools_span: Span,
-    pub user: Header<Span>,
+    pub user: HashMap<String, (Span, UserSectionValue)>,
     pub user_span: Span,
 }
 
@@ -80,6 +83,15 @@ pub enum HeaderErrorKind {
     DuplicateEntry,
     InvalidEntry(&'static str),
     ConversionError(&'static str, &'static str),
+    InvalidUserSectionValueType,
+}
+
+#[derive(Debug)]
+pub enum UserSectionValue {
+    String(String, Span),
+    Num(u64, Span),
+    Bool(bool, Span),
+    Array(Vec<UserSectionValue>, Span),
 }
 
 impl fmt::Display for HeaderErrorKind {
@@ -93,6 +105,9 @@ impl fmt::Display for HeaderErrorKind {
             }
             HeaderErrorKind::InvalidEntry(s) => &format!("Invalid entry: '{}'", s),
             HeaderErrorKind::DuplicateEntry => "Duplicate Entry",
+            HeaderErrorKind::InvalidUserSectionValueType => {
+                "Invalid value type for the %user section"
+            }
             HeaderErrorKind::ConversionError(t, err_str) => {
                 &format!("Converting header value to type '{}': {}", t, err_str)
             }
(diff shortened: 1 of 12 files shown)


cfgrammar/src/lib/header.rs
```diff
@@ -80,6 +83,15 @@ pub enum HeaderErrorKind {
...
     ConversionError(&'static str, &'static str),
+    InvalidUserSectionValueType,
+}
+
+#[derive(Debug)]
+pub enum UserSectionValue {
+    String(String, Span),
+    Num(u64, Span),
+    Bool(bool, Span),
+    Array(Vec<UserSectionValue>, Span),
```
ratmice, line 94: I wasn't sure if we wanted to go super simple e.g. `HashMap<String, String>`, or a `serde_json::Value` inspired enum like this, I went with the latter for now because it was easy enough to do, and allows us to add types later if we need?

ratmice:
I *think* that pretty much covers the approach that I was thinking would largely reuse the code we already have,
I did a `cargo semver-checks --exclude lrpar-tests` to double check that there is no semver bump needed.

lrlex/src/lib/lexer.rs
```diff
@@ -416,6 +420,7 @@ where
             start_states,
             lex_flags: DEFAULT_LEX_FLAGS,
             expected_missing_tokens: vec![],
+            user_section: HashMap::new(),
```
ratmice, line 423: This actually seems like it could be objectionable,  unlike `GrammarAST` which is never constructed from parser codegen. `CTLexerBuilder` is going to call this `from_rules` function from codegen. So this is pretty much a useless allocation from that perspective.

Maybe it should be an `Option<...>` or somewhere else entirely?

cfgrammar/src/lib/header.rs
```diff
@@ -418,83 +427,110 @@ impl<'input> GrmtoolsSectionParser<'input> {
         Self { src, required }
     }
 
+    pub fn parse(&'_ self) -> Result<(FileHeaders, usize), Vec<HeaderError<Span>>> {
+        let mut sections_lookup = HashSet::from_iter(["%grmtools", "%user"]);
```
ratmice, line 431: I suppose I was thinking about the dotted keys vs named sections
Technically it seems like we could have both a named section and a declaration of the same name,
giving users the ability to specify the name of their section (via `sections_lookup` here rather than hard coding `%user`


We could have both:
```
%foo {
  // The user specified a foo section
}

%foo later-grmtools-added-a-foo-declaration
````

The commitment would need to make is that we agree not to add sections other than `%grmtools`.
There I suppose is an obscure conflict in that it is assuming that the `%foo` *declaration* is not starting with a `{`, at the very beginning of the file at least.


This *would* reduce some duplication of the tool name
```
%user {
    nimbleparse_lsp.input_file_extension: ".foo"
    nimbleparse_lsp.grammar_path: "./foo.y"
}
```

could turn into:
```
%nimbleparse_lsp {
   input_file_extension: "foo",
   grammar_path: "./foo.y",
}
```

Actually it's probably a bit more complex than just adding it here, we'd need to ignore it for cases like `nimbleparse` which aren't expecting/passing any additional section_lookup entries. The approach taken in this patch, definitely eliminates that complexity.

cfgrammar/src/lib/header.rs
ltratt: I think this is the right way to go.

lrlex/src/lib/lexer.rs
ltratt: I think it's fine: in practise `HashMap` won't `malloc` until the first entry goes in it.

cfgrammar/src/lib/header.rs
ltratt: This raises another possibility (off the top of my head: not properly thought through etc etc!). Perhaps we only need a `%grmtools` section and we namespace _all_ directives in there (accepting, for backwards compatibility reasons, that not everything will have a `grmtools.` prefix at first).

ratmice: I think there is actually still a bit of a reason to have separate directives in that we *can* check that each declaration in the `%grmtools` section is used easily, while it's more difficult for the `%user` section (because we aren't documenting/exposing the markmap that allows marking them as used)

Well I guess check that the `grmtools.` prefixed items are used only, and filter out other prefixes in the unused check, we'd have to split the `.` during the filter (only filtering out items with a non-grmtools non-empty prefix).

ratmice: The other thing I would say is that as it is currently written, with the hashmap like API, and the string `grmtools.yacc_kind` being a different key than `yacckind`, it's definitely going to double a lot of checks in the codebase while we still have legacy entries without the namespace. It seems like it might be quite a bit.

I don't know if trying to change the way that lookup is handled to use a default prefix/search defaulting to using a grmtools namespace would be viable. But that might avoid adding a second lookup including the namespace.

ltratt: Dumb idea: imagine when we see `yacckind` (and other legacy names) we immediately converted them to `grmtools.yacc_kind` so there was no duplication. Would that solve the problem? If it doesn't, then I think I should shut up, and we can move on with the PR :)

ratmice: I assuming you mean internally in the `Header` structure, rather than in the `cargo clippy --fix` sort of rewrite the input way.  I hadn't thought of that, and it might reduce the overhead to a tolerable amount.

Because a lot of what I was worried about was e.g. we'd have to do lookups for both, but *also* every entry in the sections go through a duplicate key lookup,  So we'd have to check if there is both two `yacc_kind` and `grmtools.yacc_kind` and one prefixed and one unprefixed.

This latter idea gets rid of that little bit of combinatoric explosion and should allow us to check for unused keys still. It is probably worth experimenting with trying it, I imagine it should work.

ratmice:
Closing this for now while we try the other approach discussed, we can always reopen it if that doesn't work out.

cfgrammar/src/lib/header.rs
ltratt: > I assuming you mean internally in the Header structure, rather than in the cargo clippy --fix sort of rewrite the input way. 

Yep, that was what I was thinking. Nothing clever -- you know me by now :)

ratmice: There is one actual thing which is a little awkward we need to figure out.

Currently the `%grmtools` and `%user` section differ in that only `%grmtools` supports the `YaccKind::Grmtools(...)` and `Grmtools(...)` structure syntaxes. While these get stripped from the `%user` section, we kind of need to figure out how to handle this difference.

One thing which I was tempted to do in my original patch (but decided to not try due to doing the simplest thing) is converting the rust `constructor` like structures to back into their string representation, and adding that to the `HashMap`.

The other question is whether the `grmtools.yacc_kind` and friends should show up in the user visible hashmap?

ltratt: I don't have an immediate opinion. We could simply bork if those are passed to a key that isn't `grmtools.yacc_kind`? That's probably the easiest thing for now, in the sense that it's strict, and we can relax it later if we realise that's the right thing.

ratmice: We can do that, but the problem is mostly how we expose `grmtools.` flags through what is currently the `GrammarAST::user_section() -> HashMap<String, UserSectionValue>` once it's merged in `%grmtools` it feels like it should also return the `grmtools.*` but that means we need to either strip fields like `yacc_kind` or add a variant to `UserSectionValue` convert them to, I think we can do something simpler than `HeaderValue`. 
Like `UserSectionValue::Constructor("YaccKind::Grmtools")` and normalizing the string value.

But whether or not we allow them in user specified fields, whether and how we should/shouldn't expose the `grmtools.*` ones now that they're in the same section.

ratmice: Anyhow, I feel like adding a `Constructor` variant and converting yacc_kind back into a string *shouldn't* be that difficult, and even gives us a path to allowing user values to contain these constructor variants in the future if we're so inclined.

So I'll probably start working on it under that assumption, but curious if you have other ideas.

ltratt: > how we expose grmtools. flags through what is currently the GrammarAST::user_section() -> HashMap<String, UserSectionValue>

[As before, I haven't thought about this carefully so caveat emptor!]

Could we do something like `GrammarAST::headers_starting_with(prefix: &str) -> impl Iter<...>`? Then if you want the `grmtools` headers you do `ast.headers_starting_with("grmtools")` and so on?

ratmice: Yeah, I think we could do something like that. We might also just have `headers()` and make the user perform a `filter(|(key, val)| key.starts_with("grmtools")).

Briefly I had considered whether entries with the `grmtools` prefix should be present when they are in the `GrammarAST` but not in the `%grmtools` section, but it is hard to be consistent with that when some fields originated as values in `ctbuilder.rs` rather than `GrammarAST`. Like `warnings_are_errors` was I believe never a field of `GrammarAST`. It also just doesn't really match the `headers` notion.

PR #664, opened 2026-08-25
Add crate name to grmtools section entries

ratmice:
So here is a first foray into adding a default "tool name" for `%grmtools` section entries,
This turned out a bit more complex than I had thought, for one it seemed that since `lex_flags` largely get passed
directly into the `regex` crate, we should name those `regex.foo` rather than `grmtools.foo`.

I named this "tool name" because there is already a thing we call `namespace` in this area, and that like `YaccKind::` and the like.

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -6,7 +6,7 @@ use crate::{
     },
 };
 use regex::{Regex, RegexBuilder};
-use std::{error::Error, fmt, sync::LazyLock};
+use std::{collections::HashMap, error::Error, fmt, sync::LazyLock};
 
 /// An error regarding the `%grmtools` header section.
 ///
@@ -239,13 +239,53 @@ impl<T> Namespaced<T> {
 static RE_LEADING_WS: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[\p{Pattern_White_Space}]*").unwrap());
 static RE_NAME: LazyLock<Regex> = LazyLock::new(|| {
-    RegexBuilder::new(r"^[A-Z][A-Z_]*")
+    RegexBuilder::new(r"^[A-Z][A-Z_\.]*")
+        .case_insensitive(true)
+        .build()
+        .unwrap()
+});
+#[doc(hidden)]
+pub static RE_CRATE_DOT: LazyLock<Regex> = LazyLock::new(|| {
+    RegexBuilder::new(r"^[A-Z][A-Z_]*\.")
         .case_insensitive(true)
         .build()
         .unwrap()
 });
 static RE_DIGITS: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[0-9]+").unwrap());
 static RE_STRING: LazyLock<Regex> = LazyLock::new(|| Regex::new(r#"^\"(\\.|[^"\\])*\""#).unwrap());
+#[doc(hidden)]
+pub static CRATE_KEY_MAP: LazyLock<HashMap<&'static str, &'static str>> = LazyLock::new(|| {
+    let mut map = HashMap::new();
+    let cfgrammar = ["yacckind"];
+    let lrpar = ["recoverer", "test_files", "serialisation_format"];
+
+    let lrlex = ["lexerkind", "allow_wholeline_comments", "posix_escapes"];
+    let regex = [
+        "case_insensitive",
+        "dot_matches_new_line",
+        "multi_line",
+        "octal",
+        "swap_greed",
+        "ignore_whitespace",
+        "unicode",
+        "size_limit",
+        "dfa_size_limit",
+        "nest_limit",
+    ];
+    for s in cfgrammar {
+        map.insert(s, "cfgrammar");
+    }
+    for s in lrpar {
+        map.insert(s, "lrpar");
+    }
+    for s in lrlex {
+        map.insert(s, "lrlex");
+    }
+    for s in regex {
+        map.insert(s, "regex");
+    }
+    map
+});
 
 const MAGIC: &str = "%grmtools";
 
@@ -429,7 +469,18 @@ impl<'input> GrmtoolsSectionParser<'input> {
                 i = self.parse_ws(j);
                 while self.lookahead_is("}", i).is_none() && i < self.src.len() {
                     let (key, key_loc, val, j) = match self.parse_key_value(i) {
-                        Ok((key, key_loc, val, pos)) => (key, key_loc, val, pos),
+                        Ok((key, key_loc, val, pos)) => {
+                            let key = if !RE_CRATE_DOT.is_match(&key) {
+                                if let Some(crate_name) = CRATE_KEY_MAP.get(key.as_str()) {
+                                    format!("{crate_name}.{key}")
+                                } else {
+                                    key
+                                }
+                            } else {
+                                key
+                            };
+                            (key, key_loc, val, pos)
+                        }
                         Err(e) => {
                             errs.push(e);
                             return Err(errs);
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -118,7 +118,7 @@ impl FromStr for ASTWithValidityInfo {
         let (header, _) = GrmtoolsSectionParser::new(src, true)
             .parse()
             .map_err(|mut errs| errs.drain(..).map(|e| e.into()).collect::<Vec<_>>())?;
-        if let Some(HeaderValue(_, yk_val)) = header.get("yacckind") {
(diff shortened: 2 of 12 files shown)


cfgrammar/src/lib/header.rs
```diff
@@ -239,14 +239,52 @@ impl<T> Namespaced<T> {
...
         .unwrap()
 });
 static RE_DIGITS: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[0-9]+").unwrap());
 static RE_STRING: LazyLock<Regex> = LazyLock::new(|| Regex::new(r#"^\"(\\.|[^"\\])*\""#).unwrap());
 
+static DEFAULT_TOOLNAME: LazyLock<HashMap<&'static str, &'static str>> = LazyLock::new(|| {
+    let mut map = HashMap::new();
+    let grmtools = [
+        "yacckind",
+        "lexerkind",
```
ratmice: grmtools as the notion of a "tool name" is kind of awkward.

I was wondering if it might be better to have `lrpar.yacckind` and `lrlex.lexerkind`
instead of `grmtools.*` for all the crates?

Not sure if it is better to have uniformity across the crates, or indication of who uses it.

ratmice:
I guess I should mention this just adds default "tool names" to the existing entries.
It doesn't actually allow any user specified values yet.

cfgrammar/src/lib/header.rs
```diff
@@ -239,14 +239,52 @@ impl<T> Namespaced<T> {
...
+static RE_TOOL_NAME: LazyLock<Regex> = LazyLock::new(|| {
+    RegexBuilder::new(r"^[A-Z][A-Z_]*\.")
         .case_insensitive(true)
         .build()
         .unwrap()
 });
 static RE_DIGITS: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[0-9]+").unwrap());
 static RE_STRING: LazyLock<Regex> = LazyLock::new(|| Regex::new(r#"^\"(\\.|[^"\\])*\""#).unwrap());
 
+static DEFAULT_TOOLNAME: LazyLock<HashMap<&'static str, &'static str>> = LazyLock::new(|| {
```
ratmice: would `target name` be better than `tool name`?

cfgrammar/src/lib/header.rs
ratmice: Rather than "tool name", maybe I'll try leaning into "crate name", in which case lrpar/lrlex make sense.
Anyhow, I'm going to try that since it's the most consistent naming scheme i've thought of.

Not too worried if we end up wanting to go back to `grmtools.*` it's not hard to change.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -118,7 +118,7 @@ impl FromStr for ASTWithValidityInfo {
         let (header, _) = GrmtoolsSectionParser::new(src, true)
             .parse()
             .map_err(|mut errs| errs.drain(..).map(|e| e.into()).collect::<Vec<_>>())?;
-        if let Some(HeaderValue(_, yk_val)) = header.get("grmtools.yacckind") {
+        if let Some(HeaderValue(_, yk_val)) = header.get("lrpar.yacckind") {
```
ratmice: So, about the time I got done I remembered that `YaccKind` is actually defined in [`cfgrammar`](https://docs.rs/cfgrammar/latest/cfgrammar/yacc/enum.YaccKind.html) So i'm not sure if that isn't going to be more confusing than it is helpful?

ltratt: If we want to use the crate names then we could just call this `cfgrammar.yacckind`?

doc/src/lexextensions.md
```diff
@@ -26,26 +26,26 @@ other flags should specify their value immediately after the flag name.
...
-| `unicode`                     | bool      | &cross;  | &checkmark;   |
-| `size_limit`                  | usize     | &cross;  | &checkmark;   |
-| `dfa_size_limit`              | usize     | &cross;  | &checkmark;   |
-| `nest_limit`                  | u32       | &cross;  | &checkmark;   |
+| Flag                          | Value     | Required |
+|-------------------------------|-----------|----------|
+| `lrlex.lexerkind`                   | [LexerKind](lexcompatibility.md#lexerkinds) | &cross;  |
+| `lrlex.posix_escapes`[^†]           | bool      | &cross;  |
+| `lrlex.allow_wholeline_comments`[^‡] | bool      | &cross;  |
+| `regex.case_insensitive`            | bool      | &cross;  |
```
ltratt, line 34: The `regex` prefix hadn't occurred to me. I _think_ I like it!

ltratt:
First thoughts: this looks like it works? You've raised a really good question in the sense of "what's the right prefix?" I hadn't thought about crate names, but so far this looks like it is probably better than `grmtools.`. Indeed, perhaps we can even elevate it to a "rule" along the lines of "prefixes are derived from published crates on crates.io; user-defined headers are prefixed with `crate.`" so that there's always a way for the user to add guaranteed-non-conflicting headers of their own?

cfgrammar/src/lib/yacc/ast.rs
ratmice: Fair enough, I'm not super fond of having multiple prefixes for parser related things, but it does reflect the fact that the consumer of the field *is* `cfgrammar`.

ltratt: I agree that it does expose a bit more internalness to the user than I would ideally like, but it has the virtue of consistency. Still, this is one of those things that we might change our minds on once we've sat with it for a while!

ratmice:
Yeah, as far as working it at least passes `cargo test`.  Now that I'm waking up I think it likely I missed something in `nimbleparse` I'll check if there the build bot script would have caught anything, and have a look through the `nimbleparse` code.

ratmice:
Running through the `.buildbot.sh`, all except the `lrpar.recoverer` field were being caught by the script, however since the `recoverer` field has a default value it would fallback to that. I went looked through all uses of "recoverer" in the rest of the repo too just in case this was a problem outside nimbleparse.

ratmice:
So I *think* this one is ready in the sense that I can probably proceed to the next step of relaxing the unused key checks for unrecognized crate names, to allow user defined keys.

I'm happy to let this one simmer while we think about it though. 
I'll probably build a second branch with these commits as it's starting point.

ltratt:
What are your initial thoughts on the design?

ratmice:
Overall I'm pretty happy with how it has turned out, I like that the regex fields now all point to the regex crate explicitly.
I'm not too bothered by the cfgrammar vs lrpar prefix.  

If we keep the current behavior where the library figures out unprefixed items perhaps people can just continue using `yacckind` and not have to worry about `lrpar` vs `cfgrammar` prefixes for that.

What I mean is: with the unused key check, we can keep checking that keys prefixed with `[lrpar,regex,cfgrammar,lrlex]` are all used. Anything with without a prefix is used. Any *other* prefix is used.

I guess what I'm thinking is we really don't seem forced to deprecate `yacckind` in this process, we can reserve the `unprefixed` keys only for keys where we can infer the crate name.

Edit: Besides that I also like that the crate names gives us a way to avoid conflicting keys like you've mentioned.
And nobody is hamstrung in their ability to add keys in the future.

Edit2: I guess to summarize overall I like it, and I wonder whether the downside of separating the key prefix for `cfgrammar`/`lrpar` is actually that much of a burden in practice if we/users can continue using `yacckind` and inferring the crate prefix.

ltratt:
Yes, for backwards compatibility we should definitely allow `yacckind` (etc) to be kept as-is in user's grammars for now and we should IMHO automatically transform it to `cfgrammar.yacckind` as if that's what the user wrote in the first place. [At some point we could warn that their unprefixed use is deprecated and one day after that perhaps remove support -- but I agree that we don't have any compelling need to do that quickly and maybe we will never will. I suppose if/when someone introduces a crate called `yacckind`!]

Summary: I think we should move ahead with this approach.

ratmice:
Actually it's not even a problem if someone makes a `yacckind` crate, because of the `yacckind.` allows us to disambiguate the crate prefix from the unprefixed field. As long as we refrain from using dot in keys in ways that don't correlate to a crate name.

Edit: I mean, it would be confusing, but there is not actually any conflict, we can differentiate the two syntactically.

ratmice:
I went ahead and unmarked as draft then.

ltratt:
> Actually it's not even a problem if someone makes a yacckind crate, because of the yacckind. allows us to disambiguate the crate prefix from the unprefixed field. As long as we refrain from using dot in keys in ways that don't correlate to a crate name.

True!

In terms of the code, I'm already happy with this. Could/should we merge (subject to squashing)?

ratmice:
Sounds good to me, I went ahead and squashed it and rewrote the commit message.

PR #666, opened 2026-08-28
Simplify Header::Value

ratmice:
This is an experiment to see if we can do away with `Value<T>` and all it's friends `Setting`, `Namespaced`, etc.
Replacing it with the much simpler `GrmtoolsSectionValue<T>`. This can lead to slightly worse error messages.
e.g. because we're checking `YaccKind` as a whole including `YaccOriginalActionKind`. But I think it is pretty much limited to that. We could do better, by attempting to parse the `RustLike(string)` rather than the simple matching I've done here. (As well as derive an inner span).

Sadly this patch has ended up kind of redoing some of the case-insensitivity migration in #665
This came up because when experimenting with integrading the lookup methods with the used value checking.
This involved changing the value owned by the `GrammarAST` from a `HashMap<GrmtoolsSectionValue>` to the `Header<T>`. Then the lookup methods would have to `clone` the value and return an owned `GrmtoolsSectionValue`, because there was none to borrow anymore.

It feels like this solves that borrowing issue, but *also* is a much simpler structure, and so cleans up the code a lot?

### DIFF
--- a/cfgrammar/src/lib/header.rs
+++ b/cfgrammar/src/lib/header.rs
@@ -90,57 +90,6 @@ impl fmt::Display for HeaderErrorKind {
     }
 }
 
-impl<T> HeaderError<T> {
-    /// Returns the [SpansKind] associated with this error.
-    pub fn spanskind(&self) -> SpansKind {
-        match self.kind {
-            HeaderErrorKind::DuplicateEntry => SpansKind::DuplicationError,
-            _ => SpansKind::Error,
-        }
-    }
-}
-
-/// Indicates a value prefixed by an optional namespace.
-/// `Foo::Bar` with optional `Foo` specified being
-/// ```rust,ignore
-/// Namespaced{
-///     namespace: Some(("Foo", ...)),
-///     member: ("Bar", ...)
-/// }
-/// ```
-///
-/// Alternately just `Bar` alone without a namespace is represented by :
-/// ```rust,ignore
-/// Namespaced{
-///     namespace: None,
-///     member: ("Bar", ...)
-/// }
-/// ```
-#[derive(Debug, Eq, PartialEq)]
-#[doc(hidden)]
-pub struct Namespaced<T> {
-    pub namespace: Option<(String, T)>,
-    pub member: (String, T),
-}
-
-#[derive(Debug, Eq, PartialEq)]
-#[doc(hidden)]
-pub enum Setting<T> {
-    /// A value like `YaccKind::Grmtools`
-    Unitary(Namespaced<T>),
-    /// A value like `YaccKind::Original(UserActions)`.
-    /// In that example the field ctor would be: `Namespaced { namespace: "YaccKind", member: "Original" }`.
-    /// The field would be `Namespaced{ None, UserActions }`.
-    Constructor {
-        ctor: Namespaced<T>,
-        arg: Namespaced<T>,
-    },
-    Num(u64, T),
-    String(String, T),
-    // The two `T` values are for the spans of the open and close brackets `[`, and `]`.
-    Array(Vec<Setting<T>>, T, T),
-}
-
 /// Parser for the `%grmtools` section
 #[doc(hidden)]
 pub struct GrmtoolsSectionParser<'input> {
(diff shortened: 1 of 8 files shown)


cfgrammar/src/lib/header.rs
```diff
@@ -311,15 +207,18 @@ fn add_duplicate_occurrence<T: Eq + PartialEq + Clone>(
 }
 
 impl<'input> GrmtoolsSectionParser<'input> {
-    fn parse_setting(&'_ self, mut i: usize) -> Result<(Setting<Span>, usize), HeaderError<Span>> {
+    fn parse_setting(
+        &'_ self,
```
ratmice: Just noticing that now a few of the functions like `parse_setting` and `parse_namespaced` which 
historically made sense since they previously returned `Setting` and `Namespaced` probably need better names like `parse_rustlike` for `parse_namespaced`..

Alas I don't have a good idea for `parse_setting`, which is to the right hand side of the `:` in `value: setting`.

lrpar/src/lib/parser.rs
```diff
@@ -635,58 +640,36 @@ pub enum RecoveryKind {
...
-            RecoveryKind::CPCTPlus => Value::Setting(Setting::Unitary(Namespaced {
-                namespace: Some(("RecoveryKind".to_string(), from_loc.clone())),
-                member: ("CPCTPlus".to_string(), from_loc.clone()),
-            })),
-            RecoveryKind::None => Value::Setting(Setting::Unitary(Namespaced {
-                namespace: Some(("RecoveryKind".to_string(), from_loc.clone())),
-                member: ("None".to_string(), from_loc.clone()),
-            })),
+            RecoveryKind::CPCTPlus => {
+                GrmtoolsSectionValue::RustLike("RecoveryKind::CPCTPlus".to_string(), from_loc)
```
ratmice: I think these can probably be simplified by using the `ToTokens`/implementations to produce the string
rather than these hard coded values.

Ditto for elsewhere.

ratmice: Alas, that doesn't seem to work (it emits a fully qualified path like `:: lrpar :: RecoveryKind :: CPCTPlus`)

ratmice: The best I could come up with to simplify these further was using `Debug` [3129318](https://github.com/softdevteam/grmtools/pull/666/commits/31293181ab8435f98c5a771117b934b2d78e3c4d)

cfgrammar/src/lib/header.rs
ratmice: I renamed these `parse_value` and `parse_rustlike` in [e882220](https://github.com/softdevteam/grmtools/pull/666/commits/e88222028eaf82f077ad1c69ee63e6e66c134e26) and  [a80866d](https://github.com/softdevteam/grmtools/pull/666/commits/a80866d86aeb159c83f026b2b052feec9d9199f2)

cfgrammar/src/lib/header.rs
```diff
@@ -90,152 +90,48 @@ impl fmt::Display for HeaderErrorKind {
...
-///
-/// To be useful across diverse crates this types fields are limited to types derived from `core::` types.
-/// like booleans, numeric types, and string values.
-#[derive(Debug, Eq, PartialEq)]
-#[doc(hidden)]
-pub enum Value<T> {
-    Flag(bool, T),
-    Setting(Setting<T>),
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum GrmtoolsSectionValue<T> {
```
ltratt: I think we can still retain the "The value contained within a `Header`" comment?

ltratt: I wonder if we can find a nicer/shorter name for this. Maybe just `HeaderValue` or similar?

cfgrammar/src/lib/header.rs
```diff
@@ -392,33 +286,35 @@ impl<'input> GrmtoolsSectionParser<'input> {
...
             }
         }
     }
 
-    fn parse_namespaced(
-        &self,
-        mut i: usize,
-    ) -> Result<(Namespaced<Span>, usize), HeaderError<Span>> {
-        // Either a name alone, or a namespace which will be followed by a member.
+    fn parse_rustlike(&self, mut i: usize) -> Result<((String, Span), usize), HeaderError<Span>> {
```
ltratt: FWIW I marginally prefer "namespaced" as a name ("Rustlike" felt like it could cover a lot of things). I don't feel strongly, though.

ltratt:
It's hard not to like a PR which deletes this much stuff with such a minor impact on functionality! It suggests to me that it's heading in the right direction!

cfgrammar/src/lib/header.rs
ratmice: I believe `HeaderValue` is still in use for the `HeaderValue(key_span: Span, value: GrmtoolsSectionValue)` type, It needed to not be a tuple for orphan rules IIRC/a comment says.

In this patch we deleted a type `Value` we could just use that perhaps.

cfgrammar/src/lib/header.rs
ratmice: I don't feel strongly about it either, I don't mind switching it back.

cfgrammar/src/lib/header.rs
ratmice: Perhaps we should also add/retain a comment for the `RustLike`/`Namespaced` variant.

ratmice: I expanded upon this comment, and added docstrings for `Namespaced`/`RustLike` in [02a4bee](https://github.com/softdevteam/grmtools/pull/666/commits/02a4beefdb2528e1d2d80bc91535c3bb91725930)

cfgrammar/src/lib/header.rs
ratmice: I switched this back to `Namespaced` in [12efed9](https://github.com/softdevteam/grmtools/pull/666/commits/12efed906661dff9e2ef86b95682356b05b0b6f3)

cfgrammar/src/lib/header.rs
ratmice: Renamed these to `Value` in [fbe1160](https://github.com/softdevteam/grmtools/pull/666/commits/fbe11602b9b6b1fcbf2fea8504baa451bc81cbb7)

Now `parse_value` even corresponds correctly sense.

cfgrammar/src/lib/header.rs
```diff
@@ -150,92 +99,54 @@ pub struct GrmtoolsSectionParser<'input> {
...
-                arr_open_loc.into(),
-                arr_close_loc.into(),
-            ),
-        }
-    }
+    String(String, T),
+    Num(u64, T),
+    Bool(bool, T),
+    Array(Vec<Value<T>>, T),
+    /// A Rust like value, with an optional type namespace.
```
ltratt, line 115: I presume that "optional" here means "non-empty string"?

ltratt:
I have one further, minor, comment, but I think we're close with this one.

cfgrammar/src/lib/header.rs
ratmice: Not quite (Edit: I may have misread non-empty as "empty string"), I was about to say I should include an example, but now notice I did, by "optional type  namespace" I meant: `YaccKind::Grmtools` where `YaccKind::` is the optional type namespace,
and it's optional because the `YaccKind::` can be omitted by just specifying `Grmtools`.

Maybe I need to specify `Grmtools` is omitting the optional type namespace in the later example?

ratmice: I tried to clarify the comment in [a2661d3](https://github.com/softdevteam/grmtools/pull/666/commits/a2661d312131a79de93f98cfae5b497180d9d851)

ltratt: Ah so we expect the optional namespace to end with `::` I think?

ratmice: Yeah exactly, the actual rule (from grmtools-section.test) is a little bit more complex than just starting with an optional prefix, because it's a recursive rule.

```
    namespaced -> (String, Span): IDENT | IDENT '::' IDENT;
    val -> Value<Span>: namespaced | namespaced '(' namespaced ')' | ... ;
```

In the "Further examples" you can see one of the recursive rules with and without the optional namespace: like `Original(UserAction)`

ltratt: Got it.

ltratt:
I think this is ready for squashing. If you agree, please go ahead.

ratmice:
Squashed, I also tried to write a higher level overview in the commit message explaining it.

PR #667, opened 2026-09-11
Add API for user defined grmtools section entries in GrammarAST

ratmice:
This is a second attempt at exposing querying of entries defined by downstream crates stored in the `%grmtools` section.
The first attempt was #665, this attempt is extended to allow crates to query for unused keys defined within their namespace. And is overall simpler due to being based on the new `header::Value` type work done in #666 .

### DIFF
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -127,15 +127,18 @@ impl ASTWithValidityInfo {
         }
     }
 
-    pub fn unused_header_keys_for_crate(&self, crate_name: &str) -> Vec<String> {
+    pub fn unused_header_keys_for_crate(&self, crate_name: &str) -> Vec<(String, Span)> {
         self.grmtools_section
             .unused()
             .iter()
-            .filter(|key_name| {
+            .filter_map(|(key_name, HeaderValue(key_span, _))| {
                 let crate_prefix = format!("{crate_name}.");
-                key_name.starts_with(&crate_prefix)
+                if key_name.starts_with(&crate_prefix) {
+                    Some((key_name.clone(), *key_span))
+                } else {
+                    None
+                }
             })
-            .cloned()
             .collect::<Vec<_>>()
     }
 }
@@ -1084,7 +1087,7 @@ start -> () : "a" { () };
         }
         assert_eq!(
             ast_validity.unused_header_keys_for_crate("test"),
-            vec!["test.unused"]
+            vec![("test.unused".to_string(), src.find_span("test.unused"))]
         );
         assert_eq!(
             ast_validity.grmtools_section_value_for_crate("cfgrammar", "yacckind"),
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -14,7 +14,7 @@ use super::{
 
 use crate::{
     Span,
-    header::{GrmtoolsSectionParser, HeaderError, HeaderErrorKind, HeaderValue},
+    header::{GrmtoolsSectionParser, Header, HeaderError, HeaderErrorKind, HeaderValue, Value},
     yacc::YaccOriginalActionKind,
 };
 
@@ -178,6 +178,7 @@ pub struct GrammarAST {
     // The set of symbol names that, if unused in a
     // grammar, will not cause a warning or error.
     pub expect_unused: Vec<Symbol>,
+    pub grmtools_section: Option<Header<Span>>,
 }
 
 #[derive(Debug, Clone)]
@@ -255,6 +256,7 @@ impl GrammarAST {
             parse_generics: None,
             programs: None,
             expect_unused: Vec::new(),
+            grmtools_section: None,
         }
     }
 
@@ -542,6 +544,39 @@ impl GrammarAST {
                         }
                     }),
             )
+    }
+
+    /// Performs a lookup in the grmtools section for an entry with the key `crate_name.key_name` and returns it.
+    /// If the entry is found it marks the key as `used`, for the purposes of `unused_grmtools_section_keys_for_crate`.
+    pub fn grmtools_section_value_for_crate(
+        &mut self,
+        crate_name: &str,
+        key_name: &str,
+    ) -> Option<(Span, &Value<Span>)> {
+        let key = format!("{crate_name}.{key_name}");
+        if let Some(HeaderValue(span, value)) = self.grmtools_section.as_mut().and_then(|map| {
+            map.mark_used(&key);
+            map.get(&key)
+        }) {
+            Some((*span, value))
+        } else {
+            None
+        }
+    }
+
+    pub fn unused_grmtools_section_keys_for_crate(&self, crate_name: &str) -> Vec<String> {
+        if let Some(map) = &self.grmtools_section {
+            map.unused()
+                .iter()
+                .filter(|key_name| {
+                    let crate_prefix = format!("{crate_name}.");
+                    key_name.starts_with(&crate_prefix)
+                })
+                .cloned()
+                .collect::<Vec<_>>()
+        } else {
+            vec![]
+        }
     }
 }
 
@@ -983,5 +1018,263 @@ start -> () : "a" {$;;;; };
                 spans: vec![Span::new(32, 33)],
             }]
         );
+    }
+
+    #[test]
+    fn test_grmtools_section_values() {
+        use super::*;
+        use crate::header::Value;
+        let src = r#"
+%grmtools {
+   yacckind: Grmtools,
+   lrpar.recoverer: CPCTPlus,
+   test.Flag,
+   !test.Negative,
+   test.string: "Foo",
+   test.vec: ["Aaaa", "Bbbb"],
+   test.num: 1234,
+   test.unused: 5678
+}
+%token a
+%%
+start -> () : "a" { () };
+"#;
+        let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
+        let test_flag_span = src.find_span("test.Flag");
+        let test_neg_span = src.find_span("test.Negative");
+        let test_neg_val_span = src.find_span("!test.Negative");
+        let test_string_span = src.find_span("test.string");
+        let test_string_val_span = src.find_span("Foo");
+        let test_vec_span = src.find_span("test.vec");
+        let test_vec_a_span = src.find_span("Aaaa");
+        let test_vec_b_span = src.find_span("Bbbb");
+        let test_vec_val_span = src.find_span("[\"Aaaa\", \"Bbbb\"]");
+        let test_num_span = src.find_span("test.num");
+        let test_num_val_span = src.find_span("1234");
+        let mut test_crate_expected = HashMap::new();
+        test_crate_expected.insert(
+            "Flag".to_string(),
+            (test_flag_span, Value::Bool(true, test_flag_span)),
+        );
+        test_crate_expected.insert(
+            "Negative".to_string(),
+            (test_neg_span, Value::Bool(false, test_neg_val_span)),
+        );
+        test_crate_expected.insert(
+            "string".to_string(),
+            (
+                test_string_span,
+                Value::String("Foo".to_string(), test_string_val_span),
+            ),
+        );
+        test_crate_expected.insert(
+            "vec".to_string(),
+            (
+                test_vec_span,
+                Value::Array(
+                    vec![
+                        Value::String("Aaaa".to_string(), test_vec_a_span),
+                        Value::String("Bbbb".to_string(), test_vec_b_span),
+                    ],
+                    test_vec_val_span,
+                ),
+            ),
+        );
+        test_crate_expected.insert(
+            "num".to_string(),
+            (test_num_span, Value::Num(1234, test_num_val_span)),
+        );
+        for (key, (expected_span, expected_value)) in test_crate_expected {
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("test", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert_eq!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("test"),
+            vec!["test.unused"]
+        );
+
+        let mut cfgrammar_crate_expected = HashMap::new();
+        let yacckind_span = src.find_span("yacckind");
+        let yacckind_val_span = src.find_span("Grmtools");
+        cfgrammar_crate_expected.insert(
+            "yacckind".to_string(),
+            (
+                yacckind_span,
+                // The actual value we receive has been lower cased
+                Value::Namespaced("Grmtools".to_string(), yacckind_val_span),
+            ),
+        );
+        for (key, (expected_span, expected_value)) in cfgrammar_crate_expected {
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("cfgrammar", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("cfgrammar")
+                .is_empty()
+        );
+
+        let mut lrpar_crate_expected = HashMap::new();
+        let recoverer_span = src.find_span("lrpar.recoverer");
+        let recoverer_val_span = src.find_span("CPCTPlus");
+        lrpar_crate_expected.insert(
+            "recoverer".to_string(),
+            (
+                recoverer_span,
+                Value::Namespaced("CPCTPlus".to_string(), recoverer_val_span),
+            ),
+        );
+        for (key, (expected_span, expected_value)) in lrpar_crate_expected {
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("lrpar", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("lrpar")
+                .is_empty()
+        );
+    }
+
+    #[test]
+    fn test_grmtools_section_values2() {
+        use super::*;
+        let src = r#"
+%grmtools {
+   yacckind: Original(YaccOriginalActionKind::UserAction),
+}
+%token a
+%actiontype ()
+%%
+start: "a" { () };
+"#;
+        let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
+        let mut cfgrammar_crate_expected = HashMap::new();
+        let yacckind_span = src.find_span("yacckind");
+        let yacckind_val_span = src.find_span("Original(YaccOriginalActionKind::UserAction)");
+        cfgrammar_crate_expected.insert(
+            "yacckind".to_string(),
+            (
+                yacckind_span,
+                Value::Namespaced(
+                    "Original(YaccOriginalActionKind::UserAction)".to_string(),
+                    yacckind_val_span,
+                ),
+            ),
+        );
+        for (key, (expected_span, expected_value)) in cfgrammar_crate_expected {
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("cfgrammar", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("cfgrammar")
+                .is_empty()
+        );
+    }
+
+    #[test]
+    fn test_grmtools_section_values3() {
+        use super::*;
+        let src = r#"
+%grmtools {
+   yacckind: YaccKind::Original(YaccOriginalActionKind::UserAction),
+}
+%token a
+%actiontype ()
+%%
+start: "a" { () };
+"#;
+        let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
+        let mut cfgrammar_crate_expected = HashMap::new();
+        let yacckind_span = src.find_span("yacckind");
+        let yacckind_val_span =
+            src.find_span("YaccKind::Original(YaccOriginalActionKind::UserAction)");
+        cfgrammar_crate_expected.insert(
+            "yacckind".to_string(),
+            (
+                yacckind_span,
+                Value::Namespaced(
+                    "YaccKind::Original(YaccOriginalActionKind::UserAction)".to_string(),
+                    yacckind_val_span,
+                ),
+            ),
+        );
+        for (key, (expected_span, expected_value)) in cfgrammar_crate_expected {
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("cfgrammar", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("cfgrammar")
+                .is_empty()
+        );
+    }
+
+    #[test]
+    fn test_grmtools_section_values4() {
+        use super::*;
+        let src = r#"
+%grmtools {
+   yacckind: YaccKind::Original(UserAction),
+}
+%token a
+%actiontype ()
+%%
+start: "a" { () };
+"#;
+        let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
+        let mut cfgrammar_crate_expected = HashMap::new();
+        let yacckind_span = src.find_span("yacckind");
+        let yacckind_val_span = src.find_span("YaccKind::Original(UserAction)");
+        cfgrammar_crate_expected.insert(
+            "yacckind".to_string(),
+            (
+                yacckind_span,
+                Value::Namespaced(
+                    "YaccKind::Original(UserAction)".to_string(),
+                    yacckind_val_span,
+                ),
+            ),
+        );
+        for (key, (expected_span, expected_value)) in cfgrammar_crate_expected {
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("cfgrammar", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("cfgrammar")
+                .is_empty()
+        );
+    }
+
+    trait FindSpan {
+        fn find_span(&self, s: &str) -> Span;
+    }
+
+    impl FindSpan for &'_ str {
+        #[track_caller]
+        fn find_span(&self, s: &str) -> Span {
+            let start_pos = self.find(s).unwrap();
+            Span::new(start_pos, start_pos + s.len())
+        }
     }
 }
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -150,20 +150,20 @@ impl FromStr for ASTWithValidityInfo {
             .map_err(|mut errs| errs.drain(..).map(|e| e.into()).collect::<Vec<_>>())?;
         if let Some(HeaderValue(_, yk_val)) = header.get("cfgrammar.yacckind") {
             let yacc_kind = YaccKind::try_from(yk_val).map_err(|e| vec![e.into()])?;
-            let ast = {
+            let (ast, grmtools_section) = {
                 // We don't want to strip off the header so that span's will be correct.
                 let mut yp = YaccParser::new(yacc_kind, src);
                 yp.parse().map_err(|e| errs.extend(e)).ok();
-                let (mut ast, _) = yp.build();
+                let (mut ast, grmtools_section) = yp.build();
                 ast.complete_and_validate(Some(yacc_kind))
                     .map_err(|e| errs.push(e))
                     .ok();
-                ast
+                (ast, grmtools_section)
             };
             Ok(ASTWithValidityInfo {
                 ast,
                 errs,
-                grmtools_section: header,
+                grmtools_section,
                 yacc_kind,
             })
         } else {
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -113,7 +113,7 @@ impl ASTWithValidityInfo {
 
     /// Performs a lookup in the grmtools section for an entry with the key `crate_name.key_name` and returns it.
     /// If the entry is found it marks the key as `used`, for the purposes of `unused_header_keys_for_crate`.
-    pub fn grmtools_section_value_for_crate(
+    pub fn header_value_for_crate(
         &mut self,
         crate_name: &str,
         key_name: &str,
@@ -1082,7 +1082,7 @@ start -> () : "a" { () };
                 ),
             ),
         ] {
-            let value = ast_validity.grmtools_section_value_for_crate("test", key);
+            let value = ast_validity.header_value_for_crate("test", key);
             assert_eq!(value, Some((expected_span, &expected_value)));
         }
         assert_eq!(
@@ -1090,7 +1090,7 @@ start -> () : "a" { () };
             vec![("test.unused".to_string(), src.find_span("test.unused"))]
         );
         assert_eq!(
-            ast_validity.grmtools_section_value_for_crate("cfgrammar", "yacckind"),
+            ast_validity.header_value_for_crate("cfgrammar", "yacckind"),
             Some((
                 src.find_span("yacckind"),
                 &Value::Namespaced("Grmtools".to_string(), src.find_span("Grmtools"))
@@ -1104,7 +1104,7 @@ start -> () : "a" { () };
         );
 
         assert_eq!(
-            ast_validity.grmtools_section_value_for_crate("lrpar", "recoverer"),
+            ast_validity.header_value_for_crate("lrpar", "recoverer"),
             Some((
                 src.find_span("lrpar.recoverer"),
                 &Value::Namespaced("CPCTPlus".to_string(), src.find_span("CPCTPlus"))
@@ -1132,7 +1132,7 @@ start: "a" { () };
 "#;
         let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
         assert_eq!(
-            ast_validity.grmtools_section_value_for_crate("cfgrammar", "yacckind"),
+            ast_validity.header_value_for_crate("cfgrammar", "yacckind"),
             Some((
                 src.find_span("yacckind"),
                 &Value::Namespaced(
@@ -1162,7 +1162,7 @@ start: "a" { () };
 "#;
         let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
         assert_eq!(
-            ast_validity.grmtools_section_value_for_crate("cfgrammar", "yacckind"),
+            ast_validity.header_value_for_crate("cfgrammar", "yacckind"),
             Some((
                 src.find_span("yacckind"),
                 &Value::Namespaced(
@@ -1192,7 +1192,7 @@ start: "a" { () };
 "#;
         let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
         assert_eq!(
-            ast_validity.grmtools_section_value_for_crate("cfgrammar", "yacckind"),
+            ast_validity.header_value_for_crate("cfgrammar", "yacckind"),
             Some((
                 src.find_span("yacckind"),
                 &Value::Namespaced(
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -130,6 +130,9 @@ impl ASTWithValidityInfo {
         }
     }
 
+    /// Returns all key names given in the header specified by a `%grmtools` directive with the
+    /// `crate_name.` prefix for the given crate. If the `crate_name` is None returns any unused
+    /// keys with no crate prefix specified.
     pub fn unused_header_keys_for_crate(&self, crate_name: Option<&str>) -> Vec<(String, Span)> {
         self.grmtools_section
             .unused()
--- a/cfgrammar/src/lib/yacc/ast.rs
+++ b/cfgrammar/src/lib/yacc/ast.rs
@@ -1040,51 +1040,49 @@ start -> () : "a" {$;;;; };
 start -> () : "a" { () };
 "#;
         let mut ast_validity = ASTWithValidityInfo::from_str(src).unwrap();
-        let test_flag_span = src.find_span("test.Flag");
-        let test_neg_span = src.find_span("test.Negative");
-        let test_neg_val_span = src.find_span("!test.Negative");
-        let test_string_span = src.find_span("test.string");
-        let test_string_val_span = src.find_span("Foo");
-        let test_vec_span = src.find_span("test.vec");
-        let test_vec_a_span = src.find_span("Aaaa");
-        let test_vec_b_span = src.find_span("Bbbb");
-        let test_vec_val_span = src.find_span("[\"Aaaa\", \"Bbbb\"]");
-        let test_num_span = src.find_span("test.num");
-        let test_num_val_span = src.find_span("1234");
-        let mut test_crate_expected = HashMap::new();
-        test_crate_expected.insert(
-            "Flag".to_string(),
-            (test_flag_span, Value::Bool(true, test_flag_span)),
-        );
-        test_crate_expected.insert(
-            "Negative".to_string(),
-            (test_neg_span, Value::Bool(false, test_neg_val_span)),
-        );
-        test_crate_expected.insert(
-            "string".to_string(),
+        for (key, (expected_span, expected_value)) in [
             (
-                test_string_span,
-                Value::String("Foo".to_string(), test_string_val_span),
-            ),
-        );
-        test_crate_expected.insert(
-            "vec".to_string(),
-            (
-                test_vec_span,
-                Value::Array(
-                    vec![
-                        Value::String("Aaaa".to_string(), test_vec_a_span),
-                        Value::String("Bbbb".to_string(), test_vec_b_span),
-                    ],
-                    test_vec_val_span,
+                "Flag".to_string(),
+                (
+                    src.find_span("test.Flag"),
+                    Value::Bool(true, src.find_span("test.Flag")),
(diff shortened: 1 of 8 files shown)


cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -984,4 +1019,262 @@ start -> () : "a" {$;;;; };
...
+            let value = ast_validity
+                .ast
+                .grmtools_section_value_for_crate("test", &key);
+            assert_eq!(value, Some((expected_span, &expected_value)));
+        }
+        assert_eq!(
+            ast_validity
+                .ast
+                .unused_grmtools_section_keys_for_crate("test"),
+            vec!["test.unused"]
```
ratmice, line 1087: `CTParserBuilder` and `nimbleparse` are still using the `ast_validity.ast_grmtools_section.unused()` directly, rather than `unused_grmtools_section_keys_for_crate`.

So this usage seen in the testsuite with `test.foo` keys is likely to still trigger an error in practice, I had kind of forgotten about this until just now.
Unsure if we want to relax those errors in this patch, or a subsequent one?

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -543,6 +545,39 @@ impl GrammarAST {
...
+            map.get(&key)
+        }) {
+            Some((*span, value))
+        } else {
+            None
+        }
+    }
+
+    pub fn unused_grmtools_section_keys_for_crate(&self, crate_name: &str) -> Vec<String> {
+        if let Some(map) = &self.grmtools_section {
```
ratmice: The primary reason I marked this as draft, was I was wondering if we should try to make this return a `Span` too? (Also the function name doesn't really roll off the tongue!)

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -984,4 +1019,262 @@ start -> () : "a" {$;;;; };
...
+        let test_string_span = src.find_span("test.string");
+        let test_string_val_span = src.find_span("Foo");
+        let test_vec_span = src.find_span("test.vec");
+        let test_vec_a_span = src.find_span("Aaaa");
+        let test_vec_b_span = src.find_span("Bbbb");
+        let test_vec_val_span = src.find_span("[\"Aaaa\", \"Bbbb\"]");
+        let test_num_span = src.find_span("test.num");
+        let test_num_val_span = src.find_span("1234");
+        let mut test_crate_expected = HashMap::new();
+        test_crate_expected.insert(
```
ratmice: Perhaps this whole test can be cleaned up by using a `Vec<(key, span, value)>` instead of a `HashMap`.
That was kind of a carry-over from the prior attempt.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -543,6 +545,39 @@ impl GrammarAST {
                     }),
             )
     }
+
+    /// Performs a lookup in the grmtools section for an entry with the key `crate_name.key_name` and returns it.
+    /// If the entry is found it marks the key as `used`, for the purposes of `unused_grmtools_section_keys_for_crate`.
+    pub fn grmtools_section_value_for_crate(
+        &mut self,
+        crate_name: &str,
+        key_name: &str,
```
ratmice: I was kind of undecided whether this should do the `format!({crate_name}.{key_name})`, or
just take the key as a single string, including the crate name.

I just picked one, randomly based on the `unused` takes the crate name as a separate parameter,
but no strong opinions.

cfgrammar/src/lib/yacc/ast.rs
ratmice: I tried to do some clean up here in [eff4480](https://github.com/softdevteam/grmtools/pull/667/commits/eff44808165a9d16215938ae0840d3e267a48501)

Perhaps I went overboard reducing the let bound variables?

cfgrammar/src/lib/yacc/ast.rs
ratmice: This seems like something I need to investigate sooner rather than later, it isn't *just* that we're using `unused`, we're actually dealing with duplicate `Header` entries entirely, and there is some passing of that header value around mutably between lrlex/lrpar.

So we kind of need to move over to the new lookup API through the `GrammarAST` too.
There are still some aspects that aren't publicly exported through the `GrammarAST`, like required fields.

Edit: moved some comments here to a more relevant place.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -543,6 +545,39 @@ impl GrammarAST {
                     }),
             )
     }
+
+    /// Performs a lookup in the grmtools section for an entry with the key `crate_name.key_name` and returns it.
+    /// If the entry is found it marks the key as `used`, for the purposes of `unused_grmtools_section_keys_for_crate`.
+    pub fn grmtools_section_value_for_crate(
+        &mut self,
```
ratmice: Well, I'm concerned about this `&mut self` which is needed for the call to `mark_used`.

Given that `GrammarAST` has a lot of `pub` fields, and I don't think we currently
have any methods to obtain a `mut GrammarAST` outside of the cfgrammar crate where the tests reside.

Thus this seems like it might not work, maybe we could/should store it in the `ASTWithValidityInfo` somehow?

ratmice: I moved it in [a96e799](https://github.com/softdevteam/grmtools/pull/667/commits/a96e799d62cb5b83f8befbd47fa0f33cd24db537)

I think with that I should be able to start on the `CTParserBuilder` parts ([adb16f4](https://github.com/softdevteam/grmtools/pull/667/commits/adb16f4c6bd2c9ee6125dcad8f8b21317f845f35)).

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -124,7 +154,7 @@ impl FromStr for ASTWithValidityInfo {
                 // We don't want to strip off the header so that span's will be correct.
                 let mut yp = YaccParser::new(yacc_kind, src);
                 yp.parse().map_err(|e| errs.extend(e)).ok();
-                let mut ast = yp.build();
+                let (mut ast, _) = yp.build();
```
ratmice: We have to parse the header twice here, and we may be using the wrong instance.
The first parse is to pull out the `YaccKind`, and after that it goes through the normal route
through `YaccParser::parse`.

It probably marks entries as required in the instance returned from `YaccParser::build`, we
also use that instance in `ASTWithValidityInfo::new()` when the `YaccKind` is always known, so that would follow the same code path we use elsewhere.

ratmice: Should be fixed in [86a4457](https://github.com/softdevteam/grmtools/pull/667/commits/86a4457a22bda501212c28ee08f7800a23c8be2a)

cfgrammar/src/lib/yacc/ast.rs
ratmice: Making CTParserBuilder work should be fixed now in [adb16f4](https://github.com/softdevteam/grmtools/pull/667/commits/adb16f4c6bd2c9ee6125dcad8f8b21317f845f35)

lrpar/src/lib/codegen.rs
```diff
@@ -1294,3 +1312,74 @@ pub(crate) fn make_generics(parse_generics: Option<&str>) -> Result<Generics, Co
...
+        }
+        %%
+        start -> () : "A" { () };
+        "#;
+        let empty_header = Header::<Location>::new();
+        let src_env = ParserSrcEnv::<TestLexerTypes>::new_with_header(src, None, empty_header);
+        let build_env = src_env
+            .build_env(ParserBuildEnvArgs::new().mod_name(Some("test_module")))
+            .unwrap();
+        build_env
```
ratmice, line 1445: This may still need some work, on how to expose this check in a way that will work with downstream crates that want to use the code_generator. But this stuff is all private still anyways.

The thing to note is that this check still happens in `CTParserBuilder` so `code_generator` is not *automatically* checking for unused keys as a byproduct of code generation.

Hence having to reimplement the checks in these test cases as calls to `build_env.check_unused...`

ratmice:
It took me a while thinking about how to handle that there are multiple header instances across various crates.
for instance we have a `Header<Span>` in `cfgrammar::yacc::ASTWithValidityInfo` and a `Header<Location>` in `CTParserBuilder`/`lrpar::codegen`. These all get merged into the `Header<Location>` and each of their owners may call `mark_used` on them and then later check the unused status of some key.

I came to the conclusion that it isn't actually a problem, because these `Header<Location>` only know about keys for grmtools crates and only check for those. Further, those header instances aren't visible to downstream crates.

The key to this working is that each of these crates is going to be marking and checking the keys for the crates they know in the header they'll check, so we don't really have much of an issue with a key being marked used in one `Header` then checked for used status in another.

But I don't imagine that is either an obvious as either problem or solution. There's really two ways this could be a problem:

1. we start wanting to add ways that downstream crates can set their values via `CTParserBuilder` (beyond adding them to their parser source)
2. downstream crates perform a `check_unused` on cfgrammar/lrpar/lrlex crate entries.

The second case actually will currently fail as we're only calling mark_used in the `Header<Location>` rather than the `ast_with_validation_info.grmtools_section`, the fix that comes to mind is ensuring mark_used is called for both `Header` instances.

Edit: I'll work on a fix for this second case, and feel like we can just accept that we won't do the first?
Edit 2: There is one complication in that this means we need mutable rather than shared references to the `ASTWithValidityInfo` it doesn't seem that it is possible for e.g. `lrpar::codegen::` to satisfy the borrow checker with an `&mut ASTWithValidityInfo`.

So i'm wondering if we just perform the checks e.g. during codegen, and allow inconsistent results if a user tries to call `unused_grmtools_section_keys_for_crate("cfgrammar")` e.g. for an external crate like cfgrammar/lrpar/lrlex on their own.

lrpar/src/lib/codegen.rs
```diff
@@ -1294,3 +1312,74 @@ pub(crate) fn make_generics(parse_generics: Option<&str>) -> Result<Generics, Co
...
+            .check_unused_header_keys_for_crate(Some("cfgrammar"))
+            .unwrap();
```
ratmice: So the main problem I'm thinking about in the ginormous comment is that if we change these
`build_env.check_unused_header_keys_for_crate(Some("cfgrammar"))` calls to:

`ast_with_validity_info.unused_grmtools_section_keys_for_crate("cfgrammar")`
These start failing, because the codegen process is merging keys, and then working on the merged structure rather than the one in the AST.

Further it only has a shared AST, and it seems difficult to change that to a mutable one.

ratmice: The best thing that has come to mind is calling `mark_used` in `YaccParser::parse()` or `build` this is before the keys actually get used, but before ownership is moved.

ratmice: That is done in [fcba11b](https://github.com/softdevteam/grmtools/pull/667/commits/fcba11b571ca53d7c7e49f91f25b6aae027f2dff)

cfgrammar/src/lib/yacc/ast.rs
ratmice: Turns out there was a lot more reasons this should have been marked as draft,
but this should be fixed in [e8012c9](https://github.com/softdevteam/grmtools/pull/667/commits/e8012c93efd8c71ef2657628599a3ef2f7db97a2)

lrpar/src/lib/codegen.rs
ratmice: Regardless of whether downstream crates automatically get checking for unused keys,
they can now perform the check themselves to obtain the same results as `CTParserBuilder`?

ratmice:
Marking as ready, because I *think* I've covered all the issues I can think of as best I can.
Though this has definitely not been as smooth of a patch process as I hoped/thought it would have been.

ratmice:
I've tried a couple of times to write a sort of high level comment about design issues, but it keeps turning into a textbook/great american novel.

This was mostly talking about duplication between e.g. `ASTWithValidityInfo` and `codegen::` module, where `codegen::` currently is acting solely like a *sink*, and `ASTWithValidityInfo` is acting primarily as a *source* one is consuming entries, the other is providing them. But they overlap in providing this static analysis/checking of unused variables.

There is something to say though about the mechanisms provided by this patch, which is they currently *aren't* exposing any of this or making it usable from within `CTParserBuilder`. That deserves some high level description of what this patch is providing/background info.

In nimbleparse_lsp v2, we're basically calling `ASTWithValidityInfo::from_str` with the grammar source,
we plan on using the `codegen::` module directly, passing the prebuilt `ASTWithValidityInfo` to the code generator
with `ParserBuildEnvArgs::ast_with_validity_info`, but between those two calls we can pull out any user specified keys.

One of the unforseen problems, which I guess I am encountering now, and didn't anticipate is that the intent was for the public API to kind of not expose the `Location` type, by providing a `new()` function which just uses an empty `Header<Location>`, but that leaves the `Location` type itself still getting exposed via errors. Even though the only error that can happen are `Location::Span` ones. One thing we could consider doing is adding a generic parameter to the `codegen::` structures, so the hidden constructor `ParserSrcEnv::new_with_header` could return a `ParserSrcEnv<LexerTypesT, Location>` while `ParserSrcEnv::new()` could return a `ParserSrcEnv<LexerTypesT, Span>`, but that is a pretty significant change (and seems likely to be impossible).

It hadn't really occurred to me how/whether we plan on exposing this via `CTParserBuilder` or `codegen::` there is additional complexity with adding that, since it involves `Location` and the additional `Header<Location>` values. That is to say `codegen::` has *two* structures we could query. The one read from the file itself containing `Spans` and the one with  all the default values having been resolved and plugged in with `Locations`. So this patch hasn't attempted to really come up with any answers about this latter problem.

ratmice:
I'm going to go ahead and mark most of my waffling review comments as resolved for now.
But feel free to unresolve anything, if you feel like deserves more attention or a different approach.

It just feels like they're probably more distracting than helpful at this point.

lrpar/src/lib/codegen.rs
ratmice: Having thought about it, I'm in the coping stage where it feels like there isn't much the code generator can do about this. It doesn't know at what time all the header values have been resolved, since it gives mutable access to the header via `PaserBuildEnv::header_mut()`, currently the `check_unused_header_keys_for_crate` function is *also* on `PaserBuildEnv`.

Edit: Oops, disregard the paragraph below. I was looking at the `lrlex::codegen`, in `lrpar::codegen` we actually  need to mutate the header/mark keys as used *after* the call to `code_generator()`.
The reason we need it *after* is because of the `inspect_rt` callback which we use to implement `test_files`, it builds a `RTParserBuilder`, and marks the `test_files` entry as used.
It is worth noting though that `test_files` is explicitly marked as experimental in the book [here](https://softdevteam.github.io/grmtools/master/book/yaccextensions.html#fr-%E2%80%A0-1).

Edit2: Given the above perhaps we code do the checking in `ParserCodegen::generate` it takes a `ParserBuildEnv` and returns a `Result`. This provides a place to do the check, but still has some open questions about `Header<Location>` vs `Header<Span>`, and which header to perform the check on during codegen.

It looks like it *could* work if we actually made this a private function, and called it from `ParserBuildEnv::code_generator`. That is the next point after all header values should have been marked used. Presumably we'd need some way perhaps in `ParserBuildEnvArgs` to pass it a list of crates to check.

lrpar/src/lib/codegen.rs
```diff
@@ -1294,3 +1315,177 @@ pub(crate) fn make_generics(parse_generics: Option<&str>) -> Result<Generics, Co
...
+                .ast_with_validity_info()
+                .unused_header_keys_for_crate(Some("lrlex"))
+                .is_empty()
+        );
+        let codegen = build_env.code_generator("timestamp").unwrap();
+        let out = codegen.generate(&build_env).unwrap();
+        assert!(!out.is_empty());
+    }
+
+    trait FindSpan {
```
ratmice: This should probably just go in `test_utils.rs`

ratmice: Fixed in [9e0c164](https://github.com/softdevteam/grmtools/pull/667/commits/9e0c16401bf0c733b4474b2059f44bd2756d5928)

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -108,6 +113,48 @@ impl ASTWithValidityInfo {
             })
         }
     }
+
+    /// Performs a lookup in the grmtools section for an entry with the key `crate_name.key_name` and returns it.
+    /// If the entry is found it marks the key as `used`, for the purposes of `unused_header_keys_for_crate`.
+    pub fn header_value_for_crate(
+        &mut self,
+        crate_name: &str,
```
ltratt: Dumb question: do we even need to force two variables (`crate_name` and `key_name`) on the user? Would `key: &str` make sense alone?

lrpar/src/lib/codegen.rs
```diff
@@ -456,8 +459,26 @@ where
         self.ast_with_validity_info.yacc_kind()
     }
 
-    pub(crate) fn check_unused_header_keys(&self) -> Result<(), ParserBuildEnvError<LexerTypesT>> {
-        let unused_keys = self.header.unused();
+    /// Returns an error if any unused keys specified in a `%grmtools` directive that begin with a
+    /// `crate_name.` prefix for `crate_name` value are found. If the `crate_name` is None returns
+    ///  an error if any unused keys with no crate prefix specified are found.
```
ltratt: I wonder if instead of `crate_name` we have `prefix`, where `""` means "match all" and any non-empty string implicitly ends with `.` (we don't want "x" to match "xy.z")?

cfgrammar/src/lib/yacc/ast.rs
ratmice: We don't, I made the same comment in a review above too, for a different function.
I kind of flipped a coin, and did separate crate names and keys, due to the separate crate name elsewhere
like `check_unused_...`  where the key name isn't passed in at all.

It didn't seem like a strong argument then, happy to change it.

ltratt: I _think_ that if we use a single string, we get a nicer API in https://github.com/softdevteam/grmtools/pull/667#discussion_r4000809086 too? Warning: I might be wrong.

lrpar/src/lib/codegen.rs
ratmice: I'm not exactly certain how to interpret the suggestion,

One thing to note is that the current behavior, `None` is not actually "match all", but "match anything without any crate prefix".

Are you saying we should have 
* `Some("")` should match *all* keys.
* `None` should match keys with no prefix.
* `Some("foo")` should match "foo.*"

Or did you just mean to replace the option with just the empty string?

ltratt: Ah! OK, so I _think_ we had this discussion recently (or at least I'm hallucinating something similar). I think intuition approach is to say "everything has a prefix; for legacy keys, we add the prefix "grmtools" (or "lrpar" or ...)". So "match" then becomes "match a prefix". So `match("")` matches everything; `match("grmools")` matches grmtools' keys, and so on. Again, warning: I haven't thought deeply about this and this might not be the best API.

ratmice: Ahh, maybe we need to handle unknown unprefixed keys elsewhere then.

The way it currently works, unprefixed keys can squeek by, because the prefix is only added for *known* keys.
So if you have a key like `unknown: "foo"`, it can end up in the header because it doesn't match any "grmtools", "lrpar", or "lrlex" keys.

Perhaps that should be an error when we try to insert the key into the `Header` instead of handling it at this point where they key just happens to end up not being used by anything.

This is kind of a "lazy" way to handle that error of an unknown key with no crate prefix.

ltratt: Aha, yes, I think unprefixed-keys-that-aren't-recognised-legacy-keys are probably best banned. Forcing namespacing on users seems like a good idea (at least right now).

ratmice: Okay, so the next question is do we actually want/need the match-all behavior?
I guess I see how it is useful for users to call on their own grammars to ensure all their keys are valid.

I had been looking at it from the perspective of grmtools, and nimbleparse_lsp, where there
maybe crate specific keys that these crates can't know about. So they can't perform the checks
for external crates.

I suppose that end-user checking is enough motiviation to add it.

ltratt: I think the "match all" behaviour falls out of the hat, so unless we ban it, it Just Works (TM). I _think_ it's useful to have it, at least from grmtools internal purposes? But we could start without it (i.e. ban it), and only unban it later if we find someone has a real use case for it: I'm fine with doing that.

ratmice: I think there is a use for it, as an end-user you can typo the crate name which will get by the per-crate checks. So if you do the match-all, it'll catch that. But it's only really possible to do if you are certain all the keys have been used.

For instance if you want to add that check to *build.rs* in a grammar that contains nimbleparse_lsp keys, you'll need to mark them as used yourself (or I can provide a helper function).

That is to say this key checking becomes somewhat of a distributed problem.

ltratt: Good point -- I agree.

ratmice: So, the first part of this is done in [f3079aa](https://github.com/softdevteam/grmtools/pull/667/commits/f3079aaae2bbabfdb585a13ef2fbd001a9246b29).

I'm not certain I understand the string matching behavior.

> I wonder if instead of `crate_name` we have `prefix`, where `""` means "match all" and any non-empty string implicitly ends with `.` (we don't want "x" to match "xy.z")?

Are you saying we should assume the user passes a `"foo."` and just call `key_name.starts_with(prefix)`.
Avoiding the `format!` call, if so there is a weird problem it that can match unexpected crates sharing a prefix. We could produce an error if the prefix is non-empty and doesn't end with a dot to avoid that format? This requires changing the return type to `Result<Vec<(String, Span)>, _>`

Or did you mean something like:
```
 if string.is_empty()  { true } else { 
      key_name.starts_with(&format("{prefix}.") 
}
```

ratmice: I did manage to come up with something in [ab83ba4](https://github.com/softdevteam/grmtools/pull/667/commits/ab83ba4deb829b9ec08aef3dd9bd6e8817898dec) that doesn't accidentally match crates with the same prefix.

Let me know if you had something else in mind, or if that seems acceptable.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Should be fixed in [75127c6](https://github.com/softdevteam/grmtools/pull/667/commits/75127c6d7513f07af0110632f5b339781db2e7cb)

lrpar/src/lib/codegen.rs
ratmice: FWIW https://github.com/softdevteam/grmtools/pull/667#discussion_r3999823144 also impacts this API,
as it envisions these calls happening automatically as a side-effect of `ParserBuildEnv::code_generator`, and that the crate names get added as to a `HashSet` probably in `ParserBuildEnvArgs`.

That way downstream crates don't have to manually check for grmtools crates.
That is more of a question of how we should make this API public though.

ratmice:
So, I did think of two ways that we might be able to have a single `Header` structure across both codegen/`ASTWithValidityInfo`

1. We could make the field in `ASTWithValidityInfo` a `Header<Location>` then merge from the codegen `Header` into the one owned by `ASTWithValidityInfo` instead of the other way around (as it is now).
2. We could just abandon the whole multiple headers, and merging into the `Header<Location>` locally owned by codegen.

The latter is basically admitting defeat on the header/merge process, values that come from the builder would no longer not route through the local `Header<Location>` structure during the `CTParserBuilder` codegen process. It does also avoids making `Location` public in *codegen* errors, which I think helps. 

I'm not entirely certain that the first will work (I had difficulty with ownership, last time I tried to mutate it from CTParserBuilder, but this is early enough that it may not be a problem),

But perhaps it is worth playing around with these approaches.

Edit: I forgot that the local Header in point two *is* necessary for bootstrapping the `GrammarAST` construction via `YaccKind`, but we might be able to limit it to *just* that. Further, I was hoping I could do this experiment standalone on a separate branch, but it seems to rely on having the `Header` owned by the.

ratmice:
I did a first try at the second idea in my prior comment, that is in [0ca6955](https://github.com/softdevteam/grmtools/pull/667/commits/0ca6955033305e46c33f278ac138e1ce969c1418) 

It was kind of the minimal change needed to get everything to compile. There are still quite a few uses of `Location` in lrpar which probably need cleaning up. 

Edit: Think I forgot some fields like `SerialisationFormat` 

Should be fixed in [6cca7a6](https://github.com/softdevteam/grmtools/pull/667/commits/6cca7a6e9205c65a9f4407db01e2ea1a16f77199) along with a bunch of other things that needed to be migrated

There is still a little big of `Location`, left hanging around but primarily due to `lrlex`

cfgrammar/src/lib/yacc/ast.rs
ltratt: At first glance this does look like it improves things?

ratmice: Yeah, I think so.

Edit: Moved my other part of the coment elsewhere.

lrpar/src/lib/codegen.rs
ratmice: I think this part of the patch is a bit more of a trade-off, like I like that you can just pass in "crate_name" rather than Some("crate_name") but less enthusiastic about "" for match-all. 

But still it does seem like it really does improve things in the common case, where you're checking a crate name.

lrpar/src/lib/codegen.rs
```diff
@@ -382,11 +379,14 @@ where
         LexerTypesT: LexerTypes,
         usize: num_traits::AsPrimitive<LexerTypesT::StorageT>,
     {
-        self.merge_headers()?;
+        let (mut header, _) = GrmtoolsSectionParser::new(self.src, false).parse()?;
+        if self.yacckind.is_none() {
+            header.mark_required(&"cfgrammar.yacckind".to_string());
```
ratmice: Oops, adding this mark_required to the wrong header here.

The one we perform the check for required fields on is in `ASTWithValidityInfo`.
This should probably also be `if self.yacckind.is_none() || self.ast_with_validity_info.is_none() {`

ratmice:
If this route seems okay, perhaps as a follow up patch we could get rid of the generic type parameter `T` to `HeaderValue<T>`/`Value<T>` and just inline the `Span` argument directly?

lrpar/src/lib/codegen.rs
ratmice: I think actually we have the `MissingYaccKind` error which already covers this case, so we can just remove it.
I did that in [a4e14d5](https://github.com/softdevteam/grmtools/pull/667/commits/a4e14d5db78b1256ad2e90a3edbdff5a4d32cd0d)

lrpar/src/lib/codegen.rs
ratmice: With various changes, primarily [0ca6955](https://github.com/softdevteam/grmtools/pull/667/commits/0ca6955033305e46c33f278ac138e1ce969c1418) and [45fca2b](https://github.com/softdevteam/grmtools/pull/667/commits/45fca2b50a351233e24d03de33fc379ba20278e5)

Along with changing the place where we call `mark_used` and mutate the `Header` value have allowed these checks to be performed earlier, and largely eliminated the problem of multiple Header values, culminating in allowing this now be done as a side effect of calling `code_generator`.

So I think this should be fixed. Also, this was the last thing I was struggling with/unhappy about in this PR.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -108,6 +110,62 @@ impl ASTWithValidityInfo {
...
+        } else {
+            None
+        }
+    }
+
+    /// Returns all key names given in the header specified by a `%grmtools` directive with the
+    /// `crate_prefix.` prefix for the given crate. If the `crate_prefix` is empty returns all
+    ///  unused keys regardless of crate.
+    pub fn unused_header_keys_for_crate(&self, crate_prefix: &str) -> Vec<(String, Span)> {
+        self.grmtools_section
```
ratmice, line 129: So, one thing with this function that could be improved is that it gets called for each unused key, for each crate. It could if passed a set of crate names, do a single pass checking if the prefix of each unused key exists in the set of crate_prefixes.

So instead of doing a pass for each crate, it could do a single pass for all crates.

This wouldn't be hard to do now, I *think* it'd even be deterministic for the testsuite,
since `MarkMap` doesn't sort based on randomness.

I'm not sure if it is worth the trouble with the low number of crates and keys though?

ratmice:
If it helps I think we could split this into two patches, one that removes `Header` from `CTParserBuilder` and adds it to `ASTWithValidityInfo`, then a second patch which actually adds the unused check API?

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -108,6 +110,62 @@ impl ASTWithValidityInfo {
             })
         }
     }
+
+    /// Performs a lookup in the grmtools section for an entry with the key `crate_name.key_name` and returns it.
+    /// If the entry is found it marks the key as `used`, for the purposes of `unused_header_keys_for_crate`.
+    pub fn header_value_for_crate(&mut self, key: &str) -> Option<(Span, &Value<Span>)> {
```
ltratt: I think I might be tempted to name this `header_value_get(&mut self, key: &str)` and downplay the "crate" bit slightly. I do think that strongly encouraging users to use crate names as the prefix is a good idea, but I'm sure someone (maybe us!) will at some point have a good reason for deviating from that.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -108,6 +110,62 @@ impl ASTWithValidityInfo {
...
+            Some((*span, value))
+        } else {
+            None
+        }
+    }
+
+    /// Returns all key names given in the header specified by a `%grmtools` directive with the
+    /// `crate_prefix.` prefix for the given crate. If the `crate_prefix` is empty returns all
+    ///  unused keys regardless of crate.
+    pub fn unused_header_keys_for_crate(&self, crate_prefix: &str) -> Vec<(String, Span)> {
```
ltratt: Might this better be `iter_header_values(&self, prefix: &str) -> impl Iterator<...>`? I _think_ we can usefully avoid forcing `collect` / `Vec` on the user.

cfgrammar/src/lib/yacc/ast.rs
```diff
@@ -108,6 +110,62 @@ impl ASTWithValidityInfo {
...
+                {
+                    Some((key_name.clone(), *key_span))
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+    }
+
+    pub fn check_missing_required_keys_for_crate(&self, crate_prefix: &str) -> Vec<String> {
```
ltratt: I'm not quite sure what the purpose of this function is (at least in the external API).

cfgrammar/src/lib/yacc/parser.rs
```diff
@@ -371,8 +374,19 @@ impl YaccParser<'_> {
         }
     }
 
-    pub(crate) fn build(self) -> GrammarAST {
-        self.ast
+    pub(crate) fn build(self) -> (GrammarAST, Header<Span>) {
+        let mut header = self.header.expect("set by parse()");
```
ltratt, line 378: I'm not quite sure I've got this comment, but maybe it's because "it is difficult to do later" is the key part?

lrpar/src/lib/codegen.rs
```diff
@@ -456,17 +454,32 @@ where
         self.ast_with_validity_info.yacc_kind()
     }
 
-    pub(crate) fn check_unused_header_keys(&self) -> Result<(), ParserBuildEnvError<LexerTypesT>> {
-        let unused_keys = self.header.unused();
+    /// Causes the `code_generator()` function to check for unused entries in the grmtools section
+    /// starting for entries starting with `crate_prefix`.
+    #[allow(unused)]
+    pub(crate) fn add_value_checks_for_crate(&mut self, crate_prefix: &str) {
```
ltratt: Is this function useful if it's `unused`? [I ask this innocently!]

ltratt:
I think I get where this is going now. There's one implicit thing which I had perhaps missed earlier: we're allowing external users to mark keys as used/unused so they can then easily get the unused keys relevant to them. That's quite an interesting feature, but I must admit that my initial reaction is to wonder if we shouldn't push that functionality onto them. In other words, if I'm crate C and I define keys C.A and C.B, should it be my job to "read all keys beginning with C, and if there are any that aren't A or B, warn/error to the user"?

cfgrammar/src/lib/yacc/parser.rs
ratmice: Yeah, this comment is partially out of date now, the "local instance which merges the keys" has been basically removed in the last couple of commits on this pull request.

the "later" is referring to during `CTParserBuilder`, `mark_used` requires a mutable reference to the `Header<Local>` value, but `CTParserBuilder` has shared references to `ast_with_validity_info` which owns the `Header<Span>` that this value turns into.

So, beyond this point there wasn't really a good place where we can get a mutable reference to this `Header<Span>`.

With the `Header<Local>` gone in later patches, this ended up being *the* place where we call `mark_used`, rather than at the actual call-site where it was used. (Also because of the shared reference problem, but it simplified a few in `CTParserBuilder`  like not having to have some callbacks take a mutable reference any longer)

It is a bit weird (and perhaps a bit more error prone) to mark all the crate header keys used here in `build()` rather than marking at call-sites where they actually get used though.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Hmm, good question, I think historically there has only been one required key `YaccKind`, and
we have basically migrated all those checks to specific error checks that check explicitly for that with errors like `ParserSrcEnvError::MissingYaccKind`, rather than a check that the `yacckind` key is set in the header.

But previously `CTParserBuilder` unless it *knew* the `YaccKind` would set this as required in the header.
Which would then cause an error if there was no `%grmtools{yacckind: ...}`

ltratt: Ah, so maybe this relates to my question in https://github.com/softdevteam/grmtools/pull/667#issuecomment-5758347153. grmtools itself wants to check for unused headers with its prefix(es), so perhaps the questions is _where_ that check is.

lrpar/src/lib/codegen.rs
ratmice: This is primarily intended for external crates to call, but since we haven't made `codegen` `pub`,
essentially crates like `nimbleparse_lsp` would call `src_env.add_value_checks_for_crate("nimbleparse_lsp")` then on `build()`

it would cause unused checks for `cfgrammar`, all the crates that were registered.
maybe `register_key_prefix` or something to that effect would be a better name.

This isn't used because it was actually easier to just add `cfgrammar`, and all the default grmtools crates at once on construction. Rather than building and mutating for each crate.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Yeah, so this is the implementation of that check, the check is actually performed in the `ParserBuildEnv::code_generator()` function, returning an error if this check fails.

https://github.com/softdevteam/grmtools/pull/667/changes#diff-7a0605cd47a548bf700da1ff07486de5143ee3048757e0da3df60c6be542b852R496-R507

Edit: It's worth noting that check in `code_generator` was intended to handle *both* grmtools crates, and external crates too. There is an example of that in the test cases https://github.com/softdevteam/grmtools/pull/667/changes#diff-7a0605cd47a548bf700da1ff07486de5143ee3048757e0da3df60c6be542b852R1344-R1351

There was also a question whether this particular function should take a `HashSet`, instead of a crate name, and perform all the checks at once, leaving `codegen` to do all the cross-crate coordination

Edit: I think try and migrate to the `HashSet` api, to get rid of some accrued ugliness in this series, making this API from *ast.rs* `#[doc(hidden)]`

ratmice:
I think where the "check your own" namespace breaks down is with namespace typos, like say `cfgramma.yacckind`, ideally when the user is doing a check for crate prefix `src_env.add_value_checks_for_crate("")`, this centralized prefix checking code can catch that while it couldn't catch it if it's only checking `cfgrammar`.

But yeah, I don't really have a good playbook/map for how to best implement this kind of feature, like
I know I really *don't* like generic metadata fields where nobody has any ideas about the keys/entries,
and it's impossible to check or catch typos because of that. It's definitely possible I'm over complicating things because of my metadata table pet peeve though.

cfgrammar/src/lib/yacc/ast.rs
ratmice: I'll have to try it again, I *think* I tried to avoid the `collect`, and the iterator return value could outlive the borrowed `&self`, but maybe that was trying to get rid of the `key_name.clone()` call too.

ratmice:
Just one minor difference to note about your comment, which is about allowing external crates to mark their values.

A lot of the explicit API is what grmtools is using using internally, to mark keys as used, The `codegen` api here is largely trying to add a more implicit API.  Where as a side-effect of looking up keys, they get marked as used. And as a side effect of bulding a code generator, they get checked for being unused. The hope was that external crates wouldn't actually have to do the tracking or deal with an API surface to get the checks.

Currently in this patch they still require an explicit registration for that check via the `src_env.register_header_value_prefix("crate_name")` call, but perhaps we could also register the prefix for unused checks implicitly on key lookup too?

I think we'd still need an explicit check if we want a check that matches all namespaces. currently `src_env.register_header_value_prefix("")`

Is there a preference between implict/explicit? Also do you think the crate typo resistance is worth doing it in the crate rather than having external crates check their own namespace?

cfgrammar/src/lib/yacc/ast.rs
ratmice: It looks like that check you mention is missing from `nimbleparse`, which doesn't currently use `codegen` (not yet pub).

Though it's worth noting that you can build an `RTParserBuilder` from the result of `ParserCodegen::finish()` without calling `ParserCodegen::generate()`, so in theory we could modify nimbleparse to use that same check.

We could also perform the checks manually similar to how they are implemented in codegen though.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Also I think it should be `iter_unused_header_values`

cfgrammar/src/lib/yacc/ast.rs
ratmice: I went ahead and removed this function in [c3b3f6c](https://github.com/softdevteam/grmtools/pull/667/commits/c3b3f6cd6c27b7e2ad118687f1c5b578a608eabd) since CTParserBuilder/nimbleparse no longer needed anything like it.

Also updated to use `HashSet`, marked most of the `ast.rs` api as hidden.

cfgrammar/src/lib/yacc/ast.rs
ratmice: Should be fixed in [c3b3f6c](https://github.com/softdevteam/grmtools/pull/667/commits/c3b3f6cd6c27b7e2ad118687f1c5b578a608eabd)

I suppose, technically it returns a key rather than a value, but uniform function naming is more important?

cfgrammar/src/lib/yacc/ast.rs
ratmice: Renamed in [c3b3f6c](https://github.com/softdevteam/grmtools/pull/667/commits/c3b3f6cd6c27b7e2ad118687f1c5b578a608eabd)

lrpar/src/lib/codegen.rs
ratmice: I updated this comment in [e8e1e7f](https://github.com/softdevteam/grmtools/pull/667/commits/e8e1e7fb248c801fd0abcbaaa5c3c7b811be31c4)

Hopefully it is clearer, and better reflects the reality in the code.

nimbleparse/src/main.rs
```diff
@@ -242,6 +244,37 @@ fn main() {
...
     let yacc_kind = yk_arg.unwrap_or(yk_header_val.unwrap_or(YaccKind::Grmtools));
     let ast_validation = ASTWithValidityInfo::new(yacc_kind, &yacc_src);
+    // Note we don't expect to find any used lrlex keys, we want to produce an error if unused ones are found.
+    let crate_prefixes = HashSet::from_iter([
+        "cfgrammar".to_string(),
+        "lrpar".to_string(),
+        "lrlex".to_string(),
+    ]);
+    let unused_keys = ast_validation
+        .iter_unused_header_values(crate_prefixes)
```
ratmice: Hmm, it's worth noting I set this as hidden, so here we're calling a `doc(hidden)` function in *nimbleparse*.

Just now, after the fact I'm recalling that *nimbleparse* is supposed to be more of a working example. In which case we probably want to avoid calling hidden API?

Ideally once *codegen* is pub, this can be changed to use that and the hidden API will then be confined to being used from *codegen*.

cfgrammar/src/lib/yacc/ast.rs
ratmice: I had done this in [c3b3f6c](https://github.com/softdevteam/grmtools/pull/667/commits/c3b3f6cd6c27b7e2ad118687f1c5b578a608eabd)

cfgrammar/src/lib/yacc/parser.rs
ratmice: I updated this comment in [e8e1e7f](https://github.com/softdevteam/grmtools/pull/667/commits/e8e1e7fb248c801fd0abcbaaa5c3c7b811be31c4)

Hopefully it is clearer, and better reflects the reality in the code.

Edit: (I orginally posted this comment in the wrong thread)

lrpar/src/lib/codegen.rs
ratmice: I renamed it in [70ce6f1](https://github.com/softdevteam/grmtools/pull/667/commits/70ce6f13cd4df83aa81a70b5ef95ac0cc54958ef) hopefully that better reflects its eventual intended usage.

ratmice:
FWIW, I tried to write a relatively short document examining the API from the 3 perspectives of

1. In tree tools like nimbleparse
2. Out of tree downstream tools like nimbleparse_lsp
3. End-user crates like build.rs

Tried to keep it short, while being a bit more thorough than what I had done in comments here.

https://gist.github.com/ratmice/ff2678c672105766b9805b8eb91097f2

